PR #66, opened 2021-06-30
"should make" -> "should not make"

svmaris:


FabianKramm:
@svmaris thanks a lot for this PR! Looks good to me!

PR #89, opened 2021-07-16
Add --port flag to specify service port on the create command

### LINKED ISSUE #88: Set port to expose `vcluster create --expose`
In my cluster, using `--expose` does not work because the Service does not come up:

```
Error syncing load balancer: failed to ensure load balancer: Service cannot use 80 and 443 port
```

But if I manually create the load balancer service with port 8443, then it works. So it would come very in handy if I could set the port to be used in the `vcluster create` command.

mcbenjemaa:
Fixes #88

### DIFF
--- a/chart/templates/service.yaml
+++ b/chart/templates/service.yaml
@@ -12,7 +12,7 @@ spec:
   type: {{ .Values.service.type }}
   ports:
     - name: https
-      port: 443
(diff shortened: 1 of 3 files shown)


FabianKramm (approved):
@mcbenjemaa thanks a lot for the PR! Looks good to me!

PR #124, opened 2021-08-28
Custom resources for fake nodes

T-M-A:
Good day!
I propose to consider the option of specifying resources for fake nodes

New parameters:

--fake-nodes-cpu-count    = If enabled fake-nodes, create node with this CPU cores
--fake-nodes-mem-size   = If enabled fake-nodes, create node with this Mem capacity
--fake-nodes-storage-size = If enabled fake-nodes, create node with this storage size
--fake-nodes-hugemem-1G-count  = If enabled fake-nodes, create node with this hugeMem count pages ith 1G size
--fake-nodes-hugemem-2M-count = If enabled fake-nodes, create node with this hugeMem count pages ith 2M size

Thanks

### DIFF
--- a/cmd/vcluster/context/context.go
+++ b/cmd/vcluster/context/context.go
@@ -33,13 +33,18 @@ type VirtualClusterOptions struct {
 	ServiceNamespace     string
 	OwningStatefulSet    string
 
-	SyncAllNodes             bool
-	SyncNodeChanges          bool
-	UseFakeKubelets          bool
-	UseFakeNodes             bool
-	UseFakePersistentVolumes bool
-	EnableStorageClasses     bool
-	EnablePriorityClasses    bool
+	SyncAllNodes                  bool
+	SyncNodeChanges               bool
+	UseFakeKubelets               bool
+	UseFakeNodes                  bool
+	UseFakePersistentVolumes      bool
+	EnableStorageClasses          bool
+	EnablePriorityClasses         bool
+	FakeNodesCPUCount             string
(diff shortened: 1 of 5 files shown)


FabianKramm:
@T-M-A thanks for creating this PR! Could you elaborate a little bit why this is needed in vcluster and what your use case for this is?

T-M-A:
Good day!

We wanted to use vcluster for each development team for a specific project, setting resource limits. If I understand correctly, then the set limits on the fake-nodes will work as the cumulative limits set on the namespace

How we usually do:
Team A makes project A, they are allocated one namespace with a limit / request of 2 CPU and 5 GB of memory. At the same time, programmers have to take care of calculating and placing limits / requests for the pods themselves. This is inconvenient for them, especially in a test environment. It is also important for administrators to conserve resources for other teams as well.

How we plan use:
Team A makes project A, they are allocated one namespace without limits / requests. Administrator creates a vcluster with a fake-nodes with a total limit on the CPU and memory. Now, programmers may not specify limits for each PODs.

Sorry for my English :-)

FabianKramm:
@T-M-A thanks for the explanation! Since the scheduler of the host cluster is used for scheduling pods, I don't think setting the nodes resources has any actual effect in a vcluster except for informational purposes. So even though you would "limit" the fake nodes CPU and Memory you would still be able to create pods for an oversubscribed node as the host cluster scheduler is unaware of those limits.

T-M-A:
Good day!

Thank you. Can be closed this PR

Issue #145, opened 2021-10-10
vCluster Pods are in Pending / PVs are not created


anasalloush:
**My Environemnt:**
- 2 Vagrant Boxes (Ubuntu 20.04)
- K8s v1.20.11
- containerd://1.4.11
- Helm v3

I followed the quick guid steps to create a vCluster as in [here ](https://www.vcluster.com/docs/quickstart). but the Pods were hanging in Pending state because there are no PVs created.
As I understood from the Docs, the `--fake-persistent-volumes` option is `true` by default, meaning there is no need to creat PVs PVCs manulay right?

Here are the describe commands for the pods and their PVCs:

`kubectl describe pod/vcluster-1-0 -n host-namespace-1`

> Events:
>   Type     Reason            Age                 From               Message
>    
>   Warning  FailedScheduling  9s (x7 over 5m14s)  default-scheduler  **0/2 nodes are available: 2 pod has unbound immediate PersistentVolumeClaims.**

`kubectl describe pvc data-vcluster-1-0 -n host-namespace-1`

> Name:          data-vcluster-1-0
> Namespace:     host-namespace-1
> StorageClass:  
> Status:        Pending
> Volume:        
> Labels:        app=vcluster
>                release=vcluster-1
> Annotations:   <none>
> Finalizers:    [kubernetes.io/pvc-protection]
> Capacity:      
> Access Modes:  
> VolumeMode:    Filesystem
> Used By:       vcluster-1-0
> Events:
>   Type    Reason         Age                   From                         Message
>                                                                                                            
>   Normal     FailedBinding      14s (x26 over 6m16s)       persistentvolume-controller        **no persistent volumes available for this claim and no storage class is set**

anasalloush:
@FabianKramm  Thanks for the quick response.

Here are some extra information alongside the vcluster container logs.

```
kg nodes -o wide
NAME    STATUS   ROLES                  AGE   VERSION    INTERNAL-IP     EXTERNAL-IP   OS-IMAGE                       KERNEL-VERSION    CONTAINER-RUNTIME
kube1   Ready    control-plane,master   11m   v1.20.11   172.16.16.100   <none>        Debian GNU/Linux 10 (buster)   4.19.0-16-amd64   containerd://1.4.11
kube2   Ready    <none>                 10m   v1.20.11   172.16.16.101   <none>        Debian GNU/Linux 10 (buster)   4.19.0-16-amd64   containerd://1.4.11
```

The `PVC` is bound:

```
[...]

PR #147, opened 2021-10-12
docs(chart): ability to define a storage class

PendaGTP:
Related to: https://github.com/loft-sh/vcluster/issues/145

Looking at the default values in `values.yaml` I did not see that it was possible to modify the storage class while it is quite possible by specifying `className`.
The purpose of this PR is to add visibility on this ability.


FabianKramm:
@PendaGTP thanks for the PR! Yes good idea, looks good to me

PR #172, opened 2021-10-29
fix: service envvar resolution in dependent envvar

### LINKED ISSUE #170: Dependent Environment Variables are not getting Expanded on using vcluster
Hello,

As per Kubernetes documentation, it is possible to define interdependent environment variables

https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/ 

Based on this, we have a helm chart wherein the service port is exposed as a Generic Env using something like below:

`
MY_SERVICE_PORT: $(MY_CHART_MY_SERVICE_PORT)
`

This works fine when I installation via helm directly on the cluster.

However when I use vcluster, the Variable expansion/substitution is not happening and the value of `MY_SERVICE_PORT` is string `$(MY_CHART_MY_SERVICE_PORT)`

## Using vCluster : Not Working MY_SERVICE_PORT is not set properly
`
$ env | grep PROXY_API_SERVICE_PORT
MY_SERVICE_PORT=$(MY_CHART_MY_SERVICE_PORT)
MY_CHART_MY_SERVICE_PORT=8001
`

## Direct Installation on cluster: MY_SERVICE_PORT is set properly
`
$ env | grep PROXY_API_SERVICE_PORT
MY_SERVICE_PORT=8001
MY_CHART_MY_SERVICE_PORT=8001
`

matskiv:
Fixes #170 
In this PR I fix the order of synced environment variables to allow access to them in the dependent environment variables.
Additionally, I implemented an e2e test to cover this scenario, and a small tweak to e2e tests to improve performance (imagePullPolicy: IfNotPresent).

Kudos to @FabianKramm for pointing out where the problem might be, and which part of the code needs to be fixed :+1:

### DIFF
--- a/e2e/framework/helper.go
+++ b/e2e/framework/helper.go
@@ -44,3 +44,8 @@ func ExpectHaveKey(actual interface{}, key interface{}, explain ...interface{}) 
 func ExpectEmpty(actual interface{}, explain ...interface{}) {
 	gomega.ExpectWithOffset(1, actual).To(gomega.BeEmpty(), explain...)
 }
+
+// ExpectMatchRegexp expects the the string to match the provided regular expression
+func ExpectMatchRegexp(actual string, regexp string, explain ...interface{}) {
+	gomega.ExpectWithOffset(1, actual).To(gomega.MatchRegexp(regexp), explain...)
+}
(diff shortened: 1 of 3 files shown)


FabianKramm (approved):

PR #173, opened 2021-10-29
ci: reorganize GH workflows and improve triggers

matskiv:
I am trying to make the checks look a little less cluttered, and have them triggered only when it makes sense (based on the files that got changed in the PR).

State before:
image


FabianKramm (approved):

PR #175, opened 2021-10-30
fix: host rewrite results in incorrect result for `hostname -f`

### LINKED ISSUE #174: Hostname for statefulset pods FQDN is incorrect in vcluster >= 0.4.1
When deploying a statefulset on k8s `hostname -f` executed inside of a pod should return:
`pod-name-<ordinal>.<serviceName>` where serviceName is defined in Statefulset `spec.serviceName`

Actual behavior:
`hostname -f` returns
`pod-name-<ordinal>`

I believe the regression was added [here](https://github.com/loft-sh/vcluster/pull/135/files#diff-2b6b8762ac2665b29231b876d5d4a0263ffa01674f57c7e072a010c8522e5684R321-R328) when trying to solve for Job name behavior outlined in #131 .

moustafab:
The issue was with the ordering of the /etc/hosts file which changed in #135

Closes #174


FabianKramm (approved):
@moustafab thanks a lot for the PR! Great catch thanks a lot!

PR #186, opened 2021-11-13
Use host-namespace-1 for helm install

hasheddan:
The directions currently indicate that vcluster will be installed in
host-namespace-1, but the helm install uses namespace vcluster-1. This
updates the directions to be consistent.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>


LukasGentele:
@hasheddan Good catch. Thanks for contributing!

PR #188, opened 2021-11-15
Feat: run vcluster as non root user

### LINKED ISSUE #67: Cannot deploy vcluster when running as root is not allowed
I'm trying to deploy vcluster in a cluster that does not allow running containers as root, so the pods are not deploying with the current version of vcluster.

I have been checking the helm chart for an option to change this through values, but it looks like there isn't. 

Is running as root an unavoidable requirement for vcluster or could this be fixed? (Running k3s as non-root is possible, although still experimental.)

### LINKED ISSUE #171: vcluster on OpenShift 4
I'd like to get vcluster running on OpenShift and hit the well-known security constraints of OpenShift.

With the following `values.yaml` I got the `vcluster` Pod up and running:

```
volumes:
- name: data
  emptyDir: {}
- name: k3s-config
  emptyDir: {}
- name: kubeconfig
  emptyDir: {}
- name: roothome
  emptyDir: {}
- name: vclusterdata
  emptyDir: {}
vcluster:
  volumeMounts:
  - mountPath: /data
    name: data
  - mountPath: /k3s-config
    name: k3s-config
  - mountPath: /.kube
    name: kubeconfig
syncer:
  volumeMounts:
  - mountPath: /data
    name: data
  - mountPath: /.kube
    name: kubeconfig
  - mountPath: /root
    name: roothome
  - mountPath: /var/lib/vcluster
    name: vclusterdata
```

This solves all the filesystem permission constraints.
It's not perfect (e.g. no persistence), but enough for a PoC to at least get the vcluster Pod running.

The issue I now face is that the CoreDNS Pod which is synced from the vcluster to the host cluster isn't able to start. I get the following message in the syncer container log:

```
E1029 14:47:09.744026       1 controller.go:302] controller-runtime: manager: reconciler group  reconciler kind Pod: controller: pod-forward: name coredns-7448499f4d-rzmd7 namespace kube-system: Reconciler error pods "coredns-7448499f4d-rzmd7-x-kube-system-x-vcluster-1" is forbidden: unable to validate against any security context constraint: [provider "anyuid": Forbidden: not usable by user or serviceaccount, spec.containers[0].securityContext.capabilities.add: Invalid value: "NET_BIND_SERVICE": capability may not be added, provider "nonroot": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount, provider "machine-api-termination-handler": Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden: not usable by user or serviceaccount, provider "hostaccess": Forbidden: not usable by user or serviceaccount, provider "syn-cluster-backup-etcd": Forbidden: not usable by user or serviceaccount, provider "node-exporter": Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount]
```

That means I have to probably reconfigure the CoreDNS Pod which is started in the vcluster context. Any hints how I can this up and running?

### LINKED ISSUE #189: Add support for running vcluster with readOnlyRootFilesystem set to true
We have a PR #188 landing with support for setting the securityContext on the vcluster containers,
but while testing it I noticed that rsecurityContext.readOnlyRootFilesystem can not be set to true with the current implementation.
We are writing a vcluster kube config into container filesystem, and that fails with readOnlyRootFilesystem=true.
The fix might be as simple as adding an emptyDir volume for this purpose, or saving the file into the path where PV is mounted, but it just needs a little bit more thought and testing, which I didn't have time for when implementing #188

matskiv:
**Major changes:**
Instead of letting k3s configure CoreDNS, we install it into vcluster from manifests/coredns/coredns.yaml,
which is built into a container image and is applied on each syncer start.
This approach also allows a user to mount an alternative coredns manifest from a Configmap.

The contents of coredns.yaml are taken from k3s with only minor changes (port 1053 instead of 53, different templating syntax, templating securityContext.
A new controller keeps NodeHosts values of the CoreDNS config map up to date.

~~Vcluster kube config is written into a new location, we can't use /root/ now, but the old location is symlinked to support the old cli versions.~~
Edit: Vcluster kube config is now always written into the default secret, and the related flag creates an additional kube config secret. For backward compatibility with vcluster CLI vcluster attempts to write kube config into a file as it did before, but ignores any errors.

Additional optional Role(Binding) has been added for installation on OpenShift. It's creation is enabled with helm value `openshift.enable`.

For additional security, the data volume is mounted as read-only.
Certificate syncer has been refactored to not write into a file, this allows users to set `securityContext.readOnlyRootFilesystem=true`.

**Other:**
In the builder image, which is also used for development, the /manifests location is symlinked to synced folder.
Dockerfile contains small improvements to dev build speed (GOCACHE).
Tooling for local development has been slightly improved.
Devspace profiles now include dev-crc and deploy-crc for usage with local OpenShift 4 instance (requires CRC and Podman).

**Warning:**
A Vcluster upgrade of an instance that uses root user into one that is not using it has not been tested, and is considered unsupported for now.

Fixes #67 
Fixes #171 
Fixes #189


chart/values.yaml
```diff
@@ -119,3 +122,16 @@ ingress:
     nginx.ingress.kubernetes.io/backend-protocol: HTTPS
     nginx.ingress.kubernetes.io/ssl-passthrough: "true"
     nginx.ingress.kubernetes.io/ssl-redirect: "true"
+
+# Configure SecurityContext of the VCluster pod
+securityContext:
+  allowPrivilegeEscalation: false
+  # readOnlyRootFilesystem: true
```
matskiv, line 129: I meant to add a comment here^ that would link to #189,
readOnlyRootFilesystem: true doesn't work right now.

I'll add that comment if I'll be doing any follow-up fixes or changes.

matskiv:
After a discussion with @FabianKramm I did some changes around kube config handling, etc.
The updated code is now pushed and is ready for review and testing.
The only remaining bit is a new e2e test for the DNS setup.

pkg/server/cert/cert.go
```diff
@@ -3,14 +3,15 @@ package cert
...
-	"k8s.io/apimachinery/pkg/util/sets"
 	"net"
-	"os"
+
+	"github.com/loft-sh/vcluster/pkg/util/certhelper"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-func GenServingCerts(caCertFile, caKeyFile, certFile, keyFile, clusterDomain string, SANs []string) (bool, error) {
+func GenServingCerts(caCertFile, caKeyFile string, currentCert, currentKey *[]byte, clusterDomain string, SANs []string) (bool, error) {
```
FabianKramm, line 14: please return new key and certificate instead of replacing them in-place in the parameters

pkg/server/cert/syncer.go
```diff
@@ -3,27 +3,25 @@ package cert
...
-	"reflect"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sort"
-	"sync"
-	"time"
 )
 
 var (
-	certPath = "/var/lib/vcluster/tls"
+	certPath = "/data/var/lib/vcluster/tls"
```
FabianKramm, line 24: is this still needed?

FabianKramm (changes requested):
@matskiv thanks for the PR! Please see my comments.

chart/values.yaml
```diff
@@ -119,3 +120,23 @@ ingress:
...
+    drop:
+    - all
+  # readOnlyRootFilesystem will be set to true by default at a later release
+  # currently leaving it undefined for backwards compatibility with older vcluster cli versions
+  # readOnlyRootFilesystem: true
+
+  # To run vcluster pod as non-root uncomment runAsUser and runAsNonRoot values.
+  # Update the runAsUser value if your cluster has limitations on user UIDs.
+  # For installation on OpenShift leave the runAsUser undefined (commented out).
+  runAsUser: 12345
```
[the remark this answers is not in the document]
matskiv: Oh yeah, this was changed for testing, I forgot to comment it out again. Thanks for noticing :+1:

pkg/server/cert/syncer.go
matskiv: Nope, this is unused. I'll delete it.

pkg/coredns/coredns.go
```diff
@@ -0,0 +1,118 @@
...
+	ManifestRelativePath  = "coredns/coredns.yaml"
+	ManifestsOutputFolder = "/tmp/manifests-to-apply"
+	VarImage              = "IMAGE"
+	VarRunAsUser          = "RUN_AS_USER"
+	VarRunAsNonRoot       = "RUN_AS_NON_ROOT"
+	VarLogInDebug         = "LOG_IN_DEBUG"
+)
+
+var CoreDNSVersionMap = map[string]string{
+	"1.22": "rancher/mirrored-coredns-coredns:1.8.4",
```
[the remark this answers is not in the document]
matskiv: Yes, I forgot to update this map.
I'll test it on a few previous versions and commit the update.

pkg/util/certhelper/cert.go
```diff
@@ -110,7 +75,7 @@ func NewSignedCert(cfg Config, key crypto.Signer, caCert *x509.Certificate, caKe
 		KeyUsage:     x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
 		ExtKeyUsage:  cfg.Usages,
 	}
-	certDERBytes, err := x509.CreateCertificate(cryptorand.Reader, &certTmpl, caCert, key.Public(), caKey)
+	certDERBytes, err := x509.CreateCertificate(rand.Reader, &certTmpl, caCert, key.Public(), caKey)
```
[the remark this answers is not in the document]
matskiv: It's the same package, just imported second time under "cryptorand" name:
https://github.com/loft-sh/vcluster/blob/c09ff5de283fd313b47606f7d6b38d023839453e/pkg/util/certhelper/cert.go#L21-L22

matskiv:
I addressed the comments + added two e2e tests and made them run testing pods as non-root.

I also added a commit with a fallback code for the connect command. Please check the last commit in this PR and let me know if we should keep this, or whether it's better to keep the code clean despite breaking backward compatibility (if vcluster is using a different value of the `--out-kube-config-secret` flag).

CONTRIBUTING.md
```diff
@@ -110,9 +110,8 @@ If you are not using VSCode, configure your IDE to connect to `localhost:2346` f
 
 You can run the unit test suite with `./hack/test.sh` which should execute all the vcluster tests.  
 
-The e2e test suite can be run from the e2e folder(`cd e2e`) with this command - `VCLUSTER_SUFFIX=vcluster go test -v -ginkgo.v`.  
-Alternatively, if you [install ginkgo binary](https://github.com/onsi/ginkgo#global-installation), you can run it with `VCLUSTER_SUFFIX=vcluster ginkgo -v`.  
-If you are running vcluster with `devspace run dev` then the `VCLUSTER_SUFFIX` environment variable shouldn't be set, and you should run the tests with `go test -v -ginkgo.v` or `ginkgo -v`.
```
matskiv, line 115: FYI: I remove these mentions of the `VCLUSTER_SUFFIX` env var because the syncer and e2e test suite will now default to `vcluster` value for the suffix, which is the same as we've been using in the helm chart, so setting the env var is no longer necessary for running e2e tests.

FabianKramm (approved):
@matskiv thanks a lot! LGTM!

PR #192, opened 2021-11-23
fix: glusterfs volume endpointsName translation on sync

matskiv:
Fixes the issue reported in slack - https://loft-sh.slack.com/archives/C01N273CF4P/p1633101568199400

GlusterFS volumes use an Endpoint that is local to the pod(created in the same namespace as the pod).
The value of the `glusterfs.endpoints` field is now appropriately translated by the syncer:

image
```
$ k get endpoints -n vcluster
NAME                                ENDPOINTS                                     AGE
glusterfs-cluster-x-test-x-suffix   192.168.122.21:1,192.168.122.22:1             4m51s
```

### DIFF
--- a/pkg/controllers/resources/pods/translate/translator.go
+++ b/pkg/controllers/resources/pods/translate/translator.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


FabianKramm:

Issue #52, opened 2021-06-18
Support Pod affinity label translation


nktpro:
Currently, label names within `affinity` of Pod specs are not translated at all, which is a deal-breaker for a lot of real-world use cases. Given that Pod Topology Constraints support was recently landed, is this part of the near-term plan for the next release?

FabianKramm:
@nktpro thanks for creating this issue! You are correct, currently we do not rewrite pod affinity rules. We can add this, however currently we won't be able to support the `namespaceSelector` field, as there is no virtual namespace information available in the host cluster for now (we will change this though and sync all namespace labels to the pods that were created in the virtual namespace to select them in the host cluster to support `namespaceSelector` in pod affinity and networkpolicies)

FabianKramm:
@nktpro should work now (except `namespaceSelector`) with v0.3.0-beta.4

PR #202, opened 2021-11-25
feat: partial support for namespaceSelector in inter-pod affinity

matskiv:
Namespace selectors of the inter-pod affinity rules will now be translated into label selectors and should work for the most part.
This addresses #52, except for the edge case as documented in the code, and which we will address later.


FabianKramm (approved):

PR #208, opened 2021-11-28
test(e2e): run e2e in GH workflows for all charts

matskiv:
We should run e2e tests against all distributions/charts that are provided - k3s, k8s and k0s.
These changes will ensure that e2e tests in GH workflows(on PRs) are triggered for each distribution/chart.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -17,6 +17,10 @@ jobs:
   e2e:
     name: Execute testsuite
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        distribution: ["k3s", "k8s", "k0s"]
     steps:
     - name: Checkout repo
       uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -39,14 +43,11 @@ jobs:
       uses: loft-sh/setup-devspace@main
 
     - name: Deploy vcluster
-      run: devspace run deploy --skip-push
-
-    - name: Load built image into kind cluster
-      run: docker images | grep ghcr.io/loft-sh/loft-enterprise/dev-vcluster | head -n 1 | awk '{print $1":"$2}' | xargs kind load docker-image
+      run: devspace run deploy -p ${{ matrix.distribution}} -p kind-load --var K8S_REPLICAS=1 --skip-push
```
matskiv, line 46: I changed the replicas to one because I was hitting resource issues in the runner:
`FailedScheduling  ...  0/1 nodes are available: 1 Insufficient cpu.`
but I don't think this is ideal. Is it possible to increase CPUs for the runner?
If not, maybe creating a devspace profile with no CPU requests would be a better solution?

FabianKramm: Yeah I guess creating a different profile makes sense, testing the HA setup is definitely a good idea in the pipeline if that works.

Issue #18, opened 2021-05-18
IPv6-only clusters not supported


Frankkkkk:
Hi,
I wanted to give vcluster a try, however it seems that IPv6-only kubernetes clusters are not yet fully supported;

```
kl -n host-namespace-1 logs -f vcluster-1-0 vcluster
time="2021-05-18T09:56:40.133915808Z" level=info msg="Starting k3s v1.18.16+k3s1 (8c7dd139)"
time="2021-05-18T09:56:40.161845141Z" level=info msg="Configuring sqlite3 database connection pooling: maxIdleConns=2, maxOpenConns=0, connMaxLifetime=0s"
time="2021-05-18T09:56:40.162067426Z" level=info msg="Configuring database table schema and indexes, this may take a moment..."
time="2021-05-18T09:56:40.193732078Z" level=info msg="Database tables and indexes are up to date"
time="2021-05-18T09:56:40.200156204Z" level=info msg="Kine listening on unix://kine.sock"
time="2021-05-18T09:56:40.620849868Z" level=info msg="Active TLS secret  (ver=) (count 7): map[listener.cattle.io/cn-127.0.0.1:127.0.0.1 listener.cattle.io/cn-2001:1600:caca:1002:1d:12e:ffff:4b6d:2001:1600:caca:1002:1d:12e:ffff:4b6d listener.cattle.io/cn-2001:1600:caca:50da:1d:fffe:0:1:2001:1600:caca:50da:1d:fffe:0:1 listener.cattle.io/cn-kubernetes:kubernetes listener.cattle.io/cn-kubernetes.default:kubernetes.default listener.cattle.io/cn-kubernetes.default.svc.cluster.local:kubernetes.default.svc.cluster.local listener.cattle.io/cn-localhost:localhost listener.cattle.io/hash:58e35fc7fbc214c1d0e3c5f72f84a49a8ee8f2ec2eb4cfe34a546604d09d32f4]"
[...]

PR #209, opened 2021-11-28
Support IPv6 on service CIDR

olljanat:
First step to solve #18 

Add logic to detect if cluster does not have support for IPv4 or if vcluster is forced to use IPv6 only with policy engine.

At least combination of dual stack cluster with Kyverno policy like this:
```yaml
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: service-ipv6-only
spec:
  validationFailureAction: enforce
  background: false
  rules:
  - name: service-ipv6-only
    match:
      resources:
        kinds:
        - Service
        namespaces:
        - "vcluster-*"
    preconditions:
      all:
      - key: "{{request.operation}}"
        operator: In
        value:
        - CREATE
    mutate:
      patchStrategicMerge:
        spec:
          ipFamilyPolicy: SingleStack
          ipFamilies:
          - IPv6
```
and using `--distro k8s` works.

Pure IPv6 only env where also host cluster is IPv6 only will need more work. K0s needs this PR https://github.com/k0sproject/k0s/pull/1292 and K3s needs this PR https://github.com/k3s-io/k3s/pull/4450 before those can be used with this one.

### DIFF
--- a/cmd/vclusterctl/cmd/app/create/values/k3s.go
+++ b/cmd/vclusterctl/cmd/app/create/values/k3s.go
@@ -150,7 +150,12 @@ func getKubernetesVersion(client kubernetes.Interface, createOptions *create.Cre
 	return serverVersionString, serverMinorInt, nil
 }
 
-func GetServiceCIDR(client kubernetes.Interface, namespace string) (string, error) {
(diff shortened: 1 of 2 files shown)


FabianKramm (approved):
@olljanat thanks for creating this PR! LGTM!

Issue #167, opened 2021-10-27
k0s support beside k3s?


ThomasLohmann:
Hi everyone,

please can you also support k0s beside k3s? k0s has much more use cases (bare-metal, cloud, iot, edge etc.) compared to k3s (iot, edge). In addition, k0s is much less opinionated regarding networking, storage, ingress etc. + the size is also small (187 MB). Finally, k0s is used for conventional staging & production clusters (bare-metal or cloud) which means that dev vClusters with k0s will be much closer to staging & production. So it would be great, if you can support it. Please, see the following link: https://k0sproject.io/

Best regards,
Thomas

danmx:
I'm interested in this too. Any suggestions how the community could start contributing?

FabianKramm:
@danmx theoretically its already possible to use vcluster with other kubernetes distributions, as we do not have any hardcoded k3s dependencies anymore, however the vcluster chart is currently still hardwired to k3s, so you would need to do the wiring between vcluster syncer and k0s yourself.

We are currently working on making vcluster usable with vanilla k8s to allow easier HA, which would be then very similar to using k3s with k0s.

FabianKramm:
@ThomasLohmann @danmx we actually added a guide how to use k0s to our docs and released a new version (v0.5.0-alpha.0) that supports that

olljanat:
@FabianKramm thanks for adding support for k0s and k8s.

I did quick test on v0.5.0-alpha.3 by deploying k8s with default settings looks which to be working fine but if I enable switches `--fake-persistent-volumes=false` and `--enable-storage-classes` it fails to error:
```
I1126 13:49:32.854841       1 start.go:128] couldn't find virtual cluster kube-config, will retry in 1 seconds
```

FabianKramm:
@olljanat good catch, the problem is you overwrite the default `extraArgs` there, it should look like this in your `values.yaml`:

```yaml
syncer:
  extraArgs:
    - --request-header-ca-cert=/pki/ca.crt
    - --client-ca-cert=/pki/ca.crt
    - --server-ca-cert=/pki/ca.crt
    - --server-ca-key=/pki/ca.key
    - --kube-config=/pki/admin.conf
    - --fake-persistent-volumes=false
    - --enable-storage-classes
```

We'll fix this for the next version though so you can use the extraArgs as expected before

olljanat:
[...]

PR #216, opened 2021-12-01
fix: move k0s and k8s syncer extraArgs to allow user defined ones

matskiv:
This fixes a problem mentioned in [this comment](https://github.com/loft-sh/vcluster/issues/167#issuecomment-980006013).
Users are now free to define their own extraArgs for syncer, and the ones needed by default won't get overridden.


FabianKramm (approved):

PR #218, opened 2021-12-01
fix(cli): connect command should poll kube config Secret

matskiv:
The fallback to the legacy kube config location should not happen after just a single attempt to read the Secret.
Instead, we will try both approaches in a loop until success or timeout.


FabianKramm (approved):

PR #229, opened 2021-12-05
etcd: use pod IP instead of localhost on startup and liveness probes

olljanat:
I'm not exactly sure why this is the case but looks that #212 did broke on etcd startup totally on my env where I use Calico + BGP without any overlay or NAT.

This was the error:
```
kubectl -n vcluster-test describe pod/test-etcd-0
  Normal   Started                 72s                kubelet                  Started container etcd
  Warning  Unhealthy               9s (x6 over 59s)   kubelet                  Startup probe failed: Get "http://127.0.0.1:2381/health": dial tcp 127.0.0.1:2381: connect: connection refused
```
and if I only updated startup probe then error was:
```
kubectl -n vcluster-test describe pod/test-etcd-0
  Warning  Unhealthy               8s    kubelet                  Liveness probe failed: Get "http://127.0.0.1:2381/health": dial tcp 127.0.0.1:2381: connect: connection refused
```

Solution looks to be not defining `host:` value so pod IP gets used instead of https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#http-probes

### DIFF
--- a/charts/k8s/templates/etcd-statefulset.yaml
+++ b/charts/k8s/templates/etcd-statefulset.yaml
@@ -19,6 +19,7 @@ metadata:
 spec:
   serviceName: {{ .Release.Name }}-etcd-headless
   replicas: {{ .Values.etcd.replicas }}
+  podManagementPolicy: Parallel
   selector:
     matchLabels:
       app: vcluster-etcd
@@ -127,7 +128,6 @@ spec:
           httpGet:
             path: /health
             port: 2381
-            host: 127.0.0.1
             scheme: HTTP
           initialDelaySeconds: 10
           timeoutSeconds: 15
@@ -138,7 +138,6 @@ spec:
           httpGet:
             path: /health
             port: 2381
-            host: 127.0.0.1
             scheme: HTTP
           initialDelaySeconds: 10
           timeoutSeconds: 15


olljanat:
@matskiv FYI as #212 was your PR.

matskiv:
@olljanat thanks for letting us know about the problem and even providing a fix, it's much appreciated.
I adopted the same probe definitions as [are used in kubeadm](https://github.com/kubernetes/kubernetes/blob/0153febd9f0098d4b8d0d484927710eaf899ef40/cmd/kubeadm/app/phases/etcd/local.go#L206-L207), including [the same host](https://github.com/kubernetes/kubernetes/blob/0153febd9f0098d4b8d0d484927710eaf899ef40/cmd/kubeadm/app/util/staticpod/utils.go#L324-L328) - `127.0.0.1`, as that seemed like the safest option, and if at all possible, I would like to stick to it.

I tried reproducing your problem with minikube + Calico, which I was hoping would be close enough to your setup, but that still works fine with the `127.0.0.1` probe host. Reproducing your exact setup might be too difficult for me, so I hope you will be able to assist in exploring potential fixes later.
I then tested this PR on my minikube, but with no host being set the etcd-0 pod never gets into Ready state and the probe always responds with `{"health":"false"}` (503 HTTP status code).
All in all, I think this needs additional investigation, as the Readiness status of the pod seems to be inconsistent. Some logs I collected even suggest that the pod is Ready before the probe succeeds. I will get back to this soon.

In the meantime, would you be able to provide some additional information?
a) Kubernetes version
b) how are the probes defined on your host cluster etcd (if you use etcd of course, and have access to this information), and is it something you customize or is that just a default of the tool that you use to deploy k8s?

olljanat:
> In the meantime, would you be able to provide some additional information?
> a) Kubernetes version

I'm building multi-tenant K8s environment which need to be able to apply strict security requirements which why I'm running [RKE2](https://docs.rke2.io/) with CIS hardenings enabled. Current version is v1.22.3+rke2r1. Top of that there is dedicated workers for each vcluster and Kyverno is used to force securityContext to all *vcluster-** namespaces.

> b) how are the probes defined on your host cluster etcd (if you use etcd of course, and have access to this information), and is it something you customize or is that just a default of the tool that you use to deploy k8s?

RKE2 does run etcd as container and uses `host: 127.0.0.1` on probes **but** it is using host networking (host own IP is listed on `kubectl -n kube-system get pods -o wide`) which why it is totally different case than these vclusters.

One question which comes to my mind is that have you tested this on multi node cluster on way that vcluster will be scheduled to node which does not run host cluster's etcd? Other why your probe might end up to actually checking host clusters etcd port 2381 instead of vclusters one.


One interesting note is that I also tested to include another alpine container to same pod with etcd and from there at least curl can connect to to health endpoint.
```bash
# wget https://127.0.0.1:2381
Connecting to 127.0.0.1:2381 (127.0.0.1:2381)
ssl_client: 127.0.0.1: handshake failed: error:1408F10B:SSL routines:ssl3_get_record:wrong version number
wget: error getting response: Connection reset by peer
```
However now I wonder that how those probes actually works? Based on https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/ it is kubelet which run those probes but as it run outside of pod and `127.0.0.1` address on host and inside of pods are separate it shouldn't be possible to its connect to right service unless it used pods IP?

> I tried reproducing your problem with minikube + Calico, which I was hoping would be close enough to your setup, but that still works fine with the `127.0.0.1` probe host. Reproducing your exact setup might be too difficult for me, so I hope you will be able to assist in exploring potential fixes later. I then tested this PR on my minikube, but with no host being set the etcd-0 pod never gets into Ready state and the probe always responds with `{"health":"false"}` (503 HTTP status code). All in all, I think this needs additional investigation, as the Readiness status of the pod seems to be inconsistent. Some logs I collected even suggest that the pod is Ready before the probe succeeds. I will get back to this soon.

Minikube is one node setup so are you sure that kubelet is not just connecting host cluster's health port and assume that pod is OK?

olljanat:
> @olljanat thanks for letting us know about the problem and even providing a fix, it's much appreciated. I adopted the same probe definitions as [are used in kubeadm](https://github.com/kubernetes/kubernetes/blob/0153febd9f0098d4b8d0d484927710eaf899ef40/cmd/kubeadm/app/phases/etcd/local.go#L206-L207), including [the same host](https://github.com/kubernetes/kubernetes/blob/0153febd9f0098d4b8d0d484927710eaf899ef40/cmd/kubeadm/app/util/staticpod/utils.go#L324-L328) - `127.0.0.1`, as that seemed like the safest option, and if at all possible, I would like to stick to it.

Btw. Reason why host cluster's etcd probe is using `127.0.0.1` is that by default metrics endpoint is set to listen only that with parameter `--listen-metrics-urls=http://127.0.0.1:2381` which why it is not available on other IPs. That is not case here as vcluster's etcd is set to listed on all IPs https://github.com/loft-sh/vcluster/blob/12501506245ed157dd1e3719b2f8784096db9357/charts/k8s/templates/etcd-statefulset.yaml#L91

matskiv:
@olljanat I would hope that kubelet executes probes inside of the network namespace of a given pod, but you are probably right, host cluster etcd might be at play here. I'll have to read up some technical details about probe implementation one day... :)

> Btw. Reason why host cluster's etcd probe is using 127.0.0.1 is that by default metrics endpoint is set to listen only that with parameter --listen-metrics-urls=http://127.0.0.1:2381 which why it is not available on other IPs. That is not case here as vcluster's etcd is set to listed on all IPs 

Yeah, good point. Using pod IP (no host field) should be fine then, but I couldn't understand why it works for you, but not for me.
And I think it finally clicked :) 
You are using the default 1 etcd replica, right?
https://github.com/loft-sh/vcluster/blob/86fb09cd3904cf694158cef739a6a24fc77f4c9d/charts/k8s/values.yaml#L52-L55
But I am deploying with devspace which has override to three replicas:
https://github.com/loft-sh/vcluster/blob/86fb09cd3904cf694158cef739a6a24fc77f4c9d/devspace.yaml#L26-L27
In my case StatefulSet only creates `etcd-0`, it never passes the probe because it can't connect to peers (`etcd-1` and `etcd-2`), and it can't connect to them because the next peer will be created by the StatefulSet only after the previous one passes the probe (becomes ready).
All we need is to add `.spec.podManagementPolicy: Paralllel` on the etcd StatefulSet. Then all replicas are started at once and they can successfully establish a cluster and start reporting healthy status on the probe endpoint. If you could add this change to your PR that would be awesome, as it would make your fix work for multiple etcd replicas too. And once you push a change GH actions should trigger e2e tests, so we will have additional confirmation of it working.

olljanat:
> Yeah, good point. Using pod IP (no host field) should be fine then, but I couldn't understand why it works for you, but not for me. And I think it finally clicked :) You are using the default 1 etcd replica, right?

Yes, I use simplified setup with 1 replicas only of each component for now on purpose until I get other parts of solution working like I want to.

> All we need is to add `.spec.podManagementPolicy: Paralllel` on the etcd StatefulSet. Then all replicas are started at once and they can successfully establish a cluster and start reporting healthy status on the probe endpoint. If you could add this change to your PR that would be awesome, as it would make your fix work for multiple etcd replicas too. And once you push a change GH actions should trigger e2e tests, so we will have additional confirmation of it working.

Added (hopefully to correct place) now.

matskiv:
@olljanat yes, it's added in the right place, thank you!
CI thingy is approved now :)

FabianKramm (approved):
@olljanat thanks for the PR! LGTM!

PR #230, opened 2021-12-06
fix(ci): changes to charts should trigger e2e tests

matskiv:
I made a mistake last time I was trying to enable e2e on charts changes.
This time I am using a correct pattern, as per docs:
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#patterns-to-match-file-paths


FabianKramm (approved):

PR #232, opened 2021-12-06
feat: add NetworkPolicy support

### LINKED ISSUE #155: Network Policies don't work
I'm playing around with the vcluster and recognized that network polices don't work. Is this an known bug/not working feature?

I created (only) a simple deployment and a super stupid policy. Executing an `apt update` inside the pod is stil working. Where is my error?
```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: demo
  labels:
    app: demo
spec:
  replicas: 1
  selector:
    matchLabels:
      app: demo
  template:
    metadata:
      labels:
        app: demo
    spec:
      containers:
        - name: container
          image: ubuntu:20.10
          args:
            - sleep
            - "infinity"
---
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: demo-deny
spec:
  podSelector:
    matchLabels:
      app: demo
  policyTypes:
    - Ingress
    - Egress
```

matskiv:
This should enable NetworPolicy synchronization with all necessary rewrites for the pod label selectors and namespace label selectors.

Fixes #155

### DIFF
--- a/charts/k0s/values.yaml
+++ b/charts/k0s/values.yaml
@@ -85,6 +85,8 @@ rbac:
   role:
     # This is required for basic functionality of vcluster
     create: true
+    # Extended role permissions are required for some optional features, e.g. Networkpolicy sync
+    extended: false
 
 # The amount of replicas to run the statefulset with
 replicas: 1
--- a/charts/k3s/values.yaml
+++ b/charts/k3s/values.yaml
@@ -93,6 +93,8 @@ rbac:
   role:
     # This is required for basic functionality of vcluster
     create: true
+    # Extended role permissions are required for some optional features, e.g. Networkpolicy sync
+    extended: false
 
 # The amount of replicas to run the statefulset with
 replicas: 1
--- a/charts/k8s/values.yaml
+++ b/charts/k8s/values.yaml
@@ -132,6 +132,8 @@ rbac:
   role:
     # This is required for basic functionality of vcluster
     create: true
+    # Extended role permissions are required for some optional features, e.g. Networkpolicy sync
+    extended: false
 
 # Service configurations
 service:
--- a/devspace.yaml
+++ b/devspace.yaml
@@ -40,6 +40,8 @@ deployments:
         rbac:
           clusterRole:
             create: true
+          role:
+            extended: true
         syncer:
           readinessProbe:
             enabled: false
@@ -85,8 +87,10 @@ profiles:
     patches:
       - op: remove
         path: deployments.name=vcluster.helm.values.syncer.command
-      - op: remove
-        path: deployments.name=vcluster.helm.values.syncer.extraArgs
+      - op: replace
+        path: deployments.name=vcluster.helm.values.syncer
+        value:
+          image: ${SYNCER_IMAGE}
   - name: deploy
     parents: 
       - profile: parent-deploy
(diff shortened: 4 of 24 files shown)


matskiv:
I properly updated the code based on the recent changes in main + made this feature opt-in to allow easy installation even if the user doesn't have the permissions for network policies.
The new test will still fail, as there are at least two missing pieces here:
a) KinD cluster used for e2e tests probably needs to be configured with a different CNI plugin, one that supports network policies
b) New controller needs to be enabled in the deployments used for e2e tests. This might just be another devspace profile :thinking:

olljanat:
> a) KinD cluster used for e2e tests probably needs to be configured with a different CNI plugin, one that supports network policies

I see that you ended up disabling Calico now so recommend that you considering using K3s on network policy related e2e tests instead of KinD (and you can spin those with K3d) as it does support network policies without installing separate CNI.

charts/k0s/values.yaml
```diff
@@ -85,6 +85,9 @@ rbac:
   role:
     # This is required for basic functionality of vcluster
     create: true
+  extendedRolePermissions:
```
FabianKramm, line 88: Please rename this to `rbac.role.extended`

charts/k3s/values.yaml
```diff
@@ -93,6 +93,9 @@ rbac:
   role:
     # This is required for basic functionality of vcluster
     create: true
+  extendedRolePermissions:
```
FabianKramm, line 96: See above

charts/k8s/values.yaml
```diff
@@ -132,6 +132,9 @@ rbac:
   role:
     # This is required for basic functionality of vcluster
     create: true
+  extendedRolePermissions:
```
FabianKramm, line 135: See above

devspace.yaml
```diff
@@ -40,6 +40,8 @@ deployments:
         rbac:
           clusterRole:
             create: true
+          extendedRolePermissions:
```
FabianKramm, line 43: see above

pkg/controllers/resources/networkpolicies/translate.go
```diff
@@ -0,0 +1,128 @@
...
+	newNetworkPolicy, ok := newObj.(*networkingv1.NetworkPolicy)
+	if !ok {
+		return nil, fmt.Errorf("error converting to NetworkPolicy object named: %s", vObj.GetName())
+	}
+
+	newNetworkPolicy.Spec = *translateSpec(&vNetworkPolicy.Spec, vNetworkPolicy.GetNamespace())
+	return newNetworkPolicy, nil
+}
+
+func (s *syncer) translateUpdate(pObj, vObj *networkingv1.NetworkPolicy) (*networkingv1.NetworkPolicy, error) {
```
FabianKramm, line 31: no error is ever returned here, so please omit it

FabianKramm (changes requested):
@matskiv thanks for the PR! Please take a look at my comments, besides that it looks good!

matskiv:
@olljanat Thank you for the suggestion, I'll give it a shot :)

matskiv:
I addressed review comments.

As for the failing tests, after some investigation, it doesn't seem to be related to network policy changes. The failures happen due to a number of pod/exec calls which sometimes trigger yet unknown issue on the server that causes it to be temporarily unresponsive to new exec calls. I was not able to tell which component is causing this because multiple pods are marked as not ready when this happens, so it could be a problem on a lower network level... Anyway, since this is not caused by the network policy controller I will create a follow-up issue, and the tests will be skipped in CI for now.

matskiv:
@olljanat yes, that's the plan. Once this is merged a new beta release will be done.

FabianKramm (approved):
@matskiv LGTM! Thanks!

PR #234, opened 2021-12-07
refactor: omit * in --sync flag and enable default controllers

FabianKramm:
### Changes
- **syncer**: Default controllers are now enabled by default, there is no need to specify `*` in the `--sync` flag anymore


matskiv:
Change looks good, I think having it work this way will be more intuitive.
I tested locally - works as expected :+1:

PR #238, opened 2021-12-09
e2e: should have Endpoints and EndpointSlices pointing to API Server test

pratikjagrut:
e2e: should have Endpoints and EndpointSlices pointing to API Server test


matskiv:
Out of date.

PR #248, opened 2021-12-10
fix ingress port

1337andre:
run into this error
error validating data: ValidationError(Ingress.spec.rules[0].http.paths[0].backend.service.port): invalid type for io.k8s.api.networking.v1.ServiceBackendPort: got "integer", expected "map";


FabianKramm:
@1337andre thanks a lot for the PR! LGTM!

FabianKramm (approved):

PR #249, opened 2021-12-11
Change empty volume indention to match certs volume

kuzaxak:
To render the YAML array correctly with the persistent storage option disabled each value of the volumes 
block should be on the same indention level. 

Before fix It renders like:
```yaml
      volumes:
        - name: certs
          secret:
            secretName: t-manual-certs
      - name: data
        emptyDir: {}
```

What produces an issue in helm: 
`Error: YAML parse error on vcluster-k8s/templates/etcd-statefulset.yaml: error converting YAML to JSON: yaml: line 36: did not find expected key`


FabianKramm (approved):
@kuzaxak thanks a lot for the PR! Great find, LGTM!

PR #255, opened 2021-12-16
docs: describe network policy support

matskiv:
Let me know if I should go into more detail on how this works under the hood, or whether I should explicitly mention that cross namespaces policies are also supported.


FabianKramm (approved):

PR #256, opened 2021-12-16
Improve test stability

matskiv:
Two tests have been failing intermittently due to race conditions(IMHO). This PR should fix them.
I run the tests 50x in a loop and there was no failure, hopefully, behavior in CI is the same :).

I also updated debugging configuration to use a new flag(easier to understand) and a new container path(fixes breakpoints).


FabianKramm (approved):

PR #265, opened 2021-12-24
feat: configurable .spec.dnsPolicy of CoreDNS pod

matskiv:
In [a discussion on the Slack channel](https://loft-sh.slack.com/archives/C01N273CF4P/p1640339530234300) it became apparent that `dnsPolicy: Default` doesn't work everywhere, and we needed to set `dnsPolicy: ClusterFirst` on the vcluster CoreDNS pod to make vcluster work in a particular environment.
I was a bit on the fence with the idea of adding a new flag for such an edge case, but then I thought that adding a more general flag (`--coredns-var`) could be an acceptable fix, as it might have other uses for CoreDNS customizations. We already do templating of the CoreDNS manifest, so the new flag simply allows for the template variable overrides.

The other half of the fix is within the pod translator. When the `dnsPolicy: ClusterFirst` is set on pods within the vcluster, the translator will adjust the DNS config of the pod to point to the vcluster CoreDNS instance. This is obviously not desirable for the CoreDNS pod itself, so I added code to skip the DNS config translation for CoreDNS pods. But I am definitely open to ideas, if there is a more elegant solution for this.

### DIFF
--- a/cmd/vcluster/cmd/start.go
+++ b/cmd/vcluster/cmd/start.go
@@ -108,6 +108,8 @@ func NewStartCommand() *cobra.Command {
 
 	cmd.Flags().StringVar(&options.ClusterDomain, "cluster-domain", "cluster.local", "The cluster domain ending that should be used for the virtual cluster")
 
+	cmd.Flags().StringArrayVar(&options.CorednsVariables, "coredns-var", []string{}, "Provides custom value for the template variable used in the CoreDNS manifest(e.g. DNS_POLICY=Default). The manifest file is located in /manifests/coredns/coredns.yaml file in the syncer image, or at the same path in the source code repo.")
+
 	cmd.Flags().BoolVar(&options.LeaderElect, "leader-elect", false, "If enabled, syncer will use leader election")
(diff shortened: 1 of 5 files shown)


FabianKramm:
@matskiv thanks for creating this PR! I'm not sure if this is a good idea as the problem seems to be that `dnsPolicy: Default` might be problematic in certain environments and `dnsPolicy: ClusterFirst` might be preferable in general for core dns, so we could just switch the option and prevent vcluster from rewriting the coredns pods to point to itself.

matskiv:
@FabianKramm yes, that is certainly an option, and I am happy to update the PR if you want to go with that.
btw: I am not 100% sure what are all the cons and pros of using `dnsPolicy: ClusterFirst` for vcluster's CoreDNS. Some extra load on the hosts CoreDNS? :thinking:

GiGurra:
Thanks @matskiv and @FabianKramm for looking into this. 
I reported this issue on slack a bit ago and good to see this is moving forward.

Once this is sorted, in someway (hopefully ;) ), it will fix the current issue that vcluster does not execute for example on https://okteto.com/ (they apparently have some odd dns setup)

PR #272, opened 2022-01-05
refactor: make default helm values reusable

lizardruss:
### Changes
- refactor helm chart defaults into a reusable package
- add log interface to allow consumers to provide their own logging implementation


FabianKramm (approved):
@lizardruss thanks for the PR! LGTM!

PR #279, opened 2022-01-07
Add missing colon in OpenShift values example

tobru:
The documentation misses a colon in the example values.


FabianKramm (approved):
@tobru thanks for the PR! Good catch, LGTM!

PR #282, opened 2022-01-10
e2e: add 'should be able to deny pod and configmap creation' test

pratikjagrut:
Add e2e test


FabianKramm (approved):

Issue #190, opened 2021-11-17
[CLI] 'vcluster create' fails when the working dir has a file called 'vcluster'


enisoc:
I got a strange error when I tried to run `vcluster create`:

```console
$ vcluster create test -n test
[info]   execute command: helm upgrade test vcluster --repo https://charts.loft.sh --version 0.4.4 --kubeconfig /tmp/284868977 --namespace test --install --repository-config='' --values /tmp/863399452
[fatal]  error executing helm upgrade test vcluster --repo https://charts.loft.sh --version 0.4.4 --kubeconfig /tmp/284868977 --namespace test --install --repository-config='' --values /tmp/863399452: Release "test" does not exist. Installing it now.
Error: file '/home/enisoc/bin/vcluster' does not appear to be a gzipped archive; got 'application/octet-stream'
```

It turned out this is because I was running `vcluster` from the working directory that contained the `vcluster` binary itself. It seems that Helm prefers the local file if it exists, instead of treating `vcluster` as referring to a chart within the `charts.loft.sh` repo. Helm then tries to load the `vcluster` file as if it's a chart archive.

Here is a minimal reproduction case:

```sh
cd /tmp
touch vcluster
vcluster create test -n test
```

A directory called `vcluster` causes a similar but slightly different error:

```console
/tmp$ rm vcluster
/tmp$ mkdir vcluster
/tmp$ vcluster create test -n test
[info]   execute command: helm upgrade test vcluster --repo https://charts.loft.sh --version 0.4.4 --kubeconfig /tmp/732744325 --namespace test --install --repository-config='' --values /tmp/002659872
[fatal]  error executing helm upgrade test vcluster --repo https://charts.loft.sh --version 0.4.4 --kubeconfig /tmp/732744325 --namespace test --install --repository-config='' --values /tmp/002659872: Release "test" does not exist. Installing it now.
Error: Chart.yaml file is missing
```

LukasGentele:
Thanks for reporting this. Your assumptions about the reason for this issue are correct: https://github.com/helm/helm/issues/7862

I wish more folks would link to this issue, so Helm maintainers may understand that this is not what users expect at all.

enisoc:
Perhaps as a workaround, vcluster could execute helm with its working dir set to a temp dir whose contents are known?

FabianKramm:
@enisoc yes thats true, we could do that

LukasGentele:
[...]

PR #285, opened 2022-01-10
cli: notify user about incompatible workdir file

matskiv:
Related to #190 

### Notable changes:

- cli: notify the user about incompatible workdir file when running `vcluster create`

### DIFF
--- a/cmd/vclusterctl/cmd/create.go
+++ b/cmd/vclusterctl/cmd/create.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/loft-sh/vcluster/cmd/vclusterctl/cmd/app/create"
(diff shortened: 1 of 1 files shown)


cmd/vclusterctl/cmd/create.go
```diff
@@ -223,6 +224,14 @@ func (cmd *CreateCmd) Run(args []string) error {
 		}
 	}
 
+	workDir, err := os.Getwd()
+	if err != nil {
+		log.GetInstance().Fatal(fmt.Errorf("unable to get current work directory: %v", err))
```
FabianKramm, line 229: Please return an error here instead of killing the program directly

cmd/vclusterctl/cmd/create.go
```diff
@@ -223,6 +224,14 @@ func (cmd *CreateCmd) Run(args []string) error {
...
+	}
+	if _, err := os.Stat(filepath.Join(workDir, cmd.ChartName)); err == nil {
+		log.GetInstance().Fatalf("Aborting vcluster creation. Current working directory contains a file or a directory with the name equal to the vcluster chart name - \"%s\". Please execute vcluster create command from a directory that doesn't contain a file or directory named \"%s\".", cmd.ChartName, cmd.ChartName)
```
FabianKramm, line 232: See above

FabianKramm (changes requested):
@matskiv thanks for the PR! Please see my comments would be better to return an error instead of exiting immediately.

matskiv:
Yeah, good point. I updated the code to return the errors.

FabianKramm (approved):
@matskiv thanks! LGTM!

PR #287, opened 2022-01-11
Allow injection to Corefile via coredns-custom CM

chrisjohnson:
Same idea as https://github.com/k3s-io/k3s/pull/4397/files

If you create CM `coredns-custom` that looks like this:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: coredns-custom
  namespace: kube-system
data:
  # Must end in .server
  example.server: |
    example.org {
      log
      whoami
    }
```

then coredns will allow for Corefile injections, so we can add things like internal hostname resolvers


chrisjohnson:
Looks like the e2e tests pass, and we were successful in building a custom image and using dig to poke at it both before and after adding a coredns-custom CM and see the expected behavior.

One thing to note, due to the way the optional mount is specified, you have to kill the pod to reflect creation of the CM. We are probably going to find some spots to add a bit more documentation around this space but otherwise I think this PR is workable

FabianKramm (approved):
@chrisjohnson thanks a lot for the PR! Sounds good, LGTM!

PR #291, opened 2022-01-11
fix: cache race condition after Service creation

matskiv:
We are no longer sharing a cache between the server and controllers.
This unfortunately allowed for a race condition after Service creation.
The controller might be missing Service resource in vcluster cache,
which is causing the deletion of the Service in the host cluster
until it is recreated again by the controller based on Service resource
that gets eventually pulled into vcluster cache.
To prevent this race condition we leave the annotation
that prevents Service deletion in the host until the Service resource
in vcluster is in the expected state.

I also updated the unit test to cover this change too.
And the rest of the changes is due to go fmt :) 

### Notable changes:
- fix: syncer cache race condition after Service creation

### DIFF
--- a/pkg/controllers/resources/services/syncer_test.go
+++ b/pkg/controllers/resources/services/syncer_test.go
@@ -2,9 +2,10 @@ package services
 
 import (
 	"context"
+	"testing"
+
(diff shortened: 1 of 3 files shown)


FabianKramm (approved):

PR #292, opened 2022-01-11
fix: cache race condition after Service creation

matskiv:
Cherry-pick of the #291 for v0.5

### Notable changes:

- fix: syncer cache race condition after Service creation


FabianKramm (approved):

PR #294, opened 2022-01-13
fix: update latest supported minor version for k8s flavor

matskiv:
I added 1.23 into the version map a while back, but the condition and fallback values were not updated. 
Since 0.5.0 release notes state `cli: Added support for k8s v1.23`, should I cherry-pick this into v0.5 branch?


FabianKramm (approved):
@matskiv thanks for creating this PR! LGTM! Regarding back-porting this to v0.5.0, I don't think this is necessary, as v1.23 still works and only v1.24 would get used as v1.22 instead of v1.23, so I guess we'll release v0.6 before v1.24 comes out

PR #299, opened 2022-01-13
test (e2e): should be able to deny attaching pod

matskiv:
Adding the `should be able to deny attaching pod` test case from the conformance test suite.


FabianKramm (approved):
@matskiv thanks for creating this PR! LGTM!

PR #307, opened 2022-01-17
Sync refactor - renames and Initializer interface

matskiv:
I implemented renaming as discussed, and added Initializer interface that gets executed in parallel.


FabianKramm:
LGTM! Thanks

PR #316, opened 2022-01-19
Fix webhooks executed in the syncer

matskiv:
The test was not failing when it should have, so the first commit fixes the test and the flaw in the test suite that stayed hidden thanks to the bad test.
**Note:** This test should fail now due to a bug in the syncer. I will first let it run and fail to prove that the test is fixed, and afterward, I'll push a commit with a fix for the bug and then it should pass.

### DIFF
--- a/e2e/framework/helper.go
+++ b/e2e/framework/helper.go
@@ -1,6 +1,9 @@
 package framework
 
 import (
+	"fmt"
+
+	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
 )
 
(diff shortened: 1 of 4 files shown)


matskiv:
The e2e tests failed as expected:

> unexpected 'kubectl attach' error message. expected to contain "attaching to pod 'to-be-attached-pod' is not allowed", got "error running /usr/local/bin/kubectl --kubeconfig=/tmp/vcluster_e2e_kubeconfig_2343535401 --namespace=e2e-webhook-2-kdkuu attach --namespace=e2e-webhook-2-kdkuu to-be-attached-pod -i -c=container1:\nCommand stdout:\n\nstderr:\nIf you don't see a command prompt, try pressing enter.\nError from server (InternalError): Internal error occurred: failed calling webhook \"deny-attaching-pod.k8s.io\": failed to call webhook: the server could not find the requested resource\n\nerror:\nexit status 1"

I'll push a commit with a fix next.

FabianKramm (approved):

PR #317, opened 2022-01-19
feat: allow command execution with vcluster connect

FabianKramm:
### Changes
- **cli**: Allows command execution with vcluster context in command `vcluster connect`. For example:
```
# Retrieve vcluster namespaces
vcluster connect test -n test -- kubectl get ns

# New shell with vcluster kube context
vcluster connect test -n test -- bash
```
- **cli**: `vcluster connect` will now use a random port locally to avoid port conflicts if no `--local-port` flag is specified.
- **cli**: Added shortcut `-s` for global flag `--silent`


matskiv (approved):
I tested it with a couple of commands, works as expected, and the code looks good :+1: 
A really useful feature for scripts, CI, etc.

Issue #252, opened 2021-12-14
DaemonSet Pods stays on node even if no other workload is present


wcarlsen:
This might be intentional, so please close the issue if so and thanks for a wonderful project btw.

**What**
When spinning up a multi node cluster with e.g. minikube and deploying a nginx:latest deployment with a single replica, I only see the node where the workload is scheduled. I then scale the deployment to 10 replicas and I see more nodes, because of the 10 workload Pods being spread out by the scheduler. I then scale down to a single replica again and I see a single node after some waiting around. It is all to be expected.

But when I do the same procedure with a DaemonSet deployed, my node count never goes back down due to the DaemonSet workloads still running on those nodes even no other relevant workloads are running on those nodes.

**Expected behavior**
I would have expected the DaemonSet Pods to be terminated (after some time) on the nodes where no relevant workload is running in order to minimize unnecessary vcluster workload pressure on the "mother ship" cluster.

**Test spec**
* Vcluster version 0.4.5 (latest available version at the time)
* Minikube version v1.24.0
* Kubectl version v1.22.4

```bash
# 0) Spin up minikube test cluster
minikube start --cpus 2 --memory 2048 --nodes=3 --cni=flannel

# 1) Create vcluster
vcluster create vcluster1 -n vcluster1 --create-namespace --disable-ingress-sync

# 2) Connect to vcluster
vcluster connect vcluster1 -n vcluster1

# 3) Create deployment
kubectl --kubeconfig ./kubeconfig.yaml create deployment workload --image=nginx:latest --replicas=1

#  4) Scale deployment
kubectl --kubeconfig ./kubeconfig.yaml scale deployment workload --replicas=10

# 5) Get nodes
 kubectl --kubeconfig ./kubeconfig.yaml get nodes

# 6) Scale down
kubectl --kubeconfig ./kubeconfig.yaml scale deployment workload --replicas=1

# 7) Get nodes (OBS: you have to wait for the count to go down)
 kubectl --kubeconfig ./kubeconfig.yaml get nodes

# 8) Apply DaemonSet
cat <<EOF | kubectl --kubeconfig ./kubeconfig.yaml create -f -
apiVersion: apps/v1
kind: DaemonSet
metadata:
  labels:
    app: nginx
  name: daemonset
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx:latest
        name: nginx
EOF

[...]

PR #318, opened 2022-01-19
Filter out DaemonSet Pods when evaluating if Node is needed

wcarlsen:
Filter out DaemonSet Pods when evaluating if Node is needed, to avoid DaemonSets being left on Nodes where no other Pods run see #252

Thanks for a create project.

Signed-off-by: Willi Carlsen <carlsenwilli@gmail.com>

### DIFF
--- a/pkg/controllers/resources/nodes/fake_syncer.go
+++ b/pkg/controllers/resources/nodes/fake_syncer.go
@@ -129,7 +129,7 @@ func (r *fakeNodeSyncer) nodeNeeded(ctx *synccontext.SyncContext, nodeName strin
 		return false, err
 	}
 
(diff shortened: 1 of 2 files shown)


FabianKramm (approved):
@wcarlsen thanks a lot for the PR! Looks great, we can merge that. I guess besides having this for the fake syncer, it also would make sense to do this in the node syncer if no node-selector or sync all nodes flag is defined.

Would you mind adding that as well, its essentially just applying the filter to line 202 of the syncer.go .

wcarlsen:
@FabianKramm thanks and no problem I can try to add it.

FabianKramm (approved):
@wcarlsen thanks for the update! LGTM!

Issue #273, opened 2022-01-06
drain


kfox1111:
I've been playing around with vcluster a little bit. Very cool tool. I think there may be an unforeseen difficulty with it from the management of the underlying cluster standpoint. We typically drain each node, one at a time to upgrade the cluster. But when testing out that functionality when a vcluster was running:
```
cannot delete Pods not managed by ReplicationController, ReplicaSet, Job, DaemonSet or StatefulSet (use --force to override): default/coredns-85cb69466-cljmm-x-kube-system-x-test, kube-system/storage-provisioner
```
It looks like the owner reference is copied, but the underlying k8s seems to ignore it (perhaps since it doesnt exist). Forcing a drain is not a particularly safe operation.

I don't know if there is an annotation saying a pod is actually managed and allow it to be deleted. If such a feature doesn't exist it probably should. Maybe a feature request to k8s is in order. Maybe there are other ways of fixing it too.

Along with this, a complimentary feature would be pod disruption budget syncing if that isnt already supported. That way the over cluster could safely inform the undercluser what was safe to drain.

kfox1111:
https://github.com/kubernetes/kubernetes/issues/57049 is related.

FabianKramm:
@kfox1111 thanks for creating this issue! Yes vcluster is managing those pods, for newer versions of vcluster `v0.6.0` and onwards, you could pause the vcluster with `vcluster pause ...` and then drain the nodes, upgrade them and then resume the vcluster.

kfox1111:
Pausing wouldn't help as it would mostly have the same effect. Making the upper clusers workload unavailable. It also breaks the usage contract.

Let me go though an concrete example that might help describe the use case/problem. You might know a lot of this, but going to go through it anyway in case there are other readers.

Some organizations are starting to provide soft multitenant clusters so that groups within the organization can start using Kubernetes as users without having to learn all the complexities of managing a cluster properly. This largely relies on the separation of the cluster-admin role and admin (namespace, workload related) role. The cluster-admin is a different person (or group of people) from the workload administrators.

[...]

PR #321, opened 2022-01-20
refactor: use vcluster service as owner

FabianKramm:
### Changes
- **syncer**: Use vcluster service as owner instead of deployment / statefulset, because we can set the `controller: true` in the owner reference for it, which allows node draining (#273)

### DIFF
--- a/pkg/util/translate/translate.go
+++ b/pkg/util/translate/translate.go
@@ -16,6 +16,8 @@ var (
 	NamespaceLabel = "vcluster.loft.sh/namespace"
 	MarkerLabel    = "vcluster.loft.sh/managed-by"
 	Suffix         = "suffix"
+
+	t = true
 )
 
 var Owner client.Object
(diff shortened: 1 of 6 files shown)


pkg/util/translate/translate.go
```diff
@@ -65,6 +67,7 @@ func GetOwnerReference() []metav1.OwnerReference {
 			Kind:       typeAccessor.GetKind(),
 			Name:       Owner.GetName(),
 			UID:        Owner.GetUID(),
+			Controller: &t,
```
matskiv, line 70: I think it should be set to `true` only if the virtual pod resource has some owner reference with `controller==true`.

btw: we are already checking for almost that here - [pkg/controllers/resources/pods/translate/translator.go#L154-L168](https://github.com/loft-sh/vcluster/blob/0403fecea6d3aea9fd53fddb9f02cd1cfd116fbb/pkg/controllers/resources/pods/translate/translator.go#L154-L168) 
That^ can be simplified with the use of `metav1.GetControllerOf(&pod)`
like it's used in the kubectl drain - https://github.com/kubernetes/kubernetes/blob/2af53e9f64eee81a09fd0fd1f0fc96ce0859eb0f/staging/src/k8s.io/kubectl/pkg/drain/filters.go#L237-L244

matskiv (changes requested):

matskiv (approved):
Looks great!

matskiv (approved):

PR #326, opened 2022-01-21
refactor: get rid of shared persistent volume mutex

FabianKramm:
### Changes
- Refactored fake persistent volume & persitent volume claim controller to not use a shared mutex anymore
- `vcluster connect ... -- command` does not show a fatal anymore if the command has failed.
- Events are only updated if they actually have been changed


matskiv (approved):
LGTM
P.S: removing code without changing functionality is great! :)

PR #327, opened 2022-01-21
Remove shared mutex between pod & node controllers

FabianKramm:
### Changes
- Removes shared mutex between pod and node controllers
- Removes node service and lock factory from register context

### DIFF
--- a/pkg/controllers/resources/nodes/syncer.go
+++ b/pkg/controllers/resources/nodes/syncer.go
@@ -22,7 +22,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-func NewSyncer(ctx *synccontext.RegisterContext) (syncer.Object, error) {
+func NewSyncer(ctx *synccontext.RegisterContext, nodeService nodeservice.NodeServiceProvider) (syncer.Object, error) {
 	var err error
 	var nodeSelector labels.Selector
 	if ctx.Options.SyncAllNodes {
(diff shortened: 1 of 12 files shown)


pkg/controllers/resources/nodes/syncer.go
```diff
@@ -80,12 +82,62 @@ func (s *nodeSyncer) ModifyController(ctx *synccontext.RegisterContext, builder
...
+
+func modifyController(ctx *synccontext.RegisterContext, nodeService nodeservice.NodeServiceProvider, builder *builder.Builder) (*builder.Builder, error) {
+	// start the node service provider
+	go func() {
+		nodeService.Start(ctx.Context)
+	}()
+
+	return builder.Watches(source.NewKindWithCache(&corev1.Pod{}, ctx.PhysicalManager.GetCache()), handler.EnqueueRequestsFromMapFunc(func(object client.Object) []reconcile.Request {
+		pod, ok := object.(*corev1.Pod)
+		if !ok || pod == nil || pod.Namespace != ctx.TargetNamespace || !translate.IsManaged(pod) || pod.Spec.NodeName == "" {
```
matskiv, line 96: Nit-pick: the `pod.Namespace != ctx.TargetNamespace` part can be removed because `ctx.PhysicalManager` only watches that particular namespace.

matskiv (approved):
Great improvement of the code :+1: 
A few comments about the new code that covers some corner cases would be nice :) 
LGTM :+1:

Issue #72, opened 2021-07-02
Support Pod Disruption Budgets (PDBs)


matskiv:
Pod disruption budgets[1] are very important for high availability kubernetes services.
They can be created inside a vcluster, but if I understand vcluster architecture correctly,
they would need to be synced into host cluster to be effective.

[1] https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets

FabianKramm:
@matskiv thanks a lot for creating this issue! Yes you are correct, pod disruption budgets need to be synced to the host cluster to work effectively, we definitely should implement this

PR #331, opened 2022-01-25
Add PodDisruptionBudget syncer

pratikjagrut:
This is PR is regarding issue  #72. It adds syncer for the `PodDisruptionBudget`.

### DIFF
--- a/charts/k0s/templates/rbac/role.yaml
+++ b/charts/k0s/templates/rbac/role.yaml
@@ -22,15 +22,15 @@ rules:
   - apiGroups: ["apps"]
     resources: ["statefulsets", "replicasets", "deployments"]
     verbs: ["get", "list", "watch"]
-  - apiGroups: ["policy"]
-    resources: ["poddisruptionbudgets"]
-    verbs: ["create", "delete", "patch", "update", "get", "list", "watch"]
   {{- if .Values.rbac.role.extended }}
   - apiGroups: ["networking.k8s.io"]
     resources: ["networkpolicies"]
     verbs: ["create", "delete", "patch", "update", "get", "list", "watch"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshots"]
+    verbs: ["create", "delete", "patch", "update", "get", "list", "watch"]
+  - apiGroups: ["policy"]
+    resources: ["poddisruptionbudgets"]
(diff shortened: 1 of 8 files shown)


matskiv (approved):
Code looks great :+1:

charts/k0s/templates/rbac/role.yaml
```diff
@@ -22,6 +22,9 @@ rules:
   - apiGroups: ["apps"]
     resources: ["statefulsets", "replicasets", "deployments"]
     verbs: ["get", "list", "watch"]
+  - apiGroups: ["policy"]
```
FabianKramm, line 25: Should be moved to .Values.rbac.role.extended below

charts/k3s/templates/rbac/role.yaml
```diff
@@ -22,6 +22,9 @@ rules:
   - apiGroups: ["apps"]
     resources: ["statefulsets", "replicasets", "deployments"]
     verbs: ["get", "list", "watch"]
+  - apiGroups: ["policy"]
```
FabianKramm, line 25: Should be moved to .Values.rbac.role.extended below

charts/k8s/templates/rbac/role.yaml
```diff
@@ -22,6 +22,9 @@ rules:
   - apiGroups: ["apps"]
     resources: ["statefulsets", "replicasets", "deployments"]
     verbs: ["get", "list", "watch"]
+  - apiGroups: ["policy"]
```
FabianKramm, line 25: Should be moved to .Values.rbac.role.extended below

FabianKramm (changes requested):
@pratikjagrut thanks a lot for the PR! Looks good to me, only thing is that we should move rbac permissions to the role.extended area.

charts/k0s/templates/rbac/role.yaml
pratikjagrut: done

charts/k8s/templates/rbac/role.yaml
pratikjagrut: done

charts/k3s/templates/rbac/role.yaml
pratikjagrut: done

FabianKramm (approved):
@pratikjagrut thanks! LGTM!

PR #334, opened 2022-01-25
feat: sign container image and sha file

### LINKED ISSUE #220: sign vCluster binaries/images with cosign
Let's sign vCluster binaries with cosign, and make them verifiable for the end-users by providing clear text on the releases page about how they can do it. AFAIK, vCluster uses GitHub Actions to make a new release, so, it is easy to integrate the signing process into the workflow because there is already GitHub action for installing cosign. 

But, the question is should we consider using keyless mode or using public/private key pairs while signing binaries/images?

* https://chainguard.dev/posts/2021-12-01-zero-friction-keyless-signing
* https://github.com/sigstore/cosign-installer
cc: @Dentrax 

To complete the issue, the followings need to be done:

- [x] ~~sign container images~~
- [x] ~~sign binaries, maybe sha256 file only~~

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

Fixes #220 

👉 publish-cli: https://github.com/developer-guy/vcluster/runs/4947923465?check_suite_focus=true#step:6:12
👉 publish-images: https://github.com/developer-guy/vcluster/runs/4941268675?check_suite_focus=true#step:11:1

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -8,6 +8,8 @@ jobs:
   publish-cli:
     if: startsWith(github.ref, 'refs/tags/v') == true
     runs-on: macos-latest
+    permissions:
+      id-token: write   # This is the key for OIDC!
     steps:
       - name: Set up Go 1.17
         uses: actions/setup-go@v2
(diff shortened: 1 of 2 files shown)


FabianKramm (approved):
@developer-guy thanks a lot for this! LGTM!

Issue #330, opened 2022-01-24
Allow to set tolerations for vcluster pods


mixitgit:
### Why?
Currently it is possible to set node selector to ensure that all pods created in vcluster will be scheduled on some subset of nodes. 
But in scenarios, when there are multiple vclusters, some of which use fake nodes and some of which should use dedicated nodes, it doesn't guarantee that pods from vclusters that use fake nodes won't schedule on dedicated nodes.
### How?
This could be solved by setting tolerations on all pods in vclusters that use dedicated nodes and corresponding taints on these nodes. So in couple with node selector we can guarantee that pods will be scheduled on some node if and only if they belong to some certain vcluster. 
This could be done either by allowing to set tolerations in some separate flag, or just via `--enforce-node-selector`, because there is no case when tolerations and node selector should differ in terms of dedicated nodes

FabianKramm:
@mixitgit thanks for creating this issue! Not sure if I understand this correctly, but you want that vcluster sets tolerations to some pods automatically that are synced between vcluster and the host cluster?

mixitgit:
I want to set tolerations automatically, this would allow to create dedicated nodes for vclusters
For example:
If we set `--enforce-node-selector --node-selector=foo=bar`
Currently it will automatically add a node selector to all pods:
```
nodeSelector:
    worker.kaas.sbrf.ru/vcluster: sbl-d-01
```
If we would also add a toleration:
```
tolerations:
- key: "foo"
  operator: "Equal"
  value: "bar"
  effect: "NoSchedule"
```
We then could create a node with label `foo=bar` and a taint `foo=bar NoSchedule`, and this node will be dedicated to this vcluster (no other vclusters will schedule their pods there)
This is extremely helpful when multiple teams sit on the same host clusters and some of them should have their own dedicated nodes
This toleration won't affect other use cases (unless someone will want to label his node and at the same time have same taint on it, so the pods won't be able to schedule there, but such case seems a bit weird)

FabianKramm:
@mixitgit thanks for the explanation! Ah I see, so you want to dedicate the node completely to a vcluster, that makes sense. Yes I guess we could add a flag for this.

mixitgit:
[...]

PR #347, opened 2022-01-28
Add support for automatic tolerations

kuuji:
Hey 👋! 

My humble and naive attempt at implementing https://github.com/loft-sh/vcluster/issues/330

I couldn't find any Kubernetes function to parse tolerations.  Kubectl doesn't really support toleration, just taint, and there's technically a difference since tolerations have an operator. This does not support the operator as is

The operator will default to `Equal`` which would enforce that both key and value match the taint.

Because of the above I force the user to provide  the following format `key=value:effect` , and will fail at startup if what's provided does not match.

Might be a good idea to adjust the format to be `key=value:effect:operator` and in that case the parsing and error handling need to be a little smarter. (for example the operator can allow the value to not be provided if set to `Exists`)

Open to suggestions about the whole thing.

I have tested this in devspace with the debugger fyi. Seems to do what I expect.

### DIFF
--- a/cmd/vcluster/cmd/start.go
+++ b/cmd/vcluster/cmd/start.go
@@ -185,20 +185,9 @@ func ExecuteStart(options *context2.VirtualClusterOptions) error {
 		return err
 	}
 
-	if len(options.Tolerations) > 0 {
-		for _, toleration := range options.Tolerations {
-			eqSplit := strings.Split(toleration, "=")
-			if len(eqSplit) < 2 {
-				klog.Fatalf("Toleration: %v improperly formatted", toleration)
-				return errors.New("Toleration improperly formatted")
-			} else {
-				clSplit := strings.Split(eqSplit[1], ":")
-				if len(clSplit) < 2 {
-					klog.Fatalf("Toleration: %v improperly formatted", toleration)
-					return errors.New("Toleration improperly formatted")
-				}
-			}
-		}
+	err = validateTolerations(options.Tolerations)
+	if err != nil {
+		return err
 	}
 
 	// set suffix
@@ -336,6 +325,25 @@ func ExecuteStart(options *context2.VirtualClusterOptions) error {
 	}
 
 	<-ctx.StopChan
+	return nil
+}
+
+func validateTolerations(tolerations []string) error {
+	if len(tolerations) > 0 {
+		for _, toleration := range tolerations {
+			eqSplit := strings.Split(toleration, "=")
+			if len(eqSplit) < 2 {
+				klog.Fatalf("Toleration: %v improperly formatted", toleration)
+				return errors.New("Toleration improperly formatted")
+			} else {
+				clSplit := strings.Split(eqSplit[1], ":")
+				if len(clSplit) < 2 {
+					klog.Fatalf("Toleration: %v improperly formatted", toleration)
+					return errors.New("Toleration improperly formatted")
+				}
+			}
+		}
+	}
 	return nil
 }
 
--- a/cmd/vcluster/cmd/start.go
+++ b/cmd/vcluster/cmd/start.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
-	corev1 "k8s.io/api/core/v1"
 	"math"
 	"os"
+	"strings"
 	"time"
+
+	corev1 "k8s.io/api/core/v1"
 
 	"github.com/loft-sh/vcluster/pkg/plugin"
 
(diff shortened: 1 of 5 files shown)


cmd/vcluster/cmd/start.go
```diff
@@ -104,6 +106,7 @@ func NewStartCommand() *cobra.Command {
 
 	cmd.Flags().StringSliceVar(&options.TranslateImages, "translate-image", []string{}, "Translates image names from the virtual pod to the physical pod (e.g. coredns/coredns=mirror.io/coredns/coredns)")
 	cmd.Flags().BoolVar(&options.EnforceNodeSelector, "enforce-node-selector", true, "If enabled and --node-selector is set then the virtual cluster will ensure that no pods are scheduled outside of the node selector")
+	cmd.Flags().StringSliceVar(&options.Tolerations, "toleration", []string{}, "If set will apply the provided tolerations to all pods in the vcluster")
```
FabianKramm, line 109: Maybe we should call this `enforce-toleration` instead, as this will make sure the toleration is set.

cmd/vcluster/cmd/start.go
```diff
@@ -320,6 +328,25 @@ func ExecuteStart(options *context2.VirtualClusterOptions) error {
 	return nil
 }
 
+func validateTolerations(tolerations []string) error {
+	if len(tolerations) > 0 {
+		for _, toleration := range tolerations {
+			eqSplit := strings.Split(toleration, "=")
+			if len(eqSplit) < 2 {
+				klog.Fatalf("Toleration: %v improperly formatted", toleration)
```
FabianKramm, line 336: Fatal executes immediately, returning the error here is fine

cmd/vcluster/cmd/start.go
```diff
@@ -320,6 +328,25 @@ func ExecuteStart(options *context2.VirtualClusterOptions) error {
...
+				return errors.New("Toleration improperly formatted")
```
FabianKramm, line 337: Would be great to add the wrong toleration in the error

cmd/vcluster/cmd/start.go
```diff
@@ -320,6 +328,25 @@ func ExecuteStart(options *context2.VirtualClusterOptions) error {
...
+			} else {
```
FabianKramm, line 338: else is not needed here anymore if the branch before returns in any case

cmd/vcluster/cmd/start.go
```diff
@@ -320,6 +328,25 @@ func ExecuteStart(options *context2.VirtualClusterOptions) error {
...
+				clSplit := strings.Split(eqSplit[1], ":")
+				if len(clSplit) < 2 {
+					klog.Fatalf("Toleration: %v improperly formatted", toleration)
```
FabianKramm, line 341: same as above

pkg/controllers/resources/pods/syncer.go
```diff
@@ -70,9 +77,20 @@ func New(ctx *synccontext.RegisterContext) (syncer.Object, error) {
 
 		podTranslator: podTranslator,
 		nodeSelector:  nodeSelector,
+		tolerations:   tolerations,
 	}, nil
 }
 
+func parseToleration(toleration string) (key string, value string, effect string) {
```
FabianKramm, line 84: Currently this feels very static, could we refactor this into:

I see the difference between taints and tolerations, but can we not have a similar syntax for operations to how label expressions work with equal and exists:
- key1=value1:Effect => Operator Equals
- key1:Effect => Operator Exists

Also it would be good to support the case where effect is empty as this will match all taint effects:
- key1=value1 
- key1

There is also another special case and this just the operator `Exists` without effect or key that would match every taint, we could express this via:
- `*`

FabianKramm (changes requested):
@kuuji thanks a lot for the PR! Looks very good already, some smaller comments and I guess we should change the parse function a little bit to support more toleration cases.

cmd/vcluster/cmd/start.go
kuuji: Makes sense, and it's consistent with node selector 👍

pkg/controllers/resources/pods/syncer.go
kuuji: Oh that's a great idea. It only becomes a problem if they introduce a new operator like NoExists or something like that. But with what exists currently, no worries.

I may get a lot of inspiration from https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/taint/utils.go#L77-L118 then. It does almost everything I need logic wise and I could try and import some of the validation function they use even.

And add something separate for the operator.


> There is also a special case where just the operator Exists without effect or key can be specified, which would match every taint, we could express this via:

Oh I never thought about that, that makes a lot of sense.

Lastly, seems like that parsing function could be
1) reused for both the cmd side and the syncer
2) be a good candidate to have a good suite of unit tests given the logic involved

Should I move this to utils? Or is there a better place for this?

kuuji:
Whenever you get a chance to get another look @FabianKramm .
I did quite a bit of cleanup and completely refactored the parsing function, also added a bunch of stuff given how many different use cases there are.

pkg/controllers/resources/pods/syncer.go
FabianKramm: @kuuji sounds great, moving that to utils is a good idea!

FabianKramm (approved):
@kuuji thanks a lot! Looks good now!

PR #348, opened 2022-01-28
fix: several sync fixes

FabianKramm:
### Changes
- **syncer**: Removed `poddisruptionbudgets` as default syncer
- **syncer**: Make initial kube config secret creation non-fatal
- **syncer**: If phyiscal and virtual pod have a node assigned and they differ, delete the physical pod 
- **syncer**: Sync service nodePort from host cluster instead of virtual cluster
- **syncer**: Only write coredns manifests to local file if environment variable "DEBUG" is "true"


matskiv (approved):
Looks good :+1:

PR #349, opened 2022-01-28
refactor: delete virtual pod if node names differ

FabianKramm:
### Changes
- Delete virtual pod instead of physical if nodeNames between virtual and physical pod differ


matskiv (approved):
Good improvement :+1:

PR #350, opened 2022-01-30
Helm-Add support for exposing vCluster access via LoadBalancer

log1cb0mb:
- Updating chart for k8s distro but suppose same can be done for other options?


FabianKramm:
@log1cb0mb thanks for the PR! Currently we already LoadBalancer with `service.type=LoadBalancer`, would be great if you elaborate a little bit why this PR is needed!

log1cb0mb:
> @log1cb0mb thanks for the PR! Currently we already LoadBalancer with `service.type=LoadBalancer`, would be great if you elaborate a little bit why this PR is needed!

As discussed on Slack, updated PR to include options for `externalIPs` and `externalTrafficPolicy` configurations when `service.type=LoadBalancer` is used.

FabianKramm (approved):
@log1cb0mb thanks for the PR! LGTM!

Issue #251, opened 2021-12-13
[bug] Node tolerations are propagated to underlying k8s system


mtougeron:
When creating a deployment in vCluster I would expect the `tolerations` to apply to the 'fake' kubernetes nodes not the ones from the underlying k8s infrastructure. For example, we have an OPA policy that prohibits users from deploying workloads that tolerate running on the control plane nodes outside of the "system" namespaces. When a user tries to launch a vCluster it launches a `coredns` deployment that has the toleration of `node-role.kubernetes.io/master`. This would allow the pod to be launched on the *real* kubernetes master node(s) not just the 'fake' one.

FabianKramm:
@mtougeron thanks for creating this issue! This will be solved with the newer v0.5.0 version which is already available to try out at [v0.5.0-alpha.7](https://github.com/loft-sh/vcluster/releases/tag/v0.5.0-alpha.7).

mtougeron:
Ah great. I'll try and test that out today!

mtougeron:
I tested this out with `0.5.0-alpha.7` and the tolerations are still propagated to the underlying kubernetes system. I see where the coredns can now be modified in this repo to not have the tolerations (lmk if you want me to change that). 

I think what I was expecting was that the `tolerations` would be rewritten like the pod affinity label selectors are so that it tolerates the 'fake' nodes not the real nodes.

FabianKramm:
@mtougeron ah, I see, we have a feature called node-selector for this where you can scope the vcluster workloads to certain nodes (see https://www.vcluster.com/docs/architecture/nodes), which is similar to what you want. Regarding adding tolerations to pods to only tolerate the fake nodes, this might be a little bit of a chicken egg problem as fake nodes are created as soon as a pod is bound to a node. vcluster also has no information about the node since it does not have appropriate RBAC permissions to access node information, so vcluster will just make a node up, when fake nodes are enabled. When using node selector, vcluster will instead sync all nodes that match the selector and also sync the correct node labels, taints etc., which might then be better suited for your use case.

mtougeron:
> this might be a little bit of a chicken egg problem

I see what you mean by this. It might just need the ability to remove the tolerations from the `manifests/coredns/coredns.yaml` so the vCluster fully bootstraps and then we'll handle the rest in our ci pipeline's manipulation of the manifests.

> which might then be better suited for your use case.

[...]

PR #353, opened 2022-02-01
Don't set tolerations for coredns pods

mtougeron:
This PR removes the tolerations for the coredns deployment. In many environments there are restrictions on who/what can run on the control-plane nodes. While coredns is required for DNS inside of vcluster it is not necessary or required that it runs on the control-plane nodes.

This was discussed somewhat in https://github.com/loft-sh/vcluster/issues/251 but I'm open to feedback if this isn't an approach you'd prefer to take.


FabianKramm (approved):
@mtougeron thanks for the PR! Yes I guess its safe to merge this, lgtm

PR #356, opened 2022-02-03
Node selector can be enforced with fake nodes as well. Docs improvements.

matskiv:
#### Breaking changes:
- setting the `--node-selector` flag does not implicitly enable synchronization of real nodes anymore. Consult [the Nodes documentation page](https://www.vcluster.com/docs/architecture/nodes) to see the recommended way of enabling real node sync.

#### Changes:
- feat: the `--node-selector` and `--enforce-node-selector` flags can now be used with fake nodes
- docs: [Architecture - Pod Scheduling page](https://www.vcluster.com/docs/architecture/scheduling) has been improved and covers more use cases, including the new `--enforce-toleration` flag.


FabianKramm (approved):

PR #357, opened 2022-02-04
feat: add support for pod security standard

pratikjagrut:
This PR adds support for the pod security standards.

Any pod security standard can be enforced using `--enforce-pod-security-standard` flag. This flag accepts only one value from `privileged, baseline and restricted` and default value is `privileged`.


cmd/vcluster/cmd/start.go
```diff
@@ -138,6 +140,23 @@ func NewStartCommand() *cobra.Command {
 }
 
 func ExecuteStart(options *context2.VirtualClusterOptions) error {
+	// check the value of pod security standard
+	if options.PodSecurityStandard != "" {
+		found := func() bool {
```
FabianKramm, line 145: why create a function for this?

cmd/vcluster/cmd/start.go
```diff
@@ -124,6 +124,8 @@ func NewStartCommand() *cobra.Command {
 
 	cmd.Flags().StringVar(&options.DefaultImageRegistry, "default-image-registry", "", "This address will be prepended to all deployed system images by vcluster")
 
+	cmd.Flags().StringVar(&options.PodSecurityStandard, "enforce-pod-security-standard", "privileged", "This can be set to privileged, baseline, restricted (defaults to privileged) and vcluster would make sure during translation that these policies are enforced.")
```
FabianKramm, line 127: it should be possible to set `baseline,restricted` at the same time as those standards are additive (compare pod security standard docs)

cmd/vcluster/context/context.go
```diff
@@ -62,6 +62,8 @@ type VirtualClusterOptions struct {
 
 	DefaultImageRegistry string `json:"defaultImageRegistry"`
 
+	PodSecurityStandard string `json: enforcePodSecurityStandard`
```
FabianKramm, line 65: Wrong format, you need to do:

PodSecurityStandard string `json:"podSecurityStandard"`

pkg/controllers/podSecurity/podsecurity.go
```diff
@@ -0,0 +1,49 @@
...
+
+type PodSecurityReconciler struct {
+	Ctx *context2.ControllerContext
+	Log loghelper.Logger
+}
+
+func (r *PodSecurityReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	client := r.Ctx.VirtualManager.GetClient()
+	nsl := &corev1.NamespaceList{}
+	err := client.List(ctx, nsl)
```
FabianKramm, line 27: don't list all namespaces, rather only get the namespace from req

pkg/controllers/podSecurity/podsecurity.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+const (
+	Namespace     = "kube-system"
+	ConfigMapName = "coredns"
+	NodeHostsKey  = "NodeHosts"
+)
+
+type PodSecurityReconciler struct {
+	Ctx *context2.ControllerContext
```
FabianKramm, line 20: use virtual client and podsecuritystandard here instead of the whole context, makes it a lot easier for unit testing

pkg/controllers/register.go
```diff
@@ -157,6 +163,19 @@ func registerCoreDNSController(ctx *context.ControllerContext) error {
...
 }
 
+func registerPodSecurityController(ctx *context.ControllerContext) error {
+	err := (&podsecurity.PodSecurityReconciler{
+		Ctx: ctx,
+		Log: loghelper.New("podSecurity-controller"),
+	}).SetupWithManager(ctx.VirtualManager)
+
+	if err != nil {
+		panic(err)
```
FabianKramm, line 173: don't panic here

FabianKramm (changes requested):
@pratikjagrut thanks for the PR! Please see my comments below

cmd/vcluster/cmd/start.go
pratikjagrut: To make sure that the passed argument is valid.

cmd/vcluster/context/context.go
pratikjagrut: ahh, I changed the type and then forgot to change its format.

pkg/controllers/register.go
pratikjagrut: Oh, it was only for testing purposes.

pkg/controllers/podSecurity/podsecurity.go
```diff
@@ -0,0 +1,49 @@
...
+	pssKey := "pod-security.kubernetes.io/enforce"
+	pssValue := r.PodSecurityStandard
+
+	ns := &corev1.Namespace{}
+	err := client.Get(ctx, req.NamespacedName, ns)
+	if err != nil {
+		return ctrl.Result{RequeueAfter: time.Second}, err
+	}
+
+	labels := ns.GetLabels()
```
FabianKramm, line 31: make sure labels are non nil

pkg/controllers/podSecurity/podsecurity.go
```diff
@@ -0,0 +1,49 @@
...
+	labels[pssKey] = pssValue
+	labels["pod-security.kubernetes.io/enforce-version"] = "latest"
+
+	ns.SetLabels(labels)
+	err = client.Update(ctx, ns)
```
FabianKramm, line 36: only update if things actually have changed

cmd/vcluster/cmd/start.go
```diff
@@ -138,6 +140,17 @@ func NewStartCommand() *cobra.Command {
 }
 
 func ExecuteStart(options *context2.VirtualClusterOptions) error {
+	// check the value of pod security standard
+	var allowedPodSecurityStandards = map[string]bool{
```
FabianKramm, line 144: make this a static variable outside of the function

FabianKramm (changes requested):
@pratikjagrut please see comments below

pkg/controllers/podSecurity/podsecurity.go
pratikjagrut: I think labels are never nil. This one is ` kubernetes.io/metadata.name=` supposed to be there always. But I'll still add nil check.

pratikjagrut:
Not ready for review.

pkg/controllers/register.go
```diff
@@ -120,6 +121,12 @@ func RegisterControllers(ctx *context.ControllerContext, syncers []syncer.Object
 		return err
 	}
 
+	// register controller that maintains pod security standard check
+	err = registerPodSecurityController(ctx)
```
FabianKramm, line 125: only register controller if pod security standard is enabled

pkg/controllers/resources/pods/validatePodSecurity.go
```diff
@@ -0,0 +1,110 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	kubeinformers "k8s.io/client-go/informers"
+	"k8s.io/pod-security-admission/admission"
+	admissionapi "k8s.io/pod-security-admission/admission/api"
+	"k8s.io/pod-security-admission/api"
+	"k8s.io/pod-security-admission/metrics"
+	"k8s.io/pod-security-admission/policy"
+)
+
+func validatePodSecurityStandards(ctx context.Context, pod *corev1.Pod, s *podSyncer) (*admissionv1.AdmissionResponse, error) {
```
FabianKramm, line 19: please don't put the whole podSyncer as parameter, instead make the function part of the podSyncer with:
```
func (s *podSyncer) validatePodSecurityStandards(ctx context.Context, pod *corev1.Pod) (*admissionv1.AdmissionResponse, error) {
```

pkg/controllers/resources/pods/validatePodSecurity.go
```diff
@@ -0,0 +1,110 @@
...
+		EnforceVersion: version.String(),
+		Audit:          s.podSecurityStandard,
+		AuditVersion:   version.String(),
+		Warn:           s.podSecurityStandard,
+		WarnVersion:    version.String(),
+	}
+
+	informerFactory := kubeinformers.NewSharedInformerFactory(s.virtualClusterClient, 0 /* no resync */)
+	namespaceInformer := informerFactory.Core().V1().Namespaces()
+	namespaceLister := namespaceInformer.Lister()
```
FabianKramm, line 38: Please rewrite this so that we don't need an new informer here, rather use the ctx.VirtualClient for this (create a new struct that implements the namespace lister interface and just uses the crx.VirtualClient under the hood)

FabianKramm (changes requested):
@pratikjagrut thanks a lot, please check my comments below

pkg/controllers/register.go
pratikjagrut: pod security standard will be enabled by default with privileged.

pkg/controllers/resources/pods/validatePodSecurity.go
pratikjagrut: Okay

pkg/controllers/resources/pods/validatePodSecurity.go
pratikjagrut: Done. PTAL.

cmd/vcluster/cmd/start.go
```diff
@@ -124,6 +129,9 @@ func NewStartCommand() *cobra.Command {
 
 	cmd.Flags().StringVar(&options.DefaultImageRegistry, "default-image-registry", "", "This address will be prepended to all deployed system images by vcluster")
 
+	cmd.Flags().BoolVar(&options.EnforcePodSecurityStandard, "enforce-pod-security-standard", false, "If enabled and --pod-security-standard is set then the virtual cluster will enforce specified pod security standard")
+	cmd.Flags().StringVar(&options.PodSecurityStandard, "pod-security-standard", "privileged", "This can be set to privileged, baseline, restricted (defaults to privileged) and vcluster would make sure during translation that these policies are enforced.")
```
FabianKramm, line 133: Merge this into a single flag called `enforce-pod-security-standard` that is empty by default, but as soon as its defined its enforced at the same time

pkg/controllers/podsecurity/podsecurity.go
```diff
@@ -0,0 +1,60 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type PodSecurityReconciler struct {
+	client.Client
+	PodSecurityStandard string
+	Log                 loghelper.Logger
+}
+
+var PodSecurityExemptNamespaces = map[string]bool{
```
FabianKramm, line 20: All namespaces should be enforced

pkg/controllers/resources/pods/validatePodSecurity.go
```diff
@@ -0,0 +1,156 @@
...
+	"k8s.io/pod-security-admission/admission"
+	admissionapi "k8s.io/pod-security-admission/admission/api"
+	"k8s.io/pod-security-admission/api"
+	"k8s.io/pod-security-admission/metrics"
+	"k8s.io/pod-security-admission/policy"
+)
+
+func (s *podSyncer) isPodSecurityStandardsValid(ctx *synccontext.SyncContext, pod *corev1.Pod) (bool, error) {
+	result, err := s.validatePodSecurityStandards(ctx.Context, pod)
+	fmt.Println(result)
```
FabianKramm, line 26: remove this please

pkg/controllers/resources/pods/validatePodSecurity.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		ctx.Log.Errorf(err.Error())
+	} else if result != nil {
+		if !result.Allowed {
+			ctx.Log.Errorf("%s pod creation not allowed: %s", pod.Name, result.Result.Message)
+			pod.Status.Message = result.Result.Message
+			pod.Status.Reason = string(result.Result.Reason)
```
FabianKramm, line 33: Don't update the pod status here please as this can lead to unexpected results, emitting an event is enough for our case

pkg/controllers/resources/pods/validatePodSecurity.go
```diff
@@ -0,0 +1,156 @@
...
+			_ = ctx.VirtualClient.Status().Update(ctx.Context, pod)
+			s.EventRecorder().Eventf(pod, "Warning", "SyncError", `Pod %s is forbidden: violates PodSecurity "%s:latest"`, pod.Name, s.podSecurityStandard)
+		}
+		return result.Allowed, nil
+	}
+	return false, err
+}
+
+func (s *podSyncer) validatePodSecurityStandards(ctx context.Context, pod *corev1.Pod) (*admissionv1.AdmissionResponse, error) {
+	if _, ok := podsecurity.PodSecurityExemptNamespaces[pod.Namespace]; ok {
```
FabianKramm, line 43: there are no namespaces that should be excluded

pkg/controllers/resources/pods/validatePodSecurity.go
```diff
@@ -0,0 +1,156 @@
...
+	podSecurityDefaults := admissionapi.PodSecurityDefaults{
+		Enforce:        s.podSecurityStandard,
+		EnforceVersion: version.String(),
+		Audit:          s.podSecurityStandard,
+		AuditVersion:   version.String(),
+		Warn:           s.podSecurityStandard,
+		WarnVersion:    version.String(),
+	}
+
+	namespaceLister := &namespaceLister{
```
FabianKramm, line 59: please pass `ctx *synccontext.SyncContext` as parameter here and use `ctx.VirtualClient` here instead of `s.virtualClusterClient`

pkg/controllers/resources/pods/validatePodSecurity.go
```diff
@@ -0,0 +1,156 @@
...
+
+	if err := adm.CompleteConfiguration(); err != nil {
+		return nil, fmt.Errorf("configuration error: %w", err)
+	}
+
+	if err := adm.ValidateConfiguration(); err != nil {
+		return nil, fmt.Errorf("invalid configuration: %w", err)
+	}
+
+	nsPolicy, _ := adm.PolicyToEvaluate(map[string]string{})
```
FabianKramm, line 88: can we safely ignore the error here?

pkg/controllers/resources/pods/validatePodSecurity.go
```diff
@@ -0,0 +1,156 @@
...
+		client: s.virtualClusterClient,
+		ctx:    ctx,
+	}
+
+	evaluator, err := policy.NewEvaluator(policy.DefaultChecks())
+	if err != nil {
+		return nil, fmt.Errorf("could not create PodSecurityRegistry: %w", err)
+	}
+
+	adm := &admission.Admission{
```
FabianKramm, line 69: are all these listers required here if we just use the `adm.EvaluatePod` function?

FabianKramm (changes requested):
@pratikjagrut please take a look at my comments below

pkg/controllers/resources/pods/validatePodSecurity.go
pratikjagrut: if we remove the admission config validation method then we can remove all listers.

pkg/controllers/resources/pods/validatePodSecurity.go
pratikjagrut: I think yes. Anyway, we're using an empty map there.

pkg/controllers/resources/pods/validatePodSecurity.go
pratikjagrut: Okay

FabianKramm (approved):
@pratikjagrut thanks a lot! LGTM!

PR #371, opened 2022-02-11
feat(pod): Add annotation to overwrite phyiscal SA

### LINKED ISSUE #370: Schedule Pod with specific Serviceaccount

To use IRSA (IAM roles for serviceaccounts) on AWS EKS it is required to run a pod with a service account that has the `eks.amazonaws.com/role-arn` annotation set so EKS will mount the `AWS_WEB_IDENTITY_TOKEN_FILE` into the pod.

However, VCluster always uses the same service account it has been started with (i.e. `default`).

Provider an option that allows a user in the virtual cluster to override the service account name on the physical pod. This easiest way would be to add an annotation i.e. `vcluster.loft.sh/override-service-account-name` whose value will be used to set `spec.serviceAccountName`.

_No response_

### Additional context

_No response_

MisterMX:

/kind enhancement
/kind feature

resolves #

This adds the possibility to overwrite the `spec.serviceAccountName` of the physical pod via an optional annotation.

resolves #370 

---

### DIFF
--- a/pkg/controllers/resources/pods/translate/translator.go
+++ b/pkg/controllers/resources/pods/translate/translator.go
@@ -29,15 +29,16 @@ import (
 )
 
 const (
-	OwnerSetKind                  = "vcluster.loft.sh/owner-set-kind"
-	NamespaceAnnotation           = "vcluster.loft.sh/namespace"
-	NameAnnotation                = "vcluster.loft.sh/name"
-	LabelsAnnotation              = "vcluster.loft.sh/labels"
-	NamespaceLabelPrefix          = "vcluster.loft.sh/ns-label"
-	UIDAnnotation                 = "vcluster.loft.sh/uid"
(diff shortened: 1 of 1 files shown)


MisterMX:
@FabianKramm yes, I also thought about if/how this could be used as exploit. But since vcluster runs in its own physical namespace and SAs are namespace-scoped as well I see this as the responsibility of the host operator to not give access to SAs that have cluster-wide permssions.

If we would like to prevent virtual users from using host SAs other than default I could think about disabling the annotation ValdiationAdmissionWebhooks or via CLI argument, i.e. `--disable-service-account-override`.

What do you think?

matskiv:
The original use case raised in #370 has been addressed by #377, so I'll close this PR now.
@MisterMX Thank you for the contribution! And feel free to comment if you disagree, or if I misunderstood the intent of this PR... :)

PR #377, opened 2022-02-16
feat: add serviceaccount syncer

### LINKED ISSUE #370: Schedule Pod with specific Serviceaccount

FabianKramm:
/kind feature

resolves #370

Added a new service account syncer that makes it possible to sync service accounts from the vcluster to the host cluster with certain annotations and labels. This is useful for features such as [IAM Roles for ServiceAccounts](https://docs.aws.amazon.com/eks/latest/userguide/specify-service-account-role.html) where the service account needs a certain annotation to give AWS permissions to a pod

### DIFF
--- a/charts/k0s/templates/rbac/role.yaml
+++ b/charts/k0s/templates/rbac/role.yaml
@@ -32,6 +32,11 @@ rules:
     resources: ["volumesnapshots"]
     verbs: ["create", "delete", "patch", "update", "get", "list", "watch"]
   {{- end }}
+  {{- if or .Values.sync.serviceaccounts.enabled .Values.rbac.role.extended }}
+  - apiGroups: [""]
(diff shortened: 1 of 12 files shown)


matskiv (approved):
Looks good :+1:

PR #378, opened 2022-02-16
refactor: sync storageclasses from vcluster

FabianKramm:
/kind enhancement

vcluster will now sync storage classes from the virtual cluster to the host cluster if sync of storage classes is enabled. This will replace the current behaviour where storage classes where only synced from host to virtual cluster. 

We decided to replace the existing behaviour, because creating storage classes is a valid use case as long as the CSI driver is installed within the host cluster, but certain parameters for the CSI driver should get changed through a storage class. It also makes sense to not sync created storage classes from the host cluster anymore as this is not required to schedule persistent volume claims and currently just has informational purposes.

This is somewhat a breaking change as vclusters that currently have sync of storage classes enabled would now behave differently moving forward as changes to the host cluster storage classes are not propagated anymore. However migration should work as expected, as created storage classes within vcluster that mirrored host cluster storage classes before would just get created in the host cluster under a different name.

### DIFF
--- a/pkg/controllers/resources/persistentvolumeclaims/translate.go
+++ b/pkg/controllers/resources/persistentvolumeclaims/translate.go
@@ -2,11 +2,10 @@ package persistentvolumeclaims
 
 import (
 	"context"
-
+	"fmt"
+	"github.com/loft-sh/vcluster/pkg/constants"
 	synccontext "github.com/loft-sh/vcluster/pkg/controllers/syncer/context"
 	"github.com/loft-sh/vcluster/pkg/controllers/syncer/translator"
-
-	"github.com/loft-sh/vcluster/pkg/constants"
 	"github.com/loft-sh/vcluster/pkg/util/translate"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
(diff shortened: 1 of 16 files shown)


matskiv (changes requested):
I added two comments + docs update is needed (at least the "Synced Resources" page)

pkg/controllers/resources/persistentvolumeclaims/translate.go
```diff
@@ -15,47 +14,74 @@ import (
...
+		storageClassName = *vPvc.Spec.StorageClassName
+	} else if vPvc.Annotations != nil && vPvc.Annotations[deprecatedStorageClassAnnotation] != "" {
+		storageClassName = vPvc.Annotations[deprecatedStorageClassAnnotation]
+	}
+
+	// translate storage class if we manage those in vcluster
+	if s.storageClassesEnabled {
+		if storageClassName == "" {
+			return nil, fmt.Errorf("no storage class defined for pvc %s/%s", vPvc.Namespace, vPvc.Name)
+		}
```
matskiv, line 50: I think this condition should be removed.
Quoting [from docs](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1): `PVCs don't necessarily have to request a class. A PVC with its storageClassName set equal to "" is always interpreted to be requesting a PV with no class`

matskiv (changes requested):
One comment
+ docs update is missing (at least "Synced Resources" page)

PR #381, opened 2022-02-17
fix: add sync.* values and RBAC changes to EKS chart

matskiv:
/kind bugfix

This change brings EKS chart into sync with all other charts by adding changes done in #346 

- fix: EKS chart now accepts `.sync.*` values as documented

I also moved plugin extra rules to a template helper. This has no impact on the functionality of the charts, it just makes the chart definitions more consistent.
I will also send a PR for v0.6 branch once this one is approved.


FabianKramm (approved):

PR #383, opened 2022-02-18
Split daemonset filters into two

### LINKED ISSUE #252: DaemonSet Pods stays on node even if no other workload is present

wcarlsen:
/kind bugfix
/kind enhancement
/kind feature

resolves #252

Split daemonset filters into two, because virtual and physical pods do not share exactly the same characteristics when it comes to daemonsets, hence filtering methods needs to be different. This enables the virtual cluster to get rid of lonely daemonsets staying on nodes, where previously valid workload had run. So now a cluster can actually scale down.

### DIFF
--- a/pkg/controllers/resources/nodes/fake_syncer.go
+++ b/pkg/controllers/resources/nodes/fake_syncer.go
@@ -243,7 +243,7 @@ func filterOutVirtualDaemonSets(pl *corev1.PodList) []corev1.Pod {
 
 // Filter away physical DaemonSet Pods using annotations to enable scale down
 func filterOutPhysicalDaemonSets(pl *corev1.PodList) []corev1.Pod {
-	var podsNoDaemoSets []corev1.Pod
+	var podsNoDaemonSets []corev1.Pod
 
 	for _, item := range pl.Items {
 
--- a/pkg/controllers/resources/nodes/fake_syncer.go
+++ b/pkg/controllers/resources/nodes/fake_syncer.go
@@ -250,11 +250,8 @@ func filterOutPhysicalDaemonSets(pl *corev1.PodList) []corev1.Pod {
 	for _, item := range pl.Items {
 
 		annotations := item.GetAnnotations()
-		ownerSetKind, exists := annotations[translate.OwnerSetKind]
-		if exists {
-			if ownerSetKind != "DaemonSet" {
-				podsNoDaemonSets = append(podsNoDaemonSets, item)
-			}
+		if kind, ok := annotations[translate.OwnerSetKind]; ok && kind != "DaemonSet" {
+			podsNoDaemonSets = append(podsNoDaemonSets, item)
 		}
 	}
 	return podsNoDaemonSets
--- a/pkg/controllers/resources/nodes/fake_syncer.go
+++ b/pkg/controllers/resources/nodes/fake_syncer.go
@@ -217,8 +217,8 @@ func CreateFakeNode(ctx context.Context, nodeServiceProvider nodeservice.NodeSer
 	return nil
 }
 
-// Filter away DaemonSet Pods using OwnerReferences
-func filterOutDaemonSets(pl *corev1.PodList) []corev1.Pod {
+// Filter away  virtual DaemonSet Pods using OwnerReferences to enable scale down
+func filterOutVirtualDaemonSets(pl *corev1.PodList) []corev1.Pod {
 	var podsNoDaemonSets []corev1.Pod
 
 	for _, item := range pl.Items {
@@ -239,4 +239,19 @@ func filterOutDaemonSets(pl *corev1.PodList) []corev1.Pod {
 	}
 
 	return podsNoDaemonSets
+}
+
+// Filter away physical DaemonSet Pods using annotations to enable scale down
+func filterOutPhysicalDaemonSets(pl *corev1.PodList) []corev1.Pod {
+	var podsNoDaemoSets []corev1.Pod
+
+	for _, item := range pl.Items {
+
+		annotations := item.GetAnnotations()
+
+		if annotations["vcluster.loft.sh/owner-set-kind"] != "DaemonSet" {
+			podsNoDaemoSets = append(podsNoDaemoSets, item)
+		}
+	}
+	return podsNoDaemoSets
 }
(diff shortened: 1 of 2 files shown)


pkg/controllers/resources/nodes/fake_syncer.go
```diff
@@ -240,3 +240,18 @@ func filterOutDaemonSets(pl *corev1.PodList) []corev1.Pod {
 
 	return podsNoDaemonSets
 }
+
+// Filter away physical DaemonSet Pods using annotations to enable scale down
+func filterOutPhysicalDaemonSets(pl *corev1.PodList) []corev1.Pod {
+	var podsNoDaemoSets []corev1.Pod
```
pratikjagrut, line 246: Typo, podsNoDaemoSets -> podsNoDaemonSets.

wcarlsen: Dammit. It is now fixed :1st_place_medal:

pratikjagrut: Thank you.

wcarlsen:
@pratikjagrut I think I would need some here :sob:

pkg/controllers/resources/nodes/fake_syncer.go
```diff
@@ -240,3 +240,18 @@ func filterOutDaemonSets(pl *corev1.PodList) []corev1.Pod {
...
+
+// Filter away physical DaemonSet Pods using annotations to enable scale down
+func filterOutPhysicalDaemonSets(pl *corev1.PodList) []corev1.Pod {
+	var podsNoDaemonSets []corev1.Pod
+
+	for _, item := range pl.Items {
+
+		annotations := item.GetAnnotations()
+
+		if annotations["vcluster.loft.sh/owner-set-kind"] != "DaemonSet" {
```
FabianKramm, line 252: Please check if annotations is nil and please use the constant here (see https://github.com/loft-sh/vcluster/blob/main/pkg/controllers/resources/pods/translate/translator.go)

FabianKramm (changes requested):
@wcarlsen thanks for the PR! Please check my comment below, besides that it looks good

pkg/controllers/resources/nodes/fake_syncer.go
wcarlsen: Also I'm assuming to not append any pods to podsNoDaemonSets list if the annotation doesn't exists?

pkg/controllers/resources/nodes/fake_syncer.go
```diff
@@ -240,3 +242,20 @@ func filterOutDaemonSets(pl *corev1.PodList) []corev1.Pod {
...
 }
+
+// Filter away physical DaemonSet Pods using annotations to enable scale down
+func filterOutPhysicalDaemonSets(pl *corev1.PodList) []corev1.Pod {
+	var podsNoDaemonSets []corev1.Pod
+
+	for _, item := range pl.Items {
+
+		annotations := item.GetAnnotations()
+		ownerSetKind, exists := annotations[translate.OwnerSetKind]
```
pratikjagrut, line 253: This could be re-written as
```
if kind, ok := annotations[translate.OwnerSetKind]; ok && kind != "DaemonSet" {
	podsNoDaemonSets = append(podsNoDaemonSets, item)
}
```

pratikjagrut (changes requested):

pkg/controllers/resources/nodes/fake_syncer.go
wcarlsen: You are absolutely right :+1:  Done!

pkg/controllers/resources/nodes/fake_syncer.go
FabianKramm: Yes exactly thanks for the quick change

FabianKramm (approved):
@wcarlsen thanks LGTM!

pratikjagrut (approved):
@wcarlsen Thanks a lot! LGTM.

wcarlsen:
We made it thanks guys!

Issue #386, opened 2022-02-25
Syncer option to persist labels on physical resources 


bschwenn:
### Is your feature request related to a problem?

Mutating webhooks often use label selectors to determine which resources to process. Since vcluster mutates all the labels on synced resources, this can prevent webhooks in the physical cluster from being able to mutate the physical resources.

In particular we've encountered this issue for synced pods. 
### Which solution do you suggest?

Add a configuration option to persist the labels on resources to the syncer. The option could disable label mutation entirely for some resource types, or add both the vcluster labels and the original labels. 

### Which alternative solutions exist?

- This could be implemented by consumers via a vcluster plugin. However, this seems onerous and would require a lot of wrapping of the standard syncers.  
- Consumers could update their webhooks to catch vcluster mutated labels. However, when using vcluster only as a development workflow this would entail updating production services to handle development cases, which is a bit of an antipattern. 
- Consumers can deploy the webhooks into the vcluster. This complicates the process of spinning up clusters and could lead to quite a sprawl. 

### Additional context

My apologies if there is already a way to achieve this (or a reason it's impossible), I didn't see it in my perusal of the docs and code if so.

FabianKramm:
@bschwenn thanks for creating this issue! vcluster hashes labels to avoid conflicts between resources in the host cluster, for example it would be possible for a vcluster workload to become an endpoint of vcluster itself if we wouldn't do that, so its also a security measure.

What would be your use case for this where you would require labels to be 1:1 synced from the vcluster? Any particular pod you need to identify?

bschwenn:
Thanks for the response @FabianKramm, makes sense. 

To be more specific about our use case here, say we have a mutating webhook configured similarly to the following: 
```
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingWebhookConfiguration
metadata:
  name: pod-mutating-webhook
webhooks:
  - name: ...
    admissionReviewVersions:
      - "v1beta1"
    rules:
      - apiGroups:
          - "*"
        apiVersions:
          - "v1"
        operations:
          - "CREATE"
        resources:
          - pods
    ...
    objectSelector:
      matchExpressions:
      - key: example.com/enable-pod-mutating-webhook
        operator: In
        values:
        - "true"
```
[...]

PR #391, opened 2022-03-02
Add `--sync-labels` option

bschwenn:
/kind feature

resolves #
This resolves issue https://github.com/loft-sh/vcluster/issues/386 by adding a `--sync-labels` flag to the syncer. This flag will result in the specified labels being copied directly from virtual resources onto physical resources, in addition to the labels with the vclusted-translated key `vcluster.loft.sh/label-...`. 

Adds `--sync-label` flag to sync untranslated labels onto physical resources

- I also added a line to ignore vim swp files to `.gitignore`. 
- Tested via devspace deployment as well as a little spec addition

### DIFF
--- a/cmd/vcluster/context/context.go
+++ b/cmd/vcluster/context/context.go
@@ -64,7 +64,7 @@ type VirtualClusterOptions struct {
 
 	EnforcePodSecurityStandard string `json:"enforcePodSecurityStandard"`
 
-	SyncLabels []string `json:syncLabels`
+	SyncLabels []string `json:"syncLabels"`
 
 	// DEPRECATED FLAGS
 	DeprecatedDisableSyncResources     string
--- a/cmd/vcluster/context/context.go
+++ b/cmd/vcluster/context/context.go
@@ -63,6 +63,9 @@ type VirtualClusterOptions struct {
 	DefaultImageRegistry string `json:"defaultImageRegistry"`
 
 	EnforcePodSecurityStandard string `json:"enforcePodSecurityStandard"`
+
+	SyncLabels []string `json:syncLabels`
+
 	// DEPRECATED FLAGS
 	DeprecatedDisableSyncResources     string
 	DeprecatedOwningStatefulSet        string
--- a/pkg/controllers/syncer/translator/cluster_translator.go
+++ b/pkg/controllers/syncer/translator/cluster_translator.go
@@ -22,6 +22,7 @@ func NewClusterTranslator(ctx *context.RegisterContext, name string, obj client.
 		virtualClient:       ctx.VirtualManager.GetClient(),
 		obj:                 obj,
 		nameTranslator:      nameTranslator,
+		syncedLabels:        ctx.Options.SyncLabels,
 	}
 }
 
@@ -32,6 +33,7 @@ type clusterTranslator struct {
 	obj                 client.Object
 	nameTranslator      PhysicalNameTranslator
 	excludedAnnotations []string
+	syncedLabels        []string
 }
 
 func (n *clusterTranslator) Name() string {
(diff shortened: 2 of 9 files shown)


cmd/vcluster/context/context.go
```diff
@@ -63,6 +63,9 @@ type VirtualClusterOptions struct {
 	DefaultImageRegistry string `json:"defaultImageRegistry"`
 
 	EnforcePodSecurityStandard string `json:"enforcePodSecurityStandard"`
+
+	SyncLabels []string `json:syncLabels`
```
FabianKramm, line 67: This should be `json:"syncLabels"`

FabianKramm (changes requested):
@bschwenn thanks so much for the PR! Looks good so far, lets check what the e2e and unit tests say and then we can merge it. There is only 1 minor thing I would like you to change which I commented below.

cmd/vcluster/context/context.go
bschwenn: Good catch!

bschwenn:
@FabianKramm the k3s e2e test failure seems to be from main, I see it failing for the same reason in other PRs (e.g. https://github.com/loft-sh/vcluster/runs/4954748748?check_suite_focus=true). For the k0s failure, it doesn't look related to this PR: 
```
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/ec3a7ce113134d7a93b817d10a8272cb61118579'. Error: The request was canceled due to the configured HttpClient.Timeout of 100 seconds elapsing.
Warning: Back off 12.921 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/ec3a7ce1131[34](https://github.com/loft-sh/vcluster/runs/5404186502?check_suite_focus=true#step:1:34)d7a93b817d10a8272cb61118579'. Error: The request was canceled due to the configured HttpClient.Timeout of 100 seconds elapsing.
Warning: Back off 20.518 seconds before retry.
Error: The request was canceled due to the configured HttpClient.Timeout of 100 seconds elapsing.
```
Seems like some network flakiness that may resolve on retry.

pkg/controllers/syncer/translator/namespaced_translator.go
```diff
@@ -214,16 +216,22 @@ func translateAnnotations(vObj client.Object, pObj client.Object, excluded []str
...
+	for k, v := range vObjLabels {
 		if k == translate.NamespaceLabel {
 			newLabels[k] = v
 			continue
 		}
 
 		newLabels[ConvertLabelKey(k)] = v
 	}
+	for _, k := range syncedLabels {
+		if value, ok := vObjLabels[k]; ok {
```
FabianKramm, line 231: This panics if the virtual object has no labels as the map is nil right?

pkg/controllers/syncer/translator/cluster_translator.go
```diff
@@ -93,9 +95,15 @@ func (n *clusterTranslator) TranslateMetadataUpdate(vObj client.Object, pObj cli
 func (n *clusterTranslator) TranslateLabels(vObj client.Object) map[string]string {
 	newLabels := map[string]string{}
 	if vObj != nil {
+		vObjLabels := vObj.GetLabels()
 		for k, v := range vObj.GetLabels() {
 			newLabels[convertNamespacedLabelKey(n.physicalNamespace, k)] = v
 		}
+		for _, k := range n.syncedLabels {
+			if value, ok := vObjLabels[k]; ok {
```
FabianKramm, line 103: this panics if the virtual object has no labels as the map is nil right?

FabianKramm (changes requested):
@bschwenn yes the tests are fine, the failing test is currently unfortunately a little bit flaky as KinD has problems with multiple incoming exec connections. 

I encountered something else which might lead to a panic if the virtual cluster object labels are nil (empty), would be good if you add a nil check or verify if this would cause problems for such objects

pkg/controllers/syncer/translator/namespaced_translator.go
bschwenn: Good call again! I was assuming `GetLabels()` would not return nil but I can't verify that, so added this nil check.

FabianKramm (approved):
@bschwenn thanks for the update! LGTM!

PR #401, opened 2022-03-15
refactor: add isolated network policies

FabianKramm:
/kind enhancement

- Adds network policies to isolated mode
- Make coredns part of the helm chart to allow easier configuration of the coredns deployment
- vcluster now correctly rewrites PVC kubelet metrics

### DIFF
--- a/devspace.yaml
+++ b/devspace.yaml
@@ -22,13 +22,6 @@ deployments:
       chart:
         name: ./charts/k3s
       values:
-        isolation:
-          enabled: true
-          podSecurityStandard: ""
-          resourceQuota:
-            enabled: false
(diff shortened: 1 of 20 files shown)


devspace.yaml
```diff
@@ -69,7 +75,6 @@ dev:
         - '!/cmd'
         - '!/vendor'
         - '!/hack'
-        - '!/manifests'
```
matskiv, line 72: Would be nice to keep this^
We can add `- '/manifests/coredns'`on the next line to fix the issue with the mounted ConfigMap volume being Readonly.

matskiv (approved):
Looks good :+1:

PR #403, opened 2022-03-16
refactor: run statefulset init as non root

FabianKramm:
/kind enhancement

vcluster will now run the init container for statefulsets as non root


matskiv (approved):
LGTM :+1:

PR #408, opened 2022-03-17
Add option to skip serviceAccountToken issuing in physical pod

bschwenn:
/kind enhancement

This allows users to configure the syncer to skip override of `serviceAccountToken` `projected` volumes. In effect, this will allow the host cluster to issue the service token to the physical pod rather than having vcluster be the issuer. This is important when the host cluster uses a customized auth mechanism like [OIDC](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens). 

Add an option to prevent vcluster from issuing a `serviceAccountToken` for `projected` volumes in the host cluster.

Open to alternatives here!

### DIFF
--- a/cmd/vcluster/cmd/start.go
+++ b/cmd/vcluster/cmd/start.go
@@ -132,6 +132,7 @@ func NewStartCommand() *cobra.Command {
 
 	cmd.Flags().StringVar(&options.EnforcePodSecurityStandard, "enforce-pod-security-standard", "", "This can be set to privileged, baseline, restricted and vcluster would make sure during translation that these policies are enforced.")
 	cmd.Flags().StringSliceVar(&options.SyncLabels, "sync-labels", []string{}, "The specified labels will be synced to physical resources, in addition to their vcluster translated versions.")
+	cmd.Flags().BoolVar(&options.SyncServiceAccountToken, "sync-service-account-token", true, "Sync vcluster-issued ServiceAccountToken onto physical pod")
 
 	// Deprecated Flags
 	cmd.Flags().BoolVar(&options.DeprecatedUseFakeKubelets, "fake-kubelets", true, "DEPRECATED: use --disable-fake-kubelets instead")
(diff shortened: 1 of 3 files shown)


FabianKramm:
@bschwenn thanks for creating the PR! Not sure if we would want to support that flag in this form as it would essentially not allow any virtual pods anymore to communicate to the vcluster as service account tokens are now issued from the host and vcluster would disallow them. Usually what you want to do if you want to enable OIDC tokens for vcluster is to configure the virtual api server to use OIDC and you can then directly log into the virtual cluster with your OIDC token.

bschwenn:
@FabianKramm So to give more context, this is specifically targeted at supporting https://github.com/aws/amazon-eks-pod-identity-webhook running in the host cluster. As it stands, vcluster will override the token supplied to the physical pod via the projected volume added by the mutating webhook. 

Within the vcluster, the default service account token is distinct from the one added by this webhook (mounted as `/var/run/secrets/kubernetes.io/serviceaccount/` vs `/var/run/secrets/eks.amazonaws.com/serviceaccount/`, etc). I _believe_ the former is allowing the pod to successfully auth with the vcluster still, while the latter is allowing the pod to use IAM-based auth with the host cluster. (Admittedly this is a bit outside my primary area of expertise so it's possible I'm not groking this correctly.) Things seem to work correctly when I setup a cluster with these changes. 

We could certainly explore allowing the virtual server to use OIDC, but this would seem to add a lot of overhead to the process of cluster creation when used for developing cluster services.

FabianKramm:
@bschwenn not sure if I understand this correctly, but this should work out of the box in my opinion. The volume for the service account token is only present in the vcluster at first, then rewritten by the syncer, then the host mutating webhook adds the second volume with the EKS token, which is NOT rewritten now as the syncer already rewrote the pod there. So you would end up with 1 rewritten volume of the vcluster token and 1 not rewritten volume of the EKS cluster.

Why would you need another host cluster token there along the EKS one? Do you need to contact the host Kubernetes api server? But I'm not sure if I understand everything correctly here.

bschwenn:
@FabianKramm you're correct actually, it works! I was messing around with a vcluster plugin at the time that I made this, and it triggered an uncommon sequence of syncs which led me to believe it was necessary. 

I'll go ahead and close. Sorry for the noise! 

As per why this is needed: the second token is used to authenticate the pod against AWS APIs: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html.

PR #415, opened 2022-03-23
Add cgnat network cidrblock to networkpolicy

desponda:
Many companies use the cgnat network cidr block as kubernetes ip address space - it would be great if we could add this block to the networkpolicy

/kind enhancement

Add RFC6598 address space to network policy 


desponda:
I just realized this file exists in 3 places so need to change it there too, will add tomorrow

desponda:
Alright files have been updated

FabianKramm (approved):
@desponda thanks for the PR! Looks good to me!

PR #418, opened 2022-03-29
refactor: use vcluster service for fake nodes

FabianKramm:
/kind enhancement

Use the vcluster service label selector for creating a fake kubelet service instead of finding out the pod label selector


matskiv (approved):
The new code looks much better :+1: :+1:

PR #420, opened 2022-04-01
feat: add service selector sync

### LINKED ISSUE #281: Stop synchronizing Endpoints that belong to a Service and enhance Service sync to preserve functionality
*What:*
Currently, vcluster synchronizes Endpoints from the virtual cluster into the host cluster. The main objective of this is to ensure that [kubernetes services](https://kubernetes.io/docs/concepts/services-networking/service/) work as expected, but it should be achievable with a "translation" of the selector field of the service resources to point to the expected pods that are running in the host, and this should be implemented as part of this issue.
Going forward we should not sync Endpoints that are backing Services from the virtual cluster into the host cluster.
We will sync other Endpoints created by the user in the virtual cluster, but this will likely be an opt-in feature.

*Why:*
Reduce the permission scope of the vcluster.
In some cases, users don't have create/update/delete permissions for the endpoint resources. This will likely be more and more common, for example, due to CVE-2021-25740 affecting Kubernetes - https://github.com/kubernetes/kubernetes/issues/103675 (TLDR: mitigation of the CVE is to remove Endpoint permissions from the edit role).
This would also allow us to reduce permissions required for installation on OpenShift. Related issue - #280 

*How:*
A deeper investigation into the viability of the described changes. E.g. other needs for Endpoints edit permissions besides basic Service functionality - leader election and "control-plane.alpha.kubernetes.io/leader" annotation?
How will we be able to ignore the Service Endpoints in the endpoints controller?

FabianKramm:
/kind feature

resolves #281 

New syncer flag `--sync-service-selector` to let vcluster sync service selectors instead of endpoints

### DIFF
--- a/cmd/vcluster/cmd/start.go
+++ b/cmd/vcluster/cmd/start.go
@@ -106,7 +106,6 @@ func NewStartCommand() *cobra.Command {
 	cmd.Flags().IntVar(&options.Port, "port", 8443, "The port to bind to")
 
 	cmd.Flags().BoolVar(&options.SyncAllNodes, "sync-all-nodes", false, "If enabled and --fake-nodes is false, the virtual cluster will sync all nodes instead of only the needed ones")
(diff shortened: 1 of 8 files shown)


matskiv (approved):
Code looks good and workes as expected :+1: (tested with k3s chart)

PR #421, opened 2022-04-01
Add option to start vcluster with a scheduler

FabianKramm:
/kind feature

vcluster is now able to run a scheduler inside the virtual cluster and assign pods to nodes directly there. This allows users to taint and label nodes inside the vcluster to use pod affinity etc. without any side effect on the actual host cluster nodes.


cmd/vcluster/context/context.go
```diff
@@ -147,6 +148,11 @@ func NewControllerContext(currentNamespace string, localManager ctrl.Manager, vi
 		return nil, err
 	}
 
+	// check if nodes controller needs to be enabled
+	if (options.SyncAllNodes || options.EnableScheduler) && !controllers["nodes"] {
+		controllers["nodes"] = true
+	}
```
matskiv, line 154: I am not a fan of this kind of implicit behavior. I think we should rather document that "nodes" controller should be enabled when the user wants to enable the scheduler.

Right now, there is a permissions issue (can't get/list/watch nodes) with these values:
```
sync:
  nodes:
    enableScheduler: true
```
and it would be solved in the same way, by documenting that `sync.nodes.enable` should be set to true when the `sync.nodes.enableScheduler: true`

FabianKramm: @matskiv true, I'll change that!

PR #428, opened 2022-04-08
feat: sync service and endpoints without service by default

pratikjagrut:
/kind feature

resolves #

### DIFF
--- a/cmd/vcluster/cmd/start.go
+++ b/cmd/vcluster/cmd/start.go
@@ -4,10 +4,11 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"math"
 	"os"
 	"time"
+
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 
 	corev1 "k8s.io/api/core/v1"
 
@@ -17,7 +18,6 @@ import (
 	context2 "github.com/loft-sh/vcluster/cmd/vcluster/context"
 	"github.com/loft-sh/vcluster/pkg/apis"
 	"github.com/loft-sh/vcluster/pkg/controllers"
-	"github.com/loft-sh/vcluster/pkg/controllers/resources/endpoints"
 	"github.com/loft-sh/vcluster/pkg/controllers/resources/nodes"
 	"github.com/loft-sh/vcluster/pkg/controllers/resources/nodes/nodeservice"
 	translatepods "github.com/loft-sh/vcluster/pkg/controllers/resources/pods/translate"
@@ -114,8 +114,6 @@ func NewStartCommand() *cobra.Command {
 	cmd.Flags().StringSliceVar(&options.Tolerations, "enforce-toleration", []string{}, "If set will apply the provided tolerations to all pods in the vcluster")
 	cmd.Flags().StringVar(&options.NodeSelector, "node-selector", "", "If nodes sync is enabled, nodes with the given node selector will be synced to the virtual cluster. If fake nodes are used, and --enforce-node-selector flag is set, then vcluster will ensure that no pods are scheduled outside of the node selector.")
 	cmd.Flags().StringVar(&options.ServiceAccount, "service-account", "", "If set, will set this host service account on the synced pods")
-
-	cmd.Flags().BoolVar(&options.SyncServiceSelector, "sync-service-selector", false, "If enabled, vcluster will sync services with their selector instead of syncing the endpoints")
 
 	cmd.Flags().BoolVar(&options.OverrideHosts, "override-hosts", true, "If enabled, vcluster will override a containers /etc/hosts file if there is a subdomain specified for the pod (spec.subdomain).")
 	cmd.Flags().StringVar(&options.OverrideHostsContainerImage, "override-hosts-container-image", translatepods.HostsRewriteImage, "The image for the init container that is used for creating the override hosts file.")
@@ -469,18 +467,6 @@ func syncKubernetesService(ctx *context2.ControllerContext) error {
 
 		return errors.Wrap(err, "sync kubernetes service")
 	}
-
-	err = endpoints.SyncKubernetesServiceEndpoints(ctx.Context, ctx.VirtualManager.GetClient(), ctx.CurrentNamespaceClient, ctx.CurrentNamespace, ctx.Options.ServiceName)
-	if err != nil {
-		if kerrors.IsConflict(err) {
-			klog.Errorf("Error syncing kubernetes service endpoints: %v", err)
-			time.Sleep(time.Second)
-			return syncKubernetesService(ctx)
-		}
-
-		return errors.Wrap(err, "sync kubernetes service endpoints")
-	}
-
 	return nil
 }
 
--- a/e2e/test/k8sdefaultendpoint/k8sdefaultendpoint.go
+++ b/e2e/test/k8sdefaultendpoint/k8sdefaultendpoint.go
@@ -1,0 +1,56 @@
+package k8sdefaultendpoint
+
+import (
+	"context"
+	"reflect"
+
+	"github.com/loft-sh/vcluster/e2e/framework"
+	"github.com/onsi/ginkgo"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = ginkgo.Describe("map default/kubernetes endpoint to physical vcluster endpoint", func() {
+	var (
+		f *framework.Framework
+	)
+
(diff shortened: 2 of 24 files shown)


pkg/controllers/resources/endpoints/syncer.go
```diff
@@ -41,79 +36,23 @@ func (s *endpointsSyncer) Sync(ctx *synccontext.SyncContext, pObj client.Object,
...
-			newSubsets[i].Addresses[j].Hostname = ""
-			newSubsets[i].Addresses[j].NodeName = nil
-			newSubsets[i].Addresses[j].TargetRef = nil
-		}
-		for j := range pObj.Subsets[i].NotReadyAddresses {
-			newSubsets[i].NotReadyAddresses[j].Hostname = ""
-			newSubsets[i].NotReadyAddresses[j].NodeName = nil
-			newSubsets[i].NotReadyAddresses[j].TargetRef = nil
+			return true, nil
 		}
```
matskiv, line 48: I think we should return an error if it's of a different type than IsNotFound.

pkg/controllers/resources/endpoints/syncer.go
```diff
@@ -41,79 +36,23 @@ func (s *endpointsSyncer) Sync(ctx *synccontext.SyncContext, pObj client.Object,
 var _ syncer.Starter = &endpointsSyncer{}
 
 func (s *endpointsSyncer) ReconcileStart(ctx *synccontext.SyncContext, req ctrl.Request) (bool, error) {
```
matskiv, line 38: I think we are missing a skip (`return true, nil`) of the "kubernetes" endpoints resource in the "default" namespace. You, know the one we create with the "k8sdefaultendpoint" controller.

pkg/controllers/register.go
```diff
@@ -50,6 +52,7 @@ var ResourceControllers = map[string][]func(*synccontext.RegisterContext) (synce
 	"volumesnapshots":        newControllers(volumesnapshotclasses.New, volumesnapshots.New, volumesnapshotcontents.New),
 	"serviceaccounts":        newControllers(serviceaccounts.New),
 	"persistentvolumes,fake-persistentvolumes": newControllers(persistentvolumes.New),
+	"k8sdefaultendpoint":                       newControllers(k8sdefaultendpoint.New),
```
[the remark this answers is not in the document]
matskiv: I guess this would require a bit of a refactor to the k8sdefaultendpoint controller to make it a generic controller. We should probably discuss this option.

matskiv (changes requested):
I added a few comments. Please review them :)

pkg/controllers/k8sdefaultendpoint/syncer.go
```diff
@@ -0,0 +1,111 @@
...
+func (e *K8SDefaultEndpointReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("kubernetes_default_endpoint").
+		For(&corev1.Endpoints{}).
+		Watches(&source.Kind{
+			Type: &corev1.Endpoints{
+				ObjectMeta: v1.ObjectMeta{
+					Name:      "kubernetes",
+					Namespace: "default",
+				},
```
matskiv, line 43: This does not ensure that reconcile is only triggered for the `default/kubernetes`.
You need something like this:
https://github.com/loft-sh/vcluster/blob/7c5c470e87dc8224f41161b98f4a8ecee0d27ea4/pkg/controllers/coredns/nodehosts.go#L86-L90
+
https://github.com/loft-sh/vcluster/blob/7c5c470e87dc8224f41161b98f4a8ecee0d27ea4/pkg/controllers/coredns/nodehosts.go#L101
And then `.Watches()` can be removed.

matskiv: Also, because this controller is being registered with the host side manager (`controller.SetupWithManager(ctx.LocalManager)`) - it should care about Endpoint resource that is named the same as vcluster service, and is in the vcluster namespace, not the `default/kubernetes` one.

pkg/controllers/k8sdefaultendpoint/syncer.go
```diff
@@ -0,0 +1,111 @@
...
+		return ctrl.Result{RequeueAfter: time.Second}, err
+	}
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager adds the controller to the manager
+func (e *K8SDefaultEndpointReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("kubernetes_default_endpoint").
+		For(&corev1.Endpoints{}).
```
matskiv, line 37: Good improvement would be to also add a watch for the virtual side - trigger reconcile when the `default/kubernetes` Endpoint in the virtual cluster changes.
A predicated function(mentioned in my other comments) will be needed for that, and also a watch connected to the VirtualManager, here is an example of a watch connected to a different manager (not the same as manager passed to the `NewControllerManagedBy()`):
https://github.com/loft-sh/vcluster/blob/7c5c470e87dc8224f41161b98f4a8ecee0d27ea4/pkg/controllers/syncer/syncer.go#L167

matskiv (changes requested):
A few comments about new code.

Also, some new unit tests for the `pkg/controllers/resources/endpoints/syncer.go` would be good. And perhaps some test for the `k8sdefaultendpoint` controller too, maybe e2e is more suitable.

matskiv (approved):
Code looks really good :+1: 
It works as expected for k3s and k0s. I had a problem with k8s/eks, but it is not related to these changes, so I reported it here - https://github.com/loft-sh/vcluster/issues/433


It would be nice to add some tests too, but that can be done in a separate PR.

e2e/test/k8sdefaultendpoint/k8sdefaultendpoint.go
```diff
@@ -0,0 +1,56 @@
...
+	var (
+		f *framework.Framework
+	)
+
+	ginkgo.JustBeforeEach(func() {
+		// use default framework
+		f = framework.DefaultFramework
+	})
+
+	ginkgo.FIt("Test default/kubernetes endpoints matches with vcluster service endpoint", func() {
```
FabianKramm, line 22: Please remove `FIt`

cmd/vcluster/cmd/start.go
```diff
@@ -469,18 +467,6 @@ func syncKubernetesService(ctx *context2.ControllerContext) error {
 
 		return errors.Wrap(err, "sync kubernetes service")
 	}
-
-	err = endpoints.SyncKubernetesServiceEndpoints(ctx.Context, ctx.VirtualManager.GetClient(), ctx.CurrentNamespaceClient, ctx.CurrentNamespace, ctx.Options.ServiceName)
```
FabianKramm, line 473: Can we safely remove this? Why is this not needed anymore now?

FabianKramm (changes requested):
@pratikjagrut thanks for the PR! LGTM only two small things that need to be changed

cmd/vcluster/cmd/start.go
pratikjagrut: I think yes. Because the same thing is now taken care by the new controller.

pkg/controllers/resources/services/syncer.go
```diff
@@ -24,8 +24,7 @@ func New(ctx *synccontext.RegisterContext) (syncer.Object, error) {
 	return &serviceSyncer{
 		NamespacedTranslator: translator.NewNamespacedTranslator(ctx, "service", &corev1.Service{}),
 
-		syncServiceSelector: ctx.Options.SyncServiceSelector,
-		serviceName:         ctx.Options.ServiceName,
+		serviceName: ctx.Options.ServiceName,
 	}, nil
```
matskiv, line 28: You removed the assignment of the `syncServiceSelector` property here^
but it is still part of the `seviceSyncer` struct - https://github.com/loft-sh/vcluster/blob/cf1d01c2ddf042e8be66a9f82d686976caecf7c7/pkg/controllers/resources/services/syncer.go#L34
and more importantly, it is still checked during service translation, which prevents sync of the selector - https://github.com/loft-sh/vcluster/blob/cf1d01c2ddf042e8be66a9f82d686976caecf7c7/pkg/controllers/resources/services/translate.go#L12-L13
This definitely needs to be fixed, and it is what causes CoreDNS e2e tests to fail.

matskiv (changes requested):

FabianKramm (approved):

matskiv (approved):

PR #432, opened 2022-04-12
fix: sync pv size

FabianKramm:
/kind bugfix

Fixed an issue where vcluster would not sync the increased persistent volume size back to the virtual cluster


matskiv (approved):
All good :+1: Approved.

PR #436, opened 2022-04-14
refactor: recreate pvc if volumename is different

FabianKramm:
/kind enhancement

vcluster will now recreate a PVC inside the virtual cluster if there is a corresponding host PVC with a different volume name

### DIFF
--- a/pkg/controllers/resources/persistentvolumeclaims/syncer.go
+++ b/pkg/controllers/resources/persistentvolumeclaims/syncer.go
@@ -1,8 +1,12 @@
 package persistentvolumeclaims
 
 import (
+	"context"
+	"github.com/loft-sh/vcluster/pkg/controllers/resources/persistentvolumes"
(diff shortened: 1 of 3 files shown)


pkg/controllers/resources/persistentvolumeclaims/syncer.go
```diff
@@ -1,8 +1,12 @@
 package persistentvolumeclaims
 
 import (
+	"context"
+	"github.com/loft-sh/vcluster/pkg/controllers/resources/persistentvolumes"
 	synccontext "github.com/loft-sh/vcluster/pkg/controllers/syncer/context"
 	"github.com/loft-sh/vcluster/pkg/controllers/syncer/translator"
+	"github.com/pkg/errors"
```
matskiv, line 8: Nit-pick: this package is in maintenance mode, we could drop it completely if we use 
`fmt.Errorf("something, %s", err)` instead of `errors.Wrap`
but it's super minor :)

matskiv (approved):

PR #437, opened 2022-04-15
fix: disable k8s controller that reconciles default/kubernetes endpoints

### LINKED ISSUE #433: default/kubernetes endpoints has incorrect IP when using k8s or eks flavor
### What happened?

The default/kubernetes endpoints resource is constantly updated with wrong value by one of the k8s controllers when k8s or eks flavor is used. Other flavors k3s/k0s don't have this problem, which is because syncer and kube-api are in one pod, and share the same IP address.

```
$ kubectl get endpoints -n default -w
NAME         ENDPOINTS                           
kubernetes   172.17.0.21:8443,172.17.0.26:8443
kubernetes   172.17.0.15:8443,172.17.0.16:8443,172.17.0.20:8443
kubernetes   172.17.0.21:8443,172.17.0.26:8443      
kubernetes   172.17.0.15:8443,172.17.0.16:8443,172.17.0.20:8443
```

The endpoints defined in the default/kubernetes endpoints resource should always point to the syncer pods.

Install vcluster with k8s flavor and run `vcluster connect vcluster -- kubectl get endpoints -n default -w`

_No response_

<details>

```console
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.0", GitCommit:"ab69524f795c42094a6630298ff53f3c3ebab7f4", GitTreeState:"clean", BuildDate:"2021-12-07T18:16:20Z", GoVersion:"go1.17.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.4", GitCommit:"e6c093d87ea4cbb530a7b2ae91e54c0842d8308a", GitTreeState:"clean", BuildDate:"2022-02-16T12:32:02Z", GoVersion:"go1.17.7", Compiler:"gc", Platform:"linux/amd64"}
</details>


<details>

```
minikube
```

</details>


### vlcuster version

<details>

```console
I was using code from https://github.com/loft-sh/vcluster/pull/428 when reproducing this, but I am pretty sure previous version would have the same issue.
```

</details>

<details>

```
k8s and eks
```

</details>

### OS and Arch

<details>

```
OS: Linux, Fedora 35
Arch: x86
```
</details>

matskiv:
In this change, I disable the controller("EndpointsReconciler") in the kube-controller-manager which is updating the `default/kubernetes` Endpoints resource because it was updating it to an incorrect value.
Also, to ensure the expected outcome, the vclusters k8sdefaultendpoint controller has been updated to mimic the behavior of the original k8s "EndpointsReconciler" controller.

/kind bugfix

resolves #433 

Fixed an issue where `default/kubernetes` Endpoints object was referencing incorrect IPs in k8s and eks flavors.

### DIFF
--- a/charts/k0s/templates/secret.yaml
+++ b/charts/k0s/templates/secret.yaml
@@ -24,6 +24,7 @@ stringData:
         k0sApiPort: 9443
         extraArgs:
           enable-admission-plugins: NodeRestriction
+          endpoint-reconciler-type: none
       network:
         # Will be replaced automatically from the vcluster cli
         serviceCIDR: {{ .Values.serviceCIDR }}
--- a/charts/k3s/templates/statefulset.yaml
+++ b/charts/k3s/templates/statefulset.yaml
@@ -89,8 +89,10 @@ spec:
           {{- if not .Values.sync.nodes.enableScheduler }}
             --disable-scheduler
             --kube-controller-manager-arg=controllers=*,-nodeipam,-nodelifecycle,-persistentvolume-binder,-attachdetach,-persistentvolume-expander,-cloud-node-lifecycle
+            --kube-apiserver-arg=endpoint-reconciler-type=none
           {{- else }}
             --kube-controller-manager-arg=controllers=*,-nodeipam,-persistentvolume-binder,-attachdetach,-persistentvolume-expander,-cloud-node-lifecycle
+            --kube-apiserver-arg=endpoint-reconciler-type=none
             --kube-controller-manager-arg=node-monitor-grace-period=1h
             --kube-controller-manager-arg=node-monitor-period=1h
           {{- end }}
--- a/pkg/controllers/k8sdefaultendpoint/k8sdefaultendpoint.go
+++ b/pkg/controllers/k8sdefaultendpoint/k8sdefaultendpoint.go
@@ -2,17 +2,20 @@ package k8sdefaultendpoint
 
 import (
 	"context"
-	"encoding/json"
 	"time"
 
 	"github.com/loft-sh/vcluster/pkg/util/loghelper"
 	corev1 "k8s.io/api/core/v1"
+	discovery "k8s.io/api/discovery/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	utilnet "k8s.io/utils/net"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/source"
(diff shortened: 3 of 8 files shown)


charts/k0s/templates/secret.yaml
```diff
@@ -36,5 +36,6 @@ stringData:
           controllers: '*,-nodeipam,-persistentvolume-binder,-attachdetach,-persistentvolume-expander,-cloud-node-lifecycle'
           node-monitor-grace-period: 1h
           node-monitor-period: 1h
+          endpoint-reconciler-type: none
```
FabianKramm, line 39: Should be enabled regardless of the scheduler shouldn't it?

pkg/controllers/k8sdefaultendpoint/k8sdefaultendpoint.go
```diff
@@ -74,51 +80,155 @@ func syncKubernetesServiceEndpoints(ctx context.Context, virtualClient client.Cl
...
+		oldJSON, err := json.Marshal(vEndpoints.Subsets)
+		if err != nil {
+			return err
+		}
+		newJSON, err := json.Marshal(newSubsets)
+		if err != nil {
+			return err
+		}
+
+		if string(oldJSON) == string(newJSON) {
```
FabianKramm, line 123: Does this skip the patch?

pkg/controllers/k8sdefaultendpoint/k8sdefaultendpoint.go
```diff
@@ -74,51 +80,155 @@ func syncKubernetesServiceEndpoints(ctx context.Context, virtualClient client.Cl
...
+			Namespace: "default",
+			Name:      "kubernetes",
+		},
+	}
+
+	result, err := controllerutil.CreateOrPatch(ctx, virtualClient, vEndpoints, func() error {
+		if vEndpoints.Labels == nil {
+			vEndpoints.Labels = map[string]string{}
+		}
+		vEndpoints.Labels[discovery.LabelSkipMirror] = "true"
```
FabianKramm, line 94: Why do we need that?

pkg/controllers/k8sdefaultendpoint/k8sdefaultendpoint.go
```diff
@@ -74,51 +80,155 @@ func syncKubernetesServiceEndpoints(ctx context.Context, virtualClient client.Cl
...
+		vEndpoints.Subsets = newSubsets
+
+		return nil
+	})
+	if err != nil {
+		return nil
+	}
+
+	if result == controllerutil.OperationResultCreated || result == controllerutil.OperationResultUpdated {
+		vSlices := &discovery.EndpointSlice{
```
FabianKramm, line 136: Why do we need to do that manually instead of using the mirror controller from k8s?

charts/k3s/templates/statefulset.yaml
```diff
@@ -93,6 +93,7 @@ spec:
             --kube-controller-manager-arg=controllers=*,-nodeipam,-persistentvolume-binder,-attachdetach,-persistentvolume-expander,-cloud-node-lifecycle
             --kube-controller-manager-arg=node-monitor-grace-period=1h
             --kube-controller-manager-arg=node-monitor-period=1h
+            --kube-controller-manager-arg=endpoint-reconciler-type=none
```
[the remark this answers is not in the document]
matskiv: Yes, I'll fix that.

pkg/controllers/k8sdefaultendpoint/k8sdefaultendpoint.go
matskiv: I am just doing exactly what k8s does - https://github.com/kubernetes/kubernetes/blob/7380fc735aca591325ae1fabf8dab194b40367de/pkg/controlplane/reconcilers/lease.go#L200-L207
the mirror controller would probably work, but I am not 100% sure, so I went with this as a safer option.

pkg/controllers/k8sdefaultendpoint/k8sdefaultendpoint.go
matskiv: That's what k8s does - https://github.com/kubernetes/kubernetes/blob/7380fc735aca591325ae1fabf8dab194b40367de/pkg/controlplane/reconcilers/endpointsadapter.go#L206

pkg/controllers/k8sdefaultendpoint/k8sdefaultendpoint.go
matskiv: No. I didn't realize that when copying this piece of code :man_facepalming: 
The controllerutil.CreateOrPatch will skip the Patch call if there is no difference, so I'll just remove this^ part.

FabianKramm (approved):

PR #440, opened 2022-04-18
Allow setting imagePullSecrets for coredns

mtougeron:
/kind bugfix
/kind enhancement

Allows setting imagePullSecrets for the coredns ServiceAccount when using a private repository via the `defaultImageRegistry` value.

Fixed an issue where coredns wouldn't start due to missing imagePullSecrets when using a private repository

### DIFF
--- a/charts/eks/templates/coredns.yaml
+++ b/charts/eks/templates/coredns.yaml
@@ -10,6 +10,10 @@ data:
     metadata:
       name: coredns
       namespace: kube-system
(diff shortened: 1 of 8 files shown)


mtougeron:
Let me know if you'd rather use the existing `serviceAccount.imagePullSecrets` instead of one specific to coredns. I was on the fence for whether or not it needed its own imagePullSecrets

mtougeron:
Nevermind, closing this. Just realized it won't work the way I thought it would. I'll need to rethink this...

PR #445, opened 2022-04-20
feat: update service accounts to accommodate imagePullSecrects

### LINKED ISSUE #439: imagePullSecret or mounting secret volume should be available when using different distros to access private registries

Hi Everyone,

This is the situation I'm facing.
I've loaded all the vcluster images in to my Artifactory using the supplied scripts in the releases.
Next I've supplied in the field:
`defaultImageRegistry: myrepo.jfrog.io/docker`

the issue being: when running the vcluster create command like this:
`vcluster create danny-1 --namespace danny-1 -f values.yaml -f vcluster_args.yaml  --distro k8s --expose --debug`

It throws an error that it cannot pull the image - and it does so in the pre-install-hook-job.
https://github.com/loft-sh/vcluster/blob/main/charts/k8s/templates/pre-install-hook-job.yaml

since this job pulls the vcluster:0.7.1 image, it tries to do so using the supplied Registry and fails.
because Artifactory expects a login with credentials.
I've tried to pass volumes and extraArgs but it will not help

`vcluster:
  extraArgs:
    - imagePullSecrets: regcred
  volumes:
  - name: jfrog
    secret:
      secretName: regred
syncer:
  extraArgs:
    - imagePullSecrets: regcred`

simply because its not supported in the pre-install-hook-job.yaml.

this is currently the feature i find missing and would help a lot in AirGapped environments as well.

Allow to pass ImagePullSecrets or mount volumes to the Job from the values.yaml file.

_No response_

### Additional context

_No response_

pratikjagrut:
/kind feature

resolves #439 

Fixed an issue where vcluster now will be able to pull images from a private registry.

### DIFF
--- a/charts/eks/templates/pre-install-hook-serviceaccount.yaml
+++ b/charts/eks/templates/pre-install-hook-serviceaccount.yaml
@@ -8,4 +8,8 @@ metadata:
     "helm.sh/hook": pre-install
     "helm.sh/hook-weight": "1"
     "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+{{- if .Values.serviceAccount.imagePullSecrets }}
+imagePullSecrets:
+{{ toYaml .Values.serviceAccount.imagePullSecrets | indent 2 }}
+{{- end }}
 {{- end }}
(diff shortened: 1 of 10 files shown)


matskiv (approved):
Looks good :+1:

FabianKramm (approved):

PR #448, opened 2022-04-22
refactor: add daemon set annotation

FabianKramm:
/kind enhancement

vcluster will now set the annotation `cluster-autoscaler.kubernetes.io/daemonset-pod` on pods that belong to a daemon set inside the virtual cluster


matskiv (approved):
Looks good :+1:

PR #449, opened 2022-04-25
Add init manifest

### LINKED ISSUE #443: Ability to initialize vcluster with manifests

Currently there's no way for a user to define manifests to initialize a new `vcluster`.

We would also like to have the ability to reflect any changes made to these manifests on detection of which we re-apply the manifests.

The following enhancement proposal is suggested.

#### Goals
1. Ability to deploy resources on startup.
2. Ability to reload / re-apply any changes made to these resources.

#### Non-Goals
1. Initial iteration does not target on specifying helm charts as an argument.
2. Initial iteration does not target consuming upstream URL's to k8s manifests as an input argument.

_No response_

### Additional context

A typical use case can be a User wanting to create a new `vcluster` with some deployments like
* `nginx-ingress` as an ingress controller
* `knative` as a base layer for FaaS style development.

Since currently this is not directly supported in vcluster, User are supposed to follow roughly the following path:
1. Create a shell script that initially creates the `vcluster` itself.
2. On successful creation get the cluster (vcluster) credentials.
3. Using the obtained vcluster `kubeconfig` apply additional resource either through
    * `kubectl apply -f ...`
    * `helm install ...`
    * or Other custom tools like `istioctl`, `linkerd` etc
    
This enhancement proposes to standardize this flow to a more managable and declarative alternative with the following ideas:
1. Allow an optional `init` section in the supplies `values.yaml` file given as input to the `vcluster` cli.
2. The `init.manifests` section takes as an input a _multiline string_ which is simply the `yaml` manifest of what needs to be created/applied right after successful `vcluster` creation.
3. This section is then created as a `configmap`/`secret` in the host cluster and mounted at a designated mountpoint inside the `vcluster` – say `/etc/opt/init-manifests`
4. These manifests are then applied in much similar way as currently being done in the case of [Core DNS Manifests](https://github.com/loft-sh/vcluster/blob/main/cmd/vcluster/cmd/start.go#L354)
5. As an additional step, we also watch/pool this mountpoint and watch for any changes to the manifests, in which case, we trigger a reapply of these manifests.

As an additional step, we would like to have `helm`-style template blocks `{{ ... }}` in the `.Values.init.manifests` section that allows these resources to take advantage of inputs like `Release.Name`, `Release.Namespace` etc.

ishankhare07:
/kind enhancement
/kind feature

resolves #443 


charts/eks/templates/etcd-statefulset.yaml
```diff
@@ -73,6 +73,11 @@ spec:
         - name: data
           emptyDir: {}
       {{- end }}
+      {{- if .Values.init.manifests }}
```
FabianKramm, line 76: I think this is applied in the wrong file, you need to add this to the syncer-deployment

charts/k8s/templates/etcd-statefulset.yaml
```diff
@@ -73,6 +73,11 @@ spec:
         - name: data
           emptyDir: {}
       {{- end }}
+      {{- if .Values.init.manifests }}
```
FabianKramm, line 76: Please add this to the syncer deployment as etcd cannot do anything with the manifests

pkg/controllers/register.go
```diff
@@ -165,6 +172,21 @@ func RegisterControllers(ctx *context.ControllerContext, syncers []syncer.Object
 	return nil
 }
 
+func registerInitManifestsController(ctx *context.ControllerContext) error {
+	controller := &manifests.InitManifestsConfigMapReconciler{
+		Client:   ctx.LocalManager.GetClient(),
+		Log:      loghelper.New("initmanifests-controller"),
+		VManager: ctx.VirtualManager,
```
FabianKramm, line 179: Please rename VManager to `VirtualManager` as we are using it like this everywhere else

pkg/manifests/init.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/loft-sh/vcluster/pkg/util/applier"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	InitManifestRelativePath = "init/initmanifests.yaml"
+	DEFAULT_NAMESPACE        = "default"
```
FabianKramm, line 27: Please use `corev1.NamespaceDefault` instead of this constant

pkg/controllers/manifests/init.go
```diff
@@ -0,0 +1,71 @@
...
+		if kerrors.IsNotFound(err) {
+			r.Log.Errorf("configmap not found %v", err)
+			return ctrl.Result{}, err
+		}
+
+		return ctrl.Result{}, err
+	}
+
+	// TODO: implement better filteration through predicates
+	if !strings.Contains(cm.ObjectMeta.Name, InitManifestSuffix) {
```
FabianKramm, line 44: Instead of using the suffix here, please use the name which you can get via the ctx.Options.Name in the register function. Otherwise this might cause problems if multiple vclusters are deployed in the same namespace

pkg/controllers/manifests/init.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/loft-sh/vcluster/pkg/util/loghelper"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	InitManifestSuffix      = "-init-manifests"
+	ConfigMapDataKey        = "initmanifests.yaml"
+	DefaultNamespaceIfEmpty = "default"
```
FabianKramm, line 20: Please use `corev1.NamespaceDefault` instead

pkg/controllers/manifests/init.go
```diff
@@ -0,0 +1,71 @@
...
+		return ctrl.Result{}, err
+	}
+
+	// TODO: implement better filteration through predicates
+	if !strings.Contains(cm.ObjectMeta.Name, InitManifestSuffix) {
+		// skip
+		return ctrl.Result{}, nil
+	}
+
+	manifests, ok := cm.Data[ConfigMapDataKey]
```
FabianKramm, line 49: Instead of failing here / expecting a constant key, just loop over the data map and join all values with `\n---\n`

pkg/manifests/init.go
```diff
@@ -0,0 +1,203 @@
...
+
+	lastAppliedObjects util.UnstructuredMap
+	ErrorEmptyManifest = errors.New("empty manifest file")
+)
+
+func init() {
+	lastAppliedObjects = make(util.UnstructuredMap)
+}
+
+func ApplyInitManifests(inClusterConfig *rest.Config, namespace string) error {
```
FabianKramm, line 41: This function is never used

pkg/manifests/init.go
```diff
@@ -0,0 +1,203 @@
...
+
+		delete(lastAppliedObjects, util.UnstructuredToKObject(*obj))
+	}
+
+	lastAppliedObjects = currentlyApplyingObjects
+	klog.Infof("successfully applied all init manifest objects")
+	return nil
+}
+
+func processManifestTemplate(vars map[string]interface{}) ([]byte, error) {
```
FabianKramm, line 122: Function is also never used

pkg/manifests/init.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	if bytes.Equal(contents, []byte("---")) {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func ChangeDetected(c client.Client, namespace string) (bool, error) {
```
FabianKramm, line 170: Same here

pkg/manifests/init.go
```diff
@@ -0,0 +1,203 @@
...
+	currentHash := hexHasher(content)
+	klog.Info("current", string(content), LAST_APPLIED_MANIFEST_HASH)
+	if currentHash != LAST_APPLIED_MANIFEST_HASH {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func updateLastApplied(path string, lastAppliedManifest []byte) {
```
FabianKramm, line 193: Same here

pkg/manifests/init.go
```diff
@@ -0,0 +1,203 @@
...
+	content, _ := os.ReadFile(path)
+	LAST_APPLIED_MANIFEST_HASH = hexHasher(content)
+
+	// LAST_APPLIED_MANIFESTS = hex.EncodeToString(lastAppliedManifest[:])
+}
+
+func hexHasher(input []byte) string {
```
FabianKramm, line 200: Same here

pkg/manifests/init.go
```diff
@@ -0,0 +1,203 @@
...
+
+var (
+	LAST_APPLIED_MANIFEST_HASH string
+
+	lastAppliedObjects util.UnstructuredMap
+	ErrorEmptyManifest = errors.New("empty manifest file")
+)
+
+func init() {
+	lastAppliedObjects = make(util.UnstructuredMap)
```
FabianKramm, line 38: Please do not save this globally here and instead save it in a configmap inside the virtual cluster. Rather handle this in the controller though instead of this helper file.

pkg/manifests/init.go
```diff
@@ -0,0 +1,203 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	InitManifestRelativePath = "init/initmanifests.yaml"
+	DEFAULT_NAMESPACE        = "default"
+)
+
+var (
+	LAST_APPLIED_MANIFEST_HASH string
```
FabianKramm, line 31: Not used

pkg/manifests/init.go
```diff
@@ -0,0 +1,203 @@
...
+
+	lastAppliedObjects util.UnstructuredMap
+	ErrorEmptyManifest = errors.New("empty manifest file")
```
FabianKramm, line 34: Not used

FabianKramm (changes requested):
@ishankhare07 thanks a lot for the PR! I reviewed it and it looks good so far, but there are a couple of things we need to change. Please check my comments below.

FabianKramm (approved):

PR #450, opened 2022-04-26
fix: refresh managed endpoints

FabianKramm:
/kind bugfix

Fixed an issue where vcluster would not migrate managed endpoints to Kubernetes managed endpoints correctly.


matskiv (approved):
LGTM :+1: 
I especially like the detailed explanation in the comment, thanks for that.

PR #453, opened 2022-04-27
feat: Support CoreDNS service type configuration

olljanat:
/kind feature

Allow publish CoreDNS as load balancer service.

chart: coredns is service type and external IP can be now easily overridden through helm values

Here is full on prem example where both syncer and coredns are configured to use DNS domain `dev.local` instead of default `cluster.local`, CoreDNS is published as LoadBalancer service (so external systems can use it as forwarder for this DNS domain), Metal LB is instructed to use [address pool ip-pool](https://metallb.universe.tf/usage/#requesting-specific-ips) and Calico to do [service BGP advertisement with /32 mask](https://projectcalico.docs.tigera.io/networking/advertise-service-ips#advertising-service-ips-quick-glance)
```yaml
syncer:
  replicas: 1
  extraArgs:
    - --cluster-domain=dev.local

coredns:
  enabled: true
  replicas: 1
  config: |-
    .:1053 {
        errors
        health
        ready
        kubernetes dev.local in-addr.arpa ip6.arpa {
          pods insecure
          fallthrough in-addr.arpa ip6.arpa
        }
        hosts /etc/coredns/NodeHosts {
          ttl 60
          reload 15s
          fallthrough
        }
        prometheus :9153
        forward . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
  service:
    type: LoadBalancer
    externalTrafficPolicy: "Local"
    annotations:
      metallb.universe.tf/address-pool: ip-pool
```

### DIFF
--- a/charts/k0s/templates/coredns.yaml
+++ b/charts/k0s/templates/coredns.yaml
@@ -218,6 +218,9 @@ data:
       annotations:
         prometheus.io/port: "9153"
         prometheus.io/scrape: "true"
+        {{- if .Values.coredns.service.annotations }}
+{{ toYaml .Values.coredns.service.annotations | indent 8 }}
+        {{- end }}
       labels:
         k8s-app: kube-dns
         kubernetes.io/cluster-service: "true"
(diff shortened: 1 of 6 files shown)


FabianKramm (approved):
@olljanat thanks a lot for the PR! Looks good, makes sense to make that configurable!

PR #454, opened 2022-04-28
Fix 3 conformance tests - Service type change + default/kubernetes EndpointSlices

matskiv:
/kind bugfix

Fixes following conformance tests:
```
[sig-network] DNS [It] should provide DNS for ExternalName services
[sig-network] Services [It] should be able to change the type from ExternalName to ClusterIP
[sig-network] Services [It] should be able to change the type from ExternalName to NodePort
[sig-network] EndpointSlice [It] should have Endpoints and EndpointSlices pointing to API Server
```

fix: Service type change from ExternalName type to other types
fix: default/kubernetes EndpointSlices content should match default/kubernetes Endpoints - add the missing ports

Added unit tests are not related to the fix because originally I thought that the problem is elsewhere, but I decided to keep them anyway.


FabianKramm (approved):

PR #458, opened 2022-04-29
refactor: translate certain ingress annotations

### LINKED ISSUE #351: Nginx BasicAuth secret not synchronized
vcluster 0.5.3 does not appear to synchronize [nginx.ingress.kubernetes.io/auth-secret secrets](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#authentication).

This annotation is of course specific to the nginx ingress controller. Is this still wanted in core, or rather something for a plugin?

FabianKramm:
/kind bugfix

resolves #351 

vcluster will now rewrite certain ingress annotations

### DIFF
--- a/pkg/controllers/resources/ingresses/syncer.go
+++ b/pkg/controllers/resources/ingresses/syncer.go
@@ -4,10 +4,12 @@ import (
 	"github.com/loft-sh/vcluster/pkg/controllers/syncer"
 	synccontext "github.com/loft-sh/vcluster/pkg/controllers/syncer/context"
 	"github.com/loft-sh/vcluster/pkg/controllers/syncer/translator"
+	"github.com/loft-sh/vcluster/pkg/util/translate"
 	networkingv1 "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


pkg/controllers/resources/ingresses/syncer.go
```diff
@@ -60,10 +62,42 @@ func (s *ingressSyncer) Sync(ctx *synccontext.SyncContext, pObj client.Object, v
...
 	return translator.UniqueSlice(secrets)
 }
+
+var TranslateAnnotations = map[string]bool{
+	"nginx.ingress.kubernetes.io/auth-secret":      true,
+	"nginx.ingress.kubernetes.io/auth-tls-secret":  true,
+	"nginx.ingress.kubernetes.io/proxy-ssl-secret": true,
+}
+
+func translateAnnotations(annotations map[string]string, currentNamespace string) (map[string]string, []string) {
```
matskiv, line 81: nit-pick: `currentNamespace` variable name is slightly confusing, because AFAIK that is how we usually refer to the namespace where we create pods, etc. in the host cluster. Perhaps `ingressNamespace` would be better :)

matskiv (approved):
LGTM :+1:

pkg/controllers/resources/ingresses/syncer.go
FabianKramm: oh true, let me fix that!

PR #460, opened 2022-05-02
[Helm K3s] Add CoreDNS placement options as well as Service annotations / labels

salanki:
/kind enhancement

- Adds nodeSelector / affinity / toleration for CoreDns
- Adds ability to set label and annotation on the APIServer Service

CoreDNS will inherit the same nodeSelectors as the API Server
Labels and Annotations can now be set on the API Server service via values

It might make sense to let CoreDNs have it's own affinities, under the `coredns` section instead of inheriting the global ones. Went with global since they were not scoped under `vcluster`.

### DIFF
--- a/charts/k3s/templates/coredns.yaml
+++ b/charts/k3s/templates/coredns.yaml
@@ -121,6 +121,12 @@ data:
           priorityClassName: "system-cluster-critical"
           serviceAccountName: coredns
           nodeSelector:
+{{ toYaml .Values.nodeSelector | indent 12 }}
+          affinity:
+{{ toYaml .Values.affinity | indent 12 }}
+          tolerations:
+{{ toYaml .Values.tolerations | indent 12 }}
+          nodeSelector:
             kubernetes.io/os: linux
           topologySpreadConstraints:
             - maxSkew: 1
--- a/charts/k3s/values.yaml
+++ b/charts/k3s/values.yaml
@@ -194,6 +194,10 @@ service:
   # Configuration for LoadBalancer service type
   externalIPs: []
   externalTrafficPolicy: ""
+  # Extra Labels for the service
+  labels: {}
+  # Extra Annotations for the service
+  annotations: {}
 
 # Configure the ingress resource that allows you to access the vcluster
 ingress:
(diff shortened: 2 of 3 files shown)


charts/k3s/templates/coredns.yaml
```diff
@@ -121,6 +121,12 @@ data:
           priorityClassName: "system-cluster-critical"
           serviceAccountName: coredns
           nodeSelector:
+{{ toYaml .Values.nodeSelector | indent 12 }}
```
FabianKramm, line 124: I think we should do `.Values.coredns.nodeSelector` as people might not always want to have the same node selector for vcluster as well as coredns

charts/k3s/templates/coredns.yaml
```diff
@@ -121,6 +121,12 @@ data:
...
+          affinity:
+{{ toYaml .Values.affinity | indent 12 }}
```
FabianKramm, line 126: Same here

charts/k3s/templates/coredns.yaml
```diff
@@ -121,6 +121,12 @@ data:
...
+          tolerations:
```
FabianKramm, line 127: Same here

charts/k3s/values.yaml
```diff
@@ -194,6 +194,10 @@ service:
   # Configuration for LoadBalancer service type
   externalIPs: []
   externalTrafficPolicy: ""
+  # Extra Labels for the service
+  labels: {}
+  # Extra Annotations for the service
+  annotations: {}
```
FabianKramm, line 200: besides adding this to the k3s chart would be great if you could add this to the k0s and k8s chart as well

FabianKramm (changes requested):
@salanki thanks for creating this PR! Would be great if you could change the parts I mentioned in my review below then we can merge this.

FabianKramm:
@salanki thanks for the update, could you also add these changes to the k0s, eks and k8s as otherwise this would only be configurable for the k3s chart. Besides that you will need to rebase your branch to the current main branch as there is a conflict in the k3s/values.yaml and the e2e tests are failing. Thanks!

heiko-braun:
Outdated

PR #472, opened 2022-05-10
Adding init-manifests content

jschalz:
/kind documentation

Adding documentation for applying manifests at init, as introduced in v0.8.0.

### DIFF
--- a/docs/pages/operator/init-manifests.mdx
+++ b/docs/pages/operator/init-manifests.mdx
@@ -5,7 +5,7 @@ sidebar_label: Applying manifests on init
 
 Starting in version 0.8.0, vcluster allows users to apply manifests as soon as the vcluster is started. This can be useful for users configuring internal vcluster resources.
 
-This can be configured in under `helm` values:
+This can be configured via `helm` values:
 
 ```yaml
 init:
--- a/docs/pages/operator/init-manifests.mdx
+++ b/docs/pages/operator/init-manifests.mdx
@@ -1,0 +1,20 @@
+---
+title: Applying manifests on initialization
+sidebar_label: Applying manifests on init
+---
+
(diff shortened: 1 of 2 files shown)


FabianKramm:
@LukasGentele

docs/pages/operator/init-manifests.mdx
```diff
@@ -0,0 +1,20 @@
+---
+title: Applying manifests on initialization
+sidebar_label: Applying manifests on init
+---
+
+Starting in version 0.8.0, vcluster allows users to apply manifests as soon as the vcluster is started. This can be useful for users configuring internal vcluster resources.
+
+This can be configured in under `helm` values:
```
LukasGentele, line 8: "in under" => "via"

LukasGentele (dismissed):
Fine for now but we should start thinking about a better structure for vcluster docs. Putting everything under "Operator Guide" is not a great idea. We'll end up with a list of 100 pages under a single sidebar category. I'd prefer to have values.yaml (which effectively expresses features) and flags as separate pages somehow but we should discuss that outside this PR.

docs/pages/operator/init-manifests.mdx
jschalz: > Fine for now but we should start thinking about a better structure for vcluster docs. Putting everything under "Operator Guide" is not a great idea. We'll end up with a list of 100 pages under a single sidebar category. I'd prefer to have values.yaml (which effectively expresses features) and flags as separate pages somehow but we should discuss that outside this PR.

Agreed! I put it here for now since there's not a great place for them elsewhere, but I'd love to restructure the vcluster docs so it has a proper home.

FabianKramm (approved):

PR #473, opened 2022-05-10
add plugin hooks

### LINKED ISSUE #465: vcluster plugin hooks
Currently vcluster lets you write plugin and create completely separate controllers and syncers with the plugin SDK. However there are many use cases that do not want to exchange a syncer or create a new one from scratch and rather want to change syncing behaviour for existing syncers. We should introduce hooks that allows plugins to modify outgoing objects similar to an admission controller.

FabianKramm:
/kind feature

resolves #465

Adds plugin hooks that allow plugins to modify existing syncer behaviour. Hooks can overwrite client functions and essentially can be used similar to mutating admission controllers in the host cluster. In comparison to admission controllers, plugin hooks can also overwrite the get / list functionality of a client which allows a plugin to let vcluster view objects differently from how they are actually stored. This allows plugins to trigger certain behaviour from vcluster by either hiding or modifying certain fields.

Also a new annotation `vcluster.loft.sh/force-sync` is added for secrets and configmaps to force syncing for these resources to the host cluster. This is also interesting combined with the plugin hooks when you need to force sync a secret for example that wouldn't be synced by default from vcluster.


pkg/plugin/remote/plugin.proto
```diff
@@ -10,12 +10,44 @@ service PluginInitializer {
...
+    bool mutated = 2;
+}
+
 message LeaderInfo {
     bool leader = 1;
 }
 
 message PluginInfo {
     string name = 1;
+    string version = 2;
```
matskiv, line 50: :bulb: passing plugin SDK version here might come in handy at some point :thinking:

PR #476, opened 2022-05-13
Add 6443 port to controle plane egress NetworkPolicy

### LINKED ISSUE #475: Consider adding port 6443 to default NetworkPolicy

We are deploying `vcluster` using Terraform and use `6443` (default) for the host cluster API server:

```yaml
apiVersion: kubeadm.k8s.io/v1beta2
kind: InitConfiguration
localAPIEndpoint:
  bindPort: 6443
```

However, in `isolation.enabled=true` mode, a `NetworkPolicy` is applied that restricts Egress to ports `443` and `8443`:
https://github.com/loft-sh/vcluster/blob/main/charts/k8s/templates/networkpolicy.yaml#L54

This prevents the vcluster from coming up at all. 

We currently patch the NetworkPolicy, and work around Terraform to let the `vcluster` deployment succeed with the default. However, I am curious as to why `6443` is not currently used as a default, next to `443` and `8443`. This would save some additional work, but perhaps there are good reasons not to include that port by default.

Next to that, it would be nice to expose the ports in the Helm chart (currently only `outgoingConnections` for workloads are available in the chart).

Adding `6443` as default, and/or exposing the ports through the Helm chart.

_No response_

### Additional context

_No response_

matskiv:
/kind enhancement

resolves #475

Allows vcluster isolation NetworkPolicy to be used in a host cluster that uses 6443 port for the API server.

I decided to go with a minimal fix - adding a new port to the list, so as not to go crazy with customizations in the helm chart. If more modifications of the NetworkPolicy are required, the user can always create it separately and disable the creation via our helm chart.

### DIFF
--- a/charts/k0s/templates/networkpolicy.yaml
+++ b/charts/k0s/templates/networkpolicy.yaml
@@ -49,11 +49,12 @@ spec:
       release: {{ .Release.Name }}
   egress:
     # Allows outgoing connections to all pods with
-    # port 443 or 8443. This is needed for host Kubernetes
+    # port 443, 8443 or 6443. This is needed for host Kubernetes
     # access
     - ports:
         - port: 443
         - port: 8443
+        - port: 6443
     # Allows outgoing connections to all vcluster workloads
     # or kube system dns server
     - to:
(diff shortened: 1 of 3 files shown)


FabianKramm (approved):

PR #495, opened 2022-05-27
helmchart update with priorityClassName

### LINKED ISSUE #487: Update helm chart with priorityClassName section

The problem appears when you need to define priorityClassName value to make an order of vcluster pods rolling out. Currently it can't be done via helm chart's values. It is necessary for instance, when you have k8s vcluster distribution and want to run etcd pods first. Otherwise other pods like syncer or api will be restarted several times until etcd pods are ready. Also it is convenient to set a priority of vcluster pods inside of the host cluster. For example to increase priority of vcluster pods compared to the optional infrastructure pods.

Apply priorityClassName parameter at least to syncer, api, controller deployments, etcd stateful set and pre-install job and make it configurable separately via helm values.

_No response_

### Additional context

I think that priorityClassName is commonly used parameter among helm charts and I'm sure that it should simplify vcluster using in the future anyway.

I may create a pull request also, if it's OK for you please tell me.
_No response_

tukobadnyanoba:
/kind feature
resolves #487 

Fixed an issue where vcluster now supports priorityClassName in syncer, api, controller deployments, etcd stateful set and pre-install job through helm chart and separate configuration

### DIFF
--- a/charts/eks/templates/api-deployment.yaml
+++ b/charts/eks/templates/api-deployment.yaml
@@ -60,6 +60,9 @@ spec:
       {{- if .Values.api.volumes }}
 {{ toYaml .Values.api.volumes | indent 8 }}
       {{- end }}
+      {{- if .Values.api.priorityClassName }}
+      priorityClassName: {{ .Values.api.priorityClassName }}
+      {{- end }}
       containers:
       - name: kube-apiserver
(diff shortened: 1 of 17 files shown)


FabianKramm (approved):
@tukobadnyanoba thanks for the PR! LGTM!

PR #498, opened 2022-05-27
refactor: improve secret usage detection

FabianKramm:
/kind enhancement

Uses an index to remember what secrets belong to which pod. This should greatly improve perforamance for large vclusters as not every secrets needs to reiterate over every pod.


matskiv (approved):
Nice optimization :+1: :clap:

PR #500, opened 2022-05-31
Add coredns.resources property to externalise coredns config

### LINKED ISSUE #488: Extend coredns helm template

The problem appears when you need to roll out default coredns deployment using image from private registry, for instance because of strict company requirements. In this case you need to define imagePullSecret section for coredns deployment, but it can not be done via Helm values. Recently imagePullSecret section was added to API, controller deployments, etcd stateful using service account settings. It would be perfect if it is also available for coredns deployment.

Also it would be great to make resources section of coredns deployment configurable via Helm values. With default values it can lead to conflict with MaxLimitRequestRatio limit range installed inside a vcluster.

Update Helm coredns template with imagePullSecret parameter, make coredns deployment resources section configurable via Helm values.

_No response_

### Additional context

I may create a pull request also, if it's OK for you please tell me.

_No response_

tukobadnyanoba:
/kind feature
resolves #488 

Fixed an issue where vcluster ... now supports coredns configuration to be pushed from helm chart

### DIFF
--- a/charts/eks/values.yaml
+++ b/charts/eks/values.yaml
@@ -212,7 +212,7 @@ api:
 coredns:
   image: public.ecr.aws/eks-distro/coredns/coredns:v1.8.4-eks-1-21-8
   replicas: 1
-
+  resources: ""
 # Service account that should be used by the vcluster
 serviceAccount:
   create: true
--- a/charts/k0s/templates/coredns.yaml
+++ b/charts/k0s/templates/coredns.yaml
@@ -5,6 +5,10 @@ metadata:
   name: {{ .Release.Name }}-coredns
   namespace: {{ .Release.Namespace }}
 data:
+{{- if .Values.coredns.resources }}
+  coredns.yaml: |-
+{{ .Values.coredns.resources | indent 4 }}
+{{- else }}
   coredns.yaml: |-
     apiVersion: v1
     kind: ServiceAccount
@@ -252,4 +256,5 @@ data:
         - name: metrics
           port: 9153
           protocol: TCP
+{{- end }}
 {{- end }}
--- a/charts/k3s/templates/coredns.yaml
+++ b/charts/k3s/templates/coredns.yaml
@@ -5,6 +5,10 @@ metadata:
   name: {{ .Release.Name }}-coredns
   namespace: {{ .Release.Namespace }}
 data:
+{{- if .Values.coredns.resources }}
+  coredns.yaml: |-
+{{ .Values.coredns.resources | indent 4 }}
+{{- else }}
   coredns.yaml: |-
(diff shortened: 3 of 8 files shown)


charts/eks/values.yaml
```diff
@@ -208,7 +208,7 @@ api:
 coredns:
   image: public.ecr.aws/eks-distro/coredns/coredns:v1.8.4-eks-1-21-8
   replicas: 1
-
+  resources:
```
FabianKramm, line 211: While this works, please put an empty string here instead as it is more readable

charts/k3s/templates/coredns.yaml
```diff
@@ -62,7 +66,7 @@ data:
     data:
       Corefile: |
         {{- if .Values.coredns.config }}
-{{ .Values.coredns.config | indent 8 }}
+   {{ .Values.coredns.config | indent 8 }}
```
FabianKramm, line 69: This probably now causes issues as you are indenting the first line extra

charts/k8s/templates/coredns.yaml
```diff
@@ -219,7 +223,7 @@ data:
         prometheus.io/port: "9153"
         prometheus.io/scrape: "true"
         {{- if .Values.coredns.service.annotations }}
-{{ toYaml .Values.coredns.service.annotations | indent 8 }}
+   {{ toYaml .Values.coredns.service.annotations | indent 8 }}
```
FabianKramm, line 226: This probably now causes issues as you are indenting the first line extra

charts/k0s/templates/coredns.yaml
```diff
@@ -219,7 +223,7 @@ data:
         prometheus.io/port: "9153"
         prometheus.io/scrape: "true"
         {{- if .Values.coredns.service.annotations }}
-{{ toYaml .Values.coredns.service.annotations | indent 8 }}
+   {{ toYaml .Values.coredns.service.annotations | indent 8 }}
```
FabianKramm, line 226: This probably now causes issues as you are indenting the first line extra

charts/k0s/templates/coredns.yaml
```diff
@@ -62,7 +66,7 @@ data:
     data:
       Corefile: |
         {{- if .Values.coredns.config }}
-{{ .Values.coredns.config | indent 8 }}
+   {{ .Values.coredns.config | indent 8 }}
```
FabianKramm, line 69: This probably now causes issues as you are indenting the first line extra

FabianKramm (changes requested):
@tukobadnyanoba thanks for the PR! Please see my comments below

FabianKramm (approved):

PR #526, opened 2022-06-07
Allow context name as a flag to connect command for generated config

### LINKED ISSUE #523: Allow context name in generated kube config to be specified

`vcluster connect` creates a _kubeconfig.yaml_ with a context named `Default`:

```
...
contexts:
- context:
    cluster: local
    namespace: default
    user: user
  name: Default
current-context: Default
...
```

I'd like to merge this with my existing _kubeconfig.yaml_ which already has a context named `default`

An option to specify the generated context name, eg:
```
vcluster connect vcluster-1 --kube-config ~/.kube/vcluster.yaml --kube-config-context-name my-awesome-vcluster
```

Manually modifying the generated kubeconfig yaml

### Additional context

_No response_

tukobadnyanoba:
/kind feature

resolves #523 

Fixed an issue where vcluster ... connect command has flag to pass on the context-name which will overwrite the default context

### DIFF
--- a/cmd/vclusterctl/cmd/connect.go
+++ b/cmd/vclusterctl/cmd/connect.go
@@ -41,12 +41,13 @@ import (
 type ConnectCmd struct {
 	*flags.GlobalFlags
 
-	KubeConfig    string
-	PodName       string
-	UpdateCurrent bool
-	Print         bool
-	LocalPort     int
-	Address       string
+	KubeConfigContextName string
+	KubeConfig            string
+	PodName               string
+	UpdateCurrent         bool
+	Print                 bool
+	LocalPort             int
+	Address               string
 
 	ServiceAccount            string
 	ServiceAccountClusterRole string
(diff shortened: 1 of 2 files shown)


cmd/vclusterctl/cmd/connect.go
```diff
@@ -224,6 +226,13 @@ func (cmd *ConnectCmd) Connect(vclusterName string, command []string) error {
 		return fmt.Errorf("unexpected kube config")
 	}
 
+	if cmd.KubeConfigContextName != "Default" {
+		ctx := kubeConfig.Contexts["Default"]
```
FabianKramm, line 230: Also make sure to rename the cluster as well as auth config to the new name so that merging a kube config doesn't yield any conflicts in clusters or auths

FabianKramm (changes requested):
@tukobadnyanoba thanks for the PR! Please see comments below, besides that also make sure the context name is used if "--update-current" is true

FabianKramm (approved):
LGTM!

PR #527, opened 2022-06-07
feat: serve pprof info on port 6060

matskiv:
/kind enhancement

When DEBUG env var is true, syncer profiling information will be served via built-in API server on the /debug/pprof/ path.

To consume the information first start a port forward command in a separate terminal:
`kubectl port-forward -n vcluster-namespace $(kubectl get pods -o name -l app=vcluster -n vcluster-namespace | head -n 1) 6060:8443`

To acquire a 10s sample run:
`curl -k -o timer.prof http://localhost:6060/debug/pprof/profile?seconds=10`
it will override the previous `timer.prof` file if one exists

To view the collected data run
`go tool pprof -http=:8888 timer.prof 2>/dev/null`
and open the http://localhost:8888 URL in your browser

### DIFF
--- a/cmd/vcluster/main.go
+++ b/cmd/vcluster/main.go
@@ -1,12 +1,14 @@
 package main
 
 import (
+	"os"
+
 	"github.com/loft-sh/vcluster/cmd/vcluster/cmd"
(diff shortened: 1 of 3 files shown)


FabianKramm (approved):

PR #530, opened 2022-06-08
feat: skip port-forwarding for local k8s distros

### LINKED ISSUE #528: Simplify vcluster creation for local kubernetes clusters
Right now vcluster needs port-forwarding by default for all Kubernetes clusters if not specified otherwise. For local Kubernetes distributions such as docker-desktop, minikube etc., we could by default deploy vcluster with a NodePort service instead and then create a kubeconfig for the node port to avoid port-forwarding. This makes it a lot easier to use vcluster with local k8s. Since most distros differ in how the Kubernetes cluster is created, we need to detect first which Kubernetes distro is used and then handle each case differently. 

Distros:
- [ ] docker desktop
- [ ] minikube
- [ ] rancher desktop
- [ ] kind
- [ ] k3d
- [ ] microk8s

FabianKramm:
/kind feature

resolves #528

**BREAKING CHANGES**
- `vcluster create` now automatically does a connect. Can be disabled via `vcluster create --connect=false`
- `vcluster connect` now by default switches the current kube context to the vcluster context. Can be disabled via `vcluster connect --update-current=false`
- `vcluster delete` now by default deletes the namespace of the vcluster if it was created by vcluster before. Can be disabled via `vcluster delete --auto-delete-namespace=false`

- Refactored `vcluster create` and `vcluster connect` that will now create a nodeport service for local k8s distros such as docker-desktop & rancher-desktop, that does not need port-forwarding. 
- Refactored `vcluster` commands to work even if the current context points to a vcluster. This allows easier management of vclusters with:
```
vcluster connect cluster-1
vcluster connect cluster-2 # <- Works even though current context points to cluster-1
```
- New `vcluster disconnect` to switch back to the parent cluster context.

### DIFF
--- a/cmd/vcluster/cmd/start.go
+++ b/cmd/vcluster/cmd/start.go
@@ -184,7 +184,7 @@ func ExecuteStart(options *context2.VirtualClusterOptions) error {
 	}
 
 	// Ensure that service CIDR range is written into the expected location
-	err = wait.Poll(5*time.Second, 2*time.Minute, func() (bool, error) {
+	err = wait.PollImmediate(5*time.Second, 2*time.Minute, func() (bool, error) {
 		err = ensureServiceCIDR(inClusterClient, currentNamespace, translate.Suffix)
 		if err != nil {
 			klog.Errorf("failed to ensure that service CIDR range is written into the expected location: %v", err)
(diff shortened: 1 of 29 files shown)


cmd/vcluster/cmd/start.go
```diff
@@ -184,7 +184,7 @@ func ExecuteStart(options *context2.VirtualClusterOptions) error {
 	}
 
 	// Ensure that service CIDR range is written into the expected location
-	err = wait.Poll(5*time.Second, 2*time.Minute, func() (bool, error) {
+	err = wait.PollImmediate(5*time.Second, 2*time.Minute, func() (bool, error) {
```
matskiv, line 187: Oh, there goes my - "Hey, I improved startup time by full 5 seconds" :joy: 
jk of course! I don't why I didn't notice that it was so slow to start.
Good catch!

matskiv:
The breaking change to `vcluster delete` sounds kinda dangerous. I am not sure if the value outweighs the risk of removing something unintentionally.

matskiv (approved):
One minor comment. 
Changes look great :+1: 
Very cool new features :sunglasses:

PR #531, opened 2022-06-09
[WIP] enhancement: improve e2e testing for different configurations

### LINKED ISSUE #503: Improve e2e testing
Currently we deploy vcluster in a single configuration in parallel in 4 different jobs for each distro. While this works and smoke tests vcluster, several features and areas of vcluster remain untested due to this single deployment of vcluster. Hence I propose to deploy vcluster within those jobs multiple times in different configurations, so that we can test certain features as well.

Currently untested vcluster functionality (might be incomplete):
- [ ] default installation with namespace admin RBAC only
- [x] rootless mode
- [x] enable scheduler
- [x] sync all nodes
- [ ] vcluster in vcluster
- [x] sync nodes based on node selector
- [x] isolated mode
- [x] target namespace
- [x] pause / resume vcluster
- [ ] persistentvolumes
- [ ] storageclasses
- [ ] priorityclasses
- [ ] serviceaccounts
- [ ] plugins

pratikjagrut:
/kind enhancement
/kind test

resolves #503 


.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,211 @@
...
+        test-suite-path: ${{fromJson(needs.get-testsuites-dir.outputs.matrix)}}
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+    
+    - name: Set up kind with K8s version v1.21.1
+      uses: engineerd/setup-kind@v0.5.0
+      with:
+        version: "v0.11.1"
+        image: kindest/node:v1.21.1
```
matskiv, line 115: Would be nice to bump this to the latest versions while we are at it...
```suggestion
        version: "v0.14.0"
        image: kindest/node:v1.24.2
```

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,211 @@
...
+      run: |
+        set -x 
+        
+        sudo apt-get install -y sed
+        
+        sed -i "s|REPLACE_IMAGE_NAME|${{ env.IMAGE_NAME }}|g" ${{ matrix.test-suite-path }}/../commonValues.yaml
+        
+        docker load < vcluster_syncer
+        
+        kind load docker-image ${{ env.IMAGE_NAME }}
```
matskiv, line 145: Nice to have: `docker load` could be skipped if we use `kind load image-archive`, but it would need to be tested to check that our `vcluster_syncer` file is in a compatible format.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,211 @@
...
+        
+        chmod +x vcluster && sudo mv vcluster /usr/bin
+        
+        vcluster create ${{ env.VCLUSTER_SUFFIX }} -n ${{ env.VCLUSTER_NAMESPACE }} \
```
matskiv, line 149: Another nice to have: adding the `--debug` flag which "Prints the stack trace if an error occurs". :)

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,211 @@
...
+        -f ${{ matrix.test-suite-path }}/../commonValues.yaml \
+        -f ${{ matrix.test-suite-path }}/values.yaml
+      continue-on-error: true
+
+    - name: Wait until vcluster is ready
+      id: wait-until-vcluster-is-ready
+      run: |
+        set -x
+        kubectl wait --for=condition=ready pod -l app=${{ env.VCLUSTER_SUFFIX }} -n ${{ env.VCLUSTER_NAMESPACE }} --timeout=600s
+      continue-on-error: true
```
matskiv, line 162: For this step, I would add a condition that it only executes if the previous one didn't fail.
And we should probably reduce the timeout, 3-5mins should be enough. I was probably the one that set it to 10mins initially :sweat_smile:, but now with many more jobs this is becoming potentially more costly.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,211 @@
...
+  #   - uses: actions/github-script@v6
+  #     with:
+  #       github-token: ${{secrets.GITHUB_TOKEN}}
+  #       script: |
+  #         github.rest.issues.createComment({
+  #           issue_number: context.issue.number,
+  #           owner: context.repo.owner,
+  #           repo: context.repo.repo,
+  #           body: 'Workflow run link -> https://github.com/loft-sh/vcluster/actions/runs/${{ github.run_id }}'
+  #         })
```
matskiv, line 211: This should be removed? :thinking:

.github/workflows/unit-tests.yaml
```diff
@@ -6,7 +6,6 @@ on:
   pull_request:
     branches:
       - main
-      - v0.5
     paths:
       - "**.go"
       - "!e2e/**" # exclude changes in e2e tests
```
matskiv, line 11: This needs to be updated :) 
```suggestion
      - "!test/**" # exclude changes in e2e tests
```

hack/test.sh
```diff
@@ -4,12 +4,8 @@
 export GO111MODULE=on
 export GOFLAGS=-mod=vendor
 
-# Test if we can build the program
-echo "Building virtual cluster..."
-CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build cmd/vcluster/main.go || exit 1
```
matskiv, line 9: Why was this removed? :) 
I think it catches compilation errors nice and early and provides simpler output - easier to see compilation errors.

matskiv: And I think it wouldn't slow down the testing much because go test would reuse the build cache populated by this go build command.

test/e2e/values.yaml
```diff
@@ -0,0 +1,7 @@
+mapServices:
+  fromVirtual:
+  - from: test/test
+    to: test
+  fromHost:
+  - from: test/test
+    to: default/test
```
matskiv, line 7: Is this actually tested by any of the tests in the "e2e" suite?
If not, it still makes sense to add it, but perhaps to the `test/commonValues.yaml` instead. By adding it we would test that syncer doesn't crash when one of the mapped services is missing.

test/e2e_node/e2e_node_suite_test.go
```diff
@@ -0,0 +1,59 @@
...
+func init() {
+	_ = clientgoscheme.AddToScheme(scheme)
+	// API extensions are not in the above scheme set,
+	// and must thus be added separately.
+	_ = apiextensionsv1beta1.AddToScheme(scheme)
+	_ = apiextensionsv1.AddToScheme(scheme)
+	_ = apiregistrationv1.AddToScheme(scheme)
+}
+
+// TestRunE2ETests checks configuration parameters (specified through flags) and then runs
```
matskiv, line 38: Super minor mistake :) 
```suggestion
// TestRunE2ENodeTests checks configuration parameters (specified through flags) and then runs
```

test/e2e_node/node.go
```diff
@@ -0,0 +1,19 @@
...
+
+var _ = ginkgo.Describe("Node sync", func() {
+	f := framework.DefaultFramework
+	ginkgo.It("sync all nodes", func() {
+		hostNodes, err := f.HostClient.CoreV1().Nodes().List(f.Context, v1.ListOptions{})
+		framework.ExpectNoError(err)
+
+		virtualNodes, err := f.VclusterClient.CoreV1().Nodes().List(f.Context, v1.ListOptions{})
+		framework.ExpectNoError(err)
+		framework.ExpectEqual(len(hostNodes.Items), len(virtualNodes.Items))
```
matskiv, line 17: Just thinking out loud here. Does it make sense to check something else besides the number of nodes?
E.g. real nodes have exactly the same labels afaik.

matskiv (changes requested):
Great job! Nice to see all the improvements come together and all the jobs pass.
I added a lot of comments, but most of it is just minor things or improvement suggestions, looks nice overall :+1:

test/e2e/values.yaml
pratikjagrut: It is tested in the e2e test suite. It's only specific to that test suit.

matskiv: Oh, I didn't know. Thanks for confirming. :+1:

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,177 @@
...
+        --debug \
+        --connect=false \
+        --distro=${{ matrix.distribution }} \
+        -f ${{ matrix.test-suite-path }}/../commonValues.yaml \
+        -f ${{ matrix.test-suite-path }}/values.yaml
+      continue-on-error: true
+
+    - name: Wait until vcluster is ready
+      id: wait-until-vcluster-is-ready
+      if: steps.create-vcluster.outcome == 'failure'
```
matskiv, line 138: I think it should be executed when create vcluster did **not** fail. :)

matskiv (changes requested):

matskiv (approved):
Looks great! Thanks! :+1:

FabianKramm (approved):

PR #534, opened 2022-06-15
feat: local connect support for minikube VM driver

matskiv:
/kind enhancement

This is a follow up on #530 
`vcluster connect` can now generate kube config without port forwarding for minikube instances that use certain drivers, such as KVM on linux.


FabianKramm (approved):

PR #541, opened 2022-06-17
refactor: add plugin name as env to plugin container

FabianKramm:
/kind enhancement

Adds the plugin name from the `plugin.yaml` to the environment of the plugin container. This allows us in the plugin sdk to get rid of the name inside the code.


matskiv (approved):
Good idea! :+1:

PR #564, opened 2022-06-28
test: add e2e test for rootless mode

### LINKED ISSUE #503: Improve e2e testing

pratikjagrut:
/kind test

resolves #503

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -91,7 +91,7 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v3
     
-    - name: Set up kind with K8s version v1.21.1
+    - name: Set up kind k8s cluster
       uses: engineerd/setup-kind@v0.5.0
       with:
         version: "v0.14.0"
@@ -155,6 +155,8 @@ jobs:
         kubectl get events -n ${{ env.VCLUSTER_NAMESPACE }} --sort-by='.lastTimestamp'
         echo "======================================================================================================================"
         kubectl logs -l app=${{ env.VCLUSTER_SUFFIX }} -n ${{ env.VCLUSTER_NAMESPACE }} -c syncer --tail=500
+        echo "======================================================================================================================"
+        kubectl describe pods -n ${{ env.VCLUSTER_NAMESPACE }}
         exit 1
 
     - name: Set up Go
@@ -171,12 +173,16 @@ jobs:
         VCLUSTER_SUFFIX=${{ env.VCLUSTER_SUFFIX }} VCLUSTER_NAMESPACE=${{ env.VCLUSTER_NAMESPACE }} go test -v -ginkgo.v -ginkgo.skip='.*NetworkPolicy.*'
       continue-on-error: true
 
-    - name: Print vcluster logs if e2e tests fail
+    - name: Print logs if e2e tests fail
       if: steps.execute-e2e-tests.outcome == 'failure'
       run: |
         set -x 
+        kubectl get pods -o yaml -n ${{ env.VCLUSTER_NAMESPACE }}
+        echo "======================================================================================================================"
         kubectl get events -n ${{ env.VCLUSTER_NAMESPACE }} --sort-by='.lastTimestamp'
         echo "======================================================================================================================"
-        kubectl logs -l app=${{ env.VCLUSTER_SUFFIX }} -n ${{ env.VCLUSTER_NAMESPACE }} -c syncer --tail=5000
+        kubectl logs -l app=${{ env.VCLUSTER_SUFFIX }} -n ${{ env.VCLUSTER_NAMESPACE }} -c syncer --tail=500
+        echo "======================================================================================================================"
+        kubectl describe pods -n ${{ env.VCLUSTER_NAMESPACE }}
         exit 1
 
--- a/test/e2e_rootless/rootlessmode.go
+++ b/test/e2e_rootless/rootlessmode.go
@@ -1,0 +1,35 @@
+package e2erootlessmode
+
+import (
+	"strings"
+
+	"github.com/loft-sh/vcluster/pkg/util/podhelper"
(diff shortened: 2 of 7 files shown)


test/e2e_rootless/values.yaml
```diff
@@ -0,0 +1,9 @@
+securityContext:
+  runAsUser: 12345
+  runAsNonRoot: true
+
```
olljanat, line 4: assuming that #566 will get merged you should also add:
```yaml
fsGroup: 12345
```

test/e2e_rootless/values.yaml
```diff
@@ -0,0 +1,9 @@
...
+## this values are for k8s distro
+syncer:
+  securityContext:
+    runAsUser: 12345
+    runAsNonRoot: true
```
olljanat, line 9: assuming that #566 will get merged you should also add:
```yaml
etcd:
  fsGroup: 12345
```

olljanat (commented):
Thanks for this PR. We have been running K8s as rootless longtime already without issues. Only exception is that etcd needs fsGroup value which our policy controller adds automatically. Created PR #566 to fix it correctly in Helm charts now. 

It is also possible to harden those pods even more which why we most probably should test least privileged model at least with some variant. 

Here is maximum hardening which I have managed to get working. With those securityContexts vcluster will be able to run even on host cluster where https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted is enforced:
```yaml
syncer:
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
      - ALL
    readOnlyRootFilesystem: true
    runAsGroup: 12345
    runAsNonRoot: true
    runAsUser: 12345
    seccompProfile:
      type: RuntimeDefault

etcd:
  fsGroup: 12345
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
      - ALL
    readOnlyRootFilesystem: true
    runAsGroup: 12345
    runAsNonRoot: true
    runAsUser: 12345
    seccompProfile:
      type: RuntimeDefault

controller:
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
      - ALL
    readOnlyRootFilesystem: true
    runAsGroup: 12345
    runAsNonRoot: true
    runAsUser: 12345
    seccompProfile:
      type: RuntimeDefault

api:
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
      - ALL
      add:
      - NET_BIND_SERVICE
    readOnlyRootFilesystem: true
    runAsGroup: 12345
    runAsNonRoot: true
    runAsUser: 12345
    seccompProfile:
      type: RuntimeDefault
```
But when tested it I found that seccompProfile is actually missing from pre-install job and CoreDNS which user cannot configure so created #567 to fix that issue.

However if these are too much for you PR I can created another one later to modify those rootless tests.

pratikjagrut:
> However if these are too much for you PR I can created another one later to modify those rootless tests.

@olljanat Maybe we could have separate e2e for this because your config also has hardened security and isolation mode.

.github/workflows/e2e.yaml
```diff
@@ -171,12 +175,18 @@ jobs:
...
       run: |
         set -x 
+        kubectl get pods -o yaml -n ${{ env.VCLUSTER_NAMESPACE }}
+        echo "======================================================================================================================"
         kubectl get events -n ${{ env.VCLUSTER_NAMESPACE }} --sort-by='.lastTimestamp'
         echo "======================================================================================================================"
-        kubectl logs -l app=${{ env.VCLUSTER_SUFFIX }} -n ${{ env.VCLUSTER_NAMESPACE }} -c syncer --tail=5000
+        kubectl logs -l app=${{ env.VCLUSTER_SUFFIX }} -n ${{ env.VCLUSTER_NAMESPACE }} -c syncer --tail=500
+        echo "======================================================================================================================"
+        kubectl logs -l app=${{ env.VCLUSTER_SUFFIX }} -n ${{ env.VCLUSTER_NAMESPACE }} -c vcluster --tail 500
```
matskiv, line 188: Same as above

test/e2e_rootless/rootlessmode.go
```diff
@@ -0,0 +1,45 @@
...
+				Container: "syncer",
+				Command:   cmd,
+				Stdout:    true,
+			},
+			clientgoscheme.ParameterCodec,
+		)
+		exec, err := remotecommand.NewSPDYExecutor(f.HostConfig, "POST", req.URL())
+		framework.ExpectNoError(err)
+		var stdout bytes.Buffer
+		err = exec.Stream(remotecommand.StreamOptions{Stdout: &stdout})
```
matskiv, line 40: btw, we have util to make exec easier - https://github.com/loft-sh/vcluster/blob/main/pkg/util/podhelper/exec.go#L152-L153

matskiv (changes requested):
A change in .github/workflows/e2e.yaml is needed IMHO
The other comment is FYI only - change optional :)

matskiv (approved):
LGTM. Thanks! :+1:

FabianKramm (approved):

PR #565, opened 2022-06-28
enhancement: cancel old running workflows when the PR is updated

pratikjagrut:
/kind enhancement

The workflow is triggered when a PR is created or updated, i.e when new commits are pushed. If a commit is pushed while the old run was in progress then GH triggers a new workflow without terminating the workflow in progress. So now at this time, there are be 2 GH actions running on PR, one is on the latest commit and one is on the old commit.

The changes in this PR are to ensure that the old workflow run is terminated if the PR is updated.


FabianKramm (approved):

PR #567, opened 2022-06-29
include seccompProfile to pre-install jobs if isolation mode is enabled

olljanat:
/kind bugfix

N/A

Fixed an issue where pre-install job was not able to run in namespace where *Restricted* Pod Security Standard is enforced because of missing seccompProfile.

Related to #564 as I'm testing how tight we can harden vcluster environments.


olljanat:
@FabianKramm ah, ok that works for me. Updated PR to just include those to pre-install job if isolation mode is enabled.

FabianKramm (approved):
@olljanat great thanks! LGTM!

PR #570, opened 2022-06-29
added support for actions and conditions for alb ingress controller

### LINKED ISSUE #532: Support actions and conditions annotations for ALB ingress controller

The [actions](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/annotations/#actions) and [conditions](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/annotations/#conditions) ALB annotation are not working as expected for the Ingress resources that are synced to the host, where the AWS ALB ingress controller is installed.
There are two problems here. Firstly the annotation key suffix needs to be equal to the service name, which becomes different("translated") when synced to the host. The annotations have the following format:
```
alb.ingress.kubernetes.io/actions.${action-name}
alb.ingress.kubernetes.io/conditions.${conditions-name}
```
and as the docs state:
```
The action-name in the annotation must match the serviceName in the Ingress rules ...
The conditions-name in the annotation must match the serviceName in the Ingress rules ...
```

Second problem is in the value of the action of the `forward` type, [which references `serviceName` in the `.forwardConfig.targetGroups`](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/9c79e45dd4b8cf94b97a1c6737bf1d47306460d0/pkg/ingress/config_types.go#L73), e.g.:
```
alb.ingress.kubernetes.io/actions.forward-multiple-tg: >
      {"type":"forward","forwardConfig":{"targetGroups":[{"serviceName":"service-1","servicePort":"http" ...
```
the values of the `serviceName` fields should be translated the same way we translate the name of Service that gets synced to the host.

Implement detection of the annotations in question in the Ingress syncer.
Implement the translation for the key/value of these annotations based on the description of the problem above.

_No response_

### Additional context

_No response_

tukobadnyanoba:
/kind feature

resolves #532 

Fixed an issue where vcluster ...serviceNames are translated to vcluster way of services

### DIFF
--- a/pkg/controllers/resources/ingresses/legacy/syncer_test.go
+++ b/pkg/controllers/resources/ingresses/legacy/syncer_test.go
@@ -276,6 +276,78 @@ func TestSync(t *testing.T) {
 				assert.NilError(t, err)
 			},
 		},
+		{
+			Name: "Translate annotation with alb annotations",
+			InitialVirtualState: []runtime.Object{
+				&networkingv1beta1.Ingress{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      baseIngress.Name,
+						Namespace: baseIngress.Namespace,
+						Labels:    baseIngress.Labels,
+						Annotations: map[string]string{
+							"nginx.ingress.kubernetes.io/auth-secret":       "my-secret",
+							"alb.ingress.kubernetes.io/actions.testservice": "{\"forwardConfig\":{\"targetGroups\":[{\"serviceName\":\"nginx-service\",\"servicePort\":\"80\",\"weight\":100}]}}",
+						},
+					},
+				},
+			},
+			InitialPhysicalState: []runtime.Object{
+				&networkingv1beta1.Ingress{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      createdIngress.Name,
+						Namespace: createdIngress.Namespace,
+						Labels:    createdIngress.Labels,
+					},
+				},
+			},
+			ExpectedVirtualState: map[schema.GroupVersionKind][]runtime.Object{
+				networkingv1beta1.SchemeGroupVersion.WithKind("Ingress"): {
+					&networkingv1beta1.Ingress{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      baseIngress.Name,
+							Namespace: baseIngress.Namespace,
+							Labels:    baseIngress.Labels,
+							Annotations: map[string]string{
+								"alb.ingress.kubernetes.io/actions.testservice": "{\"forwardConfig\":{\"targetGroups\":[{\"serviceName\":\"nginx-service\",\"servicePort\":\"80\",\"weight\":100}]}}",
+								"nginx.ingress.kubernetes.io/auth-secret":       "my-secret",
+							},
+						},
+					},
+				},
+			},
+			ExpectedPhysicalState: map[schema.GroupVersionKind][]runtime.Object{
+				networkingv1beta1.SchemeGroupVersion.WithKind("Ingress"): {
+					&networkingv1beta1.Ingress{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      createdIngress.Name,
+							Namespace: createdIngress.Namespace,
+							Labels:    createdIngress.Labels,
+							Annotations: map[string]string{
+								"vcluster.loft.sh/managed-annotations":                          "alb.ingress.kubernetes.io/actions.testservice-x-test-x-suffix\nnginx.ingress.kubernetes.io/auth-secret",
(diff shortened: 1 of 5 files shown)


pkg/controllers/syncer/translator/namespaced_translator.go
```diff
@@ -216,6 +223,58 @@ func translateAnnotations(vObj client.Object, pObj client.Object, excluded []str
 	return retMap
 }
 
+func getActionOrConditionValue(annotation, actionOrCondition string) string {
```
FabianKramm, line 226: Please only rewrite those annotations for the ingress syncer.

pkg/controllers/syncer/translator/namespaced_translator.go
```diff
@@ -216,6 +223,58 @@ func translateAnnotations(vObj client.Object, pObj client.Object, excluded []str
...
+			} `json:"forwardConfig"`
+		}
+		// change k
+		action := getActionOrConditionValue(k, ActionsSuffix)
+		k = strings.Replace(k, action, translate.PhysicalName(action, vObj.GetNamespace()), 1)
+		// change v
+		var payload *actionPayload
+		err := json.Unmarshal([]byte(v), &payload)
+		if err != nil {
+			fmt.Println(err)
```
FabianKramm, line 251: Use klog instead of fmt.Println please

pkg/controllers/syncer/translator/namespaced_translator.go
```diff
@@ -216,6 +223,58 @@ func translateAnnotations(vObj client.Object, pObj client.Object, excluded []str
...
+		} else {
+			if payload != nil &&
+				&payload.ForwardConfig != nil &&
```
FabianKramm, line 254: This check is not necessary

pkg/controllers/syncer/translator/namespaced_translator.go
```diff
@@ -216,6 +223,58 @@ func translateAnnotations(vObj client.Object, pObj client.Object, excluded []str
...
+				&payload.ForwardConfig.TargetGroups != nil &&
```
FabianKramm, line 255: This check is not necessary

pkg/controllers/syncer/translator/namespaced_translator.go
```diff
@@ -216,6 +223,58 @@ func translateAnnotations(vObj client.Object, pObj client.Object, excluded []str
...
+				len(payload.ForwardConfig.TargetGroups) > 0 {
+				for _, targetGroup := range payload.ForwardConfig.TargetGroups {
+					serviceName := targetGroup["serviceName"].(string)
```
FabianKramm, line 258: Please make a safe cast here as this will panic otherwise if serviceName is not defined or if it is not a string

pkg/controllers/syncer/translator/namespaced_translator.go
```diff
@@ -216,6 +223,58 @@ func translateAnnotations(vObj client.Object, pObj client.Object, excluded []str
...
+					if len(serviceName) > 0 {
```
FabianKramm, line 259: Use serviceName != ""

pkg/controllers/syncer/translator/namespaced_translator.go
```diff
@@ -216,6 +223,58 @@ func translateAnnotations(vObj client.Object, pObj client.Object, excluded []str
...
+						targetGroup["serviceName"] = translate.PhysicalName(serviceName, vObj.GetNamespace())
+					}
+				}
+				marshalledPayload, err := json.Marshal(payload)
+				if err != nil {
+					fmt.Println(err)
```
FabianKramm, line 265: Use klog / loghelper here instead of fmt.Println

pkg/controllers/syncer/translator/namespaced_translator.go
```diff
@@ -216,6 +223,58 @@ func translateAnnotations(vObj client.Object, pObj client.Object, excluded []str
 	return retMap
 }
 
+func getActionOrConditionValue(annotation, actionOrCondition string) string {
+	i := strings.Index(annotation, actionOrCondition)
+	if i > -1 {
+		return annotation[i+len(actionOrCondition):]
+	} else {
```
FabianKramm, line 230: No else needed here

FabianKramm (changes requested):
@tukobadnyanoba thanks for the PR! Couple of comments below

matskiv (dismissed):
Since this logic is specific for Ingress resource, I think it should be moved to one of the source files in `/pkg/controllers/resources/ingresses` 

Adding unit tests would be nice.

FabianKramm (approved):

PR #571, opened 2022-06-29
fix: set more realistic resource requests

olljanat:
/kind bugfix

N/A

Fixed an issue where vcluster reserves a lot more resources that it actually needs.

I noticed that when deployed a lot of K8s based vclusters on top of same host cluster it stopped from scheduling pods with error "Insufficient cpu" when all nodes CPU usage was still under 30% (and I have not even enabled HA mode for those yet which would tribe number of those pods).

New values are based on fact that on multiple environments which I checked:
* syncer cpu was 6-8m and memory 46-52Mi
* api-server cpu was 28-32m and memory 249-282Mi
* controller cpu was 8-10m and memory 37-45Mi
* etcd cpu was 9-12m and memory 62-156Mi
* coredns cpu was 2m and memory 13-14Mi

Because Kubernetes does not use swap it is more critical to have buffer on memory than on CPU side.

Scheduler I don't use so there I'm just guessing. For same reason I didn't touched to k0s, k3s or eks other resources.

EDIT: Found now also very interesting article about topic https://sysdig.com/blog/kubernetes-resource-limits/

### DIFF
--- a/charts/eks/templates/coredns.yaml
+++ b/charts/eks/templates/coredns.yaml
@@ -139,8 +139,8 @@ data:
                 cpu: 1000m
                 memory: 170Mi
               requests:
-                cpu: 100m
-                memory: 70Mi
+                cpu: 3m
+                memory: 16Mi
             args: [ "-conf", "/etc/coredns/Corefile" ]
             volumeMounts:
             - name: config-volume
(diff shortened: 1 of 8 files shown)


FabianKramm (approved):
@olljanat thanks a lot for the research! LGTM!

PR #580, opened 2022-07-01
feat: add serviceCIDR to values only when provided

matskiv:
/kind enhancement

The `addCommonReleaseValues` is used in cluster-api-provider-vcluster, where we will no longer be setting the CIDR value upfront. This produces an unnecessary entry in the generated values:
```
serviceCIDR: null
```
which won't be present after this change.

N/A


FabianKramm (approved):

PR #586, opened 2022-07-06
fix: k8s: Disable Node authorization

olljanat:
/kind bugfix

N/A

Fixed an issue where k8s distro was allowing Node authorization which is not needed.

I was reading https://kubernetes.io/docs/reference/access-authn-authz/node/ and noticed that only reason why Node authorization exists is that kubelet needs to to authenticate with kube-apiserver so it can be safely disabled from vcluster side.

For same reason NodeRestriction adminission plugin is not needed: https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction


FabianKramm (approved):
@olljanat thanks for the PR! Good catch, looks good to me!

PR #599, opened 2022-07-15
fix: return non-zero exit code on failing command

witalisoft:
/kind bugfix

running command under vcluster kube context doesn't provide the right exit code when commands fail:
```console
$ vcluster connect vcluster -- kubectl get ns nonexistent
Error from server (NotFound): namespaces "nonexistent" not found
$ echo $?
0
```
so it's much harder to exit the ci/cd pipeline

Fixed an issue where vcluster connect doesn't return non-zero exit code of failing command


FabianKramm (changes requested):
@witalisoft thanks a lot for the PR! Great find, since ExitError provides us the correct error code, can we exit with this exact error code here as well?

FabianKramm:
@witalisoft thanks for the update, LGTM!

PR #605, opened 2022-07-19
refactor: allow node ports during service create

### LINKED ISSUE #593: Add the ability to Service Sync to Sync the NodePort - random (as from vcluster) & static

When transfering our Controller into a vCluster implementation, so far all looks Great and is working as we wish.

However, there are various scenarios where we wish for the Service to be available outside the cluster (mapServices sort of does that) - but we in various scenarios (dev scenarios) we are exposing the services as a NodePort type - For Example

```yaml
  apiVersion: v1
  kind: Service
  metadata:
    name: gui-service
    namespace: myns
  spec:
    ports:
    - name: https
      nodePort: 31005
      port: 443
      protocol: TCP
      targetPort: 8080
    selector:
      configid: gui-container
    sessionAffinity: None
    type: NodePort
```

This kind of scenario; which we also use in some of our automation flows also specifics the specific nodePort we wish to allocate; This is really important, as our dev environments are usually kind clusters where we  define extraPortMappings so that we can access the various services we expose - These port mappings are static and can't be changed on the fly in a kind cluster.

I attempted to use mapServices to solve the issue and I found out that indeed the service is mapped to the main cluster, however it is mapped/syncd as a ClusterIP (verified in the external Service definition) - that doesn't cut it for our dev Use-Case - and also for some of our automation Use-Cases.

In the vCluster config I attempted

```yaml
  mapServices:
    fromVirtual:
    - from: myns/gui-service
      to: host-gui-service 
```

Chatting with @matskiv on Slack found out that this is by design - and that we could even just use Service Sync (default enabled)-  
And Indeed I see that the services are Sync'd as defined - with the nodePort as chosen by vCluster.

What I'd like to propose is that either with the mapServices add NodPort as an option IE:

1. mapServices
    ```yaml
      mapServices:
         fromVirtual:
         - from: kwaf/waas-gui-service
           to: waas-gui-service
         nodePort: static
    ```
    

    - If the nodePort option is provided, the service should be set as type: NodePort (default random value

    
    - if nodePort is static - use the nodePort value as given in the manifest to create service - it should be up tho the user to manage the value for multiple vClusters

2. In ServiceSync - Add option to use the defined NodePort when creating the service from within the vCluster
    Obviously, only one vCluster in the Main Cluster can use that NodePort, but at least we can then use the SPECIFICALLY defined NodePort in our dev and automation Use-Cases.

_No response_

### Additional context

_No response_

FabianKramm:
/kind enhancement

resolves #593

vcluster will now sync the specific node ports of a service during creation.

### DIFF
--- a/pkg/server/filters/service.go
+++ b/pkg/server/filters/service.go
@@ -210,7 +210,6 @@ func createService(req *http.Request, decoder encoding.Decoder, localClient clie
 	}
 	newService.Annotations[services.ServiceBlockDeletion] = "true"
 	services.RewriteSelector(newService, vService)
-	services.StripNodePorts(newService)
 	err = localClient.Create(req.Context(), newService)
 	if err != nil {
 		klog.Infof("Error creating service in physical cluster: %v", err)


matskiv (approved):
I tested this locally, works as expected :+1:

Issue #606, opened 2022-07-19
Add Helm value for .spec.loadBalancerSourceRanges field of the vcluster Service


matskiv:
### Is your feature request related to a problem?

When using Service of the LoadBalancer type, a user might need to set .spec.loadBalancerSourceRanges value on the Service resource in order to secure the cluster.

### Which solution do you suggest?

Add Helm value under the `service` [field](https://github.com/loft-sh/vcluster/blob/a5036e970ece4ec26a4a853f770b7c64a82ae887/charts/k3s/values.yaml#L206-L207) and use it in the Service manifest.
Implement in Helm chart of each distro.

### Additional context

This was originally raised here - https://github.com/loft-sh/cluster-api-provider-vcluster/issues/15

PR #607, opened 2022-07-19
Allow setting the loadBalancerSourceRanges when the Service type is LoadBalancer

mtougeron:
Allow setting the loadBalancerSourceRanges when the Service type is LoadBalancer

/kind enhancement

resolves https://github.com/loft-sh/vcluster/issues/606
relates to https://github.com/loft-sh/cluster-api-provider-vcluster/issues/15

Allow setting the loadBalancerSourceRanges when the Service type is LoadBalancer in the Helm charts

### DIFF
--- a/charts/eks/templates/syncer-service.yaml
+++ b/charts/eks/templates/syncer-service.yaml
@@ -18,3 +18,11 @@ spec:
   selector:
     app: vcluster
     release: {{ .Release.Name }}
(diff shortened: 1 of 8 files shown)


matskiv (approved):
LGTM :+1: 

Would be nice to also add this to `charts/k0s` and `charts/eks`, but we can merge this and add it separately too.

matskiv (approved):
Thank you!
LGTM :)

PR #618, opened 2022-07-22
feat(charts): add helm templated init manifest

matskiv:
/kind feature

Slack thread - https://loft-sh.slack.com/archives/C01NP55P8F6/p1658489200272449
vcluster user would like to be able to pass vcluster name inside of the cluster

feat(charts): add init.manifestsTemplate field to values, which can be used to create a helm templated manifest which can use helm values passed to vcluster chart, for example vcluster name - "{{ .Release.Name }}".

This is just an idea, I open to suggestions how to implement this better, or to arguments why we shouldn't add this at all.
Also, if we are going to add this, I feel like it would be nice to add templated values for init helm charts.


FabianKramm (approved):

PR #623, opened 2022-07-26
Fix for allocatable pods sync when scheduler enabled

### LINKED ISSUE #612: The number of allocatable pods is synced incorrectly when scheduler is enabled
### What happened?

vcluster internal scheduler has been repeatedly scheduling a pod to a node that reached the allocatable limit of running pods in the host cluster. The synced pod was not able to be started on the chosen node. This caused the pod to be recreated continuously. 

This seems to also affect a cluster auto-scaler if it is installed in vcluster.

vcluster scheduler to not schedule a pod to a node that is already running the allocatable number of pods on the given node, and schedule it to a node with free capacity instead.
Node status should display accurate information about the allocatable number of pods and [other resources](https://github.com/kubernetes/api/blob/master/core/v1/resource.go).

Install vcluster [with vcluster scheduler enabled](https://www.vcluster.com/docs/architecture/scheduling#separate-vcluster-scheduler).
Create many pods in the host cluster to reach the limit of allocatable pods on your nodes.
Create enough pods in vcluster to go over the allocatable pod limit (sum of pods in vcluster and host).

This is the part of the source code that needs to be extended to cover other resources (pods and storage):
https://github.com/loft-sh/vcluster/blob/main/pkg/controllers/resources/nodes/translate.go#L189-L221

We need to subtract the number of host pods, that do not belong to this vcluster instance, from the host nodes `.status.allocatable.pods`.

N/A

N/A

### vlcuster version

<details>

```console
$ vcluster --version
vcluster version 0.10.2
```

</details>

N/A

### OS and Arch

N/A

tukobadnyanoba:
/kind bugfix

resolves #612 

Fixed an issue where vcluster ...allocatable pods has proper value

### DIFF
--- a/pkg/controllers/resources/nodes/translate.go
+++ b/pkg/controllers/resources/nodes/translate.go
@@ -3,10 +3,9 @@ package nodes
 import (
 	"context"
 	"encoding/json"
-	"github.com/loft-sh/vcluster/pkg/util/stringutil"
-
 	"github.com/loft-sh/vcluster/pkg/controllers/resources/nodes/nodeservice"
 	synccontext "github.com/loft-sh/vcluster/pkg/controllers/syncer/context"
+	"github.com/loft-sh/vcluster/pkg/util/stringutil"
 	"github.com/loft-sh/vcluster/pkg/util/translate"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/controllers/resources/nodes/translate.go
```diff
@@ -20,6 +19,8 @@ var (
 	TaintsAnnotation = "vcluster.loft.sh/original-taints"
 )
 
+const ResourcePods = "pods"
```
matskiv, line 22: nit-pick: `corev1.ResourcePods` can be used instead of defining this constant here

PR #628, opened 2022-07-27
test: add general tests for e2e_rootless testsuite

pratikjagrut:
/kind test

### DIFF
--- a/charts/eks/templates/syncer-deployment.yaml
+++ b/charts/eks/templates/syncer-deployment.yaml
@@ -57,6 +57,12 @@ spec:
       serviceAccountName: vc-{{ .Release.Name }}
       {{- end }}
       volumes:
+      {{- if or .Values.syncer.securityContext.runAsUser .Values.syncer.securityContext.runAsNonRoot }}
+        - name: helm-cache
+          emptyDir: {}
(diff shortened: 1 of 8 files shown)


charts/k8s/templates/syncer-deployment.yaml
```diff
@@ -173,6 +177,10 @@ spec:
 {{ toYaml .Values.syncer.env | indent 10 }}
         {{- end }}
         volumeMounts:
+          - name: helm-cache
```
FabianKramm, line 180: Please add a condition here similar to k0s and k3s

charts/eks/templates/syncer-deployment.yaml
```diff
@@ -168,6 +172,10 @@ spec:
 {{ toYaml .Values.syncer.env | indent 10 }}
         {{- end }}
         volumeMounts:
+          - name: helm-cache
```
FabianKramm, line 175: Please add a condition here similar to k0s and k3s

FabianKramm (changes requested):
@pratikjagrut thanks for the PR! Please take a look at my comments below

FabianKramm (approved):

PR #622, opened 2022-07-25
added emptydir volume for k3s

tukobadnyanoba:
**What issue type does this pull request address?** (keep at least one, remove the others) 
/kind bugfix
**What does this pull request do? Which issues does it resolve?** (use `resolves #<issue_number>` if possible) 
resolves #604 

**Please provide a short message that should be published in the vcluster release notes**
Fixed an issue where vcluster ...k3s with securityContext enabled won't be printing logs 'Permission denied warning' now 

**What else do we need to know?**

PR #641, opened 2022-08-02
added emptyDir to /etc/rancher path

tukobadnyanoba:
/kind bugfix

resolves #622

Fixed an issue where vcluster ...emptyDir will be mounted on /etc/rancher path


FabianKramm (approved):
@tukobadnyanoba thanks for the PR! LGTM!

PR #649, opened 2022-08-05
fix(syncer): allocatable pods calculation

matskiv:
/kind bugfix

Allocatable pods were calculated incorrectly. Calculation assumed that all pods in the targetNamespace were from vcluster, so the result was always off but at least 1 pod(vcluster itself).

Fixed an issue where nodes status would show an incorrect number of allocatable pods, which could cause scheduling problems.


FabianKramm (approved):

PR #654, opened 2022-08-08
use run-k0s only when in rootless mode

### LINKED ISSUE #648: Emptydir /run/k0s creates issues for k0s
### What happened?

After the upgrade from 0.10.2 to 0.11.0, k0s won't start anymore with the following error log:

```
Error: failed to create /data/k0s/status.sock: directory "/run/k0s" exist, but the permission is 0777. The expected permission is 755
```

Comparing to an older statefulset, the new one has an emptyDir connected to /run/k0s, while the old one does not. This causes a directory to be created with the wrong chmod.

vcluster container should just start

Install a k0s distro vcluster with version 0.11.0

_No response_

```
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.8+k0s", GitCommit:"7061dbbf75f9f82e8ab21f9be7e8ffcaae8e0d44", GitTreeState:"clean", BuildDate:"2022-03-24T22:32:09Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
```

k0s

### vlcuster version

0.11.0

k0s

### OS and Arch

-

ishankhare07:
fixes #648

/kind bugfix

resolves #648 

Fixed an issue where vcluster with the k0s distro would fail to create / upgrade to version `0.11.0` because of additional volume `/run/k0s`. We now only create this when rootless mode is enabled

### DIFF
--- a/charts/k0s/templates/statefulset.yaml
+++ b/charts/k0s/templates/statefulset.yaml
@@ -57,8 +57,10 @@ spec:
       serviceAccountName: vc-{{ .Release.Name }}
       {{- end }}
       volumes:
+        {{- if or .Values.securityContext.runAsUser .Values.securityContext.runAsNonRoot }}
         - name: run-k0s
           emptyDir: {}
+        {{- end }}
         - name: k0s-config
           secret:
             secretName: vc-{{ .Release.Name }}-config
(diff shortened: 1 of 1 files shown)


FabianKramm (approved):
@ishankhare07 thanks for the PR! LGTM!

PR #658, opened 2022-08-10
docs: add `ingressClassName` to the ingress example

hariprasadiit:
It was not clear to me as to why the ingress was not working even after setting up things as per the documentation, and it took a while to figure out that `ingressClassName` was missing from the example. 

/kind documentation

Add `ingressClassName` to the example to make it easier for the users who're new to Kubernetes


FabianKramm (approved):
@hariprasadiit thanks for the PR! Good catch, seems like we forgot to mention that in the docs!

PR #660, opened 2022-08-11
enhancement: prevent scale down of node where vcluster is running

### LINKED ISSUE #617: Prevent scale down of a node where vcluster is running

If [cluster-autoscaler](https://github.com/kubernetes/autoscaler) is installed inside the vcluster, and vcluster is not running in a highly available configuration, the autoscaler may decide to scale down the node that hosts vcluster pod and thus cause a downtime of the vcluster API server proxy and syncer.

Whenever vcluster is deployed with a single replica, a flag or environment variable with the pods node name should be added to the pod with the syncer.
The node sync controller should then add the `"cluster-autoscaler.kubernetes.io/scale-down-disabled": "true"` annotation to the node that matches the node name provided via the aforementioned flag or env var.

pratikjagrut:
/kind enhancement

resolves #617

### DIFF
--- a/charts/eks/templates/syncer-deployment.yaml
+++ b/charts/eks/templates/syncer-deployment.yaml
@@ -158,7 +158,15 @@ spec:
         securityContext:
 {{ toYaml .Values.syncer.securityContext | indent 10 }}
         env:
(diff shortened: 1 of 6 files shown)


FabianKramm (approved):

PR #662, opened 2022-08-12
Add background proxy container

tukobadnyanoba:
/kind feature

resolves # https://github.com/loft-sh/vcluster-dd-extension/issues/22

Fixed an issue where vcluster ...now starts background-proxy container with `vcluster connect` command if the flag `--background-proxy` is specified. This doesn't need users to require the terminal open after user specifies `vcluster connect --background-proxy` 


cmd/vclusterctl/cmd/connect.go
```diff
@@ -142,6 +144,16 @@ func (cmd *ConnectCmd) Connect(vclusterName string, command []string) error {
 		return err
 	}
 
+	if cmd.BackgroundProxy && localkubernetes.IsDockerInstalledAndUpAndRunning() {
```
FabianKramm, line 147: Add a check here for `cmd.Server` being empty as otherwise starting a proxy container is unnecessary

FabianKramm (changes requested):

FabianKramm (approved):

tszyszko:
Since this was merged in, you can no longer have multiple clusters in the same parent namespace. Is this intentional ? Our use-case involves multiple clusters created by CI, so multiple test clusters could be created on the fly in a single "test" namespace that has all the necessary restrictions applied in the host namespace.

PR #667, opened 2022-08-17
feat: add HA configuration for vcluster

pratikjagrut:
/kind feature

This pull request adds HA configuration for vcluster.

- The first pod-anti-affinity rule says that the scheduler should try to avoid scheduling the Pod onto a node with one or more Pods with the label app=vcluster.

- The second pod-anti-affinity rule says that the scheduler should try to avoid scheduling the Pod onto a node in the same zone as one or more Pods with the label app=vcluster.

- The pod Disruption Budget makes sure that there is a maximum 1 pod unavailable at a time.

Added an HA configuration for vcluster.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,7 @@ Fork and clone the repo:
 After adjusting the variables and you can run:
 
 ```
-devspace dev -n vcluster
+devspace dev
 ```
 
 Which should produce an output similar to:
--- a/devspace.yaml
+++ b/devspace.yaml
@@ -3,6 +3,8 @@ name: vcluster
 
 # Variables used
 vars:
+  # Make sure vcluster is the default namespace
+  DEVSPACE_FLAGS: "-n vcluster"
   SYNCER_IMAGE: ghcr.io/loft-sh/loft-enterprise/dev-vcluster
 
 # Images DevSpace will build for vcluster
@@ -151,6 +153,3 @@ pipelines:
       
       # Make sure the pvcs are deleted as well
       kubectl delete pvc --all -n ${DEVSPACE_NAMESPACE}
-
-commands:
-  dev: devspace dev -n vcluster "$@"


matskiv:
I don't know if we should add these things to the k3s and k0s charts. Could it lead users to believe that HA is supported in those too?

charts/eks/templates/podDisruptionBudget.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+    app: vcluster
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+spec:
+  maxUnavailable: {{ .Values.podDisruptionBudget.maxUnavailable }}
+  selector:
+    matchLabels:
+      app: vcluster
```
matskiv, line 15: This label selector is not sufficient to distinguish between multiple vclusters installed into the same namespace. You should add this selector too:
```
release: {{ .Release.Name }}
```

charts/eks/values.yaml
```diff
@@ -121,8 +121,30 @@ syncer:
...
+          topologyKey: "kubernetes.io/hostname"
+      # if possible avoid scheduling pod onto node that is in the same zone as one or more vcluster pods are running
+      - weight: 50
+        podAffinityTerm:
+          labelSelector:
+            matchExpressions:
+            - key: app
+              operator: In
+              values:
+              - vcluster
```
matskiv, line 146: Same issue as with label selector in PDB.
For the labelSelector in both affinity rules above.

charts/eks/values.yaml
```diff
@@ -132,15 +154,41 @@ syncer:
...
```
matskiv, line 190: etcd has the `app` label set to `vcluster-etcd` 
https://github.com/loft-sh/vcluster/blob/61419dabee3e8ad664c968f9d170cbc61003ce22/charts/eks/templates/etcd-statefulset.yaml#L25

matskiv: And same as above, you also need release: `{{ .Release.Name }}`

charts/eks/values.yaml
```diff
@@ -170,8 +218,30 @@ controller:
...
```
matskiv, line 243: Similar to etcd, controller has different value for the `app` label:
https://github.com/loft-sh/vcluster/blob/61419dabee3e8ad664c968f9d170cbc61003ce22/charts/eks/templates/controller-deployment.yaml#L37

matskiv: And same as above, you also need release: `{{ .Release.Name }}`

charts/eks/values.yaml
```diff
@@ -191,8 +261,30 @@ api:
...
```
matskiv, line 286: Same issue here^
https://github.com/loft-sh/vcluster/blob/61419dabee3e8ad664c968f9d170cbc61003ce22/charts/eks/templates/api-deployment.yaml#L37

matskiv: And you also need `release: {{ .Release.Name }}`

matskiv (changes requested):
Comments that I added for eks also apply to the k8s chart.

PR #670, opened 2022-08-18
Allow syncer to omit taints that match the enforced tolerations

### LINKED ISSUE #416: Synced nodes should ommit taints that have an enforce toleration

Yes. Tolerations are enforced on a pod level, which works nicely for Deployment and StatefulSet kinds. However, having **synched nodes** with a custom taint on every node results in 0 pods running for a given DaemonSet as the DaemonSet controller evaluates the DaemonSet with no toleration on it (the toleration would get synced on the pod level). As a consequence there are no nodes that would fulfil/tolerate the workload.

_Edited by @matskiv after discussion in the comments below._

Node syncer should not sync those taints from the host node, for which the vcluster has enforced tolerations defined.
So the `--enforce-toleration=key1=value1:NoSchedule` would mean that the `key1=value1:NoSchedule` taint would not be synced to the virtual node in vcluster.

_Original:_ ~~Enforce tolerations via a mutating webhook instead of the syncer's enforce-toleration flag or potentially allow opt-in to enforce tolerations via a mutating webhook. Adding a mutating webhook out of the box, would allow greater flexibility (support for DaemonSets) as well as transparency (toleration gets added on workload-level instead of pod-level).~~

Manually apply the toleration for DaemonSets.

### Additional context

The goal is to have vcluster operate on dedicated, real/synced nodes.

neogopher:
/kind bugfix

resolves #416 

Fixed an issue where Daemonset created in vcluster couldn't schedule pods on nodes with taints tolerated by vcluster's `--enforce-tolerations` flag.

**Issue Root Cause**
Tolerations are enforced on a pod level. It works nicely for Deployment and StatefulSet kinds, as they create the required number of pods in Pending state. These pods are then available to be scheduled on eligible nodes by K8s scheduler.
The vcluster syncer syncs these pods to host cluster and applies the tolerations specified in --enforce-toleration syncer flag to the pod spec of the pods in the host cluster. These pods thus are able to tolerate the tainted nodes and get scheduled successfully by the default kube-scheduler.

However, DaemonSet pods are created and scheduled by the DaemonSet controller instead of the default kube-scheduler. It looks at all the eligible nodes for scheduling & only creates pods for those nodes. Hence, if the DaemonSet controller is unable to find the correct number of eligible nodes it will not create expected number of pods.

**Solution**
As suggested in the issue, this PR updates the Node syncer to omit the taints tolerated by the vcluster. Since the vcluster nodes lack the taints defined with --enforce-toleration, DaemonSet controller can correctly identify the eligible nodes & create pods accordingly. Once the pods get created in vcluster, the syncer can now apply the enforeced tolerations while syncing them with the host cluster.

### DIFF
--- a/pkg/controllers/resources/nodes/syncer.go
+++ b/pkg/controllers/resources/nodes/syncer.go
@@ -8,6 +8,7 @@ import (
 	"github.com/loft-sh/vcluster/pkg/controllers/syncer"
 	synccontext "github.com/loft-sh/vcluster/pkg/controllers/syncer/context"
 	"github.com/loft-sh/vcluster/pkg/controllers/syncer/translator"
+	"github.com/loft-sh/vcluster/pkg/util/toleration"
 	"github.com/loft-sh/vcluster/pkg/util/translate"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -36,6 +37,17 @@ func NewSyncer(ctx *synccontext.RegisterContext, nodeService nodeservice.NodeSer
 		}
 	}
 
+	// parse tolerations
+	var tolerations []*corev1.Toleration
+	if len(ctx.Options.Tolerations) > 0 {
+		for _, t := range ctx.Options.Tolerations {
+			tol, err := toleration.ParseToleration(t)
+			if err == nil {
+				tolerations = append(tolerations, &tol)
+			}
+		}
+	}
+
 	return &nodeSyncer{
 		enableScheduler: ctx.Options.EnableScheduler,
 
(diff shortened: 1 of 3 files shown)


FabianKramm (approved):
@neogopher thanks a lot for the PR! Looks great, thanks for contributing!

neogopher:
Thank you so much @FabianKramm and @matskiv for giving me the opportunity to work towards my first contribution to vcluster. :)  It was indeed my pleasure and I enjoyed working with the codebase. You guys have setup the entire system for development very nicely. Looking forward to more contributions in the future. :)

PR #681, opened 2022-08-22
enhancement: add helm chart validation job

pratikjagrut:
/kind enhancement

This enhancement will help us find mistakes in the helm chart sooner than when creating vcluster. So if the test is supposed to fail because of wrong helm values or configs it will fail sooner.


heiko-braun:
Outdated

PR #687, opened 2022-08-23
Remove helm binary and use helm package

bagulm123:
/kind enhancement

resolves # This PR replaces helm binary implementation with helm package.

Fixed an issue where vcluster ... now doesn't need helm installed on user's machine.


pkg/helm/helm.go
```diff
@@ -48,243 +62,638 @@ type Client interface {
...
 	_, err = tempFile.Write(data)
 	if err != nil {
-		os.Remove(tempFile.Name())
+		_ = os.Remove(tempFile.Name())
 		return "", errors.Wrap(err, "write temp file")
 	}
 
 	// Close temp file
-	tempFile.Close()
+	_ = tempFile.Close()
```
ishankhare07, line 696: should we use a `defer` instead?

pkg/helm/values/eks.go
```diff
@@ -59,18 +57,10 @@ func getDefaultEKSReleaseValues(chartOptions *helm.ChartOptions, log log.Logger)
...
-`
-	values = strings.ReplaceAll(values, "##API_IMAGE##", apiImage)
-	values = strings.ReplaceAll(values, "##CONTROLLER_IMAGE##", controllerImage)
-	values = strings.ReplaceAll(values, "##ETCD_IMAGE##", etcdImage)
-	values = strings.ReplaceAll(values, "##COREDNS_IMAGE##", corednsImage)
+	values := ""
+	values += "api.image=" + apiImage
+	values += ",coredns.image=" + corednsImage
+	values += ",controller.image=" + controllerImage
+	values += ",etcd.image=" + etcdImage
```
ishankhare07, line 64: should we use a `Join` here instead?

pkg/helm/values/k3s.go
```diff
@@ -71,70 +61,52 @@ func getDefaultK3SReleaseValues(chartOptions *helm.ChartOptions, log log.Logger)
...
-##BASEARGS##
-`
+	values := ""
+
 	if chartOptions.Isolate {
-		values += `
-securityContext:
-  runAsUser: 12345
-  runAsNonRoot: true`
+		values += ",securityContext.runAsUser=12345,securityContext.runAsNonRoot=true"
```
ishankhare07, line 67: would a leading `,` cause any issue here?

pkg/helm/values/k3s.go
```diff
@@ -71,70 +61,52 @@ func getDefaultK3SReleaseValues(chartOptions *helm.ChartOptions, log log.Logger)
...
 	}
 
-	values = strings.ReplaceAll(values, "##IMAGE##", image)
+	values += "vcluster.image=" + image
```
ishankhare07, line 70: missing a `,` separator?

pkg/helm/values/k3s.go
```diff
@@ -71,70 +61,52 @@ func getDefaultK3SReleaseValues(chartOptions *helm.ChartOptions, log log.Logger)
...
-  nodes:
-    enabled: true`
+		values += ",sync.nodes.enabled=true"
 	}
 
 	if chartOptions.Isolate {
-		values += `
-isolation:
-  enabled: true`
+		values += ",isolation.enabled=true"
```
ishankhare07, line 107: should we move this to the block on `L66` ?

pkg/helm/values/eks.go
ishankhare07: Yea what I meant was to probably have slice of strings instead and finally join it just before using. Something like
```
values := []string{}

apiImage := "image"
corednsImage := "corednsImage"
controllerImage := "controllerImage"
etcdImage := "etcdImage"

values = append(values,
	"apiImage="+apiImage,
	"coredns.image="+corednsImage,
	"controller.image="+controllerImage,
	"etcd.image="+etcdImage,
)
...
...
...


// finally right before using values
strings.Join(values, ",")
```

that way we can avoid the complications of where a leading `,` separator is required or not

ishankhare07:
Hi @mahendrabagul , on  a side note - since this change is now also removing `helm` bin dependency for init charts, we can remove the bin from our [`Dockerfile`](https://github.com/loft-sh/vcluster/blob/main/Dockerfile#L11) as well

ishankhare07 (approved):

Issue #678, opened 2022-08-20
Status socket breaks k0s upon restart


christiaangoossens:
### What happened?

In commit https://github.com/loft-sh/vcluster/commit/61a9a2cd480b618100149c393d9858317abd4115 both /run/k0s (see my previous issue) and `--status-socket=/data/k0s/status.sock` were introduced to enable rootless mode on k0s.

However, both cause issues on restarts. Sometimes if the server shuts down improperly it will think that /data/k0s/status.sock is still in use, thus refusing to start.

### What did you expect to happen?

The server should clear /data/k0s/status.sock, or as it was previously, use a socket outside of the volumes, such that it gets cleared automatically on restart.

### How can we reproduce it (as minimally and precisely as possible)?

Run k0s vcluster in root mode and crash the container until it says that the socket is in use.

### Anything else we need to know?

Possible solution would be similar to what we did for /run/k0s => disabling it if not set to rootless (https://github.com/loft-sh/vcluster/pull/654), but this hints that this might also cause issues in general if it's run in rootless mode.

### Host cluster Kubernetes version

v1.22.8+k0s

### Host cluster Kubernetes distribution

k0s

### vlcuster version

Occurs on version 0.11 and 0.11.1

### Vcluster Kubernetes distribution(k3s(default)), k8s, k0s)

k0s

### OS and Arch

irrelevant

PR #690, opened 2022-08-26
[Helm K0s] Try to avoid leaving a status.sock around between restarts

ananace:
/kind bugfix

improves the situation for #678 - it doesn't fully resolve it, but it should actually be "fixable" with a restart of the pod with this change

Fixed an issue where a rootless install of the k0s vcluster chart could leave a socket in place, wedging the cluster until the socket is manually removed.

This will store the status socket for k0s in the emptyDir `/run/k0s`, which would mean that if an instance hits #678 then it should be enough to restart it to clear the collision.

A better fix would probably be to ensure that the socket file is removed on every startup, but I don't really want to start adding complex startup to a container that's just `k0s controller ...` at the moment

### DIFF
--- a/charts/k0s/templates/statefulset.yaml
+++ b/charts/k0s/templates/statefulset.yaml
@@ -100,7 +100,7 @@ spec:
           - {{ $f | quote }}
           {{- end }}
           {{- if or .Values.securityContext.runAsUser .Values.securityContext.runAsNonRoot }}
-          - --status-socket=/data/k0s/status.sock
+          - --status-socket=/run/k0s/status.sock
           {{- end }}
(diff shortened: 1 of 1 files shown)


FabianKramm:
@ananace thanks for the issue! We reverted the change mentioned in #678 and its now only applied for k0s rootless mode, I believe with this PR the rootless mode wouldn't work at all anymore, but we will see as soon as the tests might fail.

pratikjagrut (commented):
LGTM

ananace:
Right, I should probably make it more clear in the PR description that this is relating to rootless in particular.

FabianKramm (approved):

PR #692, opened 2022-08-29
Ensure valid kubernetes.io/hostname label for fake nodes

TylerGillson:
/kind bugfix

Resolves bug in fake node syncer where fake node hostname label values are not truncated when necessary.

Fixed an issue where vcluster fake node syncer was generating invalid values for the `kubernetes.io/hostname` label and thus failing to create fake nodes.

Nothing


FabianKramm (approved):
@TylerGillson thanks for the PR! Good catch, LGTM!

Issue #510, opened 2022-06-01
Epic: Add support for logging solutions inside vcluster


FabianKramm:
## Use Case

Currently its not possible to use common logging solutions such as FluentD or ELK stack inside vcluster, because all of them mount the node host path directly. While an acceptable solution is to provide a single logging solution in the host cluster, there are also use cases where you would want to install the solution directly inside the virtual cluster.

## Implementation

We can rewrite the host path with the correct pod uids trough symlinks in a sidecar container. We need to continously monitor changes to that folder and reflect it in the fake host path folder.

ddl-risong-na:
This is something crucial for us. Do you have an ETA on when this will be implemented?

As always: great product!

FabianKramm:
@ddl-risong-na not yet, we are still in the concept phase here.

jaredallard:
It's worth noting we solved this problem (for restic (velero)) with https://github.com/getoutreach/vcluster-fs-syncer. Of course it's not perfect but worth pointing out :D

FabianKramm:
@ddl-risong-na yes that is expected. One way to work around this is to use a node selector inside the vcluster that only syncs nodes that fullfill the selector to the virtual cluster (see the [docs](https://www.vcluster.com/docs/architecture/nodes#node-syncing-modes) for more info). If you really want to sync all nodes to the vcluster, you need to make sure that DaemonSets mostly run on the host cluster as this is a lot more efficient as you already have discovered.

Regarding the vcluster pod failing, that is probably a different issue, but you are correct it usually should withstand quite a lot of load. We did some improvements for this in the latest version v0.9.1, so if you are not using that yet would be interesting to know if that also goes into an OOM state. If yes, this is definitely worth creating a new issue for, as vcluster should be able to scale to several hundreds if not thousands per cluster.

vumdao:
I installed Elasticsearch on the host and what I got from search is following and it shows nothing related to kubernetes labels/namespace

![image](https://user-images.githubusercontent.com/37215642/177575542-33edb0cb-6841-40f1-9c02-9627f605b5b6.png)

What is a workaround for this?

sandeep540:
@FabianKramm are you looking at vcluster plug-in to solve this? Can I take a stab at it? If you can give more details it would be great

vumdao:
Hi, any update?any documents to enable logging in vcluster or from host?

PR #694, opened 2022-08-30
enable logging and backup support

ishankhare07:

/kind enhancement
/kind feature

resolves https://github.com/loft-sh/vcluster/issues/510

Fixed an issue where vcluster we need to support logging solutions like `ELK`, `fluentd`, `loki` etc. for pods inside the vcluster. Along with these we also enable support for backup solutions like velero and VM solutions like kubevirt


cmd/vcluster/cmd/hostpaths.go
```diff
@@ -0,0 +1,473 @@
...
+		// QUESTION: current mechanism cleans up container symlinks
+		// for old pods, but does not cleanup for container restarts
+		// should we be supporting that too?
+		err = cleanupOldContainerPaths(ctx, existingVPodsWithNamespace)
+		if err != nil {
+			klog.Errorf("error cleaning up old container log paths: %v", err)
+		}
+
+		// QUESTION: should we clear this immediately, would there be
+		// repercussions as we might miss some backups if we delete immediately
```
ishankhare07, line 258: @FabianKramm I would like to know your thoughts about these 2 scenarios

charts/eks/templates/hostpath-mapper-daemonset.yaml
```diff
@@ -0,0 +1,99 @@
+{{- if .Values.hostpathMapper.enabled }}
+apiVersion: apps/v1
+{{- if not .Values.hostpathMapper.dev }}
+kind: DaemonSet
+{{- else  }}
+kind: Deployment
+{{- end }}
+metadata:
+  name: {{ .Release.Name }}
```
FabianKramm, line 9: Doesn't just release name cause issues with the regular vcluster deployment?

charts/k0s/values.yaml
```diff
@@ -93,6 +93,10 @@ mapServices:
   # also create the namespace for the service.
   fromHost: []
 
+hostpathMapper:
+  enabled: false
+  dev: false
```
FabianKramm, line 98: Please remove dev here as we don't want the user to see it

pkg/controllers/resources/pods/syncer.go
```diff
@@ -212,6 +232,9 @@ func (s *podSyncer) SyncDown(ctx *synccontext.SyncContext, vObj client.Object) (
 		}
 	}
 
+	ctx.Log.Infof("checking if pod mounts any volume")
```
FabianKramm, line 235: Please remove this statement

pkg/controllers/resources/pods/syncer.go
```diff
@@ -220,6 +243,98 @@ func (s *podSyncer) SyncDown(ctx *synccontext.SyncContext, vObj client.Object) (
 	return s.SyncDownCreate(ctx, vPod, pPod)
 }
 
+func (s *podSyncer) checkAndRewriteHostPath(ctx *synccontext.SyncContext, pPod *corev1.Pod) *corev1.Pod {
+	if len(pPod.Spec.Volumes) > 0 {
+		ctx.Log.Infof("checking for hostpath volumes")
```
FabianKramm, line 248: Please remove this or make debug

cmd/vcluster/cmd/hostpaths.go
```diff
@@ -0,0 +1,473 @@
...
+	cmd.Flags().StringVar(&options.ServiceName, "service-name", "", "The service name where the vcluster proxy will be available")
+	cmd.Flags().BoolVar(&options.SetOwner, "set-owner", true, "If true, will set the same owner the currently running syncer pod has on the synced resources")
+
+	cmd.Flags().StringVar(&options.Name, "name", "vcluster", "The name of the virtual cluster")
+	cmd.Flags().StringVar(&options.BindAddress, "bind-address", "0.0.0.0", "The address to bind the server to")
+	cmd.Flags().IntVar(&options.Port, "port", 8443, "The port to bind to")
+
+	cmd.Flags().BoolVar(&options.SyncAllNodes, "sync-all-nodes", false, "If enabled and --fake-nodes is false, the virtual cluster will sync all nodes instead of only the needed ones")
+
+	cmd.Flags().StringVar(&options.ServiceAccount, "service-account", "", "If set, will set this host service account on the synced pods")
```
FabianKramm, line 89: Please remove all the flags that are not needed

cmd/vcluster/cmd/hostpaths.go
```diff
@@ -0,0 +1,473 @@
...
+}
+
+func MapHostPaths(options *context2.VirtualClusterOptions) error {
+	virtualPath = fmt.Sprintf(pods.VirtualPathTemplate, options.TargetNamespace, options.Name)
+
+	virtualKubeletPodPath = filepath.Join(virtualPath, "kubelet", "pods")
+	virtualLogsPath = filepath.Join(virtualPath, "log")
+	virtualPodLogsPath = filepath.Join(virtualLogsPath, "pods")
+
+	virtualContainerLogsPath = filepath.Join(virtualLogsPath, "containers")
```
FabianKramm, line 101: Make these part of the options instead of global variables

cmd/vcluster/cmd/hostpaths.go
```diff
@@ -0,0 +1,473 @@
...
+	SymLinkName *string
+	PhysicalPod corev1.Pod
+}
+
+var (
+	virtualPath              string
+	virtualLogsPath          string
+	virtualPodLogsPath       string
+	virtualContainerLogsPath string
+	virtualKubeletPodPath    string
```
FabianKramm, line 52: Part of options instead of global variables

cmd/vcluster/cmd/hostpaths.go
```diff
@@ -0,0 +1,473 @@
...
+		// cleanup old pod symlinks
+		err = cleanupOldPodPath(ctx, virtualPodLogsPath, existingPodsPath)
+		if err != nil {
+			klog.Errorf("error cleaning up old pod log paths: %v", err)
+		}
+
+		// cleanup old container symlinks
+		// QUESTION: current mechanism cleans up container symlinks
+		// for old pods, but does not cleanup for container restarts
+		// should we be supporting that too?
```
FabianKramm, line 251: Is this causing any problems? If yes, we definitely should support that

cmd/vcluster/cmd/hostpaths.go
FabianKramm: How is this handled in a regular cluster? Are those cleared immediately? In general the goal is here to stay as close as possible to the behaviour of a physical cluster, so what ever the physical cluster is doing here, we should try to replicate that behvaiour

FabianKramm (changes requested):

charts/k0s/templates/daemonset-hostpath-mapper.yaml
```diff
@@ -0,0 +1,99 @@
...
+      serviceAccountName: vc-{{ .Release.Name }}
+      {{- end }}
+      containers:
+      - name: hostpath-mapper
+        {{- if .Values.hostpathMapper.image }}
+        image: "{{ .Values.defaultImageRegistry }}{{ .Values.hostpathMapper.image }}"
+        {{- else }}
+        image: "{{ .Values.defaultImageRegistry }}loftsh/vcluster:{{ .Chart.Version }}"
+        {{- end }}
+        {{- if .Values.syncer.workingDir }}
```
FabianKramm, line 52: Is this needed?

FabianKramm (changes requested):

cmd/vcluster/cmd/hostpaths.go
```diff
@@ -0,0 +1,479 @@
...
+
+	mapHostPaths(ctx, localManager, virtualClusterManager)
+
+	return nil
+}
+
+func mapHostPaths(ctx context.Context, pManager, vManager manager.Manager) {
+	options := ctx.Value(optionsKey).(*context2.VirtualClusterOptions)
+
+	wait.Forever(func() {
```
matskiv, line 167: Would be nice to refactor this into a controller that would execute based on the watch + re-run at a bigger interval.
But this can be done in a future version.

matskiv:
The new pod syncer code doesn't take into account volumeMounts on ephemeral pods. But I guess it is very unlikely they would use these host paths, right?

matskiv:
The code looks really good :+1: :heart_eyes: 
I created a podman + CRI-O cluster to test it on this runtime too, I'll report how it went later today.

charts/eks/templates/daemonset-hostpath-mapper.yaml
```diff
@@ -0,0 +1,99 @@
...
+        {{- else }}
+        image: "{{ .Values.defaultImageRegistry }}loftsh/vcluster:{{ .Chart.Version }}"
+        {{- end }}
+        {{- if .Values.syncer.workingDir }}
+        workingDir: {{ .Values.syncer.workingDir }}
+        {{- end }}
+        command:
+          - /vcluster
+        args:
+          - maplogs
```
matskiv, line 58:
```suggestion
          - maphostpaths
```
This needs to be fixed for all distros

pkg/controllers/resources/pods/syncer.go
```diff
@@ -220,6 +277,124 @@ func (s *podSyncer) SyncDown(ctx *synccontext.SyncContext, vObj client.Object) (
...
+						volume.HostPath.Type,
+						KubeletPodPath,
+						PhysicalKubeletVolumeMountPath,
+						containersWithPhysicalMountPaths,
+						pPod)
+				}
+
+				if volume.HostPath.Path == LogHostpathPath {
+					pPod.Spec.Volumes[i].HostPath.Path = s.virtualLogsPath
+				}
```
matskiv, line 331: Why is the physical volume(`PhysicalLogVolumeMountPath`) not mounted in this^ case?
I think it needs to be. Have a look at my file structure when the "filebeat" pod mounts `/var/log`:

> root@elk-filebeat-filebeat-5z6kx:/var/log/containers# ls -la
> lrwxrwxrwx. 1 root root 101 Oct  6 22:16 coredns-66ffcc6b58-tsmk2_kube-system_coredns-ea86caae31738763933394f8288230572e3266882c7e0728a9055a5faacf940d.log -> /var/log/pods/kube-system_coredns-66ffcc6b58-tsmk2_82728908-ba9d-4c0d-8212-2434313cee2a/coredns/0.log
> ...
> root@elk-filebeat-filebeat-5z6kx:/var/log/containers# ls -la /var/log/pods
> lrwxrwxrwx. 1 root root 127 Oct  6 22:16 kube-system_coredns-66ffcc6b58-tsmk2_82728908-ba9d-4c0d-8212-2434313cee2a -> /var/vcluster/physical/log/pods/vcluster_coredns-66ffcc6b58-tsmk2-x-kube-system-x-vcluster_1e71df1e-5e10-4f63-9503-d681769440b8
> ...
> root@elk-filebeat-filebeat-5z6kx:/var/log/containers# ls /var
> backups  cache  lib  local  lock  log  mail  opt  run  spool  tmp
>
> root@elk-filebeat-filebeat-5z6kx:/var/log/containers# tail coredns-66ffcc6b58-tsmk2_kube-system_coredns-ea86caae31738763933394f8288230572e3266882c7e0728a9055a5faacf940d.log 
> tail: cannot open 'coredns-66ffcc6b58-tsmk2_kube-system_coredns-ea86caae31738763933394f8288230572e3266882c7e0728a9055a5faacf940d.log' for reading: No such file or directory

matskiv (changes requested):

pkg/controllers/resources/pods/syncer.go
ishankhare07: Yes I think this is because when testing against fluentd it mounts both `/var/log` and `/var/log/pods` which then won't require an explicit physical log mount but I think we need to include that. Nice catch!

ishankhare07:
> The new pod syncer code doesn't take into account volumeMounts on ephemeral pods. But I guess it is very unlikely they would use these host paths, right?

Yea I don't think is would make sense to include ephemeral pods. There might be cases for initContainers specially for kubelet paths but so far I've not seen any issues with velero or kubevirt that mount kubelet path and have initContainers

charts/eks/templates/hostpath-mapper-daemonset.yaml
ishankhare07: using `{{ .Release.Name }}-hostpath-mapper` now

cmd/vcluster/cmd/hostpaths.go
ishankhare07: we are now checking if the internal symlinks resolve. We only delete once they stop resolving

cmd/vcluster/cmd/hostpaths.go
```diff
@@ -0,0 +1,512 @@
...
+		podList := &corev1.PodList{}
+		err := pManager.GetClient().List(ctx, podList, &client.ListOptions{
+			Namespace: options.TargetNamespace,
+			FieldSelector: fields.SelectorFromSet(fields.Set{
+				NodeIndexName: os.Getenv(HostpathMapperSelfNodeNameEnvVar),
+			}),
+		})
```
matskiv, line 188: Additional small optimization here could be to use label selector to filter out pods in the namespace that are not synced by the particular vcluster instance based on the ["vcluster.loft.sh/managed-by"](https://github.com/loft-sh/vcluster/blob/e6f9bf2ddaf467869b24e51ac70e7e8bed942073/pkg/util/translate/translate.go#L18) label.
But this only makes an impact if there are multiple big vclusters in the same namespace, which is unlikely.

matskiv (approved):
I retested new code on minikube that uses podman+CRIO, with ELK (with filebeat), all worked well :+1: 
Code looks good too :+1:

FabianKramm (approved):

PR #696, opened 2022-08-30
fix: add envvar ETCD_UNSUPPORTED_ARCH=arm64 to run etcd on arm64

pratikjagrut:
/kind bugfix

It adds envvar ETCD_UNSUPPORTED_ARCH=arm64 to k0s statefulset.

Fixed an issue where vcluster creation used to fail on arm64 machine for k0s distro because etcd does not fully support arm64 yet.


FabianKramm (approved):

PR #697, opened 2022-08-31
Allow annotation of all objects created by the chart

denniskorbginski:
/kind enhancement

resolves #

Added helm value "global.annotations" to k3s chart that will add its annotations to all objects created in the chart. Annotations for specific objects have precedence over global ones.

I have a case where I want to add an annotation to all objects created in the vcluster release. I'm not super happy with using a `global` value because of its special meaning in helm, but I think if still kind of makes sense and I did not want to touch the existing top-level `annotations` value, as it's only used for the stateful set.

### DIFF
--- a/charts/k3s/templates/coredns.yaml
+++ b/charts/k3s/templates/coredns.yaml
@@ -4,6 +4,10 @@ kind: ConfigMap
 metadata:
   name: {{ .Release.Name }}-coredns
   namespace: {{ .Release.Namespace }}
+  {{- if .Values.global.annotations }}
+  annotations:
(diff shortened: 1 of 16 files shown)


FabianKramm:
@denniskorbginski thanks for creating this issue! While I like the idea, as you said I believe `global` is not a good object to use for this as it is very special in helm. Instead we could use globalAnnotations.

denniskorbginski:
@FabianKramm Yes, thats much better, I've updated the PR accordingly.

FabianKramm (approved):

PR #698, opened 2022-08-31
feat: support 1.25 version with k8s distro

matskiv:
/kind enhancement

Adds support for Kubernetes 1.25 when using k8s distro of vcluster.

I am also running v1.25 conformance tests with this change.
If the tests finish successfully before this is merged, I'll add a commit with results too.

### DIFF
--- a/charts/k8s/values.yaml
+++ b/charts/k8s/values.yaml
@@ -138,7 +138,7 @@ syncer:
 
 # Etcd settings
 etcd:
-  image: k8s.gcr.io/etcd:3.5.1-0
+  image: k8s.gcr.io/etcd:3.5.4-0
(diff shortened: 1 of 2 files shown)


charts/k8s/values.yaml
```diff
@@ -138,7 +138,7 @@ syncer:
 
 # Etcd settings
 etcd:
-  image: k8s.gcr.io/etcd:3.5.1-0
+  image: k8s.gcr.io/etcd:3.5.4-0
```
matskiv, line 141: FYI, this is based on https://github.com/kubernetes/kubernetes/blob/v1.25.0/build/dependencies.yaml#L57-L76

FabianKramm (approved):

Issue #600, opened 2022-07-16
Persistent Volume Reclaim Policy does not work


ddl-risong-na:
### What happened?

There are a few issues around persistent volumes. Here is what I did

 1. Create an EKS cluster
 2. Create ebs csi driver on the host cluster
 3. Create a vcluster and connect to it
 4. Create a storage class using ebs (The storage class can be on the host cluster or vcluster. I've tested both scenarios and the result is identical). The Reclaim Policy is `Retain`
 5. Create a PVC
 6. Create a POD
 7. The PV is created which is expected
 8. Delete the POD
 9. Delete the PVC
10. The PV is also deleted (issue1)
11. The volume on AWS is back to `Available` (issue2. If the PV is deleted, the volume should be deleted as well)

Scenario b, where I created storage class with Reclaim Policy `Delete`. Issue2 still happened - when the PV is deleted, the volume on AWS still exists

Scenario c, where I created storage class with Reclaim Policy `Delete`, after step 7
c8. Edit PV to change the Reclaim Policy to `Retain`
c9. The PV on the host cluster is still `Delete` (issue3)
c10. issue 1 and 2 also occurred

vcluster values.yaml (tested both `storageclasses` and `legacy-storageclasses`, same result)

```
serviceAccount:
  imagePullSecrets:
    - name: docker-secret
sync:
  nodes:
    enabled: true
    syncAllNodes: true
  persistentvolumes:
    enabled: true
  priorityclasses:
    enabled: true
  storageclasses:
    enabled: true
syncer:
  extraArgs: ["--override-hosts-container-image", "our-own-alpine"]
  resources:
    limits:
      memory: 15Gi
```

storage class yaml

```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: disk
parameters:
  encrypted: "true"
  type: gp3
provisioner: ebs.csi.aws.com
reclaimPolicy: Retain
volumeBindingMode: WaitForFirstConsumer
```

a sample pvc json

```
{
  "kind": "PersistentVolumeClaim",
  "apiVersion": "v1",
  "metadata": {
    "name": "claim1"
  },
  "spec": {
    "storageClassName": "disk",
    "accessModes": [
      "ReadWriteOnce"
    ],
    "resources": {
      "requests": {
        "storage": "30Gi"
      }
    }
  }
}
```

a sample pod yaml

```
apiVersion: v1
kind: Pod
metadata:
  name: pod1
[...]

PR #702, opened 2022-09-02
Delete the cloud provider volume when the PersistentVolumeReclaimPolicy is 'Delete'

maheshbaliga:
/kind bugfix

It fixes the "issue 2" in the description of issue #600.

In this issue behaviour, dynamically provisioned volumes on the cloud provider were not getting deleted, even if the PVC & PV on vcluster as well as on host cluster, got deleted. This fix makes sure the cloud provider volumes do get deleted, when PV is deleted.

**Issue Root Cause**
The infrastructure volume was getting deleted intermittently because the finalizers on the VPV was not cleared at the right time. 

**Solution**
The existing code deleted the Physical PV and later while syncing down, the finalizers were cleared from the virtual PV. This did delete the terminating virtual PV, but it interfered with the [K8s controller logic which takes care of deleting the PV & the infrastructure volume for dynamically provisioned volumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#delete). This PR refrains from deleting the PVs, letting k8s clean it up, along with the infrastructure volumes. It just deletes the finalizers on vPV while syncing, allowing it to get deleted, instead of staying in "Terminating" state.

Delete the cloud provider volume when the PersistentVolumeReclaimPolicy is 'Delete'

### DIFF
--- a/pkg/controllers/resources/persistentvolumes/syncer.go
+++ b/pkg/controllers/resources/persistentvolumes/syncer.go
@@ -127,17 +127,12 @@ func (s *persistentVolumeSyncer) Sync(ctx *synccontext.SyncContext, pObj client.
 		}
 	}
 
-	// check if objects are getting deleted
-	if vObj.GetDeletionTimestamp() != nil {
-		if pObj.GetDeletionTimestamp() == nil {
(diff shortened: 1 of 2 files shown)


ishankhare07 (approved):
LGTM

PR #705, opened 2022-09-02
Convert base64 encoded value passed in --extra-values, to values file

mahendrabagul:
/kind enhancement
resolves #

Added code to convert base64 encoded value passed in --extra-values, to values file. 

### DIFF
--- a/cmd/vclusterctl/cmd/create.go
+++ b/cmd/vclusterctl/cmd/create.go
@@ -2,13 +2,8 @@ package cmd
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
-	"os"
-	"os/exec"
-	"path/filepath"
-	"strings"
-	"time"
-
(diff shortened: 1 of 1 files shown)


cmd/vclusterctl/cmd/create.go
```diff
@@ -155,6 +155,47 @@ func (cmd *CreateCmd) Run(args []string) error {
...
+		err = tempFile.Close()
+		if err != nil {
+			return errors.Wrap(err, "close temp values file")
+		}
+		// setting new file to extraValues slice to process it further.
+		newExtraValues = append(newExtraValues, tempValuesFile)
+	}
+
+	// assuming only one base64 encoded string would be present in extraValues slice (coming from extension).
+	defer func(tempValuesFile string) {
```
FabianKramm, line 192: Move this into the loop please as there could be multiple temp values files

FabianKramm (changes requested):
@mahendrabagul thanks for the PR! Please take a look at my comments

FabianKramm (approved):

PR #706, opened 2022-09-05
refactor: improve node allocatable calculation

FabianKramm:
/kind bugfix

Improved node allocatable calculation:
* vcluster now uses an index to retrieve pods that are scheduled on a specific node
* vcluster now only calculates that information if scheduler is enabled

### DIFF
--- a/pkg/controllers/resources/nodes/syncer.go
+++ b/pkg/controllers/resources/nodes/syncer.go
@@ -2,6 +2,8 @@ package nodes
 
 import (
 	"context"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 
 	"github.com/loft-sh/vcluster/pkg/constants"
 	"github.com/loft-sh/vcluster/pkg/controllers/resources/nodes/nodeservice"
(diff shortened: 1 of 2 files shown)


pkg/controllers/resources/nodes/syncer.go
```diff
@@ -86,22 +90,40 @@ func (s *nodeSyncer) Name() string {
...
 		}
-	}()
-	podCache.WaitForCacheSync(ctx.Context)
-	s.podCache = podCache
+		// add index for pod by node
+		err = podCache.IndexField(ctx.Context, &corev1.Pod{}, indexPodByRunningNonVClusterNode, func(object client.Object) []string {
+			pPod := object.(*corev1.Pod)
+			if pPod.Status.Phase == corev1.PodSucceeded || pPod.Status.Phase == corev1.PodFailed {
+				return []string{}
+			} else if pPod.Labels != nil && pPod.Labels[translate.MarkerLabel] == translate.Suffix {
```
matskiv, line 107: I think we need to check pPod namespace too. 
This was a bug in the previous implementation to be fair.

matskiv: And a comment explaining why we are ignoring certain pods would be good. 
In the original implementation it was easier to tell from the context, but not here.

matskiv (changes requested):
I noticed one bug from earlier.
Otherwise it looks good :+1:

matskiv (approved):
Thanks for addressing the comments. LGTM :)

PR #715, opened 2022-09-08
Releases the cloud provider volume when a dynamically provisioned PV with a reclaim policy of 'Delete', is deleted.

maheshbaliga:
/kind bugfix

This fixes the "issue 2" of the issue #600. With this fix, when the PV is dynamically provisioned and its reclaim policy is Delete, we now wait for the physical PV to be deleted and in-turn the CSI to delete the cloud provider volume naturally. Once the physical PV gets deleted, the existing code takes care of clearing the finalizers on the virtual PV and the same gets deleted eventually.

Releases the cloud provider volume when a dynamically provisioned PV with a reclaim policy of 'Delete', is deleted.

### DIFF
--- a/pkg/controllers/resources/persistentvolumes/syncer.go
+++ b/pkg/controllers/resources/persistentvolumes/syncer.go
@@ -127,19 +127,20 @@ func (s *persistentVolumeSyncer) Sync(ctx *synccontext.SyncContext, pObj client.
 		}
 	}
 
-	// check if virtual persistent volume is deleted
-	if vPersistentVolume.GetDeletionTimestamp() != nil && len(vPersistentVolume.GetFinalizers()) > 0 {
-		if vPersistentVolume.Spec.ClaimRef != nil && vPersistentVolume.Spec.PersistentVolumeReclaimPolicy == corev1.PersistentVolumeReclaimDelete {
-			// executes when dynamically provisioned PV with a delete reclaim policy gets deleted
-			// it also releases the cloud provider volume.
-			vPersistentVolume.SetFinalizers(nil)
-			ctx.Log.Infof("remove virtual persistent volume %s finalizers, because virtual persistent volume is terminating", vPersistentVolume.GetName())
-			return ctrl.Result{}, ctx.VirtualClient.Update(ctx.Context, vPersistentVolume)
-		} else {
-			// executes when a Retained PV in Released status is manually deleted and when statically provisioned PVs are deleted
-			ctx.Log.Infof("remove physical persistent volume %s, because virtual persistent volume is deleted", pPersistentVolume.GetName())
-			return ctrl.Result{}, ctx.PhysicalClient.Delete(ctx.Context, pPersistentVolume)
+	// check if objects are getting deleted
+	if vObj.GetDeletionTimestamp() != nil {
+		if pObj.GetDeletionTimestamp() == nil {
+			// check if the PV is dynamically provisioned and the reclaim policy is Delete
+			if !(vPersistentVolume.Spec.ClaimRef != nil && vPersistentVolume.Spec.PersistentVolumeReclaimPolicy == corev1.PersistentVolumeReclaimDelete) {
+				ctx.Log.Infof("delete physical persistent volume %s, because virtual persistent volume is deleted", pObj.GetName())
+				err := ctx.PhysicalClient.Delete(ctx.Context, pObj)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+			}
 		}
+		ctx.Log.Infof("requeue because persistent volume %s, has to be deleted", vObj.GetName())
+		return ctrl.Result{RequeueAfter: time.Second}, nil
 	}
 
 	// check if the persistent volume should get synced
@@ -222,10 +223,6 @@ func (s *persistentVolumeSyncer) SyncUp(ctx *synccontext.SyncContext, pObj clien
 	} else if translate.IsManagedCluster(ctx.TargetNamespace, pObj) {
 		ctx.Log.Infof("delete physical persistent volume %s, because it is not needed anymore", pPersistentVolume.Name)
 		return syncer.DeleteObject(ctx, pObj)
-	} else if pPersistentVolume.GetDeletionTimestamp() != nil {
-		// executes after the sync when a Retained PV in Released status is manually deleted
-		ctx.Log.Infof("deleting physical persistent volume %s", pPersistentVolume.Name)
-		return ctrl.Result{}, nil
 	} else if sync {
 		// create the persistent volume
 		vObj := s.translateBackwards(pPersistentVolume, vPvc)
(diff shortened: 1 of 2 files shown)


pkg/controllers/resources/persistentvolumes/syncer.go
```diff
@@ -129,10 +129,17 @@ func (s *persistentVolumeSyncer) Sync(ctx *synccontext.SyncContext, pObj client.
...
 	// check if virtual persistent volume is deleted
 	if vPersistentVolume.GetDeletionTimestamp() != nil && len(vPersistentVolume.GetFinalizers()) > 0 {
-		//delete the finalizer, so that the object can be deleted
-		vPersistentVolume.SetFinalizers(nil)
-		ctx.Log.Infof("remove virtual persistent volume %s finalizers, because virtual persistent volume is terminating", vPersistentVolume.GetName())
-		return ctrl.Result{}, ctx.VirtualClient.Update(ctx.Context, vPersistentVolume)
+		if vPersistentVolume.Spec.ClaimRef != nil && vPersistentVolume.Spec.PersistentVolumeReclaimPolicy == corev1.PersistentVolumeReclaimDelete {
+			// executes when dynamically provisioned PV with a delete reclaim policy gets deleted
+			// it also releases the cloud provider volume.
+			vPersistentVolume.SetFinalizers(nil)
```
matskiv, line 135: I am not getting why the finalizer should be removed by the syncer. Could you explain that for me, please?

My initial instinct tells me to not touch the finalizers, and just wait until CSI driver or PV controller deletes the finalizer. If the reclaim policy is "Delete", that to me means that the driver needs to confirm the deletion from the filesystem(or cloud API) before deleting the PV in k8s - finalizers are key part of that process. If we remove the finalizer and PV is stuck in the deleting state, the vcluster user won't know about it, and PV in the host cluster will still exist.

maheshbaliga: Hi @matskiv. As per the steps explained in the Issue #600, the SC (Reclaim Policy = Delete), PVC and Pod are created on the virtual cluster, the PV gets dynamically provisioned. Now when the vPod and vPVC are deleted, the syncer triggers the deletion of pPVC; and host cluster initiates the deletion of the pPV and the cloud provider volume. But only the vPV is stuck in the 'Terminating' state. Hence, we are clearing the finalizers so that the same gets garbage collected.

matskiv: But if the pPV is deleted, then the SyncDown() is executed, but not the Sync(). And in SyncDown() we have finalizers removal already implemented.

maheshbaliga: Before this fix. when the vPVC was deleted, many events occured. Out of which the following can be noted:

1. the Sync() gets called first where the vPV was deleted, vPV & pPV deletion timestamp were checked and pPV was deleted using PhysicalClient.Delete(). Sync() gets called repeatedly and the reconcile request is re-queued until pPV got deleted from the host cluster.
2. the SyncDown() was called when the pPV was deleted; the vPV deletion timestamp was checked and finalizers were removed.

Even though pPV and vPV got deleted, the cloud provider volume deletion was intermittent. So, I think the call to PhysicalClient.Delete() may not be required, since the host cluster takes care of the deletion of the pPV in the case of a dynamically provisioned PV with a reclaim policy of 'Delete'. 

With this fix, when the finalizers are cleared in the Sync(), a call to SyncDown() is not required because the pPV has already been deleted by the host cluster.

matskiv: The previous behavior seems right to me. Why should the vPV be deleted before the pPV is actually gone? If the pPV deletion fails(or gets stuck), the user will never find out.

I still don't understand how deleting finalizers on the vPV helps to ensure that the cloud provider actually deletes the volume. :/

maheshbaliga: @matskiv I have pushed another commit wherein the clearing of the finalizers have been avoided in Sync() and instead we wait for the physical PV to be deleted naturally and let the SyncDown() clear the finalizers on the virtual PV. Kindly let me know of the code changes are OK.

pkg/controllers/resources/persistentvolumes/syncer.go
```diff
@@ -130,13 +130,16 @@ func (s *persistentVolumeSyncer) Sync(ctx *synccontext.SyncContext, pObj client.
 	// check if objects are getting deleted
 	if vObj.GetDeletionTimestamp() != nil {
 		if pObj.GetDeletionTimestamp() == nil {
-			ctx.Log.Infof("delete physical persistent volume %s, because virtual persistent volume is terminating", pObj.GetName())
-			err := ctx.PhysicalClient.Delete(ctx.Context, pObj)
-			if err != nil {
-				return ctrl.Result{}, err
+			// check if the PV is dynamically provisioned and the reclaim policy is Delete
+			if !(vPersistentVolume.Spec.ClaimRef != nil && vPersistentVolume.Spec.PersistentVolumeReclaimPolicy == corev1.PersistentVolumeReclaimDelete) {
```
matskiv, line 134: Could you please explain in a bit more detail why this condition is needed?
Wouldn't this mean that PVs with other reclaim policy and without a claimRef would never get deleted? :thinking: 

Answers, and the PR, are much appreciated! Thank you.

maheshbaliga: The NOT condition in this check will prevent the deletion of a PV which is dynamically provisioned and the reclaim policy is Delete. For all other conditions, it will go ahead and delete the physical PV.

matskiv: Oh damn, I need to clean my glasses :joy:. Sorry, I didn't notice the `!` right after `if` :man_facepalming: 

Okay, just so I understand completely, this will delay the deletion of the dynamically provisioned PVs with the reclaim policy is Delete. Once the vPV is deleted, the `syncUp` will be called and pPV will be deleted at that point?

matskiv: Ok, now it finally clicked. Thank you for the explanation and for your patience!! :raised_hands:

matskiv (approved):
Thank you for the PR! 
Code looks good :+1:

matskiv:
@FabianKramm can you do a final review and merge, please?

FabianKramm (approved):
Sorry for the delay. LGTM!

matskiv:
Hello @maheshbaliga :wave: 

maheshbaliga:
Hey @matskiv, unfortunately, I won't be attending KubeCon. Thanks for thinking of me though! I hope you have a great time at the conference. Wondering if there is any way I can still get the tshirt? :smile:

matskiv:
I don't think we will be able to handle the logistics of this at the moment. Maybe in the future. I hope you understand.
Thank you.

maheshbaliga:
Oh, just kidding! That's totally fine.

PR #720, opened 2022-09-09
Externalize --override-hosts flag of start command in values.yaml for eks distro

### LINKED ISSUE #717: Be able to configure to turn off Init Containers, vcluster-rewrite-hosts

No

Be able to set something like initContainer.enabled: false in the EKS Helm Chart, 
so that vcluster will not inject InitContainer, vcluster-rewrite-hosts with an image library/alpine:3.13.1,
which is not necessary for my case.

_No response_

### Additional context

_No response_

bagulm123:
/kind enhancement

resolves #717

Externalize --override-hosts flag of start command for eks distro


FabianKramm:
Since we already provide a way to configure this through our values, I don't think we need an extra option here.

PR #724, opened 2022-09-12
Provide service creation rights to the service account for correct service-cidr retrieval on vcluster-eks chart

### LINKED ISSUE #704: Init container may detect CIDR range incorrectly
### What happened?

This was originally reported in our Slack here:
https://loft-sh.slack.com/archives/C01N273CF4P/p1657916523372759

An init container for detecting CIDR range of the Service network is writing an incorrect value into the output ConfigMap.
This feature was initially added in PR https://github.com/loft-sh/vcluster/pull/497
However `vcluster get service-cidr` CLI command is detecting the CIDR range correctly.

The `vc-cidr-{{ vcluster name }}` ConfigMap should contain correct value for the `cidr` key.

All the info I have is in [the Slack thread](https://loft-sh.slack.com/archives/C01N273CF4P/p1657916523372759).

Perhaps the logic used in the init container here: https://github.com/loft-sh/vcluster/blob/605eb93a5c7dc02be6b9a18ea46644f38f95f054/pkg/util/servicecidr/servicecidr.go#L36
should import this functionality from [the CLI code](https://github.com/loft-sh/vcluster/blob/605eb93a5c7dc02be6b9a18ea46644f38f95f054/cmd/vclusterctl/cmd/get/service_cidr.go). Seems like there is some difference between the two.

<details>

```console
No info
```

</details>

<details>

```
EKS, maybe other too
```

</details>

### vlcuster version

<details>
```
0.10.2, 0.11.0, 0.11.1
```
</details>

<details>

```
Out of 3 reports, 1 was eks, the others are unspecified.
```

</details>

### OS and Arch

<details>

```
N/A
```
</details>

mabhi:
/kind bugfix

Resolves #704 
While using Helm deployment, a job runs which tries to create a fake service, an error occurs and we extract the correct service CIDR from this error message. This service CIDR gets assigned to the service-cidr configmap. Currently, the assigned value is incorrect because the service account being used did not have the rights to create the service. This causes the default CIDR value to be set to the configmap.

With this fix, the service creation rights is given to the service-account used by the job that allows the init container to get the correct CIDR.

Fixed an issue where the service-account was provided with the service resource creation rights to retrieve the correct CIDR on vcluster-eks.

### DIFF
--- a/charts/eks/templates/pre-install-hook-job-role.yaml
+++ b/charts/eks/templates/pre-install-hook-job-role.yaml
@@ -10,6 +10,6 @@ metadata:
     "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded    
 rules:
   - apiGroups: [""]
-    resources: ["secrets", "configmaps"]
+    resources: ["secrets", "configmaps","services"]
     verbs: ["create", "get", "list"]
 {{- end }}
--- a/charts/k8s/templates/pre-install-hook-job-role.yaml
+++ b/charts/k8s/templates/pre-install-hook-job-role.yaml
@@ -10,6 +10,6 @@ metadata:
     "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 rules:
   - apiGroups: [""]
-    resources: ["secrets", "configmaps"]
+    resources: ["secrets", "configmaps","services"]
     verbs: ["create", "get", "list"]
 {{- end }}


matskiv:
Good catch! :raised_hands: :+1: 

Could you also add the same change to the `charts/k8s/templates/pre-install-hook-job-role.yaml`, please?

mabhi:
Hello @matskiv , Shall I create a new PR for the k8s change ?

matskiv:
No no, just add it to this one, since it's essentially the same change. :)

mabhi:
Hi @matskiv ,
I have pushed the changes for k8s templates as well.

PR #726, opened 2022-09-12
Add loadBalancerIP; fix externalTrafficPolicy & externalIPs logic

TylerGillson:
/kind enhancement

- Adds support for configuring `loadbalancerIP`
- Fixes `externalTrafficPolicy` logic (should be included in NodePort & LoadBalancer cases)
- Fixes `externalIPs` logic. This field should not be tightly bound with the LoadBalancer case. [External IPs](https://kubernetes.io/docs/concepts/services-networking/service/#external-ips) are [typically specified in conjunction with ClusterIP services](https://medium.com/swlh/kubernetes-external-ip-service-type-5e5e9ad62fcd).

Enhanced correctness & flexibility of vcluster service configuration

N/A

### DIFF
--- a/charts/eks/templates/syncer-service.yaml
+++ b/charts/eks/templates/syncer-service.yaml
@@ -15,10 +15,21 @@ spec:
       port: 443
       targetPort: 8443
       protocol: TCP
-  selector:
-    app: vcluster
-    release: {{ .Release.Name }}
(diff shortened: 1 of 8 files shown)


FabianKramm (approved):
@TylerGillson thanks for the PR! Good catch, LGTM!

PR #730, opened 2022-09-13
Download helm binary if its not present while creating/deleting vcluster

mahendrabagul:
/kind enhancement

resolves #

install helm binary if its not present on user's machine while creating/deleting vcluster.

vcluster doesn't need helm installed anymore, it will now download and install helm binary automatically while vcluster is getting created/deleted.


pkg/util/command/prefixed_saver.go
```diff
@@ -0,0 +1,83 @@
...
+// and the last N bytes written to it. The Bytes() methods reconstructs
+// it with a pretty error message.
+type prefixSuffixSaver struct {
+	N         int // max size of prefix or suffix
+	prefix    []byte
+	suffix    []byte // ring buffer once len(suffix) == N
+	suffixOff int    // offset to write into suffix
+	skipped   int64
+
+	// TODO(bradfitz): we could keep one large []byte and use part of it for
```
matskiv, line 18: This^ indicates that the code was copied from somewhere.
Could you please add a comment at the start of the file that links to the source file, ideally with a permanent link to the current commit? This simplifies the review and might be useful for future reference. Please do the same on all files or code chunks that were copied.

matskiv (approved):
Looks good :+1:

PR #731, opened 2022-09-13
feat(chart): helm values to set labels on vclusters pods

matskiv:
/kind enhancement

This allows users to set labels on the pods that are created from the helm chart, including the coredns pod (subject to the usual translation unless `--sync-labels` syncer flag is used).
This is useful for example when Istio sidecar injection is enabled in the namespace where vcluster is installed. Then the user can disable it on the vcluster pods with the `sidecar.istio.io/inject: "false"` label.

chart: Added a helm value to set labels on the vcluster pods.


FabianKramm (approved):

PR #733, opened 2022-09-13
fix: increase minimum coredns resource requests

carlmontanari:
/kind bugfix

resolves N/A

Fixed an issue where vcluster resource limits for coredns were too low.

In GKE we encountered an error `Error syncing to physical cluster: pods "coredns-749456dcbf-8d729-x-kube-system-x-vc-cm-123" is forbidden: [minimum cpu usage per Container is 10m, but request is 3m, minimum memory usage per Container is 32Mi, but request is 16Mi]` which caused coredns pod to never get deployed.

### DIFF
--- a/charts/eks/values.yaml
+++ b/charts/eks/values.yaml
@@ -221,10 +221,12 @@ coredns:
   resources:
     limits:
       cpu: 1000m
-      memory: 170Mi
+      memory: 512Mi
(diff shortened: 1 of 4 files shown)


FabianKramm (approved):
@carlmontanari thanks a lot for the PR! LGTM!

PR #735, opened 2022-09-15
fix: vcluster create error - illegal base64 data at input byte x

matskiv:
/kind bugfix

Solves this error:
```
$ vcluster create vcluster -n vcluster -f values_1.25.yaml
info   Detected local kubernetes cluster minikube. Will deploy vcluster with a NodePort & sync real nodes
info   officially unsupported host server version 1.25, will fallback to virtual cluster version v1.24
fatal  illegal base64 data at input byte 6
```

N/A - this was a regression from the last beta build, I don't think it's necessary to mention a fix for an issue which was not present in the previous GA build.

### DIFF
--- a/cmd/vclusterctl/cmd/create.go
+++ b/cmd/vclusterctl/cmd/create.go
@@ -4,6 +4,12 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
 	"github.com/loft-sh/vcluster/cmd/vclusterctl/cmd/app/localkubernetes"
 	"github.com/loft-sh/vcluster/cmd/vclusterctl/cmd/find"
 	"github.com/loft-sh/vcluster/cmd/vclusterctl/log/survey"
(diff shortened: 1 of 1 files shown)


cmd/vclusterctl/cmd/create.go
```diff
@@ -170,7 +171,7 @@ func (cmd *CreateCmd) Run(args []string) error {
 	for _, value := range cmd.ExtraValues {
 		decodedString, err := getBase64DecodedString(value)
 		if err != nil {
-			if strings.Contains(err.Error(), "illegal base64 data at input byte 0") {
+			if strings.Contains(err.Error(), "illegal base64 data at input byte") {
```
FabianKramm, line 174: Instead of fatal when the base64 fails, can we always add it as a values file and only handle the case if there is no error to continue with the base64 file?

FabianKramm (changes requested):

cmd/vclusterctl/cmd/create.go
matskiv: Makes sense. Done :+1:

FabianKramm (approved):

PR #751, opened 2022-09-22
sync loadbalancer status from host service

### LINKED ISSUE #732: Sync the .status.loadBalancer field of the mapped Services.

Services created by vcluster based on the [configured service mapping](https://www.vcluster.com/docs/architecture/networking#mapping-services-between-vcluster-and-host-cluster) are lacking status, which means that it is not possible to read for example the IP of the assigned load balancer.

Sync the `.status.loadBalancer` field of [the mapped Services](https://www.vcluster.com/docs/architecture/networking#mapping-services-between-vcluster-and-host-cluster).

No viable solution. A workaround is to read the `.status.loadBalancer` from the source Service, but that requires additional knowledge, access, and manual effort from the user.

### Additional context

This is the controller that executes synchronization of the mapped Services:
[https://github.com/loft-sh/vcluster/blob/main/pkg/controllers/servicesync/servicesync.go](https://github.com/loft-sh/vcluster/blob/main/pkg/controllers/servicesync/servicesync.go)

rahulii:
/kind feature
resolves #732 

Services created by vcluster based on the configured service mapping are lacking status. This PR sync the loadbalancer status from host service to service created in vcluster

### DIFF
--- a/pkg/controllers/servicesync/servicesync.go
+++ b/pkg/controllers/servicesync/servicesync.go
@@ -2,6 +2,8 @@ package servicesync
 
 import (
 	"context"
+	"strings"
+
 	"github.com/loft-sh/vcluster/pkg/controllers/resources/services"
 	"github.com/loft-sh/vcluster/pkg/util/loghelper"
 	"github.com/loft-sh/vcluster/pkg/util/translate"
@@ -15,7 +17,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
-	"strings"
 )
 
 type ServiceSyncer struct {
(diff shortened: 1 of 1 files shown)


pkg/controllers/servicesync/servicesync.go
```diff
@@ -195,6 +196,12 @@ func (e *ServiceSyncer) syncServiceAndEndpoints(ctx context.Context, fromService
 		return ctrl.Result{}, nil
 	}
 
+	// sync the loadbalancer status
+	if fromService.Spec.Type == corev1.ServiceTypeLoadBalancer && !apiequality.Semantic.DeepEqual(fromService.Status.LoadBalancer, corev1.LoadBalancerStatus{}) && !apiequality.Semantic.DeepEqual(fromService.Status.LoadBalancer, toService.Status.LoadBalancer) {
```
matskiv, line 200: I am curious to know why is part of the condition needed?
```
!apiequality.Semantic.DeepEqual(fromService.Status.LoadBalancer, corev1.LoadBalancerStatus{})
```

I would expect that if the state changes from having something in the `.Status.LoadBalancer` to having nothing, then we would like to have that reflected in the virtual Service resource too.

rahulii: ah, that makes sense!

rahulii:
@matskiv do we also need tests for this?

matskiv (approved):
LGTM, thank you!

matskiv:
@rahulii The tests would be nice, but since we don't have any tests set up for this controller yet, I don't think it should be mandatory.

rahulii:
> this

cool! Thank for all the help

rahulii:
> @rahulii The tests would be nice, but since we don't have any tests set up for this controller yet, I don't think it should be mandatory.

cool, thanks for all the help :)

matskiv:
Hello @rahulii  :wave: 

PR #435, opened 2022-04-13
introduce isolation.targetNamespace in k0s, k3s, k8s chart values

ishankhare07:
also fix their usage in the chart templates
fixes #425

Signed-off-by: Ishan Khare <me@ishankhare.dev>

**What issue type does this pull request address?** (keep at least one, remove the others) 
/kind bugfix

**What does this pull request do? Which issues does it resolve?** (use `resolves #<issue_number>` if possible) 
resolves #425 

**Please provide a short message that should be published in the vcluster release notes**
fix `isolation.targetNamespace` usage in the chart templates

**What else do we need to know?**

PR #760, opened 2022-09-25
Add option for simpler target namespace configuration

### LINKED ISSUE #659: Enhance Longhorn compability

I have been investigating Longhorn issue where pods was failing to write to disk because volumes was in read only mode https://github.com/longhorn/longhorn/issues/3325#issuecomment-1207912287

However now I did find out that Longhorn have already recovery logic for that situation [documentation](https://longhorn.io/docs/1.3.0/high-availability/recover-volume/) and [actual code](https://github.com/longhorn/longhorn-manager/blob/93063e1b3df7ca3716afef45ecb9c5ed258ce983/controller/kubernetes_pod_controller.go#L163-L185)

**but** that only applies to situation where pod `ownerReferences` is either `kind: Deployment` or `kind: StatefulSet` and currently pods created by vcluster looks to have `kind: Service`

Make pods using vclusters statefulset as owner or alternatively create pause container deployment with zero replicas and make pods created by vcluster pointing to that one.

Alternative would be to add support for this on Longhorn side but I'm not sure how eager they are doing so.

### Additional context

_No response_

olljanat:
/kind feature

resolves #659

Add option `syncer.targetNamespace` for simpler target namespace configuration.

I did find out that Longhorn [has logic](https://github.com/longhorn/longhorn-manager/blob/93063e1b3df7ca3716afef45ecb9c5ed258ce983/controller/kubernetes_pod_controller.go#L284-L303) to remove pods which are failing because of volume mounting issues (which happens example during Longhorn upgrade) but it only works if pods has `ownerReferences` configured which is not case when `--target-namespace` is used.

That why I added logic which create headless service to target namespace which is used as owner for pods.

Additionally `isolation.namespace` setting was deprecated and replaced by `syncer.targetNamespace` (breaking chage) and e2e tests updated to cover these scenarios.

FYI @ishankhare07 because you added `isolation.namespace` in #435 and @pratikjagrut because you added target namespace tests in #626

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -134,6 +134,8 @@ jobs:
         kind load image-archive vcluster_syncer
 
         chmod +x vcluster && sudo mv vcluster /usr/bin
+
+        kubectl create namespace vcluster-workload
         
         vcluster create ${{ env.VCLUSTER_SUFFIX }} -n ${{ env.VCLUSTER_NAMESPACE }} \
         --create-namespace \
--- a/charts/eks/templates/limitrange.yaml
+++ b/charts/eks/templates/limitrange.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: LimitRange
 metadata:
   name: {{ .Release.Name }}-limit-range
-  namespace: {{ .Values.isolation.namespace | default .Release.Namespace }}
+  namespace: {{ .Values.syncer.targetNamespace | default .Release.Namespace }}
 spec:
   limits:
   - default:
--- a/charts/eks/templates/networkpolicy.yaml
+++ b/charts/eks/templates/networkpolicy.yaml
@@ -3,7 +3,7 @@ apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
   name: {{ .Release.Name }}-workloads
-  namespace: {{ .Values.isolation.namespace | default .Release.Namespace }}
+  namespace: {{ .Values.syncer.targetNamespace | default .Release.Namespace }}
 spec:
   podSelector:
     matchLabels:
--- a/charts/eks/templates/rbac/target-ns-role.yaml
+++ b/charts/eks/templates/rbac/target-ns-role.yaml
@@ -1,0 +1,75 @@
+{{- if .Values.syncer.targetNamespace }}
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ .Release.Name }}
+  namespace: {{ .Values.syncer.targetNamespace }}
+  labels:
+    app: vcluster
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+rules:
+  - apiGroups: [""]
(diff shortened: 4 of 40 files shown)


olljanat:
Oh, I forgot one typo to test code when consolidated this to single commit after testing it in my fork. Now it should pass tests.

ishankhare07:
LGTM

ishankhare07 (approved):
LGTM

matskiv:
@olljanat Hi, could you describe your use case for the "target namespace" feature, please?
Currently, this feature is undocumented, and we have not encouraged users to use it(no support in the chart). IMHO, without a strong use case, it would make sense to deprecate it.

matskiv:
@olljanat Thank you for describing your use case. I don't want to break your setup for no good reason, and we don't have any concrete plans for the deprecation of this, so no need to worry yet.
I think that from the technical point of view, this use case is doable with label selectors and a single namespace. So I am still not a fan of encouraging the use of this feature, but I'll see what my other colleagues have to say about this. cc @FabianKramm 

If we ultimately decide not to merge this PR, perhaps your changes could be made intto a separate helm chart which would have vcluster chart as a dependency. How are you deploying your customizations currently?

olljanat:
> I think that from the technical point of view, this use case is doable with label selectors and a single namespace.

On some level yes but checking things like what is status of vcluster it self and its workloads from host cluster would be harder then. Same applies to role delegation, logging, etc where there is need to be able to separate host and vcluster views.

> So I am still not a fan of encouraging the use of this feature

There is already `isolation.namespace` which is useful only if targetNamespace is used and I just made it more generic. However one way to do this would be of course just leave out that `targetNamespace` settings from configuration examples. Then only those who really look how these helm charts are made would find them.

> How are you deploying your customizations currently?

We use updated version of [this](https://github.com/k8s4u/gitops/blob/main/_helm-export/K8s4uHelm.psm1) script which copies Helm charts to our GIT and store them with version number. Then in parallel with those we have another `chart-addons` folder which contains files which gets copied to `templates` folder before deployment. In there we have added those role and role binding creations. Target namespace setting `--target-namespace=` is set just in values.yaml.

> If we ultimately decide not to merge this PR, perhaps your changes could be made intto a separate helm chart which would have vcluster chart as a dependency.

Challenge on that approach is that when ever new features are added or existing refactored in vcluster we would need update our charts also. That has happened couple of times which was fine when this project was in earlier beta phase but now when things are more stable and there are good e2e tests in here I would prefer to have this scenario included to those. That way it will be at least known thing when old settings are deprecated and replaced with new ones, etc.

Also that `ownerReferences` issue cannot be solved without code changes here which would need more logic if vcluster cannot trust that there is that headless service available. Or it would need to be handled with separate plugin which does exactly same thing than vcluster's own logic but only in this scenario. Both of those are of course doable but feels a bit hack IMO

matskiv:
@olljanat I discussed this PR with another maintainer, and for now, we decided to not merge this PR. We will reconsider if more users are making use of this feature, or once additional strong use cases are made known to us.
Also, we noticed that the chart complexity has grown quite a bit, so we must be mindful of that. Once we improve how we manage the chart changes and testing, we will be able to consider more new features for the charts.

But just to reassure you - we currently don't plan to deprecate the "target namespace" feature.

PR #766, opened 2022-09-28
Remove one case of PV provisioning workaround 

### LINKED ISSUE #754: A pod might become unschedulable when using a PV with nodeAffinity
### What happened?

A dynamically provisioned PV that has nodeAffinity set (e.g. because the host cluster spans multiple AZs) might become unusable by the pod that requested it, making the pod unschedulable.
This happens because of our workaround in vcluster, which sets the `volume.kubernetes.io/selected-node` annotation incorrectly. A node from the wrong AZ gets selected, which prevents scheduling. Responsible code is here - https://github.com/loft-sh/vcluster/blob/1d77744033469912abe52a5f46fa5ea074ed37d4/pkg/controllers/resources/persistentvolumeclaims/translate.go#L112-L149

Seems like the workaround code above does not take into account nodeAffinity of the PV, perhaps that would be sufficient to fix the issue. But ideally, the node selection should be done by the scheduler([here](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/volumebinding/binder.go#L410) I belive), so we should investigate what prevents scheduler from doing this correctly, and as a result, forces us to maintain this workaround.

A PV of a StorageClass with `volumeBindingMode: WaitForFirstConsumer` should get dynamically provisioned based on a PVC, and then mounted to a successfully scheduled Pod even in a cluster with topology constraints (e.g. multiple AZs).

Ideally, a cluster with nodes spanning across multiple AZs should be used to reproduce this. But it might be possible with a local multi-node cluster with hostpath provisioned too.
A StorageClass with `volumeBindingMode: WaitForFirstConsumer` should be created. It should be used in a PVC that is used by some Pod.

This was reported on EKS host, and the reporter stated that the issue happens when CSI Addon (aws-ebs-csi-driver [v1.10.0-eksbuild]) is enabled.

Reported on EKS, but applies to all distributions I believe.

### vlcuster version
0.12.0

rohantmp:
Signed-off-by: Rohan CJ <rohan.joseph@loft.sh>

/kind bugfix

A dynamically provisioned PV that has nodeAffinity set (e.g. because the host cluster spans multiple AZs) might become unusable by the pod that requested it, making the pod unschedulable.
This happens because of our workaround in vcluster, which sets the volume.kubernetes.io/selected-node annotation incorrectly. A node from the wrong AZ gets selected, which prevents scheduling.

This PR disables the workaround for when the scheduler is enabled and storageclass sync is on. The workaround is no longer necessary in this case as the scheduler correctly sets the annotation.

resolves #754 

### DIFF
--- a/pkg/controllers/resources/persistentvolumeclaims/syncer.go
+++ b/pkg/controllers/resources/persistentvolumeclaims/syncer.go
@@ -2,6 +2,7 @@ package persistentvolumeclaims
 
 import (
 	"context"
+
 	"github.com/loft-sh/vcluster/pkg/controllers/resources/persistentvolumes"
 	synccontext "github.com/loft-sh/vcluster/pkg/controllers/syncer/context"
 	"github.com/loft-sh/vcluster/pkg/controllers/syncer/translator"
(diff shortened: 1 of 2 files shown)


devspace.yaml
```diff
@@ -42,6 +42,8 @@ deployments:
 
   vcluster-k8s:
     helm:
+      valuesFiles:
+        - tpjsm/values.yaml
```
matskiv, line 46: I think the changes to devspace.yaml were pushed by mistake.
Highlighting just so we don't merge this as is :)

rohantmp:
@matskiv tested the changes, and they should only be active in the verified path :)

pkg/controllers/resources/persistentvolumeclaims/syncer.go
```diff
@@ -35,10 +36,15 @@ const (
 )
 
 func New(ctx *synccontext.RegisterContext) (syncer.Object, error) {
+	storageClassesEnabled := ctx.Controllers["storageclasses"]
+	excludedAnnotations := []string{bindCompletedAnnotation, boundByControllerAnnotation, storageProvisionerAnnotation}
+	if !storageClassesEnabled && ctx.Options.EnableScheduler {
```
matskiv, line 41: I think it would also make sense to check for the `ctx.Controllers["legacy-storageclasses"]` here and in the translateUpdate(). When "legacy-storageclasses" is enabled the StorageClasses will by synced from the host to the vcluster.
But right now we can merge it as is to fix the most pressing problem.

matskiv (approved):
LGTM :+1:

PR #777, opened 2022-10-05
feat(syncer): sync csi objects when scheduler enabled

### LINKED ISSUE #773: Sync all objects required by virtual scheduler when enabled, remove workaround.

There is a workaround that uses custom scheduling logic [here](https://github.com/loft-sh/vcluster/blob/86603ca11282e12dd25ecf7c8649dfc93bf196eb/pkg/controllers/resources/persistentvolumeclaims/translate.go#L113).

The workaround is necessary if the storageclass isn't synced as the scheduler cannot make a decision without the storageclass. 

I'll post a follow up comment with a design proposal.

### Additional context

Other objects might need to be synced for the scheduler to make an informed decision. Scheduler plugins register watches on certain resources [here](https://github.com/kubernetes/kubernetes/blob/ef56451ae52407af16647214865e4818eb2ac958/pkg/scheduler/framework/plugins/volumebinding/volume_binding.go#L95).

rohantmp:
Signed-off-by: Rohan CJ <rohan.joseph@loft.sh>

See description in #773 

/kind bugfix
/kind feature

resolves #773 

Automatically syncs some storage related objects when scheduler is enabled.


rohantmp:
The e2e I want to do is:
- when scheduler enabled:
  - pods PVCs  that use WaitForFirstConsumer (default with kind) should be scheduled fine
  -  when storageclass sync enabled:
    - storageclass should not be upsynced
  - when storageclass sync disabled:
    - storageclass should be upsyced


I have manually tested all these cases, but e2e is to prevent future breakage.
Testing different configurations of deployed vcluster will require additional e2e scaffolding where the setup/teardown is done in ginkgo. This will take longer than expected and I have other pressing tasks.


So currently, I'm just going to cover this case:

- when scheduler enabled:
  - pods PVCs  that use WaitForFirstConsumer (default with kind) should be scheduled fine

rohantmp:
Added the small e2e I mentioned

FabianKramm (approved):

cmd/vcluster/context/context.go
```diff
@@ -162,6 +172,18 @@ func NewControllerContext(currentNamespace string, localManager ctrl.Manager, vi
...
 
+	// enable additional controllers required for scheduling with storage
+	if options.EnableScheduler && controllers["persistentvolumeclaims"] {
+		klog.Infof("persistentvolumeclaim syncing and scheduler enabled, enabling required controllers: %q", schedulerRequiredControllers)
+		for _, ctr := range schedulerRequiredControllers {
+			controllers[ctr] = true
+		}
+		if !(controllers["storageclasses"] || controllers["legacy-storageclasses"]) {
+			klog.Info("persistentvolumeclaim syncing and scheduler enabled, but storageclass sync not enabled. Syncing host storageclasses to vcluster(legacy-storageclasses)")
+			controllers["legacy-storageclasses"] = true
```
matskiv, line 183: Oh, I wish we hadn't named it "legacy-storageclasses", because it feel odd to default to that :laughing: 
But it makes sense.

pkg/controllers/resources/csidrivers/syncer_test.go
```diff
@@ -0,0 +1,147 @@
...
+	})
+}
+
+func int64Ref(i int64) *int64 {
+	return &i
+}
+
+func boolRef(b bool) *bool {
+	return &b
+}
```
matskiv, line 147: btw, there are these utility functions for the same purpose, and we already import them in vcluster - https://pkg.go.dev/k8s.io/utils/pointer#pkg-index
This is just FYI, I am not asking for code change :)

pkg/controllers/resources/csidrivers/translate.go
```diff
@@ -0,0 +1,40 @@
...
+		updated = newIfNil(updated, vObj)
+		updated.Labels = updatedLabels
+		updated.Annotations = updatedAnnotations
+	}
+
+	// for each csidriver in phyical, translate and copy
+
+	if !equality.Semantic.DeepEqual(vObj.Spec, pObj.Spec) {
+		updated = newIfNil(updated, vObj)
+		pObj.Spec.DeepCopyInto(&updated.Spec)
```
matskiv, line 29: I think this is not necessary because `obj.DeepCopy()` used in `newIfNil()` will copy Spec too.

pkg/controllers/resources/csinodes/syncer.go
```diff
@@ -0,0 +1,48 @@
...
+	return &csinodeSyncer{
+		Translator: translator.NewMirrorPhysicalTranslator("csinode", &storagev1.CSINode{}),
+	}, nil
+}
+
+type csinodeSyncer struct {
+	translator.Translator
+}
+
+// TODO, only sync selected nodes
```
matskiv, line 22: Would be nice to create an issue about this and link it from the comment if you are going to do other edits.
This issue could probably be marked as a "good-first-issue", we don't have many of those, so it would be nice.

test/e2e_scheduler/waitforfirstconsumer.go
```diff
@@ -0,0 +1,90 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	ginkgo.It("Wait for Statefulset to become ready", func() {
+		// TODO: initialize a controller-runtime client in the framework
```
matskiv, line 67: There is `f.VclusterClient` which you could use here.

matskiv (approved):
I added some observations, and I'll add one comment with a question separately.
Generally, all looks good :+1:

matskiv:
I noticed that `CSIStorageCapacity` is referencing the `StorageClassName` in the spec. This means that the current implementation is relevant only when we sync StorageClasses from the host to virtual cluster (the "legacy" way), which is now default if not set otherwise. But if the opposite StorageClass sync is enabled, from virtual to host, then the CSIStorageCapacity syncer should ideally also sync from virtual to host + translate the `.spec.StorageClassName`. @rohantmp wdyt?

pkg/controllers/resources/csidrivers/syncer_test.go
rohantmp: Damn it, I'd looked for these and not found them :)

pkg/controllers/resources/csidrivers/translate.go
rohantmp: Oh nice, will change!

pkg/controllers/resources/csinodes/syncer.go
rohantmp: I should get this into this PR, this was a TODO for me :) Good catch!

test/e2e_scheduler/waitforfirstconsumer.go
rohantmp: I think it's a subconscious leftover from porting over controller code that used uncached clients to cached clients. Yeah, should be totally fine to use the client-go client in e2e

rohantmp:
> I noticed that CSIStorageCapacity is referencing the StorageClassName in the spec. This means that the current implementation is relevant only when we sync StorageClasses from the host to virtual cluster (the "legacy" way), which is now default if not set otherwise. But if the opposite StorageClass sync is enabled, from virtual to host, then the CSIStorageCapacity syncer should ideally also sync from virtual to host + translate the .spec.StorageClassName. @rohantmp wdyt?

Good catch! I'll update the PR

rohantmp:
Will push some changes

rohantmp:
Just got a cluster running that actually populates CSIStoragecapacities (using topolvm on minikube with disks attached), looking into properly translating storageClassName.

matskiv:
@rohantmp If I am not mistaken, besides the storageClassName translation this will require an additional syncer for the CSIStorageCapacity. Similar to what we have for StorageClasses. The current implementation would be used with the "legacy-storageclasses" sync (or whatever is the new name - #783 ), and the new one would be used with "storageclasses" sync mode (from virtual to host). I think these syncers should have the same prefix as "storageclasses" one.

rohantmp:
@matskiv , the name is not changing at all, only one of the fields, since none of the identifying fields need to be translated differently, couldn't we simply translate the storageClassName differently depending on which controller is enabled in RegisterContext.Controllers?

Another speedbump I hit is that I didn't realise CSIStorageCapacities are namespaced as I needed to find a CSIDriver that uses them to deploy. 
So they'll be in the namespace that their csi driver is in usually. However, our operator's physical manager has a namespaced cache. So I'm exploring adding a new cache in SourceKindWithCache that watches all namespaces for CSIStorageCapacities in the syncer's ModifyController func.

matskiv:
> the name is not changing at all, only one of the fields, since none of the identifying fields need to be translated differently, couldn't we simply translate the storageClassName differently depending on which controller is enabled in RegisterContext.Controllers?

@rohantmp oh, I misunderstood how the CSIStorageCapacity resource gets created. Yes, this^ makes sense :+1: 
btw: sort of thinking out loud here, when syncing the CSIStorageCapacity resource up, should we filter out all unrelated resources when syncing StorageClasses from the virtual to the host? By unrelated I mean the ones where storageClassName doesn't reference a synced StorageClass. I suppose this may get implemented as a side effect of not being able to reverse the storageClassName to the value used in vcluster if the StorageClass is not coming from the vcluster.

rohantmp:
Since I have more pressing tasks, and I keep missing my estimates on this PR.
I set myself a time limit on how long I'd take to fix this up. A couple of times. 

I realized midway through my efforts that I can't just glue together pieces of the downward namespaced translator, I'll have to write a new one. So, I'll be delaying the few remaining changes required in this PR until I can get my other tasks done. I might sneak in some work here and there though :)

pkg/controllers/resources/csidrivers/translate.go
rohantmp: It won't copy if the obj to be updated is not nil

pkg/controllers/resources/csinodes/syncer.go
rohantmp: Just updated the PR, but forgot this. Will update.

test/e2e_scheduler/waitforfirstconsumer.go
rohantmp: After having worked with the regular client-go client a bit more while I was away from this PR. I now remember why I like the controller-runtime client so much more. It's generic and much more ergonomic!

rohantmp:
- [ ] merge conflict
- [ ] node filter
- [ ] typos
all teensie things, getting to this in to morning.

Estimate grew because I got carried away and unit-tested the controller validation too, and this required a bit of refactoring :monkey:

pkg/controllers/resources/csinodes/syncer.go
rohantmp: Updated the PR with this :)

pkg/controllers/resources/csidrivers/translate.go
matskiv: oh, good point!!

hack/assets/main.go
```diff
@@ -5,18 +5,19 @@ import (
...
-	"github.com/loft-sh/vcluster/pkg/controllers/resources/pods/translate"
 )
 
 func main() {
 	images := []string{}
 
 	// loft
 	images = append(images, "loftsh/vcluster:"+cleanTag(os.Args[1]))
-	images = append(images, translate.HostsRewriteImage)
+	images = append(images, syncercontext.DefaultHostsRewriteImage)
```
rohantmp, line 20: moved this to context along with all the other flag definitions

cmd/vcluster/cmd/start.go
```diff
@@ -86,70 +85,7 @@ func NewStartCommand() *cobra.Command {
 			return ExecuteStart(options)
 		},
 	}
-
-	cmd.Flags().StringVar(&options.KubeConfigContextName, "kube-config-context-name", "", "If set, will override the context name of the generated virtual cluster kube config with this name")
```
rohantmp, line 90: moved this to context/flag_options for unit testing

cmd/vcluster/context/context.go
```diff
@@ -98,50 +19,11 @@ type ControllerContext struct {
 	CurrentNamespace       string
 	CurrentNamespaceClient client.Client
 
-	Controllers map[string]bool
+	Controllers sets.String
 	Options     *VirtualClusterOptions
 	StopChan    <-chan struct{}
 }
 
-var ExistingControllers = map[string]bool{
```
rohantmp, line 106: moved this to new file: `enable_controllers.go` and unit-tested the functionality

cmd/vcluster/context/context.go
```diff
@@ -182,75 +54,6 @@ func NewControllerContext(currentNamespace string, localManager ctrl.Manager, vi
 	}, nil
 }
 
-func parseControllers(options *VirtualClusterOptions) (map[string]bool, error) {
```
rohantmp, line 185: moved to enable_controllers.go and unit tested

pkg/constants/indices.go
```diff
@@ -2,6 +2,7 @@ package constants
 
 const (
 	IndexByPhysicalName  = "IndexByPhysicalName"
+	IndexByVirtualName   = "IndexByVirtualName"
```
rohantmp, line 5: CSIStorageCapacity is namespaced and we sync from Host->Virtual, so we need to look up by virtual name

pkg/controllers/resources/csistoragecapacities/syncer.go
```diff
@@ -0,0 +1,109 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+func New(ctx *synccontext.RegisterContext) (syncer.Object, error) {
+	fmt.Printf("\n\n\nstorageclasssync: %+v\n\n\n", ctx.Controllers["storageclasses"])
```
rohantmp, line 22: caught, this removing

pkg/controllers/resources/csistoragecapacities/translator.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/loft-sh/vcluster/pkg/util/translate"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ syncer.Syncer = &csistoragecapacitySyncer{}
+
+// var _ translator.Translator = &csistoragecapacitySyncer{}
```
rohantmp, line 21: caught this, removing

pkg/controllers/resources/persistentvolumeclaims/translate.go
```diff
@@ -114,45 +113,45 @@ func (s *persistentVolumeClaimSyncer) translateUpdate(ctx *synccontext.SyncConte
...
-					found = true
-					break
-				}
-			}
-			if !found {
-				return nil, fmt.Errorf("couldn't find any virtual nodes in cluster matching storage class topologies")
-			}
-		}
-	}
+	// if !s.storageClassesEnabled && /* the scheduler can make the right decision and set selectedNodeAnnotation if the storageClass is synced */
```
rohantmp, line 116: caught this, removing

rohantmp (commented):
using github to go over things, caught some commented code and debug statements

pkg/controllers/syncer/translator/namespaced_translator.go
```diff
@@ -157,18 +156,18 @@ func (n *namespacedTranslator) TranslateMetadataUpdate(vObj client.Object, pObj
 }
 
 func TranslateMetadataUpdate(vObj client.Object, pObj client.Object, syncedLabels []string, excludedAnnotations ...string) (bool, map[string]string, map[string]string) {
-	updatedAnnotations := translateAnnotations(vObj, pObj, excludedAnnotations)
-	updatedLabels := translateLabels(vObj, pObj, syncedLabels)
+	updatedAnnotations := TranslateAnnotations(vObj, pObj, excludedAnnotations)
```
rohantmp, line 159: exported this for use in `csistoragecapacities/translator.go`

cmd/vcluster/context/context.go
matskiv: We will actually rename it :) - #783

rohantmp:
removed a commented out block
```
-       // if !storageClassesEnabled && ctx.Options.EnableScheduler {
-       //      excludedAnnotations = append(excludedAnnotations, selectedNodeAnnotation)
-       // }
```

pkg/controllers/syncer/testing/testing.go
```diff
@@ -34,6 +37,9 @@ type SyncTest struct {
 	ExpectedPhysicalState map[schema.GroupVersionKind][]runtime.Object
 	ExpectedVirtualState  map[schema.GroupVersionKind][]runtime.Object
 
+	ExpectNotFoundPhysical map[schema.GroupVersionKind][]types.NamespacedName
+	ExpectNotFoundVirtual  map[schema.GroupVersionKind][]types.NamespacedName
```
matskiv, line 41: Is this necessary? :thinking: 
I thought that if the fake client returns resources that are not in ExpectedPhysicalState( or ExpectedVirtualState) then the test will fail with a "... State mismatch ..." message. Some other test cases might already rely on this.

test/e2e_scheduler/waitforfirstconsumer.go
matskiv: I like client-go in the tests because then tests are a bit easier to read, they look a bit closer to the "BDD test cases".
But this is pretty subjective, so no problem with keeping it as is :)

test/e2e_scheduler/waitforfirstconsumer.go
```diff
@@ -0,0 +1,90 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = ginkgo.Describe("Schedule a Statefulset with WaitForFirstConsumer PVCs", func() {
+	// the default kind storageclass is WaitForFirtConsumer
```
matskiv, line 20: > // the default kind storageclass is WaitForFirtConsumer

I don't think we should rely on this^. For example, in minikube, the default StorageClass has `volumeBindingMode: Immediate`, so if someone runs the tests locally, or if we switch the k8s used in the CI, or the KinD default changes - the test would always pass(with potential false positive).

matskiv (changes requested):
The code looks pretty good. Great work!

One missing piece is docs update - ["Synced resources" page](https://www.vcluster.com/docs/architecture/synced-resources) and maybe also [the chapter about vcluster scheduler](https://www.vcluster.com/docs/architecture/scheduling#separate-vcluster-scheduler).
Other than that, I added a few questions and some nit picks, but the docs are main reason for "Requesting changes".

test/e2e_scheduler/waitforfirstconsumer.go
rohantmp: Makes sense! I can fetch and make a copy of the default storageclass that has the correct binding mode.

cmd/vcluster/context/enable_controllers_test.go
```diff
@@ -0,0 +1,135 @@
...
+			desc:           "default case",
+			optsModifier:   func(v *VirtualClusterOptions) {},
+			expectEnabled:  DefaultEnabledControllers.List(),
+			expectDisabled: ExistingControllers.Difference(DefaultEnabledControllers).List(),
+			expectError:    false,
+		},
+		{
+			desc: "scheduler with pvc enabled, nodes not enabled",
+			optsModifier: func(v *VirtualClusterOptions) {
+
```
rohantmp, line 32: I'll remove this newline

cmd/vcluster/context/enable_controllers_test.go
```diff
@@ -0,0 +1,135 @@
...
+				v.Controllers = []string{"persistentvolumeclaims"}
+				v.EnableScheduler = true
+			},
+			expectError: true,
+		},
+		{
+			desc: "scheduler with pvc enabled, storageclasses not enabled",
+			optsModifier: func(v *VirtualClusterOptions) {
+
```
rohantmp, line 41: extra newline

pkg/controllers/resources/csistoragecapacities/syncer.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil || shouldSkip {
+		return ctrl.Result{}, err
+	}
+	ctx.Log.Infof("create CSIStorageCapacity %s, because it does not exist in virtual cluster", vObj.Name)
+	return ctrl.Result{}, ctx.VirtualClient.Create(ctx.Context, vObj)
+}
+
+func (s *csistoragecapacitySyncer) Sync(ctx *synccontext.SyncContext, pObj client.Object, vObj client.Object) (ctrl.Result, error) {
+	// check if there is a change
+	updated, shouldSkip, err := s.translateUpdateBackwards(ctx, pObj.(*storagev1.CSIStorageCapacity), vObj.(*storagev1.CSIStorageCapacity))
```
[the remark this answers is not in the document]
rohantmp: Ah yes, will change!

pkg/controllers/resources/csistoragecapacities/translate.go
```diff
@@ -0,0 +1,95 @@
...
+		err := clienthelper.GetByIndex(ctx.Context, ctx.VirtualClient, sc, constants.IndexByPhysicalName, physName)
+		if s.storageClassSyncEnabled && errors.IsNotFound(err) {
+			return "", true, nil
+		}
+		return sc.Name, false, nil
+	}
+	return physName, false, nil
+}
+
+func (s *csistoragecapacitySyncer) translateBackwards(ctx *synccontext.SyncContext, pObj *storagev1.CSIStorageCapacity) (*storagev1.CSIStorageCapacity, bool, error) {
```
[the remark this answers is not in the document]
rohantmp: Looking into this

test/e2e_scheduler/waitforfirstconsumer.go
matskiv: Yeah, that would be nice. But this is not a blocker, so if you are busy with other things now, we can follow up in a separate PR.

rohantmp:
Refactored, still making the other changes

pkg/controllers/syncer/testing/testing.go
rohantmp: It doesn't check unexpected objects. Just compares the state of expected ones. (code is identical for ExpectedVirtualState)

```
	// Compare states
	if s.ExpectedPhysicalState != nil {
		for gvk, objs := range s.ExpectedPhysicalState {
			err := compareObjs(t, s.Name+" physical state", ctx, pClient, gvk, scheme, objs, s.Compare)
			if err != nil {
				t.Fatalf("%s - Physical State mismatch: %v", s.Name, err)
			}
		}
	}
```

pkg/controllers/resources/csistoragecapacities/translate.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/loft-sh/vcluster/pkg/util/clienthelper"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+)
+
+// returns virtual scname, shouldSkip
+func (s *csistoragecapacitySyncer) fetchVirtualStorageClass(ctx *synccontext.SyncContext, physName string) (string, bool, error) {
+	if s.storageClassSyncEnabled && s.legacyStorageClassSyncEnabled {
+		return "", false, fmt.Errorf("either storageclass or legacy-storageclasses controller should be enabled")
```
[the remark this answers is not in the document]
rohantmp: Yeah, removing

pkg/controllers/resources/csistoragecapacities/translate.go
```diff
@@ -0,0 +1,95 @@
...
+	} else if s.storageClassSyncEnabled {
+		sc := &storagev1.StorageClass{}
+		// the csistorage capacity being synced to the virtual cluster needs the name of the virtual storage cluster
+		err := clienthelper.GetByIndex(ctx.Context, ctx.VirtualClient, sc, constants.IndexByPhysicalName, physName)
+		if s.storageClassSyncEnabled && errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
rohantmp: Yeah, removing

pkg/controllers/syncer/testing/testing.go
matskiv: Yes, you are right. But that would cover our use case if we declare that our expected state for some gvk is no resources.
The `compareObjs` that is called there^ first checks [if the object count matches](https://github.com/loft-sh/vcluster/blob/3a209556835064b76104223bfa0eee429c1f24ee/pkg/controllers/syncer/testing/testing.go#L113), and when it does it will look for every expected obj and [fail if it is not found](https://github.com/loft-sh/vcluster/blob/3a209556835064b76104223bfa0eee429c1f24ee/pkg/controllers/syncer/testing/testing.go#L173-L175).
So I think it should work as long as you include an empty map entry in the ExpectedVirtualState, e.g:
```
ExpectedVirtualState: map[schema.GroupVersionKind][]runtime.Object{
    storagev1.SchemeGroupVersion.WithKind(kind): {},
}
```
so this^ would verify that no resource of storagev1 `kind` exists in the virtual state.

pkg/controllers/resources/csistoragecapacities/translate.go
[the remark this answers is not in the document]
rohantmp: Done

pkg/controllers/resources/csistoragecapacities/syncer.go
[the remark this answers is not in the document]
rohantmp: Done

pkg/controllers/syncer/testing/testing.go
rohantmp: Changed to this :)

rohantmp:
> One missing piece is docs update - ["Synced resources" page](https://www.vcluster.com/docs/architecture/synced-resources) and maybe also [the chapter about vcluster scheduler](https://www.vcluster.com/docs/architecture/scheduling#separate-vcluster-scheduler).
Other than that, I added a few questions and some nit picks, but the docs are main reason for "Requesting changes".

Added this

test/e2e_scheduler/waitforfirstconsumer.go
rohantmp: Fixed!

matskiv (approved):
LGTM! :+1: 
Thank you for addressing all comments :raised_hands:

FabianKramm (approved):

PR #778, opened 2022-10-08
Update cli.mdx

salaboy:
The download URL was wrong

/kind bugfix

resolves #

Fixed an issue where vcluster installing CLI docs


FabianKramm:
@salaboy thanks for the PR! Are you sure this is correct, the new urls don't work for me unfortunately. What error do you get with the existing URLs?

rohantmp:
The old urls are what work for me too, closing.

PR #790, opened 2022-10-16
Fix wrong indentation on init-manifest template

xunleii:
/kind bugfix

Fix the wrong indentation inside the `init-manifest` template

Fixed the wrong indentation inside the `init-manifest` template


FabianKramm (approved):
@xunleii thanks for the PR! LGTM!

PR #792, opened 2022-10-17
Allow templating with chart values inside the init-configmap

xunleii:
/kind enhancement

This PR does not solve any issue, it is rather a suggestion. The goal is to allow the use of `vcluster` as a subchart by adding some templates to the Helm values. With this we can use the actual values to inject them into the Helm map that will be deployed when the vcluster is created.

Allow templating with chart values inside the init-configmap

### DIFF
--- a/charts/eks/templates/init-configmap.yaml
+++ b/charts/eks/templates/init-configmap.yaml
@@ -33,9 +33,10 @@ data:
       {{- if .insecure }}
       insecure: true
       {{- end}}
-      {{- if .values }}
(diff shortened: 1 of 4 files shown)


matskiv (approved):
We already have `.init.manifestsTemplate`, and this addition will complement init helm charts in a similar way. Thank you for contributing.
LGTM :+1:

matskiv:
Hello @xunleii :wave: 

xunleii:
Hi @matskiv
I will be attending this Kubecon, so thank you! :D

PR #793, opened 2022-10-18
add a new flag --ignore-not-found to vclusterctl

### LINKED ISSUE #771: `vcluster delete --ignore-not-found` just like `kubectl`

I'm using vcluster in CI, and I add `vcluster delete` as a cleanup step of my pipeline. Sometimes the pipeline fails before the vcluster gets created, and therefore `vcluster delete` fails with `couldn't find vcluster`.

A new flag like `vcluster delete --ignore-not-found`, similar to `kubectl delete --ignore-not-found`. It would prevent returning a non-zero status code in case the vcluster does not exist already. Also, it would still attempt to delete other resources if told to, like `--delete-namespace` and the PVC, but also would not fail if the PVC or Namespace is not found.

A wrapper for `vcluster delete`:

```
#!/bin/bash

set -eux

code=0

# Ignore when vcluster isn't found
if ! output=$(vcluster delete vcluster --namespace my-vcluster 2>&1) &&
    [[ "${output}" != *"couldn't find vcluster"* ]]; then
    code=1
fi
echo "${output}"

kubectl delete namespace my-vcluster --ignore-not-found || code=1

exit "${code}"
```

### Additional context

*No response*

rahulii:
/kind feature

resolves #771 

Added a new flag `--ignore-not-found` that will not error out in case the vcluster is not found while deleting it

### DIFF
--- a/cmd/vclusterctl/cmd/delete.go
+++ b/cmd/vclusterctl/cmd/delete.go
@@ -34,6 +34,7 @@ type DeleteCmd struct {
 	restConfig *rest.Config
 	kubeClient *kubernetes.Clientset
 	log        log.Logger
+	IgnoreNotFound bool
 }
 
 // NewDeleteCmd creates a new command
@@ -64,6 +65,7 @@ vcluster delete test --namespace test
 	}
 
 	cobraCmd.Flags().BoolVar(&cmd.KeepPVC, "keep-pvc", false, "If enabled, vcluster will not delete the persistent volume claim of the vcluster")
+	cobraCmd.Flags().BoolVar(&cmd.IgnoreNotFound, "ignore-not-found", false, "If enabled, vcluster will not error out in case vcluster does not exists")
 	cobraCmd.Flags().BoolVar(&cmd.DeleteNamespace, "delete-namespace", false, "If enabled, vcluster will delete the namespace of the vcluster")
 	cobraCmd.Flags().BoolVar(&cmd.AutoDeleteNamespace, "auto-delete-namespace", true, "If enabled, vcluster will delete the namespace of the vcluster if it was created by vclusterctl")
 	return cobraCmd
@@ -102,6 +104,12 @@ func (cmd *DeleteCmd) Run(cobraCmd *cobra.Command, args []string) error {
 	cmd.log.Infof("Delete vcluster %s...", args[0])
 	err = helm.NewClient(cmd.rawConfig, cmd.log, helmBinaryPath).Delete(args[0], cmd.Namespace)
 	if err != nil {
+		if cmd.IgnoreNotFound {
+			if kerrors.IsNotFound(err){
+				cmd.log.Info("vcluster not found,--ignore-not-found flag set")
+				return nil
+			}
+		}
 		return err
 	}
 	cmd.log.Donef("Successfully deleted virtual cluster %s in namespace %s", args[0], cmd.Namespace)
--- a/cmd/vclusterctl/cmd/delete.go
+++ b/cmd/vclusterctl/cmd/delete.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
+	"strings"
 
 	"github.com/loft-sh/vcluster/cmd/vclusterctl/cmd/app/localkubernetes"
 	"github.com/loft-sh/vcluster/cmd/vclusterctl/cmd/find"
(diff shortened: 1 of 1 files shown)


FabianKramm:
@rahulii thanks for the PR! Looks like you need to rebase the PR on the current master, would be great if you could do that!

rahulii:
@FabianKramm @matskiv @mahendrabagul  , if the vcluster is not found (it returns deom here https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/delete.go#L160). Here I check if --ignore-not-found flag is set and the error is "Not Found error" to not error out. But from here,  In case user says `vcluster delete my-cluster --namespace my-namespace --delete-namespace --ignore-not-found` the kubeClient is nil , and hence it panics. 
My question is: in case --ignore-not-found is set and vcluster is not found, do we construct a kubeclient from current context  using https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/find/find.go#L42?

matskiv:
@rahulii I think you will need to modify [the prepare function](https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/delete.go#L157) to initialize the cmd.kubeClient before it calls find.GetVCluster

cmd/vclusterctl/cmd/delete.go
```diff
@@ -144,7 +156,11 @@ func (cmd *DeleteCmd) Run(cobraCmd *cobra.Command, args []string) error {
 		err = client.CoreV1().Namespaces().Delete(context.Background(), cmd.Namespace, metav1.DeleteOptions{})
 		if err != nil {
 			if !kerrors.IsNotFound(err) {
-				return errors.Wrap(err, "delete namespace")
+				if cmd.IgnoreNotFound{
```
matskiv, line 159: Same as above - I don't think this is necessary as the line above ignores the "NotFound" error.

rahulii:
> @rahulii I think you will need to modify [the prepare function](https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/delete.go#L157) to initialize the cmd.kubeClient before it calls find.GetVCluster

@matskiv not sure how is that possible? here the [restConfig](https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/delete.go#L174) is fetched using vCluster.ClientFactory.ClientConfig().

rahulii:
> @rahulii I think you will need to modify [the prepare function](https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/delete.go#L157) to initialize the cmd.kubeClient before it calls find.GetVCluster

@matskiv here is the approach I was thinking:
func (cmd *DeleteCmd) prepare(vClusterName string) error {
if cmd.IgnoreNotFound {
   context := CurrentContext()
   // construct kubeclient using this context
} else {
  // current approach 
}

}

matskiv:
@rahulii I think you don't even need to check the flag in the prepare function. Just move this part to the very top of the prepare function, before GetVCluster is called:
```
kubeClient, err := kubernetes.NewForConfig(restConfig)
if err != nil {
  return err
}
cmd.kubeClient = kubeClient
```

rahulii:
> restConfig

rahulii:
> @rahulii I think you don't even need to check the flag in the prepare function. Just move this part to the very top of the prepare function, before GetVCluster is called:
> 
> ```
> kubeClient, err := kubernetes.NewForConfig(restConfig)
> if err != nil {
>   return err
> }
> cmd.kubeClient = kubeClient
> ```

@matskiv umm to create a kubeClient it needs a restConfig which is derived as:
```restConfig, err := vCluster.ClientFactory.ClientConfig()
	if err != nil {
		return err
	}```
here - https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/delete.go#L178

**Comment by matskiv** [comment]
@rahulii yes, good point, I overlook that. But I think we can still do it, we just need to provide the restConfig like so:
```
restConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(clientcmd.NewDefaultClientConfigLoadingRules(), &clientcmd.ConfigOverrides{}).ClientConfig()
```
that^ is pretty much equivalent to the `vCluster.ClientFactory.ClientConfig()`.

matskiv:
Hello @rahulii 
Do you plan to finish this PR or should we unassign you from the issue and close the PR for now? If we close, are you okay with someone else cherry-picking your changes into a new PR and adding the remaining code?

rahulsawra98:
> Hello @rahulii Do you plan to finish this PR or should we unassign you from the issue and close the PR for now? If we close, are you okay with someone else cherry-picking your changes into a new PR and adding the remaining code?

hey @matskiv I will be completing it very soon. Apologies for delay!

rahulii:
> Hello @rahulii Do you plan to finish this PR or should we unassign you from the issue and close the PR for now? If we close, are you okay with someone else cherry-picking your changes into a new PR and adding the remaining code?

Hey @matskiv I am on it. Really sorry for delays!

heiko-braun:
Outdated

PR #794, opened 2022-10-18
feat: add option to disable IngressClasses sync

### LINKED ISSUE #755: Conditionally enable IngressClasses syncing separately from Ingresses

In an environment that uses Ingress resources, you may want to enable ingress syncing. Doing this on v0.12.0 results in the vcluster pods producing the following logout put before retstarting:

```
E0922 15:04:36.385006       1 reflector.go:140] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:262: Failed to watch *v1.IngressClass: failed to list *v1.IngressClass: ingressclasses.networking.k8s.io is forbidden: User "system:serviceaccount:customer1:vc-vcluster" cannot list resource "ingressclasses" in API group "networking.k8s.io" at the cluster scope
```

This is because the serviceaccount doesn't have a ClusterRole allowing it to interact with IngressClasses, which is a cluster-scoped resource.

In our ideal solution, vcluster is entirely contained in it's namespace and doesn't need a clusterrole for IngressClasses that won't be used.

This idea was proposed but was then changed to it's current state in #676.

Ingress and IngressClasses should be conditionally enabled separately to maintain least-privileges.

Creating and binding a clusterrole to the vcluster service account will workaround the issue, but it unnecessarily increases exposure for those who don't use IngressClasses.

### Additional context

vcluster still seems to be usable in this state, but the vcluster pods are regularly restarted.

matskiv:
/kind enhancement

- resolves #755
- removes Ingress permissions from the Role when Ingress sync is not enabled
- updates the Docs to reflect changes made in this PR and in v0.12.0

Chart: Added the sync.ingressclasses.enabled value to allow the user to enable or disable IngressClasses sync independently of Ingress sync configuration.

Part of the code is from the [0ac870bc98c3cb26551278dab6a5c0e4fa508eae](https://github.com/loft-sh/vcluster/commit/0ac870bc98c3cb26551278dab6a5c0e4fa508eae) commit. I had some tooling problems when resolving the merge conflicts, so I apologize for not providing the credit by cherry-picking it properly.


matskiv:
I've tested that this behaves as expected:
```
$ helm template vcluster charts/eks/ | grep ingressclasses
          - --sync=-ingressclasses

$ helm template vcluster charts/eks/ --set sync.ingresses.enabled=true | grep ingressclasses
    resources: ["ingressclasses"]

$ helm template vcluster charts/eks/ --set sync.ingresses.enabled=true --set sync.ingressclasses.enabled=false | grep ingressclasses
          - --sync=-ingressclasses
```

FabianKramm (approved):

PR #798, opened 2022-10-27
fix: correctly support Pods ephemeral volumes

matskiv:
/kind bugfix

Fixes issues arising from incorrect sync of the ephemeral volumes. Previously, an ephemeral volume sync would work under some (lucky) conditions, but it would produce two PVC in the host and the one mounted could have been of the wrong StorageClass. 

**Steps to reproduce**
Deploy the following manifest into a vcluster that has StorageClass sync enabled. You might need to tweak the provisioner field of the storage class. This pod would fail to start without the fix in this PR, and two PVC would exist in the host cluster.
``` yaml
kind: Pod
apiVersion: v1
metadata:
  name: my-app
spec:
  containers:
    - name: my-frontend
      image: busybox:1.28
      volumeMounts:
      - mountPath: "/scratch"
        name: scratch-volume
      command: [ "sleep", "1000000" ]
  volumes:
    - name: scratch-volume
      ephemeral:
        volumeClaimTemplate:
          metadata:
            labels:
              type: my-frontend-volume
          spec:
            accessModes: [ "ReadWriteOnce" ]
            resources:
              requests:
                storage: 1Gi
---
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
  name: my-sc
provisioner: k8s.io/minikube-hostpath
reclaimPolicy: Delete
volumeBindingMode: Immediate
```


yinglunfeng:
亲，您的邮件我已收到，我看到后会尽快恢复您的！

matskiv:
@yinglunfeng If translation is correct, it seems like you have some email auto-reply configured. Please configure it in a way which will not reply to GitHub notifications. Thank you!

FabianKramm (approved):

PR #801, opened 2022-10-29
Add Node Sync on Labels and Pod Workloads

brandonrjacobs:
/kind enhancement

Adds an enhancement to allow nodes to be synced based on the NodeSelector or the pod workloads being run. 

Adds an enhancement to sync nodes with either the nodeSelector, or running pods. Previously if you supplied a nodeSelector it would ignore any nodes needed by a pod. 

The Test Suite should provide adequate coverage of the changes, but since I'm new to your project, it could require some modifications or additions.

### DIFF
--- a/pkg/controllers/resources/nodes/syncer.go
+++ b/pkg/controllers/resources/nodes/syncer.go
@@ -55,6 +55,7 @@ func NewSyncer(ctx *synccontext.RegisterContext, nodeService nodeservice.NodeSer
 	return &nodeSyncer{
 		enableScheduler: ctx.Options.EnableScheduler,
 
+		enforceNodeSelector: ctx.Options.EnforceNodeSelector,
 		nodeServiceProvider: nodeService,
 		nodeSelector:        nodeSelector,
 		useFakeKubelets:     !ctx.Options.DisableFakeKubelets,
@@ -69,8 +70,9 @@ func NewSyncer(ctx *synccontext.RegisterContext, nodeService nodeservice.NodeSer
 type nodeSyncer struct {
 	enableScheduler bool
 
-	nodeSelector    labels.Selector
-	useFakeKubelets bool
+	enforceNodeSelector bool
+	nodeSelector        labels.Selector
+	useFakeKubelets     bool
 
 	physicalClient client.Client
 	virtualClient  client.Client
@@ -291,10 +293,11 @@ func (s *nodeSyncer) shouldSync(ctx context.Context, pObj *corev1.Node) (bool, e
 			ls = labels.Set{}
 		}
 
-		if !s.nodeSelector.Matches(ls) {
+		matched := s.nodeSelector.Matches(ls)
+		if !matched && !s.enforceNodeSelector {
 			return isNodeNeededByPod(ctx, s.virtualClient, s.physicalClient, pObj.Name)
 		}
-		return true, nil
+		return matched, nil
 	}
 
 	return isNodeNeededByPod(ctx, s.virtualClient, s.physicalClient, pObj.Name)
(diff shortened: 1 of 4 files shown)


matskiv:
@brandonrjacobs Thank you for submitting your first pull request for vcluster! Could you please describe your use case for this change?

This nodeSelector is often used with the intention to limit where the vcluster pods will run, so we need to be careful while making changes in this area to maintain the same level of security as expected by the current users.
Also, the documentation explicitly states - `vcluster will only sync nodes that match the given label selector`, so I am pretty hesitant about introducing this change.

salanki:
The `--enforce-node-selector` handles the enforcement, and that functionality is not changed with this PR. Currently, if you do not use `--enforce-node-selector ` you can still schedule on non labeled nodes, they will just not show up in `kubectl get nodes` which is confusing to a user.

I do understand that the docs state `vcluster will only sync nodes that match the given label selector`, and this therefor could be conceived as a change in behavior / breaking change. To not change existing behavior, we could put the addition of syncing dynamic nodes under another flag that can be added as an option to the `nodeSelector`.

matskiv:
You have a good point with the `--enforce-node-selector`. This PR kinda fixes the behavior when `--enforce-node-selector=false` - one would expect to see the node in such case, it's just that the docs don't reflect this option at all. But rather than adding another flag, I would probably go with documenting `--enforce-node-selector=false` in the docs. I'll check with other maintainers for second opinion :slightly_smiling_face: 

I am still curious about the use case. Maybe it is solvable with set-based syntax - `nodeSelector: "nodeKey in (goodValue, alsoGoodValue)"`?

pkg/controllers/resources/nodes/syncer.go
```diff
@@ -291,7 +291,10 @@ func (s *nodeSyncer) shouldSync(ctx context.Context, pObj *corev1.Node) (bool, e
 			ls = labels.Set{}
 		}
 
-		return s.nodeSelector.Matches(ls), nil
+		if !s.nodeSelector.Matches(ls) {
```
matskiv, line 294: The value of [EnforceNodeSelector option (--enforce-node-selector)](https://github.com/loft-sh/vcluster/blob/74ac1af61f884dc46ef14e66f1b203c05ed80aeb/cmd/vcluster/cmd/start.go#L116) should be checked too, we should allow nodes that don't match the selector unless this option is explicitly set to false.

matskiv (changes requested):
We can merge once my code comment is addressed.
An update to docs would be nice to have, to capture the case of nodeLabel + "--enforce-node-selector=false" :)

pkg/controllers/resources/nodes/syncer.go
brandonrjacobs: Thanks for the feedback, making those changes.

salanki: I agree this should be done.
 
There are only two scenarios I can see where a node would end up being included where it shouldn't with `--enforce-node-selector`:
- Syncer was started without `--enforce-node-selector`, or a different set of labels to match on, and then alter restarted with another set of labels while there are already pods running in the cluter.
- A node was re-labeled.

FabianKramm (approved):
@brandonrjacobs thanks for the PR! LGTM!

pkg/controllers/resources/nodes/syncer.go
matskiv: @salanki yes, you described the edge cases well. I don't this we need to solve these, so the current implementation is good! :)

matskiv (approved):
All looks good :+1: 
Thank you for the contribution, and additional thanks for the tests! :)

PR #818, opened 2022-11-11
Fix Chart generation experience for: `devspace dev` and macos 

rohantmp:
/kind bugfix

Fixed an issue whereL

- You needed to run `go generate ./...` within the devspace syncer
- hack/embed-charts depended on realpath (not available by default on MacOS)


matskiv (approved):
LGTM!
Thank you for the PR.

PR #823, opened 2022-11-15
fix: version format for the embeded chart

matskiv:
/kind bugfix

A different approach to fix the same problem that we tried to fix with #819
The charts that we usually create with the release pipeline don't have the `v` prefix for the chart version. The chart version is used internally for the syncer image tag. Using the version with the `v` prefix caused ImagePullBackOff errors.

N/A


FabianKramm (approved):

PR #825, opened 2022-11-16
Docs and dev enhancments

matskiv:
/kind bugfix
/kind documentation

The `devspace dev` was occasionally failing due to coredns manifests being mounted as ReadOnly volume.
The CPU limits are slowing down development when using devspace, so I bumped that up to 100 (I think 0 or empty didn't work when I tried it, but it was some time ago).
Added `&values` anchor to reuse parts of the config (namely helm values) and avoid duplication (yes, I am too lazy to scroll down through duplicate lines :joy: ).

Added regex example to the generic crd sync plugin page because someone mentioned that it would be nice to have :) 

N/A ?

### DIFF
--- a/devspace.yaml
+++ b/devspace.yaml
@@ -41,10 +41,6 @@ deployments:
             extended: true
         syncer:
           image: ${SYNCER_IMAGE}
-          resources:
-            limits:
-              cpu: 100 # to prevent CPU throttling during compilation
-              memory: 2Gi # to prevent OOM kills which can happen during first compilation due to high number of deps
 
   vcluster-k3s:
     helm:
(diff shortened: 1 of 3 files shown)


devspace.yaml
```diff
@@ -63,28 +41,24 @@ deployments:
             extended: true
         syncer:
           image: ${SYNCER_IMAGE}
+          resources:
+            limits:
```
FabianKramm, line 45: Not sure if we want to set these as the cpu limits can lead to the pod not getting scheduled at all and the memory limits could potentially crash the process

matskiv: Ok, I've set no limit now. I wasn't working via values, so I used:
```
dev:
  vcluster:
    resources:
      limits: ...
```

FabianKramm (approved):

PR #837, opened 2022-11-28
add flag to enable hostpath mapping

ishankhare07:
this is wrt the syncer component

/kind bugfix
/kind enhancement

resolves an issue where the hostpaths in pod volumes and volume mounts were being rewritten even when the hestpath mapper component was not enabled from the helm charts.

Fixed an issue where the hostpaths in pod volumes and volume mounts were being rewritten even when the hestpath mapper component was not enabled from the helm charts.


FabianKramm (approved):
@ishankhare07 thanks a lot for the PR! LGTM!

PR #867, opened 2022-12-20
allow setting clusterDomain in k0s helm chart

### LINKED ISSUE #397: clusterDomain hardcoded to `cluster.local` in coredns config
### What happened?

When using the `--cluster-domain` command line parameter to set a custom clusterDomain, this is passed down to the kubelet and certificates but the coredns configuration does not change. This leads to failing lookups inside the cluster using that customDomain.

The coredns configMap should be updated accordingly. Currently its [hardcoded](https://github.com/loft-sh/vcluster/blob/main/manifests/coredns/coredns.yaml#L61), but it should be dynamic like in the [k3s coredns configmap](https://github.com/k3s-io/k3s/blob/master/manifests/coredns.yaml#L60)

start a vcluster with `--custom-domain=my.domain.local` and try a `ping kube-dns.kube-system.svc.my.domain.local` in any pod

There's [https://github.com/loft-sh/vcluster/pull/287](https://github.com/loft-sh/vcluster/pull/287) allowing to add custom imports to the coredns file but this doesn't help as you can't override the existing block.

does not apply

does not apply

### vlcuster version

0\.6.0

all, as the coredns configmap is defined by vcluster

### OS and Arch

any

an-tex:
/kind feature

Partly resolves #397 . In comparison to k3s, the cluster domain cannot be passed as an argument to the [k0s controller](https://docs.k0sproject.io/v1.25.4+k0s.0/cli/k0s_controller/). instead it needs to be defined in the [k0s config](https://docs.k0sproject.io/v1.25.4+k0s.0/configuration/#specnetwork). This PR exposes the `spec.network.clusterDomain` configuration field as a helm value

clusterDomain can be passed to k0s

In case coredns managed by vcluster, a custom coredns.yaml must still be provided as described in https://github.com/loft-sh/vcluster/issues/397

### DIFF
--- a/charts/k0s/templates/secret.yaml
+++ b/charts/k0s/templates/secret.yaml
@@ -36,6 +36,9 @@ stringData:
         serviceCIDR: CIDR_PLACEHOLDER
         {{- end }}
         provider: custom
+        {{- if .Values.vcluster.clusterDomain }}
+        clusterDomain: {{ .Values.vcluster.clusterDomain }}
+        {{- end}}
       controllerManager:
         extraArgs:
           {{- if not .Values.sync.nodes.enableScheduler }}
--- a/charts/k0s/values.yaml
+++ b/charts/k0s/values.yaml
@@ -150,6 +150,7 @@ vcluster:
       cpu: 200m
       memory: 256Mi
   priorityClassName: ""
+  # clusterDomain: cluster.local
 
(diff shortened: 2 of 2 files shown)


FabianKramm (approved):
@an-tex thanks a lot for the PR! LGTM!

PR #870, opened 2022-12-22
add custom clusterDomain to coredns

an-tex:
/kind enhancement

resolves https://github.com/loft-sh/vcluster/issues/397 for k0s 

custom clusterDomain is added to coredns config for k0s clusters

This builds upon https://github.com/loft-sh/vcluster/pull/867 to additionally add the custom clusterDomain to the coredns config. There's no [workaround to build a custom coredns config](https://github.com/loft-sh/vcluster/issues/397#issuecomment-1062870644) necessary any more. Affects only k0s

### DIFF
--- a/charts/k0s/templates/coredns.yaml
+++ b/charts/k0s/templates/coredns.yaml
@@ -74,7 +74,7 @@ data:
             health
             ready
             rewrite name regex .*\.{{ .Release.Name }}\.{{ .Release.Namespace }}\.nodes\.vcluster\.com kubernetes.default.svc.cluster.local
-            kubernetes cluster.local in-addr.arpa ip6.arpa {
+            kubernetes {{- if .Values.vcluster.clusterDomain }} {{ .Values.vcluster.clusterDomain }} {{- end }} cluster.local in-addr.arpa ip6.arpa {
               pods insecure
               fallthrough in-addr.arpa ip6.arpa
             }


charts/k0s/templates/coredns.yaml
```diff
@@ -73,7 +73,7 @@ data:
             errors
             health
             ready
-            kubernetes cluster.local in-addr.arpa ip6.arpa {
+            kubernetes {{- if .Values.vcluster.clusterDomain }} {{ .Values.vcluster.clusterDomain }} {{- end }} cluster.local in-addr.arpa ip6.arpa {
```
FabianKramm, line 76: Why is the `-` needed here in `{{-`, wouldn't `{{ if ` be better here?

FabianKramm:
@an-tex why isn't `coredns.config` an option here? You could just define your own coredns config

charts/k0s/templates/coredns.yaml
an-tex: yep, from the template doc: [`{{- (with the dash and space added) indicates that whitespace should be chomped left, while -}} means whitespace to the right should be consumed.`](https://helm.sh/docs/chart_template_guide/control_structures/#controlling-whitespace). i've tested the output with `helm template` and spacing looked correct for both cases

an-tex:
> @an-tex thanks for the PR! Why isn't `coredns.config` an option here? You could just define your own coredns config

Thanks for looking into this @FabianKramm and sorry for the late reply. Additionally providing an own coredns config is certainly an option but it seems an unnecessary step to me if I already explicitly set the cluster domain as a helm value. This might also be confusing for users with k0s knowledge as when setting the cluster domain for a k0s cluster, the coredns configuration (which is overwritten by vcluster) is adjusted accordingly too.

FabianKramm (approved):

matskiv:
Hello @an-tex :wave: 

an-tex:
Hi @matskiv ! Sorry for the late reply I've been on holidays. Thanks for the offer but I won't be attending. Have fun at the conference :)

PR #878, opened 2022-12-28
feat(nodes): Add filtering to Node Labels on Node Sync

brandonrjacobs:
/kind enhancement

Adds functionality to filter the labels on nodes as they get synced down. Prevents information about the host cluster nodes being leaked.

Add capability to filter out specific labels from Nodes synced into vcluster. 

Could possibly tie this to nodes synced enabled.


brandonrjacobs:
@FabianKramm Good Question, and please help me understand if i'm incorrect as I need to dive into the code more but ideally this situation is where we require all of the "real" node labels and annotations to be present, yet need to hide very select labels that don't apply to those who are using vcluster. This is specifically for multi-tenant environments where we have handful of tenants using the same set of nodes, all using real nodes yet we want to obfuscate certain specifics of the node. 

This is mainly for cases where we are using schedulers inside vcluster.

heiko-braun:
Outdated

PR #882, opened 2022-12-29
[ENHANCEMENT] Exposing hostpathMapper.resources to values.yaml

### LINKED ISSUE #881: Add resource section for hostpath-mapper to values.yaml

I am required to specify resources (requests and limits) in my kubernetes cluster, which I think is not possible for the hostpath-mapper unless I download and edit the helm chart (templates/daemonset-hostpath-mapper.yaml).

Would it maybe make sense to include the resources into the values.yaml?

_No response_

### Additional context

_No response_

cinimins:
/kind enhancement

resolves #881 

hostpathMapper.resources added to values.yaml

I am completely new to helm, kubernetes and vCluster. I am very open to suggestions and improvements.
I have seen commented resource blocks with suggestions for cpu and memory, which seems really helpful for beginners like me. If someone has a suggestion which values are suitable, please edit or comment. The values I added are random and just a placeholder.

### DIFF
--- a/charts/eks/templates/daemonset-hostpath-mapper.yaml
+++ b/charts/eks/templates/daemonset-hostpath-mapper.yaml
@@ -78,6 +78,8 @@ spec:
             mountPath: /tmp/vcluster/{{ .Release.Namespace }}/{{ .Release.Name }}/kubelet/pods
           - name: kubeconfig
             mountPath: /data/server/tls
+        resources:
+{{ toYaml .Values.hostpathMapper.resources | indent 10 }}
       volumes:
         - name: logs
           hostPath:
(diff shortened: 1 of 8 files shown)


matskiv:
Hello, thanks a lot for the PR!

I did some quick load "testing", and with 200 pods deployed into vcluster the memory consumption grew to 35MiB. So something like 40-50 MiB for the requests should be fine for the majority of use cases. Double the memory for the limit, as you did, is a good approach here. The CPU usage was under 10mi most of the time, so 20mi is a good suggestion :+1: 
You can update the comments with memory suggestions based on my testing if you want, but I am happy to merge as is also - just flip it from Draft state to normal (I think there is "Ready for review" button for that). :)

cinimins:
Thank you for the quick test. I updated the resource suggestions and marked the PR as ready for review :)

matskiv (approved):
Thanks again for your PR. LGTM! :)

PR #886, opened 2023-01-03
Include vcluster service name in K3s cert SANs by default

TylerGillson:
/kind enhancement

Ensures that a K3s-based vcluster's Service can be reached by other applications within the host K8s cluster via its DNS name without encountering certificate issues.

Fixed an issue where the vcluster Service name was missing from the K3s certificate SANs

The vcluster Service name is automatically included as a cert SAN when using CNCF K8s. Therefore, for the sake of consistency, it should work the same way with K3s. Otherwise we need to rely on the ClusterIP or manually adding a `--tls-san=<my_vcluster_name>` to the K3s extraArgs every time.

### DIFF
--- a/charts/k3s/templates/statefulset.yaml
+++ b/charts/k3s/templates/statefulset.yaml
@@ -172,6 +172,7 @@ spec:
           {{- else }}
             --service-cidr=$(SERVICE_CIDR)
           {{- end }}
+            --tls-san={{ .Release.Name }}
           {{- range $f := .Values.vcluster.extraArgs }}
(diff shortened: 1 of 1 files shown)


FabianKramm:
Hey @TylerGillson thanks for creating this PR! Shouldn't this already be possible because of https://github.com/loft-sh/vcluster/blob/main/pkg/server/cert/syncer.go#L84 ? If not, we could just add there `translate.Suffix` which is the `{{ .Release.Name }}` and it would work for the other distros as well.

PR #900, opened 2023-01-17
fix: always render the (cluster because multinamespace)role when multi ns enabled

carlmontanari:
/kind bugfix

N/A

Fixed an issue where vcluster doesn't render role needed for multinamespace mode

think I didn't realize before in testing since I copied values I use for home dev vclusters which included ingress sync, which caused that template logic to be true, which made things work.

would it make sense to have a (cluster)role especially for multinamespace mode instead of tacking on extra conditionals to existing roles making things a bit more confusing?


matskiv (approved):
LGTM :) :+1:

PR #903, opened 2023-01-20
chore: update k8s dependencies to v1.26

neogopher:
/kind enhancement

Updated vcluster dependencies to Kubernetes v1.26

Closes ENG-892


pkg/server/filters/nodechanges.go
```diff
@@ -56,7 +56,7 @@ func WithNodeChanges(h http.Handler, uncachedLocalClient, uncachedVirtualClient
 						return
 					}
 
-					responsewriters.WriteObjectNegotiated(s, negotiation.DefaultEndpointRestrictions, corev1.SchemeGroupVersion, w, req, http.StatusOK, updatedNode)
+					responsewriters.WriteObjectNegotiated(s, negotiation.DefaultEndpointRestrictions, corev1.SchemeGroupVersion, w, req, http.StatusOK, updatedNode, false)
```
neogopher, line 59: **responsewriters.WriteObjectNegotiated** now accepts an extra argument _listGVKInContentType_ that controls whether the object is serialized with GVK info.
https://github.com/kubernetes/apiserver/blob/ea4e287db4aaada47d30b9d7b18f831a9729a93e/pkg/endpoints/handlers/responsewriters/writers.go#L289

pkg/util/blockingcacheclient/client.go
```diff
@@ -186,7 +186,16 @@ type CacheStatusClient struct {
 	Cache *CacheClient
 }
 
-func (c *CacheStatusClient) Update(ctx context.Context, obj client.Object, opts ...client.UpdateOption) error {
+func (c *CacheStatusClient) Create(ctx context.Context, obj client.Object, subResource client.Object, opts ...client.SubResourceCreateOption) error {
```
neogopher, line 189: We need to implement the _Create_ method for the **CacheStatusClient** struct for it to comply to the SubResourceWriter interface. as per [alias](https://github.com/kubernetes-sigs/controller-runtime/blob/84c5c9f4e29bfd3cead5c89552c9bc7dc90c45ad/pkg/client/interfaces.go#L132) and [interface](https://github.com/kubernetes-sigs/controller-runtime/blob/84c5c9f4e29bfd3cead5c89552c9bc7dc90c45ad/pkg/client/interfaces.go#L140). 

Just wanted to know, whether it is valid here to have a _Create_ for Status subresource, as we usually avoid that and let the APIServer handle status subresource creation.

matskiv: I don't know if there are cases where this would actually get called, but since we need to fulfill the interface... :man_shrugging:

matskiv (approved):
LGTM

PR #904, opened 2023-01-23
fix: ensure multinamespace role name not broken by truncation

carlmontanari:
/kind bugfix

N/A

Fixed an issue where vclusters in multinamespace mode with long release names would not get the second cluster role created

basically just a better version of the previous "fix", this time hopefully for real :D


matskiv (approved):
Thanks @carlmontanari !

PR #914, opened 2023-01-31
fix: make sure multi namespace suffix is unique

FabianKramm:
/kind bugfix

Fixed an issue where vcluster multi namespace suffixes where not unique


matskiv (approved):
Good catch :+1:

PR #929, opened 2023-02-10
docs: improve steps for using Helm values with CAPI provider

matskiv:
/kind documentation

IMHO significant improvement over the previous command - should work on OSX (which was missing -z flag on sed), supports double quotes inside the values file, and probably covers edge cases I am not even aware of.

N/A - not necessary IMHO

### DIFF
--- a/docs/pages/operator/cluster-api-provider.mdx
+++ b/docs/pages/operator/cluster-api-provider.mdx
@@ -45,7 +45,7 @@ clusterctl generate cluster ${CLUSTER_NAME} \
 
 :::tip
 You can populate the HELM_VALUES variable from a `values.yaml` file using this command:
-```export HELM_VALUES=$(cat values.yaml | sed -z 's/\n/\\n/g')```
+```export HELM_VALUES=$(cat values.yaml | python -c 'import yaml,sys; print(yaml.dump(sys.stdin.read()).strip()[1:-1])')```
 :::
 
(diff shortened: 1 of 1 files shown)


docs/pages/operator/cluster-api-provider.mdx
```diff
@@ -45,7 +45,7 @@ clusterctl generate cluster ${CLUSTER_NAME} \
 
 :::tip
 You can populate the HELM_VALUES variable from a `values.yaml` file using this command:
-```export HELM_VALUES=$(cat values.yaml | sed -z 's/\n/\\n/g')```
+```export HELM_VALUES=$(cat values.yaml | python -c 'import yaml,sys; print(yaml.dump(sys.stdin.read()).strip()[1:-1])')```
```
matskiv, line 48: Just to document this a little bit - the goal is to output values.yaml in a format that can be placed into another yaml file as a single string that will be wrapped in double quotes, essentially - `values: "${HELM_VALUES:=}"`
So the `yaml.dump` will escape all quotes and new lines, but it also wraps the result into quotes (sometimes `"` and sometimes `'`) and adds a new line in the end - these things are solved by `.strip()[1:-1]`.

carlmontanari: LGTM! the only downside I can think of is python == 2.x python but I sincerely hope that isn't the case for most folks (and even if it is they can obviously just set the env var w/ content themselves)

PR #942, opened 2023-02-27
chore: add more security harden settings

sandangel:
Signed-off-by: San Nguyen <vinhsannguyen91@gmail.com>

/kind enhancement

resolves #


charts/k0s/templates/coredns.yaml
```diff
@@ -132,6 +132,7 @@ data:
         spec:
           priorityClassName: "system-cluster-critical"
           serviceAccountName: coredns
+          automountServiceAccountToken: true
```
matskiv, line 135: I noticed that in this file you have not set the `automountServiceAccountToken: false` on the ServiceAccount

charts/k8s/templates/coredns.yaml
```diff
@@ -132,6 +132,7 @@ data:
         spec:
           priorityClassName: "system-cluster-critical"
           serviceAccountName: coredns
+          automountServiceAccountToken: true
```
matskiv, line 135: I noticed that in this file you have not set the `automountServiceAccountToken: false` on the ServiceAccount

charts/k8s/templates/etcd-statefulset.yaml
```diff
@@ -96,7 +96,7 @@ spec:
 {{ toYaml .Values.etcd.nodeSelector | indent 8 }}
       tolerations:
 {{ toYaml .Values.etcd.tolerations | indent 8 }}
-      automountServiceAccountToken: false
+      automountServiceAccountToken: true
```
matskiv, line 99: I don't think etcd requires ServiceAccount token

charts/k8s/templates/scheduler-deployment.yaml
```diff
@@ -89,7 +89,7 @@ spec:
 {{ toYaml .Values.scheduler.nodeSelector | indent 8 }}
       tolerations:
 {{ toYaml .Values.scheduler.tolerations | indent 8 }}
-      automountServiceAccountToken: false
+      automountServiceAccountToken: true
```
matskiv, line 92: I don't think the scheduler requires host ServiceAccount token

charts/k3s/templates/statefulset.yaml
```diff
@@ -194,11 +197,24 @@ spec:
...
+        readinessProbe:
+{{ toYaml .Values.vcluster.readinessProbe | indent 10 }}
+        {{- end }}
         securityContext:
 {{ toYaml .Values.securityContext | indent 10 }}
         volumeMounts:
           - name: config
             mountPath: /etc/rancher
+          - name: kube
+            mountPath: /.kube
```
matskiv, line 217: The tests are failing because the `kube` volume is available only sometimes - `{{- if or .Values.securityContext.runAsUser .Values.securityContext.runAsNonRoot .Values.securityContext.readOnlyRootFilesystem }}` 
This needs to be fixed.

btw, would be nice to add the .kube volume to other charts too, but not a blocker.

matskiv (changes requested):
See review comments below. Main problem is the .kube volume that fails the tests.

charts/k3s/templates/statefulset.yaml
sandangel: I don't think other charts need .kube volume, because they are not generating kubeconfig. I have tried k8s but it doesn't have this issue.

sandangel:
@matskiv thanks for the review comments. I made the changes as required.

heiko-braun:
Outdated

PR #943, opened 2023-02-27
Set runAsGroup field of securityContext for 'vcluster-rewrite-hosts' initContainer

neogopher:
/kind bugfix

Fixed an issue where vcluster would not correctly sync securityContext for StatefulSet workload


sandangel:
Can we also add readonlyrootfilesystem?

FabianKramm:
@sandangel since the container needs to rewrite the /etc/hosts file I don't think that is possible

FabianKramm (approved):

PR #952, opened 2023-03-09
add import capability for cluster scoped CRDs

ishankhare07:
/kind enhancement
/kind feature

resolves ENG-887

Adds capability in vcluster to import cluster scoped resources

### DIFF
--- a/test/e2e_generic/clusterscope/ingressclasses.go
+++ b/test/e2e_generic/clusterscope/ingressclasses.go
@@ -65,10 +65,10 @@ var _ = ginkgo.Describe("Generic sync cluster scoped resources", func() {
 			_, err = f.HostClient.NetworkingV1().IngressClasses().Get(ctx, IngressClassName, metav1.GetOptions{})
 			if err != nil {
 				if kerrors.IsNotFound(err) {
-					return false, nil
+					return false, err
 				}
 
-				return false, err
+				return false, nil
 			}
 
 			return true, nil


pkg/controllers/syncer/syncer.go
```diff
@@ -217,12 +218,19 @@ func (r *syncerController) enqueuePhysical(obj client.Object, q workqueue.RateLi
...
 
 	name := r.syncer.PhysicalToVirtual(obj)
 	if name.Name != "" {
 		q.Add(reconcile.Request{NamespacedName: name})
+	} else if r.options.IsClusterScopedCRD {
+		q.Add(reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Name: obj.GetName(),
+			},
+		})
```
matskiv, line 233: Why is this branch needed? When would the `name.Name` be empty?
Also, considering how import syncer implements `PhysicalToVirtual()` when `r.options.IsClusterScopedCRD` is true, the outcome would be the same as in the branch above :thinking:

pkg/controllers/generic/import_syncer.go
```diff
@@ -72,23 +83,36 @@ func CreateImporters(ctx *context2.ControllerContext, config *config.Config) err
...
+	}
+
+	if scopeAndSubresource, ok := gvkRegister[gvk]; ok {
+		if scopeAndSubresource.IsClusterScoped {
+			syncerOptions.IsClusterScopedCRD = true
+		}
+
+		if scopeAndSubresource.HasStatusSubresource {
+			syncerOptions.HasStatusSubresource = true
+		}
```
matskiv, line 101: Could this be simplified or am I missing something?
```suggestion
		syncerOptions.IsClusterScopedCRD = scopeAndSubresource.IsClusterScoped
		syncerOptions.HasStatusSubresource = scopeAndSubresource.HasStatusSubresource
```

matskiv (changes requested):
Few minor comments, overall looks good, but let's discuss how this syncer should work before merging. I am mainly concerned with the annotation that we add to pObj and the pObj.Delete calls.

pkg/controllers/syncer/syncer.go
ishankhare07: Yea, that's true, I guess I added this later on inside `PhysicalToVirtual()` and missed removing this `else if`

pkg/controllers/generic/import_syncer.go
```diff
@@ -365,6 +401,11 @@ func (s *importer) updateVirtualAnnotations(a map[string]string) map[string]stri
 }
 
 func (s *importer) addAnnotationsToPhysicalObject(ctx *synccontext.SyncContext, pObj client.Object) error {
+	if s.syncerOptions.IsClusterScopedCRD {
+		// do not add annotations to physical object
+		return nil
+	}
```
matskiv, line 407: nit-pick: would be better to check this before calling this function, and leave the function concerned with just adding the annotations

pkg/controllers/generic/import_syncer.go
```diff
@@ -341,6 +372,11 @@ func (s *importer) TranslateMetadata(pObj client.Object) client.Object {
 	nn := s.PhysicalToVirtual(pObj)
 	vObj.SetName(nn.Name)
 	vObj.SetNamespace(nn.Namespace)
+
+	if s.syncerOptions.IsClusterScopedCRD {
+		vObj.SetLabels(nil)
+	}
```
matskiv, line 378: Q: What is the reason for removing the Labels?
I could imagine them being useful, e.g. resources that are exported might have LabelSelectors that would be selecting on these cluster-scoped imported resources. AFAIK Prometheus uses this.

pkg/controllers/generic/import_syncer.go
```diff
@@ -244,7 +265,7 @@ func (s *importer) Sync(ctx *synccontext.SyncContext, pObj client.Object, vObj c
 
 	// check if either object is getting deleted
 	if vObj.GetDeletionTimestamp() != nil || pObj.GetDeletionTimestamp() != nil {
-		if pObj.GetDeletionTimestamp() == nil {
+		if pObj.GetDeletionTimestamp() == nil && !s.syncerOptions.IsClusterScopedCRD {
```
matskiv, line 268: Could we also add the same condition around the `ctx.PhysicalClient.Delete()` call in the SyncUp() ?
I know that it will not be called because the annotation is not added, but since we don't have any tests, adding a condition could decrease the chance we allow the Deletion to happen after some future code changes.
Actually, now that I mention tests, we should probably have some at least to ensure that Delete / Update does not happen, as the consequences could be quite bad. Wdyt?

matskiv (commented):
Looks good overall, just two questions and nit-pick :see_no_evil: :)

pkg/controllers/generic/import_syncer.go
ishankhare07: actually this function is getting called from more than 1 places, at-least 2 if I recall correctly
hence for the sake of simplicity I put this check here. But if you strongly feel we should do it the other way I can change it

pkg/controllers/generic/import_syncer.go
ishankhare07: > * Only support 1:1 name mappings without any label selectors  

ref: point 1 from ENG-887

matskiv: Syncing labels on the resources is not the same as syncing based on the labels. This code is not related to label selectors as far as I can tell.

pkg/controllers/generic/import_syncer.go
matskiv: I don't feel that strongly about this :)

test/e2e_generic/clusterscope/ingressclasses.go
```diff
@@ -0,0 +1,97 @@
...
+		err := f.VclusterClient.NetworkingV1().IngressClasses().Delete(ctx, IngressClassName, metav1.DeleteOptions{})
+		framework.ExpectNoError(err)
+
+		// should not delete the physical ingress class
+		err = wait.PollImmediate(time.Millisecond*500, framework.PollTimeout, func() (bool, error) {
+			_, err = f.HostClient.NetworkingV1().IngressClasses().Get(ctx, IngressClassName, metav1.GetOptions{})
+			if err != nil {
+				if kerrors.IsNotFound(err) {
+					return false, nil
+				}
```
matskiv, line 69: Why is this check required? :thinking:

ishankhare07: Yea I think it should be the otherway around
```
if kerrors.IsNotFound(err) {
    return false, err   // error for not aligning with assert condition
}

return false, nil   // some other error while fetching, simply retry getting till timeout
```

matskiv: Yes, I agree

matskiv (approved):
LGTM. Thanks for the e2e tests! :+1:

PR #961, opened 2023-03-16
Allow accessing host services in vcluster without cluster role

### LINKED ISSUE #951: Allow map host namespace svc to vcluster without cluster role

- Cluster role is required for service mapping from host namespce to vcluster, even with the service inside the same namespace with vcluster.

- Allow using Role instead of cluster role to map service in host namespace where vcluster is installed.

- Alternatively, I think we can configure coredns to resolve dns request directly, or forward it to the host cluster dns server. This might replace the approach of current service mapping.

### Additional context

pods in vcluster can not access istio mesh on

- My vcluster pods have istio-proxy sidecar injected by 
- Vcluster translate the address of service in the mesh to IP address when pod try to looks for 
- pods on vcluster can not connect to that service because the traffic should be routed through the sidecar  istio-proxy

neogopher:
/kind enhancement

resolves #951

Allow accessing host services in vcluster without cluster role if the source service lies in the vcluster namespace

Closes ENG-1093

### DIFF
--- a/charts/eks/templates/_helpers.tpl
+++ b/charts/eks/templates/_helpers.tpl
@@ -43,7 +43,7 @@ Whether to create a cluster role or not
 {{- define "vcluster.createClusterRole" -}}
 {{- if or
     (not
-        (empty (include "vcluster.serviceMapping.fromHost" . )))
+        (empty (include "vcluster.serviceMapping.fromHost.otherNamespace" . )))
     (not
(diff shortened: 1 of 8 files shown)


Interaze:
Would love to see this. Especially w.r.t SyncAllNodes, and Sync Ingresses

heiko-braun:
Outdated

PR #964, opened 2023-03-19
fix: cert expiry should check for cert's expiry date

neurodrone:
/kind bugfix

The `IsCertExpired` check that is used to determine if certificates need to be regenerated is currently broken. The current check doesn't check for cert's expiration date. This change adds a fix to that part of the code while also adding a test to make sure any future refactors don't cause regression.

Fixed an issue where vcluster has a bug for checking expiration of certs.

Hopefully the added test should catch if this issue should arise in future.


FabianKramm (approved):
@neurodrone thanks a lot for the PR! Good catch!

matskiv:
Hello @neurodrone 

neurodrone:
Hi @matskiv, thanks for the offer! I'd have loved to but currently I have no plans to travel in April. Hopefully the next Kubecon though! :)

PR #970, opened 2023-03-24
enable optional generic resource syncing

ishankhare07:
/kind enhancement

resolves ENG-997 

Fixed an issue where vcluster would fail in case the specified resource in the import/export generic config does not exist on host cluster. We now have the concept of `optional` resources


FabianKramm (approved):
@ishankhare07 thanks a lot for the PR! LGTM!

PR #978, opened 2023-03-29
images: replace docker.io references with ghcr.io

rohantmp:
Will rebase and test after we do a beta release with pushed images.

TODO: review docs

Signed-off-by: Rohan CJ <rohantmp@gmail.com>

/kind enhancement

resolves #

### DIFF
--- a/assets/push-images.sh
+++ b/assets/push-images.sh
@@ -75,7 +75,7 @@ for i in "${linux_images[@]}"; do
         image_name="${reg}/${i}"
         ;;
     *)
-        image_name="${reg}/loftsh/${i}"
+        image_name="${reg}/ghcr.io/loft-sh/${i}"
         ;;
     esac
 
--- a/charts/eks/templates/daemonset-hostpath-mapper.yaml
+++ b/charts/eks/templates/daemonset-hostpath-mapper.yaml
@@ -47,7 +47,7 @@ spec:
         {{- if .Values.hostpathMapper.image }}
         image: "{{ .Values.defaultImageRegistry }}{{ .Values.hostpathMapper.image }}"
         {{- else }}
-        image: "{{ .Values.defaultImageRegistry }}loftsh/vcluster:{{ .Chart.Version }}"
+        image: "{{ .Values.defaultImageRegistry }}ghcr.io/loft-sh/vcluster:{{ .Chart.Version }}"
         {{- end }}
         {{- if .Values.syncer.workingDir }}
         workingDir: {{ .Values.syncer.workingDir }}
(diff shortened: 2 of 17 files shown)


charts/eks/templates/daemonset-hostpath-mapper.yaml
```diff
@@ -47,7 +47,7 @@ spec:
         {{- if .Values.hostpathMapper.image }}
         image: "{{ .Values.defaultImageRegistry }}{{ .Values.hostpathMapper.image }}"
         {{- else }}
-        image: "{{ .Values.defaultImageRegistry }}loftsh/vcluster:{{ .Chart.Version }}"
+        image: "{{ .Values.defaultImageRegistry }}ghcr.io/loft-sh/vcluster:{{ .Chart.Version }}"
```
matskiv, line 50: I am not sure this will work correctly when `{{ .Values.defaultImageRegistry }}` is set :thinking: 
When it is set, the `ghcr.io/` should probably be omitted. 
And when the `{{ .Values.defaultImageRegistry }}` is set, should this be backward compatible - "loftsh" org without a dash? @FabianKramm wdyt?

assets/push-images.sh
```diff
@@ -75,7 +75,7 @@ for i in "${linux_images[@]}"; do
         image_name="${reg}/${i}"
         ;;
     *)
-        image_name="${reg}/loftsh/${i}"
+        image_name="${reg}/ghcr.io/loft-sh/${i}"
```
matskiv, line 78: similar to my other comment, are we sure that "ghcr.io" part should be here? Will it work with all image registries?

matskiv (changes requested):

charts/eks/templates/daemonset-hostpath-mapper.yaml
rohantmp: Yeah, was initially confused by this as well, but the image `ghcr.io/loft-sh/vcluster ` when used with `push-images.sh --registry ghcr.io/rohantmp` will be pushed to `ghcr.io/rohantmp/ghcr.io/loft-sh/vcluster` and that is by design.

assets/push-images.sh
rohantmp: This is for consistency with the other images that hit the other case above. TBH, none of the images we list will match this regex since they all have a slash.

rohantmp:
Docker docs for image name specs: https://docs.docker.com/engine/reference/commandline/tag/#description

FabianKramm (approved):

PR #983, opened 2023-03-29
reimplement fake kubelet services for node InternalIP addresses

ishankhare07:
this is needed for proper resolution of kubelet metrics targets for prometheus

/kind bugfix

resolves ENG-1197

Fixed an issue where prometheus was not able to get kubelet metrics because the fake nodes did not have `Addresses` of type `InternalIP`

### DIFF
--- a/cmd/vcluster/context/flag_options.go
+++ b/cmd/vcluster/context/flag_options.go
@@ -39,6 +39,7 @@ type VirtualClusterOptions struct {
 	SyncAllNodes        bool     `json:"syncAllNodes,omitempty"`
 	EnableScheduler     bool     `json:"enableScheduler,omitempty"`
 	DisableFakeKubelets bool     `json:"disableFakeKubelets,omitempty"`
+	FakeKubeletIPs      bool     `json:"fakeKubeletIPs,omitempty"`
 	ClearNodeImages     bool     `json:"clearNodeImages,omitempty"`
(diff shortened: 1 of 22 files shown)


cmd/vcluster/context/flag_options.go
```diff
@@ -121,6 +122,7 @@ func AddFlags(flags *pflag.FlagSet, options *VirtualClusterOptions) {
 	flags.BoolVar(&options.SyncAllNodes, "sync-all-nodes", false, "If enabled and --fake-nodes is false, the virtual cluster will sync all nodes instead of only the needed ones")
 	flags.BoolVar(&options.EnableScheduler, "enable-scheduler", false, "If enabled, will expect a scheduler running in the virtual cluster")
 	flags.BoolVar(&options.DisableFakeKubelets, "disable-fake-kubelets", false, "If disabled, the virtual cluster will not create fake kubelet endpoints to support metrics-servers")
+	flags.BoolVar(&options.FakeKubeletIPs, "fake-kubelet-ips", true, "If enabled, virtual cluster will assign fake ips of type NodeInternalIP to fake the kubelets")
```
ishankhare07, line 125: @FabianKramm should this be enabled but default or should it be explicitly enabled? I'll set the value accordingly in the chart values as well

FabianKramm (approved):

PR #990, opened 2023-04-04
docs: Mention cloud costs for load balancer

kostis-codefresh:

/kind documentation

resolves #

It adds a warning about costs for load balancers

I don't think it needs to be mentioned in the release notes.

Following the documentation right now can be very expensive for people that use Cloud vendors and use multiple virtual clusters.


rohantmp (approved):
LGTM! Thank you for the PR :)

PR #993, opened 2023-04-07
Disable node/proxy functionality in Isolated mode

### LINKED ISSUE #937: Disable node/proxy functionality in Isolated mode and add a flag to optionally enable it

The `node/proxy` functionality in Kubernetes can potentially allow circumvention of the security measures introduced by the Isolated mode of vcluster. AFAIK it is mitigated when "fake nodes" are used, but not with the "real nodes" sync.

The `node/proxy` RBAC permission should not be added to the ClusterRole in the Helm template whenever isolated mode is enabled.
A new Helm value should be added to optionally allow original behavior even in the isolated mode.

Manually adjusting the ClusterRole to remove the `node/proxy` RBAC permission.

### Additional context

This can be considered a breaking change and should not be included only in a new minor release + it should be highlighted in the release notes.

sigmabaryon:
/kind feature

resolves #937

Remove the node/proxy permission from the cluster role when isolated mode is enabled.
Add a new Helm value to optionally allow original behavior even in the isolated mode.

### DIFF
--- a/charts/eks/templates/rbac/clusterrole.yaml
+++ b/charts/eks/templates/rbac/clusterrole.yaml
@@ -18,7 +18,9 @@ rules:
     resources: ["nodes", "nodes/status"]
     verbs: ["get", "watch", "list"]
   - apiGroups: [""]
-    resources: ["pods", "nodes/proxy", "nodes/metrics", "nodes/stats"]
+    resources: [ "pods",{{- if .Values.isolation.enabled }}
+      {{- if .Values.isolation.nodeProxyPermission.enabled}} "node/proxy",{{- end}}
+      {{- else}} "node/proxy",{{- end }} "nodes/metrics", "nodes/stats"]
     verbs: ["get", "watch", "list"]
   {{- end }}
   {{- if or (and .Values.sync.nodes.enabled .Values.sync.nodes.syncNodeChanges) .Values.rbac.clusterRole.create }}
--- a/charts/eks/templates/rbac/clusterrole.yaml
+++ b/charts/eks/templates/rbac/clusterrole.yaml
@@ -18,9 +18,12 @@ rules:
     resources: ["nodes", "nodes/status"]
     verbs: ["get", "watch", "list"]
   - apiGroups: [""]
-    resources: [ "pods",{{- if .Values.isolation.enabled }}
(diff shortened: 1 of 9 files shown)


charts/eks/templates/rbac/clusterrole.yaml
```diff
@@ -18,7 +18,9 @@ rules:
     resources: ["nodes", "nodes/status"]
     verbs: ["get", "watch", "list"]
   - apiGroups: [""]
-    resources: ["pods", "nodes/proxy", "nodes/metrics", "nodes/stats"]
+    resources: [ "pods",{{- if .Values.isolation.enabled }}
+      {{- if .Values.isolation.nodeProxyPermission.enabled}} "node/proxy",{{- end}}
+      {{- else}} "node/proxy",{{- end }} "nodes/metrics", "nodes/stats"]
```
matskiv, line 23: I know the PR is still in draft state, but adding my 2cents to potentially save one review iteration :) 
I'd rather see `"nodes/proxy"` permission in a separate rule, which can then be wrapped in conditions. That would be more readable.

matskiv:
@sigmabaryon Thank you for the PR. Could you please leave a comment on the related issue just so I can assign it to you? :)

charts/eks/templates/rbac/clusterrole.yaml
sigmabaryon: Have added a new rule for node/proxy permission for readability.

matskiv:
I saw this error in the failing check for k3s:
```
 fatal  error executing helm upgrade vcluster ./charts/k3s --kubeconfig /tmp/1434243732 --namespace vcluster --install --repository-config='' --values /tmp/3549946504 --values ./test/commonValues.yaml --values ./test/e2e_node/values.yaml: Release "vcluster" does not exist. Installing it now.
Error: YAML parse error on vcluster/templates/rbac/clusterrole.yaml: error converting YAML to JSON: yaml: line 15: mapping values are not allowed in this context
```
Other charts probably have the same problem. You can check the "Create vcluster" step to see the error.

sigmabaryon:
@matskiv I've fixed the issue related to the error reported in the tests.

FabianKramm:
@sigmabaryon thanks, but it seems e2e tests are still failing, would you mind taking another look?

Issue #968, opened 2023-03-22
Move vcluster service account token from pod annotation to secret


FabianKramm:
Right now we are writing the internal vcluster service account token into a pod annotation on the host pod manifest. From within the vcluster this is not viewable, but under some circumstances this could potentially lead to problems in the host cluster as users in the host cluster that can list pods could view this token directly or it could show up unintentionally inside an audit log or similar on the host cluster.

Instead we will create a separate secret for each pod that holds this token and users on the host cluster would need additional permissions to view this token. We will also add an owner reference to the pod to the created secret to make sure its auto cleaned up when the corresponding pod is deleted.

Other things we need to take care of:

* Make sure this secret is not synced back in multi namespace mode when importing secrets from the host cluster. We probably can use an annotation on the secret there to exclude it from syncing.
* The logic of creating this secret should be taken care of within the pod syncer and we don't want a separate controller.
* If the pod defines multiple projected service account volumes, we should save all these tokens within a single secret to avoid cluttering the host cluster with secrets.
* For pods that are still using the old approach, we should not create a secret. Only if they are recreated a secret should get created as well.

PR #996, opened 2023-04-10
move service account token to secret

ishankhare07:
add to index for secret syncer

/kind bugfix
/kind enhancement

resolves  ENG-1172 #968

Fixed an issue where vcluster service account token was mounted using downward API and annotation. We are now instead doing it with a dedicated secret containing all projected service account tokens

### DIFF
--- a/pkg/controllers/generic/import_syncer.go
+++ b/pkg/controllers/generic/import_syncer.go
@@ -176,14 +176,20 @@ var _ syncer.UpSyncer = &importer{}
 
 func (s *importer) SyncUp(ctx *synccontext.SyncContext, pObj client.Object) (ctrl.Result, error) {
 	// check if annotation is already present
-	if pObj.GetAnnotations() != nil &&
-		pObj.GetAnnotations()[translate.ControllerLabel] == s.Name() &&
-		!s.syncerOptions.IsClusterScopedCRD { // only delete pObj if its not cluster scoped
-		err := ctx.PhysicalClient.Delete(ctx.Context, pObj)
-		if err != nil && !kerrors.IsNotFound(err) {
-			return ctrl.Result{}, err
+	if pObj.GetAnnotations() != nil {
+		if pObj.GetAnnotations()[translate.ControllerLabel] == s.Name() &&
+			!s.syncerOptions.IsClusterScopedCRD { // only delete pObj if its not cluster scoped
+			err := ctx.PhysicalClient.Delete(ctx.Context, pObj)
+			if err != nil && !kerrors.IsNotFound(err) {
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{}, nil
 		}
-		return ctrl.Result{}, nil
+
+		// check if back sync is disabled eg. for service account token secrets
+		if pObj.GetAnnotations()[translate.SkipBacksyncInMultiNamespaceMode] == "true" {
+			return ctrl.Result{}, nil
+		}
 	}
 
 	// add annotation to physical resource to mark it as controlled by this syncer
@@ -264,6 +270,9 @@ func (s *importer) Sync(ctx *synccontext.SyncContext, pObj client.Object, vObj c
 	if err != nil {
 		return ctrl.Result{}, err
 	} else if !managed {
+		return ctrl.Result{}, nil
+	} else if pObj.GetAnnotations() != nil &&
+		pObj.GetAnnotations()[translate.SkipBacksyncInMultiNamespaceMode] == "true" {
 		return ctrl.Result{}, nil
 	}
 
--- a/pkg/controllers/generic/import_syncer.go
+++ b/pkg/controllers/generic/import_syncer.go
@@ -152,6 +152,12 @@ func (s *importer) ExcludePhysical(pObj client.Object) bool {
 }
 
 func (s *importer) excludeObject(obj client.Object) bool {
+	// check if back sync is disabled eg. for service account token secrets
+	if obj.GetAnnotations() != nil &&
+		obj.GetAnnotations()[translate.SkipBacksyncInMultiNamespaceMode] == "true" {
+		return true
+	}
+
 	if obj.GetLabels() != nil &&
 		obj.GetLabels()[translate.ControllerLabel] != "" {
 		return true
@@ -183,11 +189,6 @@ func (s *importer) SyncUp(ctx *synccontext.SyncContext, pObj client.Object) (ctr
 			if err != nil && !kerrors.IsNotFound(err) {
 				return ctrl.Result{}, err
 			}
-			return ctrl.Result{}, nil
-		}
-
-		// check if back sync is disabled eg. for service account token secrets
-		if pObj.GetAnnotations()[translate.SkipBacksyncInMultiNamespaceMode] == "true" {
 			return ctrl.Result{}, nil
 		}
 	}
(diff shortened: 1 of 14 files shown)


pkg/controllers/generic/import_syncer.go
```diff
@@ -176,14 +176,20 @@ var _ syncer.UpSyncer = &importer{}
...
+			if err != nil && !kerrors.IsNotFound(err) {
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{}, nil
+		}
+
+		// check if back sync is disabled eg. for service account token secrets
+		if pObj.GetAnnotations()[translate.SkipBacksyncInMultiNamespaceMode] == "true" {
+			return ctrl.Result{}, nil
 		}
```
matskiv, line 192: I think we could move this to `IsManaged()`, and the resource wouldn't even get put into reconcile queue. Wdyt?

pkg/controllers/resources/pods/translate/sa_token_secret.go
```diff
@@ -0,0 +1,82 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	SkipSATokenSecretBacksyncAnnotation = "vcluster.loft.sh/skip-sa-secret-backsync"
```
matskiv, line 17: This is unused?

pkg/controllers/resources/pods/translate/sa_token_secret.go
```diff
@@ -0,0 +1,82 @@
...
+						Name:       vPod.Name,
+						UID:        vPod.UID,
+					},
+				},
+			},
+			Type:       corev1.SecretTypeOpaque,
+			StringData: tokens,
+		}
+
+		err := vClient.Create(ctx, secret)
```
matskiv, line 73: Why are we creating the secret in virtual cluster instead of creating directly in the host?

matskiv (changes requested):
This works, but I have a few questions and potential improvement suggestions.

pkg/controllers/resources/pods/translate/sa_token_secret.go
ishankhare07: good catch, I think we can directly create this in host cluster

pkg/controllers/generic/import_syncer.go
ishankhare07: yea, I guess I'll just move it to `excludeObject` and then its called in `IsManaged`

pkg/controllers/resources/pods/translate/sa_token_secret.go
matskiv: Yeah, let's do that. Hiding the secret from the user has a few small benefits, and this will be a bit less overhead on the virtual API server.

ishankhare07: also I was worried about this case which I'm reminded of now, what if the vcluster used deleted the secret (there's no controller reconciling on it here). But creating the secret directly on physical solves this challenge

pkg/controllers/resources/pods/translate/translator.go
```diff
@@ -414,7 +431,22 @@ func (t *translator) translateVolumes(pPod *corev1.Pod, vPod *corev1.Pod) error
...
 }
 
-func (t *translator) translateProjectedVolume(projectedVolume *corev1.ProjectedVolumeSource, pPod *corev1.Pod, vPod *corev1.Pod) error {
+func (t *translator) shouldCreateTokenSecret(vPod *corev1.Pod) bool {
+	// if the automount field is not set at all (which means true by default)
+	// or if set, then should be true, skip otherwise
+	if vPod.Spec.AutomountServiceAccountToken != nil &&
+		!*vPod.Spec.AutomountServiceAccountToken {
+		return false
+	}
```
matskiv, line 440: I am not 100% sure if this is a reliable way to check if a SA token is needed.
It would be best to check if there is a projected volume of `serviceAccountToken` type.
AFAIK, when `AutomountServiceAccountToken != false` the pod has a serviceAccountToken projected volume added in spec, but it might be added for a different reason too.

pkg/controllers/resources/pods/translate/sa_token_secret.go
```diff
@@ -0,0 +1,144 @@
...
+				return err
+			}
+			secret.SetOwnerReferences([]metav1.OwnerReference{
+				{
+					APIVersion: typeAccessor.GetAPIVersion(),
+					Kind:       typeAccessor.GetKind(),
+					Name:       translate.Owner.GetName(),
+					UID:        translate.Owner.GetUID(),
+				},
+			})
```
matskiv, line 79: nit-pick: can be done with the util function [here](https://github.com/loft-sh/vcluster/blob/29acb7ccdb845fd5f6c7738a5eb9aacc1fa5cd2f/pkg/util/translate/translate.go#L41)

```suggestion
			secret.SetOwnerReferences(translate.GetOwnerReference(nil))
```

pkg/controllers/resources/pods/translate.go
```diff
@@ -52,6 +53,24 @@ func (s *podSyncer) getK8sIPDNSIPServiceList(ctx *synccontext.SyncContext, vPod
 }
 
 func (s *podSyncer) translateUpdate(pObj, vObj *corev1.Pod) (*corev1.Pod, error) {
+	ctrlRuntimeClient, err := client.New(s.physicalClusterConfig, client.Options{})
+	if err != nil {
+		return nil, err
+	}
```
matskiv, line 59: Is this intentionally a non-cached client?
If not, then we should just pass the `ctx.PhysicalClient` to this function from the `(s *podSyncer) Sync` where it gets called.

pkg/controllers/resources/pods/translate/sa_token_secret.go
```diff
@@ -0,0 +1,144 @@
...
+
+		vclusterServiceOwnerReference = &metav1.OwnerReference{
+			APIVersion: typeAccessor.GetAPIVersion(),
+			Kind:       typeAccessor.GetKind(),
+			Name:       translate.Owner.GetName(),
+			UID:        translate.Owner.GetUID(),
+		}
+
+		for i, owner := range owners {
+			if equality.Semantic.DeepEqual(owner, vclusterServiceOwnerReference) {
```
matskiv, line 119: nit-pick, we could just check `owner.UID == translate.Owner.GetUID()` to save ten lines above :joy:

pkg/controllers/resources/pods/translate/sa_token_secret.go
```diff
@@ -0,0 +1,144 @@
...
+		// pod not set as owner update accordingly
+		secret.ObjectMeta.OwnerReferences = append(secret.ObjectMeta.OwnerReferences, podOwnerReference)
+	}
+
+	err := pClient.Update(ctx, secret)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
matskiv, line 143: nit-pick
```suggestion
	return pClient.Update(ctx, secret)
```

matskiv (changes requested):
Two important comments and a few low-priority nit-picks :) 
Overall looks great :+1:

pkg/controllers/resources/pods/translate.go
ishankhare07: my bad, I only looked at `podSyncer.physicalClient` which was a `kubernetes.Interface`. I didn't realize we had a `client.Client` from `context`

pkg/controllers/resources/pods/translate/sa_token_secret.go
ishankhare07: 🤦🏻‍♂️

matskiv (approved):
LGTM :+1: 
Thank you for addressing all the comments :)

PR #997, opened 2023-04-11
Fix statefulset broken tamplate

### LINKED ISSUE #122: Helm Chart: statefulset template, broken logic
[https://github.com/loft-sh/vcluster/blob/main/chart/templates/statefulset.yaml#L32](https://github.com/loft-sh/vcluster/blob/main/chart/templates/statefulset.yaml#L32)

spec.volumeClaimTemplates.0.spec.storageClassName by default renders as null instead of a value or empty string. This breaks kustomize validation.

```
  {{- if .Values.storage.persistence }}  <== default renders as to IF TRUE
  {{- if not .Values.storage.volumeClaimTemplates }}  <== defaults renders to IF NOT FALSE
  volumeClaimTemplates:
    - metadata:
        name: data
      spec:
        accessModes: [ "ReadWriteOnce" ]
        storageClassName: {{ .Values.storage.className }} <== undefined in values file
        resources:
          requests:
            storage: {{ .Values.storage.size }}
  {{- else }}
  volumeClaimTemplates:
{{ toYaml .Values.volumeClaimTemplates | indent 4 }}
  {{- end }}
  {{- end }}
```

I'm actually surprised Helm doesn't complain about a null value to the key, seems like a break of yaml spec.

K8s doesn't care if the field is defined or not, and the docs simply state to use *a string*.
*storageClassName: "" # Empty string must be explicitly set otherwise default StorageClass will be set*
[https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1)
[https://v1-21.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#persistentvolumeclaimspec-v1-core](https://v1-21.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#persistentvolumeclaimspec-v1-core)

Proposed:

* Define the key 'storageClassName' only if a name is defined in values OR
* Define a default in values "Filesystem" is the inferred value when undefined per: [https://github.com/kubernetes/api/blob/release-1.21/core/v1/types.go#L489](https://github.com/kubernetes/api/blob/release-1.21/core/v1/types.go#L489)

hiteshwani29:
/kind enhancement

resolves #122 

Fixed an issue where vcluster helm storageClass template is passing nil value to storageClassName field if className is not set/passed in the values.yaml

### DIFF
--- a/charts/eks/templates/etcd-statefulset.yaml
+++ b/charts/eks/templates/etcd-statefulset.yaml
@@ -33,7 +33,11 @@ spec:
         name: data
       spec:
         accessModes: [ "ReadWriteOnce" ]
+        {{- if .Values.etcd.storage.className }}
         storageClassName: {{ .Values.etcd.storage.className }}
+        {{- else }}
+        storageClassName: ""
+        {{- end }}
         resources:
           requests:
             storage: {{ .Values.etcd.storage.size }}
--- a/charts/eks/templates/etcd-statefulset.yaml
+++ b/charts/eks/templates/etcd-statefulset.yaml
@@ -35,8 +35,6 @@ spec:
         accessModes: [ "ReadWriteOnce" ]
         {{- if .Values.etcd.storage.className }}
         storageClassName: {{ .Values.etcd.storage.className }}
-        {{- else }}
-        storageClassName: ""
         {{- end }}
         resources:
           requests:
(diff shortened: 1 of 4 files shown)


FabianKramm (approved):

FabianKramm:
@hiteshwani29 seems like your change causes all e2e tests to fail, would you mind to retake a look?

hiteshwani29:
@FabianKramm Did changes to not set the storageClass at all. Can you relook at PR again?

FabianKramm:
@hiteshwani29 thanks, LGTM!

matskiv:
Hello @hiteshwani29 :wave: 
Thank you for the PR!

hiteshwani29:
@matskiv will not be able to present at the kubeCon conference.

Issue #992, opened 2023-04-05
Syncer should handle disabled coredns gracefully


matskiv:
### What happened?

syncer logs multiple errors when coredns is disabled via helm values (coredns.enabled: false)

> I0405 10:29:34.933648   10883 start.go:354\] Failed to apply CoreDNS configuration from the manifest file: unable to parse /manifests/coredns/coredns.yaml: open /manifests/coredns/coredns.yaml: no such file or directory

### What did you expect to happen?

A single warning or info message logged instead of multiple errors.

### How can we reproduce it (as minimally and precisely as possible)?

```
coredns:
  enabled: false
```

### Anything else we need to know?

*No response*

### Host cluster Kubernetes version

N/A

### Host cluster Kubernetes distribution

N/A

### vlcuster version

<details>
v0.14.2
</details>

### Vcluster Kubernetes distribution(k3s(default)), k8s, k0s)

N/A

### OS and Arch

N/A

matskiv:
We should probably just check if the coredns manifest file exists before going into [the retry loop that applies it](https://github.com/loft-sh/vcluster/blob/19dcb355f8e672d80fbbb05858858baeb4a2268e/cmd/vcluster/cmd/start.go#L339C3-L347). 
Other ideas can be considered too, but please no new flags :joy: :joy:

PR #998, opened 2023-04-11
Bug Fix: Add check if coredns files exists before trying to apply

brandonrjacobs:
/kind bugfix

**What does this pull request do? 
resolves https://github.com/loft-sh/vcluster/issues/992

Fixed an issue where disabling coredns results in an error loop during syncer start up. The following PR adds a check for if the file exists and if not we exit the coredns loop and skip the apply of it.


FabianKramm:
@brandonrjacobs thanks a lot for the PR! LGTM!

PR #1023, opened 2023-05-04
fix: increase timeout for retrieving working vcluster pods

neurodrone:
/kind bugfix

This change increases the timeout for retrieving a list of vcluster pods in a running state. It was observed that when running a command such as `vcluster create --upgrade`, when changes deployed incurred redeploy of vcluster pods, it was seen that we hit a failure such as follows because it takes ~15s or so for the pods to come back:

```
fatal    finding vcluster pod: timed out waiting for the condition - can't find a running vcluster pod in namespace vcluster-testvc
```

Fixed an issue where vcluster would timeout on retrieving a list of running `vcluster` pods for a given cluster when they were recently redeployed.

The timeouts don't occur when `--connect=false` is set because it follows a different codepath.

### DIFF
--- a/cmd/vclusterctl/cmd/connect.go
+++ b/cmd/vclusterctl/cmd/connect.go
@@ -314,7 +314,7 @@ func (cmd *ConnectCmd) getVClusterKubeConfig(vclusterName string, command []stri
 	var err error
 	podName := cmd.PodName
 	if podName == "" {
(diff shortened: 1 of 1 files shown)


matskiv (approved):
LGTM

matskiv:
Thank you for your contribution!

PR #1027, opened 2023-05-09
add metrics server proxy docs

ishankhare07:
/kind documentation

resolves ENG-1001 Docs - add documentation for metrics proxy


matskiv:
I feel like the metrics proxy could be the recommended option. So rather than showing how the user should install metric-server, we should first suggest they try metrics-proxy. Wdyt @FabianKramm and @ishankhare07 ?

ishankhare07:
I think you're refering to the section above this – which is more about installing metrics server inside vcluster – in that case yes I agree we can put this section above as a recommended option with an 'info' block saying it requires metrics server on the host cluster

FabianKramm:
@matskiv yep I agree, we should recommend the user this feature if they want metrics inside vcluster

matskiv:
@ishankhare07 yes, exactly!

matskiv (approved):
LGTM

PR #1032, opened 2023-05-11
feat: headless syncer deployment

FabianKramm:
/kind enhancement

Allows to deploy vcluster in headless mode which doesn't deploy any vcluster control plane and only the needed permissions and service for it.


matskiv:
I noticed that a few things that could be considered part of the control plane are still created in headless mode:
```
*/templates/coredns.yaml (mounted as volume)
*/templates/init-configmap.yaml ( not mounted, only read via API)
k0s/templates/secret.yamll (mounted as volume)
```
should these^ really get created in headless mode?

PR #1033, opened 2023-05-12
Fix: vclusterctl - Wrong connect logic for the list command

rafaeljusto:
/kind bugfix

The `list` command doesn't correctly return if the virtual cluster is connected when applying the global context filtering (option `--context`) in the `table` output format.

The `json` output format was also missing the `connected` and `age` fields available in the `table` format.

The solution for the connected logic is to run the routine to detect the current context and use that as a reference when looping over the existing virtual clusters. This is required because `cmd.Context` could contain the filtering context instead of the connected one.

For the `json` output, the type `VCluster`, created on the same file, was used to build the correct output format. Missing fields were added to it.

Fixed an issue where `vcluster list` was returning the wrong logic for `connected` flag.

None.

### DIFF
--- a/cmd/vclusterctl/cmd/list.go
+++ b/cmd/vclusterctl/cmd/list.go
@@ -21,7 +21,9 @@ type VCluster struct {
 	Namespace  string
 	Created    time.Time
 	AgeSeconds int
+	Context    string
 	Status     string
+	Connected  bool
 }
 
 // ListCmd holds the login cmd flags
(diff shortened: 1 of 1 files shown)


ishankhare07 (approved):

rafaeljusto:
Thanks for reviewing this @ishankhare07 . 🙂 The merge is still blocked here, probably restricted to repository owners. Do I need to do something else or you guys will take care from here?

ishankhare07:
@rafaeljusto should be merged soon 🙂

Issue #1025, opened 2023-05-08
Proxy Metrics API - not 100% k8s compliant


famarting:
### What happened?

I was testing the new proxy metrics API in vcluster. For my usecase I have an application that queries the metrics API for all the namespaces and using label selectors.

I noticed this usecase is not working. The metrics API is returning 0 results when it should return data.

Here is an example using label selectors and another example without them. Notice the query with label selectors returns 0 items when it should return something

```
~  $ kubectl get --raw "/apis/metrics.k8s.io/v1beta1/namespaces/dapr-system/pods?labelSelector=app%3Ddapr-operator" | jq '.'
{
  "kind": "PodMetricsList",
  "apiVersion": "metrics.k8s.io/v1beta1",
  "metadata": {},
  "items": []
}
~  $ kubectl get --raw "/apis/metrics.k8s.io/v1beta1/namespaces/dapr-system/pods" | jq '.' 
{
  "kind": "PodMetricsList",
  "apiVersion": "metrics.k8s.io/v1beta1",
  "metadata": {},
  "items": [
    {
      "metadata": {
        "name": "dapr-operator-5bb677d4fb-fc6nf",
        "namespace": "dapr-system",
        "creationTimestamp": "2023-05-04T16:09:10Z",
        "labels": {
          "app": "dapr-operator",
          "app.kubernetes.io/component": "operator",
```

Something else I also noticed, but don't have an example for it, is that the native k8s metrics API supports querying by all namespaces. The API paths looks like `/apis/metrics.k8s.io/v1beta1/pods`. In this case the server returns an error.

### What did you expect to happen?

I expected the metrics API to have the same behavior as the native k8s metrics server.

### How can we reproduce it (as minimally and precisely as possible)?

Run vcluster with the new metrics API enabled

```
proxy:
  metricsServer:
    nodes:
      enabled: true
    pods:
      enabled: true
```

Connect to the vcluster and deploy something to it.
i.e `helm upgrade --install redis bitnami/redis --set architecture=standalone`

Try to query the metrics API with and without setting label selectors
example command `kubectl get --raw "/apis/metrics.k8s.io/v1beta1/namespaces/redis/pods"`

Switch context to the host k8s cluster to verify the API paths supported by the k8s native metrics server

### Anything else we need to know?

*No response*

### Host cluster Kubernetes version

<details>

```
$ kubectl version
[...]

PR #1056, opened 2023-05-24
Translate labelselectors metrics proxy

ishankhare07:
/kind bugfix

resolves ENG-1371 #1025 

Fixed an issue where vcluster metrics server proxy was not translating the label selectors


matskiv (approved):
The code looks good. Thanks!

PR #1072, opened 2023-06-09
fix: disable CGO when building CLI via build-cli.sh

matskiv:
/kind bugfix

resolves https://loft-sh.slack.com/archives/C01N273CF4P/p1686308243191459

Fixed an issue where vcluster v0.15.1 CLI would fail in environments with older glibc version.

v0.15.1 uses dynamically linked libc:
```
$  ldd $(which vcluster)
	linux-vdso.so.1 (0x00007ffed80a9000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f7443e25000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f7444017000)
```
Building with this change produces static binary, just as it was in v0.15.0:
```
$ ldd ./release/vcluster-linux-amd64
	not a dynamic executable
```

### DIFF
--- a/hack/build-cli.sh
+++ b/hack/build-cli.sh
@@ -69,7 +69,7 @@ for OS in ${VCLUSTER_BUILD_PLATFORMS[@]}; do
     fi
 
     echo "Building for ${OS}/${ARCH}"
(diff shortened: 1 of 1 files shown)


ishankhare07 (approved):

rohantmp (approved):
LGTM!

PR #1086, opened 2023-06-17
set vcluster-rewrite-hosts resource requests equal to limits

### LINKED ISSUE #1085: Make vcluster-rewrite-hosts resources configurable or set requests equal to limits
In order to use the static CPU policy in Kubernetes properly, resource requests need to equal limits in all container in a pod for a pod to fall into the `Guaranteed` QoS class. Since the `vcluster-rewrite-hosts` doesn't have the resource requests equal to the limits, this makes it so no pods scheduled in a vcluster can take advantage of the static CPU policy properly.

For more information on the static CPU policy, check out https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/

Here is where the resources are hardcoded: https://github.com/loft-sh/vcluster/blob/a1185ebbe5eb6c9aca4956f7bac121a2b9b9975b/pkg/controllers/resources/pods/translate/hosts.go#L45C1-L52

It would be great if either:
- We made these resources configurable
- Just set requests equal to limits

I don't have a strong preference either way, but considering this init container is super minimal and not resource intensive, just setting requests equal to limits is probably fine and the easiest solution here rather than plumbing through new configurable values.

Thanks!

com6056:
/kind bugfix

resolves #1085

Fixed an issue where vcluster was setting resource requests not equal to limits for the `vcluster-rewrite-hosts` init container, preventing proper usage of the static CPU policy in Kubernetes.

Full writeup of the issue can be found in https://github.com/loft-sh/vcluster/issues/1085

### DIFF
--- a/pkg/controllers/resources/pods/translate/hosts.go
+++ b/pkg/controllers/resources/pods/translate/hosts.go
@@ -47,8 +47,8 @@ func rewritePodHostnameFQDN(pPod *corev1.Pod, defaultImageRegistry, hostsRewrite
 					corev1.ResourceMemory: resource.MustParse("64Mi"),
 				},
 				Requests: map[corev1.ResourceName]resource.Quantity{
-					corev1.ResourceCPU:    resource.MustParse("10m"),
-					corev1.ResourceMemory: resource.MustParse("32Mi"),
+					corev1.ResourceCPU:    resource.MustParse("30m"),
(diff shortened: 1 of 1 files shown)


FabianKramm (approved):
@com6056 thanks for the PR! LGTM!

matskiv (approved):
Thanks for the PR! LGTM

PR #1098, opened 2023-06-23
Trim slash suffix from HostPath path

matskiv:
/kind bugfix

Trim slash suffix from HostPath path

### DIFF
--- a/pkg/controllers/resources/pods/translate/hostpath.go
+++ b/pkg/controllers/resources/pods/translate/hostpath.go
@@ -2,8 +2,9 @@ package translate
 
 import (
 	"fmt"
-	corev1 "k8s.io/api/core/v1"
 	"strings"
+
+	corev1 "k8s.io/api/core/v1"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/controllers/resources/pods/translate/hostpath.go
```diff
@@ -75,41 +76,54 @@ func (t *translator) rewriteHostPaths(pPod *corev1.Pod) {
...
+							PodLoggingHostPath,
+							PhysicalPodLogVolumeMountPath,
+							podLogMountPath,
+							pPod,
+						)
+					}
 				}
 
-				if volume.HostPath.Path == KubeletPodPath &&
+				if strings.TrimSuffix(volume.HostPath.Path, "/") == KubeletPodPath &&
```
matskiv, line 94: This is a bug fix. If you don't merge the PR, at least get this fix into the codebase.

matskiv: There are probably other places that have this issue.

ThomasK33 (approved):

PR #1113, opened 2023-07-08
Fix secret syncer when both ingresses and all secret syncing are enabled

aleoli:
/kind bugfix

resolves #

Fixed an issue where vcluster did not correctly sync all secrets when both all secrets and ingress sync were enabled.

The all-secret synchronization flag was ignored when the ingresses sync was enabled.


FabianKramm (approved):
@aleoli thanks a lot for the PR! LGTM!

PR #1122, opened 2023-07-27
deprecate hostpath mapper component

ishankhare07:
/kind enhancement

resolves ENG-1730 ENG-1779

deprecate hostpath mapper component, ensure flag for physical mount and mount propagation stays hostpath rewriting is still part of core pod translate to toggle from external component


charts/k0s/templates/statefulset.yaml
```diff
@@ -199,9 +199,6 @@ spec:
           {{- if .Values.sync.nodes.nodeSelector }}
           - --node-selector={{ .Values.sync.nodes.nodeSelector }}
           {{- end }}
-          {{- if .Values.hostpathMapper.enabled }}
-          - --rewrite-host-paths=true
```
FabianKramm, line 203: Isn't this still needed for the central host path mapper?

cmd/vcluster/context/flag_options.go
```diff
@@ -159,7 +163,8 @@ func AddFlags(flags *pflag.FlagSet, options *VirtualClusterOptions) {
 	flags.StringVar(&options.HostMetricsBindAddress, "host-metrics-bind-address", "0", "If set, metrics for the controller manager for the resources managed in the host cluster will be exposed at this address")
 	flags.StringVar(&options.VirtualMetricsBindAddress, "virtual-metrics-bind-address", "0", "If set, metrics for the controller manager for the resources managed in the virtual cluster will be exposed at this address")
 
-	flags.BoolVar(&options.RewriteHostPaths, "rewrite-host-paths", false, "If enabled, syncer will rewite hostpaths in synced pod volumes")
```
FabianKramm, line 162: Can we keep the old flag, but just change the behaviour? Also add a new flag to restore the old behaviour

FabianKramm (changes requested):

charts/k0s/templates/statefulset.yaml
ishankhare07: there's no explicit flag needed now for central HPM

cmd/vcluster/context/flag_options.go
ishankhare07: sure, this is moved here now: https://github.com/loft-sh/vcluster/pull/1122/files#diff-c4c21e3be46aa3e2374bdeb86c95b1e5fc10dd95987f14d653c7fb734155ca46R178

cmd/vcluster/context/flag_options.go
```diff
@@ -159,7 +164,8 @@ func AddFlags(flags *pflag.FlagSet, options *VirtualClusterOptions) {
...
+	flags.BoolVar(&options.MountPhysicalHostPaths, "mount-physical-host-paths", false, "If enabled, syncer will rewite hostpaths in synced pod volumes")
+	flags.BoolVar(&options.EnsureMountPropagation, "ensure-mount-propagation", true, "If enabled, syncer will check certain hostpaths in synced pod volumes for mount propagation and enable correctly")
```
ishankhare07, line 168: @FabianKramm the default value here is true which means this will always check mount propagation and rewrite host paths to `/tmp/...` for select containers. However it does not effect because without the hpm component (wether central or daemon) its not gonna work anyway. Also if we were to not do this rewrite its not going to deviate in terms of behavior. What it enables us to do is - no config is needed when central HPM comes into picture.

ishankhare07: Also we can get rid of this flag entirely, but I choose to keep it in case someone wants to explicitly turn off hostpath rewrite and mount propagation even for those select containers

ThomasK33 (approved):

FabianKramm (approved):

PR #1125, opened 2023-07-28
fix: race condition, cache and metrics proxy

matskiv:
/kind bugfix

Should resolve the following error:
> start.go:381] Error registering metrics apiservice: the cache is not started, can not read objects

Fixed an issue where vcluster would not initialize metrics proxy due to a race condition.


rohantmp (approved):
LGTM :)

ishankhare07:
Fixes ENG-1796

PR #1132, opened 2023-08-01
Docs: Add "--update-current=false" to "Exposing vcluster"

kai-wei-mo:
/kind documentation

The flag `--update-current=false` was probably intended to be here because immediately after it, a code snippet suggests the use of `kubectl --kubeconfig ...`. This code snippet would not work without `--update-current=false`.


FabianKramm:
@kai-wei-mo thanks for the PR! Not sure if I understand that correctly, but there is already a `--update-current=false` present in the command

FabianKramm:
I'll close this for now, if there is anything else, feel free to reopen

PR #1135, opened 2023-08-07
Add TLS support to Ingress

maze88:
/kind enhancement
/kind feature

Adds optional TLS configuration for Ingress in `values.yaml` and the respective change in `templates/ingress.yaml` of all 4 distro charts. 

Added TLS support when exposing VCluster's with Ingress reources.

- Although including standard comments in the `values.yaml` file, this may need to be documented in the products documentation (`docs/`).
- I only tested with the default k3s distro chart. Here is a quick command to checkout the template:
```bash
helm template my-vcluster ./charts/k3s \
  --set ingress.enabled=true \
  --set ingress.host=vcluster.maze.omg.lol \
  --set ingress.tls[0].hosts[0]=vcluster.maze.omg.lol \
  --set ingress.tls[0].secretName=tls-vcluster.maze.omg.lol \
  | grep -A 25 'kind: Ingress'
```

### DIFF
--- a/charts/eks/templates/ingress.yaml
+++ b/charts/eks/templates/ingress.yaml
@@ -24,4 +24,8 @@ spec:
                   name: https
             path: /
             pathType: {{ .Values.ingress.pathType }}
+  {{- with .Values.ingress.tls }}
+  tls:
+    {{- toYaml . | nindent 4 }}
+  {{- end -}}
 {{- end }}
--- a/charts/eks/templates/ingress.yaml
+++ b/charts/eks/templates/ingress.yaml
@@ -2,7 +2,10 @@
 apiVersion: {{ .Values.ingress.apiVersion }}
 kind: Ingress
 metadata:
-{{- $annotations := merge .Values.ingress.annotations .Values.globalAnnotations  }}  
+{{- $annotations := merge .Values.ingress.annotations .Values.globalAnnotations  }}
+  {{- if .Values.ingress.tls }}
+    {{- $annotations = omit $annotations "nginx.ingress.kubernetes.io/ssl-passthrough" }}
+  {{- end }}
   {{- if $annotations }}
   annotations:
   {{- toYaml $annotations | nindent 4 }}
--- a/charts/eks/values.yaml
+++ b/charts/eks/values.yaml
@@ -367,6 +367,11 @@ ingress:
     nginx.ingress.kubernetes.io/backend-protocol: HTTPS
     nginx.ingress.kubernetes.io/ssl-passthrough: "true"
     nginx.ingress.kubernetes.io/ssl-redirect: "true"
+  # Ingress TLS configuration
+  tls: []
+    # - secretName: tls-vcluster.local
+    #   hosts:
+    #     - vcluster.local
 
 # Set "enable" to true when running vcluster in an OpenShift host
 # This will add an extra rule to the deployed role binding in order
--- a/charts/k0s/templates/ingress.yaml
+++ b/charts/k0s/templates/ingress.yaml
@@ -3,6 +3,9 @@ apiVersion: {{ .Values.ingress.apiVersion }}
 kind: Ingress
 metadata:
   {{- $annotations := merge .Values.ingress.annotations .Values.globalAnnotations }}
+  {{- if .Values.ingress.tls }}
+    {{- $annotations = omit $annotations "nginx.ingress.kubernetes.io/ssl-passthrough" }}
+  {{- end }}
   {{- if $annotations }}
   annotations:
   {{- toYaml $annotations | nindent 4 }}
--- a/charts/k0s/templates/ingress.yaml
+++ b/charts/k0s/templates/ingress.yaml
@@ -24,4 +24,8 @@ spec:
                   name: https
             path: /
             pathType: {{ .Values.ingress.pathType }}
+  {{- with .Values.ingress.tls }}
+  tls:
+    {{- toYaml . | nindent 4 }}
+  {{- end -}}
 {{- end }}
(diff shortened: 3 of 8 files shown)


ThomasK33:
Hey @maze88 👋,

Thanks for the PR. I'm assuming that this PR is referring to exposing a vcluster [without SSL-Passthrough](https://www.vcluster.com/docs/operator/external-access#ingress-without-ssl-passthrough)?

If that's the case, you'd need to additionally remove the `nginx.ingress.kubernetes.io/ssl-passthrough: "true"` annotation conditionally from the ingress object/template if TLS certificates are passed in values.

---

Alternatively, if this is not referring to the ingress without the SSL-Passthrough section of the docs, in that case, one could consider configuring TLS on the k3s/k0s/k8s/EKS level and configure the ingress controller to enable SSL-Passthrough.

ThomasK33:
Hey @maze88 👋,

First of all, I appreciate your initiative to enhance the helm chart.

Suppose you have enabled SSL-passthrough at the Nginx Ingress Controller level (using `controller.extraArgs`) and added the SSL-passthrough annotation to your VCluster's Ingress as recommended in the documentation. In that case, the SSL traffic is essentially forwarded directly to the backend service without the Ingress controller decrypting it. This means that the TLS configuration at the ingress level becomes unnecessary, as the decryption would be handled at the service (VCluster) level.

With SSL-Passthrough correctly configured, any TLS configuration for the Ingress itself would be redundant since the traffic merely passes through the Ingress without being decrypted.

If your main goal is to expose VCluster with SSL-Passthrough, the changes made in this PR may not be necessary. Instead, we may need to ensure users know how to correctly configure SSL-Passthrough in both the ingress controller and object (which you have already done). We can also extend our documentation to help clarify this aspect further for users.

Please let me know what you think, and if you have any more questions about SSL-Passthrough, I'd be happy to help.

maze88:
Thanks for the explanation @ThomasK33 ...!
This leaves me with two questions:
1. What's the best practice for exposing a VCluster that is both secure (TLS), cost-efficient (Ingress), and provides a seamless/transparent experience for tenants (_my_ customers/students)?
2. If one opts for the Ingress exposure method (with SSL-Passthrough) and doesn't configure TLS at the Ingress level, then how and where is the outward-facing hostname and TLS managed?

I agree with that fact that the documentation may require a few clarifications or additions about:
- Best practice for exposing VCluster with TLS and Ingress?
- What is the user-experience degradation, described in the [Ingress method's first line](https://www.vcluster.com/docs/operator/external-access#ingress)?

maze88:
Thanks for the explanation @ThomasK33 ...!
This leaves me with the following questions (which I feel also warrant documentation clarifications/improvements):
1. What's the best practice for exposing a VCluster that is both secure (TLS), cost-efficient (Ingress), and provides a seamless/transparent experience for tenants (_my_ customers/students) with or without SSL-Passthrough?
2. In the Nginx Ingress documentation they warn that [SSL-Passthrough introduces a non-negligible _performance penalty_](https://kubernetes.github.io/ingress-nginx/user-guide/tls/#ssl-passthrough), whereas in the [VCluster documentation you state that with SSL-Passthrough will provide the _best user experience_](https://www.vcluster.com/docs/operator/external-access#ingress). Could you explain this contradiction?

ThomasK33:
In response to your first question, there are many best practices here, yet there is no silver bullet, as it depends on your needs and solution architecture.

If you're looking to supply a custom CA certificate to the vcluster, that process involves a lot of automation and distro-specific glue code. Vcluster itself consists of a syncer container and a Kubernetes distro.
When starting up the Kubernetes distro, the TLS setup is specific to the distribution. For instance, k3s generates the cluster CA certificates and all other required certificates during startup if they are not located in a particular file system location. 

Suppose you're looking to expose a vcluster via an ingress. In that case, it is considered best practice to expose it with SSL-Passthrough, directly pointing towards the given vcluster, and to provide users with a KUBECONFIG file containing the vcluster's CA certificate + user certificates. However, suppose SSL-Passthrough cannot be configured on the ingress level. In that case, the certificate-based authentication will not work. As mentioned in the docs, SA-based authentication is the only option because the Kubernetes distro runs with a different set of certificates than the ingress controller. 
(Alternatively, a commercial offering called [Loft](https://loft.sh) automates this process and provides a streamlined experience of setting up and configuring those mentioned above.)

As for your second question, I don't think the issue is necessarily a contradiction. According to Nginx, Nginx performs SNI-based routing and does not terminate the TLS connection, which introduces a bit of overhead and can result in a performance penalty. However, if the setup itself is small, this statement is negligible.

For the best user experience using ingress, we recommended exposing the kube apiserver via ingress with SSL-Passthrough, as it mimics a standalone Kubernetes installation as closely as possible. The interactions are the same as if K3s were set up on a VM and exposed, without limitations such as the SA-only authentication in the non-SSL-Passthrough setup.

maze88:
> Nginx performs SNI-based routing and does not terminate the TLS connection, which introduces a bit of overhead and can result in a performance penalty. However, if the setup itself is small, this statement is negligible.

What counts as small? I am planning to run about 50 VClusters on an EKS host cluster.

> For the best user experience using ingress, we recommended exposing the kube apiserver via ingress with SSL-Passthrough, as it mimics a standalone Kubernetes installation as closely as possible.

This answers my question (thank you!) - and was *not* clear based on the documentation. As a user, I suggest that this gets added and emphasized in the documenation! (:

ThomasK33:
> What counts as small? for example - I am planning to run about 50 VClusters on an EKS host cluster.

I'd recommend benchmarking for anything related to performance, as there's no right or wrong answer/setup on paper.

> and was not clear based on the documentation

I'll add it to the docs in the next update run.
To be clear, though, the recommendation only holds up when exposing via Ingress.
Exposing via dedicated LoadBalancers, NoderPorts, or other means is equally valid and recommended.

ThomasK33 (changes requested):
Regarding the actual PR: 

Please conditionally remove the `nginx.ingress.kubernetes.io/ssl-passthrough: "true"` from the ingresses if a TLS config is set.

ThomasK33 (approved):

ThomasK33:
Cool, thanks for the PR @maze88! 👍

maze88:
cheers (:

PR #1169, opened 2023-08-22
Upstream/docs restructure

ishankhare07:
/kind documentation

resolves ENG-1808

Restructuring vcluster docs along with added PRO docs sections


lizardruss:
This really needs to be rebased. I've attempted it with #1249, but ideally we would update this original PR

PR #1172, opened 2023-08-23
Add v1.28 support k8s distro

neogopher:
/kind enhancement

Adds support for Kubernetes 1.28 when using k8s distro of vcluster.

Closes ENG-1933


cmd/vclusterctl/log/baselog_logger.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (l baseloggerLogSink) Info(level int, msg string, kvs ...interface{}) {
+	l.logger.Infof("%s\t", msg)
+	for k, v := range l.keyValues {
+		l.logger.Infof("%s: %+v  ", k, v)
+	}
+	for i := 0; i < len(kvs); i += 2 {
+		l.logger.Infof("%s: %+v  ", kvs[i], kvs[i+1])
+	}
+}
```
ThomasK33, line 31: This needs to be a proper implementation of the logger interface.

In logr, a message is an error, info, or info with a verbosity level (debug messages).
Per convention in loft + Kubernetes, infos level messages will have a level of 0, while any debug message will be higher than 0.

Also, the key-value pairs should be part of the log message, not separate log messages.

I'll update the github.com/loft-sh/log library to provide a method to convert a logger to a logr interface.

ThomasK33 (changes requested):

ThomasK33:
I've updated the BaseLogger interface to include a function for logr compatibility.
Refactored this PR to use that standard method instead of rewriting it in this repo, as it will find use in other projects, too.

If you're fine with it, I'd go ahead and merge it.

FabianKramm (approved):

neogopher:
Hi @ThomasK33,
Thanks a bunch for this :slightly_smiling_face:. Yeah was really missing this conversion method to reuse existing loggers. Looks all good. We can merge it. :+1:

ThomasK33 (approved):

PR #1180, opened 2023-08-25
chore: update k8s dependencies to v1.28

neogopher:
/kind enhancement

Updated vcluster Kubernetes dependencies to  v1.28

Closes ENG-1963


pkg/server/server.go
```diff
@@ -354,7 +354,7 @@ func DefaultBuildHandlerChain(apiHandler http.Handler, c *server.Config) http.Ha
 	if c.FlowControl != nil {
 		workEstimatorCfg := flowcontrolrequest.DefaultWorkEstimatorConfig()
 		requestWorkEstimator := flowcontrolrequest.NewWorkEstimator(
-			c.StorageObjectCountTracker.Get, c.FlowControl.GetInterestedWatchCount, workEstimatorCfg)
+			c.StorageObjectCountTracker.Get, c.FlowControl.GetInterestedWatchCount, workEstimatorCfg, c.FlowControl.GetMaxSeats)
```
neogopher, line 357: Based on similar update in k8s.io/apiserver/pkg/server - [Link](https://github.com/kubernetes/apiserver/blob/a78d1b9b03fbae72f895dc85f1bf793e77e7e9c5/pkg/server/config.go#L916)

pkg/leaderelection/leaderelection.go
```diff
@@ -48,7 +48,7 @@ func StartLeaderElection(ctx *context2.ControllerContext, scheme *runtime.Scheme
 
 	// Lock required for leader election
 	rl, err := resourcelock.New(
-		resourcelock.ConfigMapsLeasesResourceLock,
+		resourcelock.LeasesResourceLock,
```
neogopher, line 51: Based on [this](https://github.com/kubernetes/client-go/blob/4dd144a93b873d7597a1002ab06d763a3edb0ac3/tools/leaderelection/resourcelock/interface.go#L185)
``

go.mod
```diff
@@ -37,37 +37,37 @@ require (
...
-	k8s.io/cli-runtime v0.27.2
-	k8s.io/client-go v0.27.2
-	k8s.io/component-helpers v0.27.2
+	k8s.io/api v0.28.0
+	k8s.io/apiextensions-apiserver v0.28.0
+	k8s.io/apimachinery v0.28.0
+	k8s.io/apiserver v0.28.0
+	k8s.io/cli-runtime v0.28.0
+	k8s.io/client-go v0.28.0
+	k8s.io/component-helpers v0.28.0
```
ThomasK33, line 46: I just checked the lib versions, and there's already a `v0.28.1`.

It might make sense to update to `v0.28.1` here.

pkg/leaderelection/leaderelection.go
ThomasK33: Does this require additional or changed rbac permissions?

go.mod
neogopher: I was just trying to place safe and update to the bare minimum. :wink: Will try that out now.

pkg/leaderelection/leaderelection.go
neogopher: I tried to search more on maybe some migration procedures but couldn't find anything except that log.

FabianKramm (changes requested):
@neogopher thanks for the PR! We need to update controller-runtime as well to v0.16 as otherwise this might cause issues

pkg/leaderelection/leaderelection.go
FabianKramm: Yeah we need to double check that this isn't causing issues, also is this really completely removed?

neogopher: [This](url) is the commit where the removal has happened.

neogopher: I found more on this as below.

This [#117558](https://github.com/kubernetes/kubernetes/pull/117558) is the PR in kubernetes that introduces the current change and describes it. It all started in [#80289](https://github.com/kubernetes/kubernetes/issues/80289) with the intention to migrate leader election to use Lease API. The migration path was decided as
endpoints/configmap -> endpointsleases/configmapleases -> lease.

I followed our changes in vcluster where we went from [configmap](https://github.com/loft-sh/vcluster/commit/a6eead1f44d3eb956f90e4faf9f237972c2e9173) lock to [configmapleases](https://github.com/loft-sh/vcluster/commit/a0716b6bddd490121d9acac13c90eefd673b8f45#diff-2b56d264d4ca0994b896902728fb1a43c9524c2e0a39fdc9c10edb4447a4fa02) lock and now we are actually in the last phase of the migration i.e. to go from configmapleases to leases.

As for the RBAC, I see that the necessary rules for the "coordination.k8s.io"/leases were already added in [this](https://github.com/loft-sh/vcluster/commit/a0716b6bddd490121d9acac13c90eefd673b8f45#diff-2b56d264d4ca0994b896902728fb1a43c9524c2e0a39fdc9c10edb4447a4fa02) commit.

FabianKramm (approved):

PR #1192, opened 2023-09-07
fix: coredns deploy psp cap

sandangel:
Signed-off-by: San Nguyen <vinhsannguyen91@gmail.com>

/kind bugfix

resolves #

Fixed an issue where coredns deployment breaks the Security policy for dropping ALL caps and only add NET_BIND_SERVICE 

This is a regression. We added NET_BIND_SERVICE when updating coredns image but removed the drop: ALL in securityContext.capabilities.


FabianKramm (approved):
Hey @sandangel ! Thanks a lot for the PR, LGTM!

PR #1212, opened 2023-09-18
fix: bundle should not start a new list entry

### LINKED ISSUE #1211: Bundled chart deployment fails for more than one bundled chart
### What happened?

I used the example from the docs:  https://www.vcluster.com/docs/operator/init-manifests#chart-bundle-mode
to create a deployment with two bundled charts:

```
  init:
    helm:
      - bundle: <base64 encoded tar file 1>
        values: |-
          args:
          - -d
        release:
          name: chart1
          namespace: chart1
      - bundle: <base64 encoded tar file 2>
        values: |-
          args:
          - -v
        release:
          name: chart2
          namespace: chart2
```
The first chart applies fine but the second chart fails with an error complaining that the resources from the first chart already exist, which is true, but the second chart shouldn't care.

One clue was in the syncer logs:

```
error executing helm install chart2 /tmp/-.tgz
```

which seemed like a really odd name for a temporary tar file.

Looking at the code:  https://github.com/loft-sh/vcluster/blob/main/pkg/controllers/manifests/init.go#L492
it expects to use chart.name, chart.repo and chart.version in the tar file name for bundles or non-bundles., and clearly chart.name and chart.version are empty in this case, as they are in the example in the docs.

So okay, I'll add the chart info to the inputs:

```
  init:
    helm:
      - bundle: <base64 encoded tar file 1>
        chart:
          name: chart1
          version: 1.0
          repo: dummyrepo
        values: |-
          args:
          - -d
        release:
          name: chart1
          namespace: chart1
      - bundle: <base64 encoded tar file 2>
        chart:
          name: chart2
          version: 1.1
          repo: dummyrepo
        values: |-
          args:
          - -v
        release:
          name: chart2
          namespace: chart2
```
This fails with errors complaining that it can't download the charts from dummyrepo, which is true because the charts are bundled, they don't live in a repo.

Looking at the init configmap it becomes more clear:

```
- name: crossplane
  repo: foo
  version: 1.13.2
- bundle: <base64 encoded tar file 1>
  values: |-
    args:
    - -d
  release:
    name: chart1
    namespace: chart1
- name: vxpresources
  repo: foo
  version: 0.1.0
- bundle: <base64 encoded tar file 2>
  values: |-
     args:
     - -v
  release:
    name: chart2
    namespace: chart2
```

the init configmap template is adding a `-` to the `bundle` and forming a new entry so we end up with four charts instead of two.

I think the solution is to remove the `-` from the bundle entry and update the documentation to show that that chart name, and version are required for bundled charts.

I'll open a PR

The two bundled charts should have been applied correctly.

Specify two bundled charts in the init data

_No response_

<details>

```console
$ kubectl version
1.26 client, 1.24 server
```

</details>

<details>

```
Openshift
```

</details>

### vlcuster version

<details>

```console
$ vcluster --version
vcluster version 0.15.7

```

</details>

<details>

```
k3s
```

</details>

### OS and Arch

<details>

```
OS:  Linux
Arch: amd64
```
</details>

bobh66:
/kind bugfix
/kind documentation

resolves #1211 

This fixes an issue where the `bundle` attribute was starting a new list entry and cutting off any provided `chart` information which is needed for extracting the tar file into a specific filename.

Also updated the documentation example and values.yaml files to provide more information on the `init` helm chart entries.

Fixed an issue where vcluster could not apply more than one bundled helm chart.

The `init` code assumes that the `chart.name`, `chart.repo` and `chart.version` are populated for both `bundle` and `repo` charts.  The `repo` is not technically needed for bundled charts and the filename can be generated without it.  Requiring the name and version for a bundled chart may seem a little odd, but it's the way the tar file name is determined, so unless that code gets refactored the `chart.name` and `chart.version` should be specified even for bundled charts.

### DIFF
--- a/charts/eks/templates/init-configmap.yaml
+++ b/charts/eks/templates/init-configmap.yaml
@@ -11,7 +11,7 @@ metadata:
     heritage: "{{ .Release.Service }}"
   {{- if .Values.globalAnnotations}}
   annotations:
-{{ toYaml .Values.globalAnnotations | indent 4 }}    
+{{ toYaml .Values.globalAnnotations | indent 4 }}
   {{- end }}
 data:
   manifests: |-
@@ -36,7 +36,7 @@ data:
       {{- end}}
   {{- end }}
   {{- if .bundle }}
-    - bundle: {{ .bundle }}
+      bundle: {{ .bundle }}
   {{- end }}
   {{- if or .chart .bundle }}
       {{- if or .values .valuesTemplate }}
--- a/charts/eks/values.yaml
+++ b/charts/eks/values.yaml
@@ -389,6 +389,22 @@ init:
   # this allows you to use helm values inside, e.g.: {{ .Release.Name }}
   manifestsTemplate: ''
   helm: []
+    # - bundle: <string> - base64-encoded .tar.gz file content (optional - overrides chart.repo)
+    #   chart:
+    #     name: <string>  REQUIRED
+    #     version: <string>  REQUIRED
+    #     repo: <string>  (optional when bundle is used)
+    #     username: <string>   (if required for repo)
+    #     password: <string>   (if required for repo)
+    #     insecure: boolean    (if required for repo)
+    #   release:
+    #     releaseName: <string> REQUIRED
+    #     releaseNamespace: <string> REQUIRED
+    #     timeout: number
+    #   values: |-  string YAML object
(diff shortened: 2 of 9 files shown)


ThomasK33 (approved):
Hey @bobh66 👋 

Thanks for the PR, LGTM

PR #1216, opened 2023-09-20
docs: add CLI docs generation

lizardruss:
/kind documentation

resolves #

Added generated CLI documentation


Justfile
```diff
@@ -68,6 +68,9 @@ copy-assets:
 generate-vcluster-images version="0.0.0":
   go run -mod vendor ./hack/assets/main.go {{ version }} > ./release/vcluster-images.txt
 
+generate-cli-docs version="0.0.0":
```
ThomasK33, line 71: I don't think we need the version argument here.

```suggestion
generate-cli-docs:
```

ThomasK33 (changes requested):
Looks good to me, tiny change desired.

PR #1240, opened 2023-09-25
syncer: handle conflict errors correctly by requeueing

sttts:
/kind bugfix

resolves #

Conflict errors mean that another actor touched an object on the kube-apiserver in parallel. It's part of the optimistic concurrency prototcol and should be handled silently. This PR put a condition around the error code that create a `SyncError` events. Instead `ctrl.Result{Requeue: true}` is returned.

Fixed an issue where vcluster creates `SyncError` events on (expected) kube-apiserver conflicts instead of silently requeueing.


pkg/controllers/generic/export_syncer.go
```diff
@@ -142,6 +142,9 @@ func (f *exporter) SyncDown(ctx *synccontext.SyncContext, vObj client.Object) (c
 	pObj, err := f.patcher.ApplyPatches(ctx.Context, vObj, nil, f.config.Patches, f.config.ReversePatches, func(vObj client.Object) (client.Object, error) {
 		return f.TranslateMetadata(ctx.Context, vObj), nil
 	}, &virtualToHostNameResolver{namespace: vObj.GetNamespace(), targetNamespace: translate.Default.PhysicalNamespace(vObj.GetNamespace())})
+	if !kerrors.IsConflict(err) {
```
FabianKramm, line 145: Isn't the logic inverted here? Shouldn't this be `if kerrors.IsConflict(err) {`? Same goes for the other if statements below

sttts: indeed. Copy and paste. Sec.

sttts: fixed.

pkg/controllers/generic/export_syncer.go
```diff
@@ -243,6 +249,9 @@ func (f *exporter) Sync(ctx *synccontext.SyncContext, pObj client.Object, vObj c
 
 			return ctrl.Result{}, nil
 		}
+		if kerrors.IsConflict(err) {
+			return ctrl.Result{Requeue: true}, nil
+		}
```
sttts, line 254: @FabianKramm please double check this one. Above you have `ReplaceOnConflict`? Is that intentional and correct?

FabianKramm: @sttts thanks for the pointer, the `IsInvalid` is definitely required for resources that have immutable fields and need to be replaced if updated on one side as otherwise this will get stuck in an error retry loop.

You probably know this better than me, but are there any other conflict errors in Kubernetes except the `the object has been modified; please apply your changes to the latest version and try again`? If yes, I believe we should actually check its this exact error instead of just checking for if its a conflict error

sttts: A patch or update will fail if some precondition (resource version, uid) does not hold. For a patch in controller-runtime it's usually `client.MergeFromWithOptimisticLock{}` as patch parameter. Both cases should be fixed through a requeue. `IsInvalid` is different maybe because multiple actors could get stuck when applying changes from both sides. With replace you can then force your view of the world. I guess that's the purpose of the code here.

FabianKramm: @sttts that makes sense, thanks for the explanation! I think we should remove the `kerrors.IsConflict` then from the condition above and just leave `kerrors.IsInvalid`

sttts: done

FabianKramm (approved):
Great, thanks a lot @sttts

PR #1259, opened 2023-09-29
fix node labels syncing

### LINKED ISSUE #1001: Node label kubernetes.io/hostname should contain correct value
### What happened?

When "fake nodes" sync is used, we set a fake hostname (fake-$NAME) in `kubernetes.io/hostname` Node label.

We should set the value of `kubernetes.io/hostname` to the `$NAME.nodes.vcluster.com`([*GetNodeHost()*](https://github.com/loft-sh/vcluster/blob/c0cfca75b2fe547663f4e1a04407a38b22a5913a/pkg/controllers/resources/nodes/fake_syncer.go#L323)) instead.
This should also be done for the "real nodes sync".

```
vcluster create vc -n vc
# in vcluster kube context:
kubectl get nodes --label-columns="kubernetes.io/hostname"
```

*No response*

N/A

N/A

### vlcuster version

v0.15.0-beta.0

N/A

### OS and Arch

N/A

ishankhare07:
/kind bugfix
/kind enhancement
/kind test

resolves #1001 ENG-1289

Fixed an issue where vcluster node were syncing as `fake-$NAME`. Use `$NAME.nodes.vcluster.com` instead

### DIFF
--- a/pkg/controllers/resources/nodes/translate.go
+++ b/pkg/controllers/resources/nodes/translate.go
@@ -100,7 +100,9 @@ func (s *nodeSyncer) translateUpdateBackwards(pNode *corev1.Node, vNode *corev1.
 		}
 	}
 
-	labels["kubernetes.io/hostname"] = GetNodeHost(vNode.Name)
+	if s.useFakeKubelets {
+		labels["kubernetes.io/hostname"] = GetNodeHost(vNode.Name)
+	}
 
 	// Omit those taints for which the vcluster has enforced tolerations defined
 	if len(s.enforcedTolerations) > 0 && len(translatedSpec.Taints) > 0 {
--- a/pkg/controllers/resources/nodes/translate.go
+++ b/pkg/controllers/resources/nodes/translate.go
@@ -98,7 +98,10 @@ func (s *nodeSyncer) translateUpdateBackwards(pNode *corev1.Node, vNode *corev1.
 		} else {
 			delete(annotations, TaintsAnnotation)
 		}
+		labels, annotations = translate.ApplyMetadata(pNode.Annotations, vNode.Annotations, pNode.Labels, vNode.Labels)
 	}
+
+	labels["kubernetes.io/hostname"] = GetNodeHost(vNode.Name)
 
 	// Omit those taints for which the vcluster has enforced tolerations defined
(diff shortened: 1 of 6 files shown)


FabianKramm (changes requested):

pkg/controllers/resources/nodes/translate.go
```diff
@@ -90,6 +90,7 @@ func (s *nodeSyncer) translateUpdateBackwards(pNode *corev1.Node, vNode *corev1.
 		}
 	} else {
 		labels, annotations = translate.ApplyMetadata(pNode.Annotations, vNode.Annotations, pNode.Labels, vNode.Labels)
+		labels["kubernetes.io/hostname"] = GetNodeHost(vNode.Name)
```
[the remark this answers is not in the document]
ishankhare07: Hi @FabianKramm actually the original issue also mentioned
> This should also be done for the "real nodes sync".

And yea I think I added the above line at the wrong place, I will move it out of the if-else block – we should not be ignoring the `enableScheduler` path, thanks for pointing out

pkg/controllers/resources/nodes/translate.go
```diff
@@ -98,8 +98,11 @@ func (s *nodeSyncer) translateUpdateBackwards(pNode *corev1.Node, vNode *corev1.
 		} else {
 			delete(annotations, TaintsAnnotation)
 		}
+		labels, annotations = translate.ApplyMetadata(pNode.Annotations, vNode.Annotations, pNode.Labels, vNode.Labels)
```
FabianKramm, line 101: Why is that needed here?

FabianKramm (changes requested):

pkg/controllers/resources/nodes/translate.go
```diff
@@ -100,6 +100,8 @@ func (s *nodeSyncer) translateUpdateBackwards(pNode *corev1.Node, vNode *corev1.
 		}
 	}
 
+	labels["kubernetes.io/hostname"] = GetNodeHost(vNode.Name)
```
FabianKramm, line 103: I think we should only do this if `s.useFakeKubelets` is true as otherwise we won't rewritte the hostname at all

FabianKramm (changes requested):

FabianKramm (approved):

PR #1282, opened 2023-10-06
struct for mapping values

ishankhare07:
/kind enhancement
/kind feature

resolves ENG-2132

Fixed an issue where vcluster helm chart values can be validated against a set of go struct values

### DIFF
--- a/cmd/vcluster/cmd/values/v1/base_values.go
+++ b/cmd/vcluster/cmd/values/v1/base_values.go
@@ -6,7 +6,7 @@ import (
 	policyv1 "k8s.io/api/policy/v1"
 )
 
-type Helm struct {
+type BaseHelm struct {
 	GlobalAnnotations    map[string]interface{} `json:"globalAnnotations,omitempty"`
 	Pro                  bool                   `json:"pro,omitempty"`
 	EnableHA             bool                   `json:"enableHA,omitempty"`
@@ -17,7 +17,6 @@ type Helm struct {
 	FallbackHostDNS      bool                   `json:"fallbackHostDns,omitempty"`
 	MapServices          MapServices            `json:"mapServices,omitempty"`
 	Proxy                ProxyValues            `json:"proxy,omitempty"`
-	Syncer               SyncerValues           `json:"syncer,omitempty"`
 	Vcluster             VclusterValues         `json:"vcluster,omitempty"`
 	Storage              StorageValues          `json:"storage,omitempty"`
 	Volumes              []corev1.Volume        `json:"volumes,omitempty"`
@@ -56,6 +55,24 @@ type Helm struct {
 	MultiNamespaceMode EnabledSwitch    `json:"multiNamespaceMode,omitempty"`
 	Telemetry          TelemetryValues  `json:"telemetry,omitempty"`
 	NoopSyncer         NoopSyncerValues `json:"noopSyncer,omitempty"`
+}
+
+type K3s struct {
+	BaseHelm
+	Image  string           `json:"image,omitempty"`
+	Syncer BaseSyncerValues `json:"syncer,omitempty"`
+}
+
+type BaseSyncerValues struct {
+	ExtraArgs             []string                    `json:"extraArgs,omitempty"`
+	Env                   []corev1.EnvVar             `json:"env,omitempty"`
+	LivenessProbe         EnabledSwitch               `json:"livenessProbe,omitempty"`
+	ReadinessProbe        EnabledSwitch               `json:"readinessProbe,omitempty"`
+	VolumeMounts          []corev1.VolumeMount        `json:"volumeMounts,omitempty"`
+	ExtraVolumeMounts     []corev1.VolumeMount        `json:"extraVolumeMounts,omitempty"`
+	Resources             corev1.ResourceRequirements `json:"resources,omitempty"`
+	KubeConfigContextName string                      `json:"kubeConfigContextName,omitempty"`
+	ServiceAnnotations    map[string]interface{}      `json:"serviceAnnotations,omitempty"`
 }
 
 type SyncValues struct {
@@ -131,18 +148,6 @@ type MetricsProxyServerConfig struct {
 	Pods  EnabledSwitch `json:"pods,omitempty"`
 }
 
-type SyncerValues struct {
-	ExtraArgs             []string                    `json:"extraArgs,omitempty"`
-	Env                   []corev1.EnvVar             `json:"env,omitempty"`
-	LivenessProbe         EnabledSwitch               `json:"livenessProbe,omitempty"`
-	ReadinessProbe        EnabledSwitch               `json:"readinessProbe,omitempty"`
-	VolumeMounts          []corev1.VolumeMount        `json:"volumeMounts,omitempty"`
-	ExtraVolumeMounts     []corev1.VolumeMount        `json:"extraVolumeMounts,omitempty"`
-	Resources             corev1.ResourceRequirements `json:"resources,omitempty"`
-	KubeConfigContextName string                      `json:"kubeConfigContextName,omitempty"`
-	ServiceAnnotations    map[string]interface{}      `json:"serviceAnnotations,omitempty"`
-}
-
 type VclusterValues struct {
 	Image             string                      `json:"image,omitempty"`
 	Command           []string                    `json:"command,omitempty"`
@@ -152,6 +157,9 @@ type VclusterValues struct {
 	VolumeMounts      []corev1.VolumeMount        `json:"volumeMounts,omitempty"`
 	Env               []corev1.EnvVar             `json:"env,omitempty"`
 	Resources         corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// this is only provided in context of k0s right now
+	PriorityClassName string `json:"priorityClassName,omitempty"`
 }
 
 type StorageValues struct {
--- a/cmd/vcluster/cmd/values/v1/base_values.go
+++ b/cmd/vcluster/cmd/values/v1/base_values.go
@@ -3,7 +3,7 @@ package values
 import (
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
-	policyv1 "k8s.io/api/policy/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 type BaseHelm struct {
@@ -17,14 +17,14 @@ type BaseHelm struct {
 	FallbackHostDNS      bool                   `json:"fallbackHostDns,omitempty"`
 	MapServices          MapServices            `json:"mapServices,omitempty"`
 	Proxy                ProxyValues            `json:"proxy,omitempty"`
-	Vcluster             VclusterValues         `json:"vcluster,omitempty"`
+	VCluster             VClusterValues         `json:"vcluster,omitempty"`
 	Storage              StorageValues          `json:"storage,omitempty"`
 	Volumes              []corev1.Volume        `json:"volumes,omitempty"`
 	ServiceAccount       struct {
 		Create bool `json:"create,omitempty"`
 	} `json:"serviceAccount,omitempty"`
 	WorkloadServiceAccount struct {
-		Annotations map[string]interface{} `json:"annotations,omitempty"`
+		Annotations map[string]string `json:"annotations,omitempty"`
 	} `json:"workloadServiceAccount,omitempty"`
 	Rbac                RBACValues          `json:"rbac,omitempty"`
 	Replicas            uint32              `json:"replicas,omitempty"`
@@ -36,7 +36,7 @@ type BaseHelm struct {
 	PodLabels           map[string]string   `json:"podLabels,omitempty"`
 	Annotations         map[string]string   `json:"annotations,omitempty"`
 	PodAnnotations      map[string]string   `json:"podAnnotations,omitempty"`
-	PodDisruptionbudget PDBValues           `json:"podDisruptionbudget,omitempty"`
+	PodDisruptionBudget PDBValues           `json:"podDisruptionBudget,omitempty"`
 	ServerToken         struct {
 		Value        string                   `json:"value,omitempty"`
 		SecretKeyRef corev1.SecretKeySelector `json:"secretKeyRef,omitempty"`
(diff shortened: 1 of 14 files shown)


FabianKramm:
@ishankhare07 thanks a lot for the PR! Can you add a test that parses the `chart/k3s/values.yaml`, `chart/k0s/values.yaml` etc. into this new golang schema and makes sure there are no problems with it? That would really help in future when adding new fields to the schema

ishankhare07:
Hey @FabianKramm we already have the tests here - https://github.com/loft-sh/vcluster/pull/1282/files#diff-a8711a7442370873ac6f18d8c609b0e4951129bc7533aa25134cc426df67cb52

cmd/vcluster/cmd/values/v1/base_values.go
```diff
@@ -0,0 +1,344 @@
+package values
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	policyv1 "k8s.io/api/policy/v1"
+)
+
+type BaseHelm struct {
+	GlobalAnnotations    map[string]interface{} `json:"globalAnnotations,omitempty"`
```
ThomasK33, line 10: Shouldn't this be:

```suggestion
	GlobalAnnotations    map[string]string `json:"globalAnnotations,omitempty"`
```

cmd/vcluster/cmd/values/v1/base_values.go
```diff
@@ -0,0 +1,344 @@
...
+	Pro                  bool                   `json:"pro,omitempty"`
+	EnableHA             bool                   `json:"enableHA,omitempty"`
+	Headless             bool                   `json:"headless,omitempty"`
+	DefaultImageRegistry string                 `json:"defaultImageRegistry,omitempty"`
+	Plugin               map[string]interface{} `json:"plugin,omitempty"`
```
ThomasK33, line 15: Would turning this into a `map[string]corev1.Container` make sense?

We could import the [corev1.Container definition](https://pkg.go.dev/k8s.io/api/core/v1#Container).

cmd/vcluster/cmd/values/v1/base_values.go
```diff
@@ -0,0 +1,344 @@
...
+type BaseSyncerValues struct {
+	ExtraArgs             []string                    `json:"extraArgs,omitempty"`
+	Env                   []corev1.EnvVar             `json:"env,omitempty"`
+	LivenessProbe         EnabledSwitch               `json:"livenessProbe,omitempty"`
+	ReadinessProbe        EnabledSwitch               `json:"readinessProbe,omitempty"`
+	VolumeMounts          []corev1.VolumeMount        `json:"volumeMounts,omitempty"`
+	ExtraVolumeMounts     []corev1.VolumeMount        `json:"extraVolumeMounts,omitempty"`
+	Resources             corev1.ResourceRequirements `json:"resources,omitempty"`
+	KubeConfigContextName string                      `json:"kubeConfigContextName,omitempty"`
+	ServiceAnnotations    map[string]interface{}      `json:"serviceAnnotations,omitempty"`
```
ThomasK33, line 75:
```suggestion
	ServiceAnnotations    map[string]string      `json:"serviceAnnotations,omitempty"`
```

cmd/vcluster/cmd/values/v1/base_values.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+type EnabledSwitch struct {
+	Enabled bool `json:"enabled,omitempty"`
+}
+
+type MapServices struct {
+	FromVirtual []map[string]interface{} `json:"fromVirtual,omitempty"`
+	FromHost    []map[string]interface{} `json:"fromHost,omitempty"`
+}
```
ThomasK33, line 140:
```suggestion
type MapServices struct {
	FromVirtual []ServiceMapping `json:"fromVirtual,omitempty"`
	FromHost    []ServiceMapping `json:"fromHost,omitempty"`
}

type ServiceMapping struct {
	From string `json:"from"`
	To   string `json:"to"`
}
```

cmd/vcluster/cmd/values/v1/base_values.go
```diff
@@ -0,0 +1,344 @@
...
+	Record    Record       `json:"record"`
+	Target    Target       `json:"target"`
+	AllowedOn []FilterSpec `json:"allowedOn"`
+	ExceptOn  []FilterSpec `json:"exceptOn"`
+}
+
+type Record struct {
+	RecordType RecordType
+	FQDN       *string `json:"fqdn"`
+	Service    *string `json:"service"`
```
ThomasK33, line 236: Are those missing the omitempty on purpose/do we want to serialize nil values explicitly?

cmd/vcluster/cmd/values/v1/base_values.go
```diff
@@ -0,0 +1,344 @@
...
+
+type DNSMappings struct {
+	Record    Record       `json:"record"`
+	Target    Target       `json:"target"`
+	AllowedOn []FilterSpec `json:"allowedOn"`
+	ExceptOn  []FilterSpec `json:"exceptOn"`
+}
+
+type Record struct {
+	RecordType RecordType
```
ThomasK33, line 234: Why is RecordType a custom type, when there are no other enum values for it?

cmd/vcluster/cmd/values/v1/base_values.go
```diff
@@ -0,0 +1,344 @@
...
+	FQDN       *string `json:"fqdn"`
+	Service    *string `json:"service"`
+	Namespace  *string `json:"namespace"`
+}
+
+type RecordType string
+type TargetMode string
```
ThomasK33, line 241: Why is this a custom type mapping instead of an inlined string?

cmd/vcluster/cmd/values/v1/base_values.go
```diff
@@ -0,0 +1,344 @@
...
+
+type Target struct {
+	Mode      TargetMode `json:"mode"`
+	Vcluster  *string    `json:"vcluster"`
+	URL       *string    `json:"url"`
+	Service   *string    `json:"service"`
+	Namespace *string    `json:"namespace"`
```
ThomasK33, line 248: Is omitempty missing here on purpose?

cmd/vcluster/cmd/values/v1/k8s_values.go
```diff
@@ -0,0 +1,91 @@
...
+	SecurityContext    corev1.SecurityContext    `json:"securityContext,omitempty"`
+	PodSecurityContext corev1.PodSecurityContext `json:"podSecurityContext,omitempty"`
+}
+
+type APIServerValues struct {
+	CommonValues
+	SyncerExORCommonValues
+	ControlPlaneCommonValues
+	SecurityContext    corev1.SecurityContext `json:"securityContext,omitempty"`
+	ServiceAnnotations map[string]interface{} `json:"serviceAnnotations,omitempty"`
```
ThomasK33, line 28:
```suggestion
	ServiceAnnotations map[string]string `json:"serviceAnnotations,omitempty"`
```

ThomasK33 (changes requested):

cmd/vcluster/cmd/values/v1/base_values.go
ishankhare07: I think the plugin definition is more than just the container. For eg. `rbac.role` etc.

cmd/vcluster/cmd/values/v1/base_values.go
ishankhare07: Actually all these record related values are basically coming from pro code, and there we do have things like enum defined for these types. I just kept them same to have an easier parity amount these values across the repos

cmd/vcluster/cmd/values/v1/base_values.go
ThomasK33: Can we create a new type embedding all the necessary fields and options in this case?

ishankhare07: I see, yeah that would be better. We can do that

cmd/vcluster/cmd/values/v1/base_values.go
```diff
@@ -0,0 +1,349 @@
...
+	Pro                  bool                   `json:"pro,omitempty"`
+	EnableHA             bool                   `json:"enableHA,omitempty"`
+	Headless             bool                   `json:"headless,omitempty"`
+	DefaultImageRegistry string                 `json:"defaultImageRegistry,omitempty"`
+	Plugin               map[string]interface{} `json:"plugin,omitempty"`
+	Sync                 SyncValues             `json:"sync,omitempty"`
+	FallbackHostDNS      bool                   `json:"fallbackHostDns,omitempty"`
+	MapServices          MapServices            `json:"mapServices,omitempty"`
+	Proxy                ProxyValues            `json:"proxy,omitempty"`
+	Vcluster             VclusterValues         `json:"vcluster,omitempty"`
```
FabianKramm, line 20: Can we rename all the Vcluster stuff into VCluster? Vcluster looks really awkward

cmd/vcluster/cmd/values/v1/base_values.go
```diff
@@ -0,0 +1,349 @@
...
+	Role struct {
+		Create               bool     `json:"create,omitempty"`
+		Extended             bool     `json:"extended,omitempty"`
+		ExcludedAPIResources []string `json:"excludedAPIResources,omitempty"`
+	} `json:"role,omitempty"`
+}
+
+type PDBValues struct {
+	Enabled bool `json:"enabled,omitempty"`
+	policyv1.PodDisruptionBudgetSpec
```
FabianKramm, line 189: What exactly can be configured here?

cmd/vcluster/cmd/values/v1/base_values.go
```diff
@@ -0,0 +1,349 @@
...
+	Service        CoreDNSServiceValues        `json:"service,omitempty"`
+	Resources      corev1.ResourceRequirements `json:"resources,omitempty"`
+	Manifests      string                      `json:"manifests,omitempty"`
+	PodAnnotations map[string]string           `json:"podAnnotations,omitempty"`
+	PodLabels      map[string]string           `json:"podLabels,omitempty"`
+}
+
+type CoreDNSPluginValues struct {
+	Enabled bool
+	Config  []DNSMappings
```
FabianKramm, line 228: Do those values not need a json field?

pkg/telemetry/types/types.go
```diff
@@ -5,7 +5,7 @@ package types
  */
 
 type SyncerTelemetryConfig struct {
-	Disabled           string `json:"disabled,omitempty"`
+	Disabled           *bool  `json:"disabled,omitempty"`
```
FabianKramm, line 8: Why is this a pointer?

FabianKramm (changes requested):

cmd/vcluster/cmd/values/v1/base_values.go
```diff
@@ -0,0 +1,349 @@
...
+	MapServices          MapServices            `json:"mapServices,omitempty"`
+	Proxy                ProxyValues            `json:"proxy,omitempty"`
+	Vcluster             VclusterValues         `json:"vcluster,omitempty"`
+	Storage              StorageValues          `json:"storage,omitempty"`
+	Volumes              []corev1.Volume        `json:"volumes,omitempty"`
+	ServiceAccount       struct {
+		Create bool `json:"create,omitempty"`
+	} `json:"serviceAccount,omitempty"`
+	WorkloadServiceAccount struct {
+		Annotations map[string]interface{} `json:"annotations,omitempty"`
```
ThomasK33, line 27:
```suggestion
		Annotations map[string]string `json:"annotations,omitempty"`
```

cmd/vcluster/cmd/values/v1/k8s_values.go
```diff
@@ -0,0 +1,91 @@
...
+type EtcdValues struct {
+	CommonValues
+	SyncerExORCommonValues
+	ControlPlaneCommonValues
+	Storage struct {
+		Persistence bool   `json:"persistence,omitempty"`
+		Size        string `json:"size,omitempty"`
+	} `json:"storage,omitempty"`
+	SecurityContext    corev1.SecurityContext `json:"securityContext,omitempty"`
+	ServiceAnnotations map[string]interface{} `json:"serviceAnnotations,omitempty"`
```
ThomasK33, line 53:
```suggestion
	ServiceAnnotations map[string]string `json:"serviceAnnotations,omitempty"`
```

cmd/vcluster/cmd/values/v1/base_values.go
ishankhare07: okay so an update on this, since the plugin name key can be a variable in this case, it would mean we can't just have a straight forward struct for this. We would need an overlapping struct OR need to implement a custom UnmarshalJSON function that extracts the name of plugin and then uses reflect to set that as the struct tag so that the plugin name se correctly set. I think we should just go with a map for this case and address it in future in case we really face customer demands about validation for plugin fields

ThomasK33: Why doesn't turning this into a `map[string]PluginSpec` work?

And `PluginSpec` could be something along the lines of


```go
type PluginSpec struct {
  corev1.Container
  rbacv1.Role
}
```

pkg/telemetry/types/types.go
ishankhare07: I think its was initially set as string->bool ([here](https://github.com/loft-sh/vcluster/blob/main/pkg/telemetry/collect.go#L44)) and since since I've converted it to a bool now, keeping it a pointer would help us distinguish it from not set vs. explicitly set to `false` [here](https://github.com/ishankhare07/vcluster/blob/values-struct-and-validation/pkg/telemetry/collect.go#L44-L45)

cmd/vcluster/cmd/values/v1/base_values.go
ishankhare07: changed this to a struct with only 
```
MinAvailable   *intstr.IntOrString `json:"minAvailable,omitempty"`
MaxUnavailable  *intstr.IntOrString `json:"minAvailable,omitempty"`
```

cmd/vcluster/cmd/values/v1/base_values.go
FabianKramm: Yes a map is the correct way here

pkg/telemetry/types/types.go
FabianKramm: But we just need to check if its `true` right? Or do we have any check where we explicitly want to distinguish between `false` and `unset`?

FabianKramm (approved):

ThomasK33 (approved):

PR #1297, opened 2023-10-12
fix: ingress secret is deleted by vcluster

### LINKED ISSUE #1296: Ingress secret is deleted by vCluster when ingress sync is enabled
### What happened?

Some users are reporting that when ingress sync is enabled and a tls  enabled ingress is created, the ingress secretName is rewritten  (to avoid conflicts while mapping multiple namespaces in the vcluster to single namespaces in the host cluster). 

An secret of this name is created in the host cluster by cert-manager or other entity, but this secret is eventually deleted by the syncer. 

The syncer should not delete objects it does not manage

- start a vcluster with ingress sync enabled
- install cert-manager, provision an issuer
- create an ingress in the vcluster with tls enabled
- wait for the secret to be created for the certificate in the host cluster
- restart the vCluster pod to trigger sync

I'm not sure if this is all it takes to repro, will update as I know more

_No response_

<details>

```console
$ kubectl version
# paste output here
```

</details>

<details>

```
# Write here
```

</details>

### vlcuster version

<details>

```console
$ vcluster --version
# paste output here
```

</details>

<details>

```
# Write here
```

</details>

### OS and Arch

<details>

```
OS: 
Arch:
```
</details>

rohantmp:
Signed-off-by: Rohan CJ <rohantmp@gmail.com>

/kind bugfix

resolves #1296 
resolves ENG-2226

Fixed an issue where vcluster deletes an ingress secret.


rohantmp:
Oops, left a debug statement in there

rohantmp:
Also need to check how the object gets enqueued at all

FabianKramm:
@rohantmp I think the approach is not bad, why haven't we done this before? Is there any issues with this approach you can see?

rohantmp:
Hmm, so the only thing I think might need to be handled is when an unmanaged physical object  exists and a virtual object wants to be synced by that name.

rohantmp:
In that case, should we just throw a syncer error and call it a naming conflict?

FabianKramm:
@rohantmp good question, yes I think returning an error as well as a emitting a syncer error is a good here.

pkg/controllers/syncer/syncer.go
```diff
@@ -105,14 +106,23 @@ func (r *syncerController) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-	// this is to distinguish generic and plugin syncers with the core syncers
-	if pObj != nil && r.excludePhysical(pObj) {
-		return ctrl.Result{}, nil
+	if pObj != nil {
+		// check if physical object is actually managed by vcluster or just coincidentally named
+		isManaged, err := r.syncer.IsManaged(ctx, pObj)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to check if physical object is managed: %w", err)
+		}
+		if !isManaged {
```
FabianKramm, line 119: This should be an `else if !isManaged {`

pkg/controllers/syncer/syncer.go
```diff
@@ -105,14 +106,23 @@ func (r *syncerController) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			return ctrl.Result{}, nil
+		}
+		// this is to distinguish generic and plugin syncers with the core syncers
+		if r.excludePhysical(pObj) {
```
FabianKramm, line 123: I think we should move this check up and do it before IsManaged to preserve the old behaviour as much as possible

rohantmp:
Verified the discussed event is returned when there is a conflict: 
image

pkg/controllers/syncer/syncer.go
rohantmp: done

rohantmp:
New message:
image

pkg/controllers/syncer/syncer.go
FabianKramm: @rohantmp yes but this is semantically still part of the above statement and looks cleaner

FabianKramm: @rohantmp any updates here?

PavelGloba:
It would be super cool if this was included in a 16.4 🙏

FabianKramm (approved):

rohantmp:
@PavelGloba 
It will be!

dimq:
Nice, thank you very much all of you for your work, we've just seen the same problem here !!! Wonderful work !!!!

PR #1317, opened 2023-10-25
refactor: remove k8s & eks helm hooks & cleanup charts

FabianKramm:
/kind enhancement

Fixes ENG-2300


nourspace:
Amazing work! We were having trouble with the etcd as we had to scale it down to a single replica and force it to create a new cluster of single member. Reverting the replicas and removing the force flag `--force-new-cluster` resulted in issues with SSL connectivity between etcd members. I was trying to somehow mimic the pre-install hook and saw that you just merged the PR 🙏 

When is this expected to land in a release?

PR #1357, opened 2023-11-17
Prevent pause failures due to missing pods 

Piroddi:
/kind bugfix

resolves #

Currently the pause process will fail if a pod is not found. Example:

````
Pausing app
info   Scale down statefulSet vcluster/app...
info   Delete 49 vcluster workloads
fatal  pods "kafka-zookeeper-0-x-datastores-x-app" not found
````

Since pods are ephemeral, anything can happen to a pod from the time the list is created to when the pod is deleted in the for loop. If there are a large number of pods associated with an environment, there are higher chances of a pod being removed due to other factors before vcluster is able to delete it. 

 This failure prevents the remaining pods in the environment from being cleaned up. 
 
 The proposed solution is to not fail, if client-go returns a not found error, and continue on deleting the remaining pods

Fixed an issue where vcluster pause fails due to a pod missing


FabianKramm (approved):
@Piroddi thanks a lot for the PR! LGTM!

PR #1380, opened 2023-11-28
k0s is now run from the syncer container

facchettos:
/kind enhancement

resolves eng-2478

k0s is now run from inside the syncer container

### DIFF
--- a/charts/k0s/templates/statefulset.yaml
+++ b/charts/k0s/templates/statefulset.yaml
@@ -115,7 +115,7 @@ spec:
       {{- end }}
       initContainers:
       - image: {{ .Values.defaultImageRegistry }}{{ .Values.vcluster.image }}
-        name: init
+        name: vcluster
         command:
           - /bin/sh
         args:
@@ -129,8 +129,6 @@ spec:
         volumeMounts:
           - name: k0s-binary
             mountPath: /k0s-binary
-          - name: k0s-config
-            mountPath: /k0s-config
       containers:
       {{- if not .Values.syncer.disabled }}
       - name: syncer
@@ -258,11 +256,6 @@ spec:
           {{- end }}
           - name: ETCD_UNSUPPORTED_ARCH
             value: arm64
-          - name: CONFIG_READY
-            valueFrom:
-              secretKeyRef:
-                name: "vc-{{ .Release.Name }}-config"
-                key: CONFIG_READY
           - name: VCLUSTER_COMMAND
             value: |-
               command:
@@ -297,6 +290,10 @@ spec:
             mountPath: /.cache/helm
           - name: k0s-binary
             mountPath: /k0s-binary
+          - name: k0s-config
+            mountPath: /etc/k0s
+          - mountPath: /data
+            name: data
           - name: run-k0s
             mountPath: /run/k0s
           - name: tmp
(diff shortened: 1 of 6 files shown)


charts/k0s/templates/statefulset.yaml
```diff
@@ -111,50 +113,25 @@ spec:
 {{ toYaml . | indent 8 }}
       {{- end }}
       {{- end }}
-      containers:
-      {{- if not .Values.vcluster.disabled }}
+      initContainers:
       - image: {{ .Values.defaultImageRegistry }}{{ .Values.vcluster.image }}
-        name: vcluster
```
FabianKramm, line 117: Leave the name at vcluster, some programs expect it to be that name still

charts/k0s/templates/statefulset.yaml
```diff
@@ -276,6 +253,35 @@ spec:
...
+          {{- if .Values.vcluster.env }}
+{{ toYaml .Values.vcluster.env | indent 10 }}
+          {{- end }}
+          - name: ETCD_UNSUPPORTED_ARCH
+            value: arm64
+          - name: CONFIG_READY
+            valueFrom:
+              secretKeyRef:
+                name: "vc-{{ .Release.Name }}-config"
+                key: CONFIG_READY
```
FabianKramm, line 265: Remove that env var it shouldn't be needed anymore

charts/k0s/templates/statefulset.yaml
```diff
@@ -111,50 +113,25 @@ spec:
...
-          - name: run-k0s
-            mountPath: /run/k0s
-{{ toYaml .Values.vcluster.volumeMounts | indent 10 }}
-        resources:
-{{ toYaml .Values.vcluster.resources | indent 10 }}
-      {{- end }}
+          - name: k0s-binary
+            mountPath: /k0s-binary
+          - name: k0s-config
+            mountPath: /k0s-config
```
FabianKramm, line 133: k0s-config shouldn't be needed

charts/k0s/templates/statefulset.yaml
```diff
@@ -359,6 +369,10 @@ spec:
         volumeDevices:
 {{ toYaml $container.volumeDevices | indent 10 }}
         volumeMounts:
+          - name: run-k0s
+            mountPath: /run/k0s
+          - name: k0s-binary
+            mountPath: /k0s-binary
```
FabianKramm, line 375: these are plugin mounts, so we don't need that here

charts/k0s/templates/statefulset.yaml
```diff
@@ -289,6 +295,10 @@ spec:
         volumeMounts:
           - name: helm-cache
             mountPath: /.cache/helm
+          - name: k0s-binary
+            mountPath: /k0s-binary
+          - name: run-k0s
+            mountPath: /run/k0s
```
FabianKramm, line 301: I guess we will need the k0s-config mount and data mount here instead of the values, also would be good to add the vcluster.extraMounts here

charts/k0s/values.yaml
```diff
@@ -146,7 +146,8 @@ syncer:
   volumeMounts:
     - mountPath: /data
       name: data
-      readOnly: true
+    - mountPath: /etc/k0s
+      name: k0s-config
```
FabianKramm, line 150: move this and /data to the manifest itself

FabianKramm (changes requested):

charts/k0s/templates/statefulset.yaml
```diff
@@ -289,14 +288,24 @@ spec:
...
           - name: tmp
             mountPath: /tmp
         {{- if .Values.coredns.enabled }}
           - name: coredns
             mountPath: /manifests/coredns
             readOnly: true
         {{- end }}
-{{ toYaml .Values.syncer.volumeMounts | indent 10 }}
+        {{- if .Values.syncer.extraVolumeMounts }}
```
FabianKramm, line 306: Just add the `.Values.vcluster.volumeMounts` here

FabianKramm (changes requested):

FabianKramm (approved):

PR #1381, opened 2023-11-28
set default ephemeral storage for the syncer container

ishankhare07:
/kind bugfix
/kind enhancement

resolves ENG-2416

Fixed an issue where vcluster under isolation mode and very less limitrange set for `ephemeral-storage` would crash

### DIFF
--- a/charts/eks/values.yaml
+++ b/charts/eks/values.yaml
@@ -150,8 +150,10 @@ syncer:
     enabled: true
   resources:
     limits:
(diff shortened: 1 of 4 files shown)


FabianKramm (changes requested):

FabianKramm (approved):

joaocc:
@FabianKramm would this also be something that would be useful/needed on the eks distribution? Thx

PR #1418, opened 2023-12-15
now writes the config to disk to avoid race condition with secret update

facchettos:
/kind enhancement

resolves eng-2582

Fixed an issue where k0s config wasn't updated in time when updating the secret


FabianKramm (changes requested):
can we remove the mount of the secret completely?

FabianKramm (approved):

PR #1440, opened 2024-01-03
Merge k8s api-server, controller-manager, scheduler into syncer container

facchettos:
/kind feature

resolves eng-2540

Merged apiservers, controller and syncer into a single container


pkg/certs/ensure.go
```diff
@@ -47,7 +47,7 @@ func EnsureCerts(
 	}
 	cfg.Networking.ServiceSubnet = serviceCIDR
 	cfg.Networking.DNSDomain = clusterDomain
-	cfg.ControlPlaneEndpoint = vClusterName + apiSuffix
+	cfg.ControlPlaneEndpoint = "127.0.0.1:6443"
```
FabianKramm, line 50: Can we remove apiSuffix from this function completely?

pkg/constants/distro.go
```diff
@@ -19,5 +20,10 @@ func GetVClusterDistro() string {
 		return K3SDistro
 	}
 
-	return K8SDistro
+	_, err = os.Stat("/k8s-binaries")
+	if err == nil {
+		return K8SDistro
+	}
+
+	return EKSDistro
```
FabianKramm, line 28: Can we check here for eks-binaries?

pkg/k8s/k8s.go
```diff
@@ -0,0 +1,171 @@
...
+
+	controllerCommand := &command{}
+	controllerEnv, ok := os.LookupEnv(controllerCmd)
+	if ok {
+		err := yaml.Unmarshal([]byte(controllerEnv), controllerCommand)
+		if err != nil {
+			return fmt.Errorf("parsing controller command %s: %w", controllerEnv, err)
+		}
+		eg.Go(func() error {
+			return retry.OnError(backoff, retryAlways, func() error {
```
FabianKramm, line 74: Why retry here? I think we should fatal if the command fatals

pkg/setup/initialize.go
```diff
@@ -123,14 +124,31 @@ func initialize(
...
+				// loop in Initialize
+
+				err := k8s.StartK8S(parentCtx, apiUp)
+
+				if err != nil {
+					klog.Fatalf("Error running k8s: %v", err)
+				}
+			}()
+			klog.Info("waiting for the api to be up")
+			<-apiUp
```
FabianKramm, line 147: Why do we need to wait here? We also don't wait for the other distros do we?

FabianKramm (changes requested):

pkg/setup/initialize.go
facchettos: yes, otherwise the controller crashes, and then we get error outputs on the cli. Since we know this is happening at the start, we should wait for it

pkg/k8s/k8s.go
facchettos: yeah, I forgot to remove it, because before I was not waiting for the api to be up and it would always crash

pkg/setup/initialize.go
```diff
@@ -123,14 +124,31 @@ func initialize(
...
+				return err
+			}
+			apiUp := make(chan struct{})
+			// start k8s
+			go func() {
+				// we need to run this with the parent ctx as otherwise this context will be cancelled by the wait
+				// loop in Initialize
+
+				err := k8s.StartK8S(parentCtx, apiUp, options.Name)
+
```
FabianKramm, line 141: Can we remove new line here?

pkg/setup/initialize.go
```diff
@@ -123,14 +124,31 @@ func initialize(
...
+			err = GenerateK8sCerts(ctx, currentNamespaceClient, vClusterName, currentNamespace, serviceCIDR, certificatesDir, options.ClusterDomain)
+			if err != nil {
+				return err
+			}
+			apiUp := make(chan struct{})
+			// start k8s
+			go func() {
+				// we need to run this with the parent ctx as otherwise this context will be cancelled by the wait
+				// loop in Initialize
+
```
FabianKramm, line 139: Can we remove new line here?

charts/k8s/values.yaml
```diff
@@ -153,7 +153,7 @@ syncer:
     limits:
       ephemeral-storage: 8Gi
       cpu: 1000m
-      memory: 512Mi
+      memory: 1024Mi
```
FabianKramm, line 156: Can we increase to 2Gi? Same for syncer in eks

FabianKramm (approved):

PR #1441, opened 2024-01-04
added how to eneable-ssl-passthrough so users can avoid leaving the docs

mpetason:
/kind documentation

keeps the user in our docs when configuring ingress so they won't have to leave and look at the nginx documentation. troubleshooting this took me longer than it should have and I had to end up referring to an old blog post that I wrote a year ago to remember how I got it to work.

(docs) added the exact config change needed to enable-ssl-passthrough


FabianKramm (approved):
@mpetason thanks a lot for the PR! LGTM!

PR #1458, opened 2024-01-17
feat(cli): add ignore-not-found flag

### LINKED ISSUE #771: `vcluster delete --ignore-not-found` just like `kubectl`

mariuskimmina:
/kind feature

resolves #771 

Added --ignore-not-found flag to prevent errors when the virtual cluster or the namespace was not found

This was partially based on #989 which got closed to due to inactivity

### DIFF
--- a/cmd/vclusterctl/cmd/delete.go
+++ b/cmd/vclusterctl/cmd/delete.go
@@ -39,6 +39,7 @@ type DeleteCmd struct {
 	DeleteNamespace     bool
 	DeleteConfigMap     bool
 	AutoDeleteNamespace bool
+	IgnoreNotFound      bool
 
 	rawConfig  *clientcmdapi.Config
 	restConfig *rest.Config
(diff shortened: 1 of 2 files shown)


cmd/vclusterctl/cmd/delete.go
```diff
@@ -94,18 +96,28 @@ func (cmd *DeleteCmd) Run(cobraCmd *cobra.Command, args []string) error {
...
 	if err != nil {
-		return err
+		if !cmd.IgnoreNotFound {
+			return err
+		}
+		var errorNotFound *find.ErrorNotFoundVcluster
+		if !errors.As(err, &errorNotFound) {
+			return err
+		}
+		vclusterFound = false
```
FabianKramm, line 110: Why do we need to keep going here? Can we not just return nil and be done with it?

mariuskimmina: I thought about that but then a command like 

```
vcluster delete my-cluster --namespace my-namespace --delete-namespace --ignore-not-found
```

would not delete the namespace if the namespace exists but the vcluster does not.
I wanted to make sure that the namespace would also get cleaned up in such a case

FabianKramm: I think its okay to not delete the namespace if the vcluster doesn't exist, since that operation clearly was either aborted or the user messed somehow differently with the vcluster deployment.

FabianKramm: If this should become a problem, we could add that later then as well

FabianKramm (approved):

mariuskimmina:
@FabianKramm fixed the linter issue and also squashed the commits

PR #1462, opened 2024-01-19
added cli info command

facchettos:
/kind feature

resolves eng-2693

adds a new command to display relevant infos about cli and platform

### DIFF
--- a/cmd/vclusterctl/cmd/info.go
+++ b/cmd/vclusterctl/cmd/info.go
@@ -19,7 +19,7 @@ type cliInfo struct {
 	InstanceID string `yaml:"instanceID,omitempty"`
 }
 
-// NewCreateCmd creates a new command
+// NewInfoCmd creates a new info command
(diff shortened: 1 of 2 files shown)


cmd/vclusterctl/cmd/info.go
```diff
@@ -0,0 +1,64 @@
...
+	"runtime"
+	"text/template"
+
+	"github.com/loft-sh/log"
+	"github.com/loft-sh/vcluster/pkg/pro"
+	"github.com/loft-sh/vcluster/pkg/telemetry"
+	"github.com/spf13/cobra"
+)
+
+const cliInfoTemplate = `CLI Info:
```
FabianKramm, line 14: Can we use yaml instead of this? Something like:
```
info:
  version: ...
  os: ...
  arch: ...
  machineID: ...
  instanceID: ...
```

FabianKramm (changes requested):

FabianKramm (approved):

PR #1474, opened 2024-01-25
now handles both deprecated `replicas` and `syncer.replicas`

facchettos:
/kind feature

part of eng-2715 

deprecates `replicas` in favour of `syncer.replicas` to converge different distros configs

### DIFF
--- a/charts/k0s/templates/_migrate.tpl
+++ b/charts/k0s/templates/_migrate.tpl
@@ -2,5 +2,5 @@
   handles both replicas and syncer.replicas
 */}}
 {{- define "vcluster.replicas" -}}
-{{ if .Values.syncer.replicas }}{{ .Values.syncer.replicas }}{{ else }}{{ .Values.replicas }}{{ end }}
+{{ if .Values.replicas }}{{ .Values.replicas }}{{ else }}{{ .Values.syncer.replicas }}{{ end }}
 {{- end }}
--- a/charts/k0s/templates/_migrate.tpl
+++ b/charts/k0s/templates/_migrate.tpl
@@ -1,0 +1,6 @@
+{{/*
+  handles both replicas and syncer.replicas
+*/}}
+{{- define "vcluster.replicas" -}}
+{{ if .Values.syncer.replicas }}{{ .Values.syncer.replicas }}{{ else }}{{ .Values.replicas }}{{ end }}
+{{- end }}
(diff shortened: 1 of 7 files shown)


charts/k0s/templates/_migrate.tpl
```diff
@@ -0,0 +1,6 @@
+{{/*
+  handles both replicas and syncer.replicas
+*/}}
+{{- define "vcluster.replicas" -}}
+{{ if .Values.syncer.replicas }}{{ .Values.syncer.replicas }}{{ else }}{{ .Values.replicas }}{{ end }}
```
FabianKramm, line 5: Shouldn't the if check .Values.replicas here as we set .Values.syncer.replicas now by default?

charts/k3s/templates/_migrate.tpl
```diff
@@ -0,0 +1,6 @@
+{{/*
+  handles both replicas and syncer.replicas
+*/}}
+{{- define "vcluster.replicas" -}}
+{{ if .Values.syncer.replicas }}{{ .Values.syncer.replicas }}{{ else }}{{ .Values.replicas }}{{ end }}
```
FabianKramm, line 5: Same as above

charts/k3s/templates/rbac/role.yaml
```diff
@@ -50,7 +50,7 @@ rules:
     resources: ["endpoints"]
     verbs: ["create", "delete", "patch", "update"]
   {{- end }}
-  {{- if or (gt (int .Values.replicas) 1) .Values.rbac.role.extended }}
+  {{- if or (gt (int .Values.replicas) 1) (gt (int .Values.syncer.replicas) 1) .Values.rbac.role.extended }}
```
FabianKramm, line 53: Can we not use the template here?

FabianKramm (changes requested):

charts/k3s/values.yaml
```diff
@@ -253,7 +254,8 @@ rbac:
       # - pods/exec
 
 # The amount of replicas to run the statefulset with
-replicas: 1
+# THIS IS DEPRECATED. Please use syncer.replicas instead
+# replicas: 1
```
FabianKramm, line 258: just remove this comment here completely

FabianKramm (changes requested):

FabianKramm (approved):

PR #1476, opened 2024-01-25
removed unused values since the merge into a single container

facchettos:
/kind enhancement

part of eng-2715 

removed unused values since the merge into single containers 


charts/k8s/templates/etcd-statefulset.yaml
```diff
@@ -66,7 +66,7 @@ spec:
       {{- if .Values.etcd.affinity }}
       affinity:
 {{ toYaml .Values.etcd.affinity | indent 8 }}
-      {{- else if .Values.enableHA }}
+      {{- else }}
```
FabianKramm, line 69: Is this correct? Will this make problems in future?

FabianKramm (approved):

PR #1506, opened 2024-02-01
Fix failing  Conformance test - evicts pods with minTolerationSeconds

neogopher:
/kind bugfix

Fixed an issue where vcluster would not evict pods correctly when scheduler was enabled


FabianKramm:
@neogopher thanks for the PR, is this also a problem for the other distros?

neogopher:
I am testing the same right now with k3s.

neogopher:
Confirmed. This is working fine already with other distros.

FabianKramm (approved):

PR #1514, opened 2024-02-02
Fix: Update fake node arch the same as host node

yeahdongcn:
/kind bugfix

I'm running vCluster in a cluster consisting of ARM64 nodes. I encountered an arch label mismatch when migrating applications from the host cluster to vCluster. However, the issue was resolved by updating the fake node architecture to match that of the host node.

Resolved an issue where running applications in vCluster on ARM64 nodes were encountering architecture label mismatches.

* just build-snapshot -> ok
* go test ./pkg/controllers/resources/nodes/ -> ok

### DIFF
--- a/pkg/controllers/resources/nodes/fake_syncer.go
+++ b/pkg/controllers/resources/nodes/fake_syncer.go
@@ -3,6 +3,7 @@ package nodes
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


FabianKramm (approved):
@yeahdongcn great catch and thanks for the PR! LGTM!

PR #1516, opened 2024-02-02
renamed kubelet-config to worker-config as it is removed in k0s 1.29

facchettos:
/kind bugfix

resolves 

Fixed an issue where k0s wouldn't start because we were disabling an option that was removed in the k0s 1.29 


facchettos:
```
Note: As of k0s 1.26, the kubelet-config component has been replaced by the worker-config 
component. k0s will issue a warning when the old component name is being used. It is 
scheduled for removal in k0s 1.27. Please update to the new component name.
```

PR #1518, opened 2024-02-05
added support for embedded etcd

facchettos:
/kind feature

part of eng-2695

added support for embedded Etcd in k0s

### DIFF
--- a/charts/k0s/templates/syncer.yaml
+++ b/charts/k0s/templates/syncer.yaml
@@ -160,6 +160,15 @@ spec:
           - --server-ca-cert=/data/k0s/pki/ca.crt
           - --server-ca-key=/data/k0s/pki/ca.key
           - --kube-config=/data/k0s/pki/admin.conf
+          {{- if and .Values.embeddedEtcd.enabled .Values.pro }}
+          - --etcd-embedded
+          - --etcd-replicas={{ .Values.syncer.replicas }}
+          {{- end }}
+          {{- if (gt (int .Values.syncer.replicas ) 1) }}
+          - --leader-elect=true
+          {{- else }}
+          - --leader-elect=false
+          {{- end }}
           {{- include "vcluster.legacyPlugins.args" . | indent 10 }}
           {{- include "vcluster.serviceMapping.fromHost" . | indent 10 }}
           {{- include "vcluster.serviceMapping.fromVirtual" . | indent 10 }}
@@ -256,6 +265,10 @@ spec:
         {{- include "vcluster.plugins.config" . | indent 10 }}
           - name: VCLUSTER_DISTRO
             value: k0s
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
         {{- if eq ( ( include "vcluster.replicas" . ) | toString | atoi) 1 }}
           - name: VCLUSTER_NODE_NAME
             valueFrom:
--- a/charts/k0s/values.yaml
+++ b/charts/k0s/values.yaml
@@ -15,6 +15,11 @@ pro: false
 # secrets within the cluster.
 proLicenseSecret: ""
 
+# Embedded etcd settings
+embeddedEtcd:
+  # If embedded etcd should be enabled, this is a PRO only feature
+  enabled: false
+
 # If true, will deploy vcluster in headless mode, which means no deployment
 # or statefulset is created.
 headless: false
--- a/pkg/certs/ensure.go
+++ b/pkg/certs/ensure.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
-	"time"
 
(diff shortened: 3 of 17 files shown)


charts/k0s/templates/syncer.yaml
```diff
@@ -18,6 +18,7 @@ metadata:
 {{ toYaml $annotations | indent 4 }}
   {{- end }}
 spec:
+  publishNotReadyAddresses: true
```
FabianKramm, line 21: I think this is wrong here

FabianKramm (changes requested):
Besides that, don't we need some sort of token or how do we make sure in a multi-replica k0s the certificates are not different between the different replicas? k3s is using the token option, while k8s & eks are using a central certificate secret.

charts/k0s/templates/syncer.yaml
facchettos: I don't believe it is, it would not create a working vcluster without it because we are waiting on etcd to be available before starting the api, and etcd would not start because it would not have any record for `vcluster-headless` service (or something like that)
we have the same in k3s https://github.com/loft-sh/vcluster/blob/52324ca83eaec37c35d25cc202e2c59db29d6526/charts/k3s/templates/statefulset-service.yaml#L18 I believe for the same reason

facchettos:
@FabianKramm  
> Besides that, don't we need some sort of token or how do we make sure in a multi-replica k0s the certificates are not different between the different replicas? k3s is using the token option, while k8s & eks are using a central certificate secret.

No, I reuse the `EnsureK3sCerts` function in my vcluster-pro pr, which stores the certificates in a secret, the token doesn't really have much to do with it

FabianKramm (approved):

charts/k0s/values.yaml
```diff
@@ -14,6 +14,10 @@ pro: false
 # namespace, please enable clusterRole.create and define an extra clusterRole.extraRules to allow vCluster to retrieve
 # secrets within the cluster.
 proLicenseSecret: ""
+# Embedded etcd settings
```
FabianKramm, line 17: Can you add a new line between proLicenseSecret and the comment?

pkg/certs/ensure.go
```diff
@@ -96,26 +112,108 @@ func EnsureCerts(
...
+			if err != nil {
+				return nil, err
+			}
+			files[name] = b
+		}
+	}
+	return files, err
+}
+
+func WaitForK0sFiles(ctx context.Context, certDir string) (map[string][]byte, error) {
```
FabianKramm, line 183: Why is this function part of this package?

pkg/setup/initialize.go
```diff
@@ -100,6 +106,17 @@ func initialize(
 				klog.Fatalf("Error running k0s: %v", err)
 			}
 		}()
+		files, err := certs.WaitForK0sFiles(ctx, "/data/k0s/pki")
+		if err != nil {
+			klog.Info(err)
+			return err
+		}
+		err = updateSecret(ctx, currentNamespaceClient, currentNamespace, vClusterName, files)
```
FabianKramm, line 114: Add new line above

pkg/setup/initialize.go
```diff
@@ -100,6 +106,17 @@ func initialize(
...
+		if err != nil {
+			klog.Info(err)
+			return err
+		}
+
```
FabianKramm, line 119: Remove new line

pkg/setup/initialize.go
```diff
@@ -181,3 +197,21 @@ func GenerateK8sCerts(ctx context.Context, currentNamespaceClient kubernetes.Int
...
+	if secret.Data == nil {
+		secret.Data = make(map[string][]byte)
+	}
+	for fileName, content := range files {
+		secret.Data[fileName] = content
+	}
+
+	// if any error we will retry from the poll loop
+	_, err = currentNamespaceClient.CoreV1().Secrets(currentNamespace).Update(ctx, secret, metav1.UpdateOptions{})
+
```
FabianKramm, line 215: remove new line

pkg/setup/initialize.go
```diff
@@ -181,3 +197,21 @@ func GenerateK8sCerts(ctx context.Context, currentNamespaceClient kubernetes.Int
 
 	return nil
 }
+
+func updateSecret(ctx context.Context, currentNamespaceClient kubernetes.Interface, currentNamespace, vclusterName string, files map[string][]byte) error {
```
FabianKramm, line 201: This is a big race condition, multiple replicas can be at this function at the same time and will all one after another override this secret with new files

FabianKramm (changes requested):

pkg/certs/ensure.go
facchettos: because we are waiting for the certs to be created?

FabianKramm (approved):

PR #1532, opened 2024-02-09
Version auto update

facchettos:
/kind documentation

resolves eng-2754

### DIFF
--- a/known_issues.yaml
+++ b/known_issues.yaml
@@ -1,0 +1,4 @@
+#k3s:
+#  1.29:
+#    rancher/k3s:v1.20.9-k3s1: this is an example issue, until we have real known issues
+
--- a/.github/workflows/compatibility.yaml
+++ b/.github/workflows/compatibility.yaml
@@ -1,0 +1,27 @@
+name: Matrix compatibility
+
+on:
+  release:
+    types: [created]
+  pull_request:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  go-unit-test:
+    name: Generate the compatibility matrix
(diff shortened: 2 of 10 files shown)


.github/workflows/compat-matrix.yaml
```diff
@@ -0,0 +1,39 @@
...
+        uses: actions/checkout@v4
+      - name: Execute unit tests
+        run: go run hack/compat-matrix/main.go docs/pages/deploying-vclusters/compat-matrix.mdx
+      - name: commit change if needed
+        run: |
+          git diff --quiet && git diff --staged --quiet || git commit -am 'updated compatibility matrix'
+      - name: Push changes
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TO:KEN }}
+        run: git push
```
ThomasK33, line 39: Please create a PR instead of directly pushing to main. 

See this for reference: https://github.com/loft-sh/vcluster/blob/d6187b08543c2b81e2dca794cff6baed3c3f048b/.github/workflows/go-licenses.yaml#L37-L47

known_issues.yaml
ThomasK33, line 1: Do we want to keep this in the root dir or the `hack/compat-matrix/` directory?

ThomasK33 (changes requested):
Please squash all the commits into one.

known_issues.yaml
facchettos: I'd keep it in root, there's no reason to hide it

ThomasK33: In that case, we might want to remove the commented-out entries.
Someone not versed in yaml might think that those aren't working 😅

facchettos: yeah maybe, I updated the error to say that it was an example issue until we have some real ones, it's just a placeholder to have the example format and it will have real issues soon (hopefully not too soon)

ThomasK33 (approved):

PR #1554, opened 2024-02-23
WIP feat: add structs for vcluster configuration defaults & validation

lizardruss:
/kind feature

resolves ENG-2879

Added structured validation and defaults for vcluster configuration

WIP


pkg/config/schema/schema.go
facchettos, line 1: This is already defined in the vcluster-values repo

facchettos (changes requested):
This should be done in vcluster-values

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+				"cpu":    {Format: "100m"},
+				"memory": {Format: "256Mi"},
+			},
+		}
+	}
+}
+
+type DistributionsK8s struct {
+	APIServer         *corev1.Container `yaml:"apiServer,omitempty" json:"apiServer,omitempty"`
+	ControllerManager *corev1.Container `yaml:"controllerManager,omitempty" json:"controllerManager,omitempty"`
```
facchettos, line 238: this should not be a container, since it's included in the syncer container, there should only be an image here

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+			Requests: map[corev1.ResourceName]resource.Quantity{
+				"cpu":    {Format: "100m"},
+				"memory": {Format: "256Mi"},
+			},
+		}
+	}
+}
+
+type DistributionsK8s struct {
+	APIServer         *corev1.Container `yaml:"apiServer,omitempty" json:"apiServer,omitempty"`
```
facchettos, line 237: this should not be a container, since it's included in the syncer container, there should only be an image here

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+	CSINodes             Enablement `yaml:"csiNodes,omitempty" json:"csiNodes,omitempty"`
+	CSIStorageCapacities Enablement `yaml:"csiStorageCapacities,omitempty" json:"csiStorageCapacities,omitempty"`
+	IngressClasses       Enablement `yaml:"ingressClasses,omitempty" json:"ingressClasses,omitempty"`
+	Events               Enablement `yaml:"events,omitempty" json:"events,omitempty"`
+	StorageClasses       Enablement `yaml:"storageClasses,omitempty" json:"storageClasses,omitempty"`
+}
+
+type Enablement struct {
+	Enabled bool `yaml:"enabled,omitempty" json:"enabled,omitempty" default:"true"`
+}
```
facchettos, line 73: this is weird to have all the stuff enabled by default

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+	Version string `yaml:"version,omitempty" json:"version,omitempty"`
+}
+
+type ControlPlane struct {
+	Distro           Distributions                `yaml:"distro,omitempty" json:"distro,omitempty"`
+	BackingStore     BackingStore                 `yaml:"backingStore,omitempty" json:"backingStore,omitempty"`
+	HostPathMapper   HostPathMapper               `yaml:"hostPathMapper,omitempty" json:"hostPathMapper,omitempty"`
+	CoreDNS          CoreDNS                      `yaml:"coredns,omitempty" json:"coredns,omitempty"`
+	VirtualScheduler *Enablement                  `yaml:"virtualScheduler,omitempty" json:"virtualScheduler,omitempty" product:"pro"`
+	Proxy            *ControlPlaneProxy           `yaml:"proxy,omitempty" json:"proxy,omitempty"`
```
facchettos, line 202: why are those pointers?

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+
+type Distributions struct {
+	K3S      *DistributionsK3s    `yaml:"k3s,omitempty" json:"k3s,omitempty"`
+	K8S      *DistributionsK8s    `yaml:"k8s,omitempty" json:"k8s,omitempty"`
+	K0S      *corev1.Container    `yaml:"k0s,omitempty" json:"k0s,omitempty"`
+	Advanced DistributionAdvanced `yaml:"advanced,omitempty" json:"advanced,omitempty"`
+}
+
+type DistributionsK3s struct {
+	corev1.Container `yaml:",inline" json:",inline"`
```
facchettos, line 217: imo this should not be containers because that lets too much freedom and it's going to be super hard to support if everyone can modify all the settings of every container

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+	Service          ControlPlaneService          `yaml:"service,omitempty" json:"service,omitempty"`
+	Ingress          ControlPlaneIngress          `yaml:"ingress,omitempty" json:"ingress,omitempty"`
+	HighAvailability ControlPlaneHighAvailability `yaml:"highAvailability,omitempty" json:"highAvailability,omitempty"`
+	Advanced         ControlPlaneAdvanced         `yaml:"advanced,omitempty" json:"advanced,omitempty"`
+}
+
+type Distributions struct {
+	K3S      *DistributionsK3s    `yaml:"k3s,omitempty" json:"k3s,omitempty"`
+	K8S      *DistributionsK8s    `yaml:"k8s,omitempty" json:"k8s,omitempty"`
+	K0S      *corev1.Container    `yaml:"k0s,omitempty" json:"k0s,omitempty"`
```
facchettos, line 212: imo this should not be containers because that lets too much freedom and it's going to be super hard to support if everyone can modify all the settings of every container

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+	ServerCACert        string `yaml:"serverCACert,omitempty" json:"serverCACert,omitempty" default:"/data/server/tls/server-ca.crt"`
+	ClientCACert        string `yaml:"clientCACert,omitempty" json:"clientCACert,omitempty" default:"/data/server/tls/client-ca.crt"`
+	RequestHeaderCACert string `yaml:"requestHeaderCACert,omitempty" json:"requestHeaderCACert,omitempty" default:"/data/server/tls/request-header-ca.crt"`
+}
+
+type BackingStore struct {
+	SQLite       Enablement    `yaml:"sqlite,omitempty" json:"sqlite,omitempty"`
+	EmbeddedEtcd *EmbeddedEtcd `yaml:"embeddedEtcd,omitempty" json:"embeddedEtcd,omitempty" product:"pro"`
+	ExternalEtcd *ExternalEtcd `yaml:"externalEtcd,omitempty" json:"externalEtcd,omitempty"`
+}
```
facchettos, line 258: why is sqlite not the same kind of data than the etcds?

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+
+type EmbeddedEtcd struct {
+	Enablement  `yaml:",inline" json:",inline"`
+	MigrateFrom EmbeddedEtcdMigrateFrom `yaml:"migrateFrom,omitempty" json:"migrateFrom,omitempty"`
+}
+
+type EmbeddedEtcdMigrateFrom struct {
+	SQLite bool `yaml:"sqlite,omitempty" json:"sqlite,omitempty" default:"true"`
```
facchettos, line 266: this is not needed because we will detect it automatically

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.21.5
 require (
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/blang/semver/v4 v4.0.0
+	github.com/creasty/defaults v1.7.0
```
facchettos, line 8: I would rather use mergo for this, the default would just be a struct instead of all these annotations

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+	Endpoints              Enablement      `yaml:"endpoints,omitempty" json:"endpoints,omitempty"`
+	Ingresses              Enablement      `yaml:"ingresses,omitempty" json:"ingresses,omitempty"`
+	PriorityClasses        Enablement      `yaml:"priorityClasses,omitempty" json:"priorityClasses,omitempty"`
+	NetworkPolicies        Enablement      `yaml:"networkPolicies,omitempty" json:"networkPolicies,omitempty"`
+	VolumeSnapshots        Enablement      `yaml:"volumeSnapshots,omitempty" json:"volumeSnapshots,omitempty"`
+	PodDisruptionBudgets   Enablement      `yaml:"podDisruptionBudgets,omitempty" json:"podDisruptionBudgets,omitempty"`
+	ServiceAccounts        Enablement      `yaml:"serviceAccounts,omitempty" json:"serviceAccounts,omitempty"`
+	StorageClasses         Enablement      `yaml:"storageClasses,omitempty" json:"storageClasses,omitempty"`
+	PersistentVolumes      Enablement      `yaml:"persistentVolumes,omitempty" json:"persistentVolumes,omitempty"`
+	PersistentVolumeClaims Enablement      `yaml:"persistentVolumeClaims,omitempty" json:"persistentVolumeClaims,omitempty"`
```
facchettos, line 56: this should probably be a list

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+	Pods                   SyncPods        `yaml:"pods,omitempty" json:"pods,omitempty"`
+}
+
+type SyncFromHost struct {
+	CSIDrivers           Enablement `yaml:"csiDrivers,omitempty" json:"csiDrivers,omitempty"`
+	CSINodes             Enablement `yaml:"csiNodes,omitempty" json:"csiNodes,omitempty"`
+	CSIStorageCapacities Enablement `yaml:"csiStorageCapacities,omitempty" json:"csiStorageCapacities,omitempty"`
+	IngressClasses       Enablement `yaml:"ingressClasses,omitempty" json:"ingressClasses,omitempty"`
+	Events               Enablement `yaml:"events,omitempty" json:"events,omitempty"`
+	StorageClasses       Enablement `yaml:"storageClasses,omitempty" json:"storageClasses,omitempty"`
```
facchettos, line 68: this should be a list

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+}
+
+type Enablement struct {
+	Enabled bool `yaml:"enabled,omitempty" json:"enabled,omitempty" default:"true"`
+}
+
+type SyncAllResource struct {
+	Enablement `yaml:",inline" json:",inline"`
```
facchettos, line 76: this is not needed, since we already have the bool field right below

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+
+	SyncLabelsTaints bool              `yaml:"syncLabelsTaints,omitempty" json:"syncLabelsTaints,omitempty"`
+	ClearImageStatus bool              `yaml:"clearImageStatus,omitempty" json:"clearImageStatus,omitempty"`
+	Selector         map[string]string `yaml:"selector,omitempty" json:"selector,omitempty"`
+}
+
+type SyncPseudoNodes struct {
+	Enablement `yaml:",inline" json:",inline"`
+}
```
facchettos, line 111: this should be only a bool imo

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+	TranslateImage         map[string]string `yaml:"translateImage,omitempty" json:"translateImage,omitempty"`
+	EnforceTolerations     []string          `yaml:"enforceTolerations,omitempty" json:"enforceTolerations,omitempty"` // validate format
+	UseSecretsForSATokens  bool              `yaml:"useSecretsForSATokens,omitempty" json:"useSecretsForSATokens,omitempty" default:"true"`
+	RewriteHosts           *SyncRewriteHosts `yaml:"rewriteHosts,omitempty" json:"rewriteHosts,omitempty"`
+}
+
+type SyncRewriteHosts struct {
+	Enabled            bool `yaml:"enabled,omitempty" json:"enabled,omitempty" default:"true"`
+	InitContainerImage bool `yaml:"initContainerImage,omitempty" json:"initContainerImage,omitempty" default:"true"`
+}
```
facchettos, line 94: wouldn't `InitContainerImage ` be enough here

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+
+type Observability struct {
+	ServiceMonitor Enablement           `yaml:"serviceMonitor,omitempty" json:"serviceMonitor,omitempty"`
+	Metrics        ObservabilityMetrics `yaml:"metrics,omitempty" json:"metrics,omitempty"`
+}
+
+type ObservabilityMetrics struct {
+	Proxy ObservabilityMetricsProxy `yaml:"proxy,omitempty" json:"proxy,omitempty"`
+}
```
facchettos, line 120: why is it nested?

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+
+type ObservabilityMetricsProxy struct {
+	Nodes Enablement `yaml:"nodes,omitempty" json:"nodes,omitempty"`
+	Pods  Enablement `yaml:"pods,omitempty" json:"pods,omitempty"`
```
facchettos, line 124: would probably be better with a bool

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+	// TODO Docs, Logs
+}
+
+type Networking struct {
+	ReplicateServices ReplicateServices  `yaml:"replicateServices,omitempty" json:"replicateServices,omitempty"`
```
facchettos, line 129: it should be a array here no?

pkg/config/schema/schema.go
facchettos: also here, this should probably set under `apiserverimage` to be clear what that sets

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+	NetworkPolicy       NetworkPolicy    `yaml:"networkPolicy,omitempty" json:"networkPolicy,omitempty"`
+	AdmissionControl    AdmissionControl `yaml:"admissionControl,omitempty" json:"admissionControl,omitempty" product:"pro"`
+}
+
+type ResourceQuota struct {
+	Enabled       bool                        `yaml:"enabled,omitempty" json:"enabled,omitempty" default:"false"`
+	Quota         corev1.ResourceList         `yaml:"quota,omitempty" json:"quota,omitempty"`
+	ScopeSelector ScopeSelector               `yaml:"scopeSelector,omitempty" json:"scopeSelector,omitempty"`
+	Scopes        []corev1.ResourceQuotaScope `yaml:"scopes,omitempty" json:"scopes,omitempty"`
+}
```
facchettos, line 417: could we get rid of the enabled switch here?

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+	Scheduling             ControlPlaneScheduling             `yaml:"scheduling,omitempty" json:"scheduling,omitempty"`
+	ServiceAccounts        ControlPlaneServiceAccounts        `yaml:"serviceAccounts,omitempty" json:"serviceAccounts,omitempty"`
+	WorkloadServiceAccount ControlPlaneWorkloadServiceAccount `yaml:"workloadServiceAccount,omitempty" json:"workloadServiceAccount,omitempty"`
+	Probes                 ControlPlaneProbes                 `yaml:"probes,omitempty" json:"probes,omitempty"`
+	Security               ControlPlaneSecurity               `yaml:"security,omitempty" json:"security,omitempty"`
+	Metadata               ControlPlaneMetadata               `yaml:"metadata,omitempty" json:"metadata,omitempty"`
+}
+
+type ControlPlanePersistence struct {
+	Enablement            `yaml:",inline" json:",inline"`
```
facchettos, line 349: we don't need the switch here since it's defined by the size

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+
+type ControlPlaneService struct {
+	LabelsAndAnnotations `yaml:",inline" json:",inline"`
+
+	Name string              `yaml:"name,omitempty" json:"name,omitempty"`
+	Spec *corev1.ServiceSpec `yaml:"spec,omitempty" json:"spec,omitempty"`
+}
+
+type ControlPlaneIngress struct {
+	Enablement           `yaml:",inline" json:",inline"`
```
facchettos, line 322: we don't need the switch here

pkg/config/schema/schema.go
```diff
@@ -0,0 +1,607 @@
...
+type DistributionPaths struct {
+	KubeConfig          string `yaml:"kubeConfig,omitempty" json:"kubeConfig,omitempty" default:"/data/server/cred/admin.kubeconfig"`
+	ServerCAKey         string `yaml:"serverCAKey,omitempty" json:"serverCAKey,omitempty" default:"/data/server/tls/server-ca.key"`
+	ServerCACert        string `yaml:"serverCACert,omitempty" json:"serverCACert,omitempty" default:"/data/server/tls/server-ca.crt"`
+	ClientCACert        string `yaml:"clientCACert,omitempty" json:"clientCACert,omitempty" default:"/data/server/tls/client-ca.crt"`
+	RequestHeaderCACert string `yaml:"requestHeaderCACert,omitempty" json:"requestHeaderCACert,omitempty" default:"/data/server/tls/request-header-ca.crt"`
+}
+
+type BackingStore struct {
+	SQLite       Enablement    `yaml:"sqlite,omitempty" json:"sqlite,omitempty"`
```
facchettos, line 255: this shouldn't be defined, since it's either the default for k3s, either not used

go.mod
lizardruss: My thought using this would be to also set the jsonschema generated defaults by reading these tags. jsonschema also has tags for defaults, so may ultimately drop this library if it winds up being difficult to use these for both purposes.

pkg/config/schema/schema.go
FabianKramm: Yeah since we are using this in other projects, such as loft, its better to have this separate

lizardruss:
Seems there was a miscommunication about the purpose behind this PR. Closing.

PR #1558, opened 2024-02-26
feat: Added value to suppress creation of ClusterRole and ClusterRole…

db3f:
We have users using vCluster who experience problems when trying to uses certain features implying the creation au a ClusterRole/ClusterRoleBinding, since those users don't habe enough rights to do this. Our solution is to provide them with a Service account that has all RBAC rights needed to run vCluster. However, the vCluster helm chart wants to create a ClusterRole, ClusterRoleBinding anyhow. Therefore I added a "flag" suppressClusterRoleCreation to values.yaml  to prevent the ClusterRole/ClusterRoleBinding from being created. If this flag  is not set or set to false, CLusterRole creation will happen as before.

Gives the option to suppress ClusterRole cration.

Added a flag to values.yaml to optionally prevent the Creation auf a ClusterRole

### DIFF
--- a/charts/eks/templates/_helpers.tpl
+++ b/charts/eks/templates/_helpers.tpl
@@ -55,7 +55,7 @@ Whether to create a cluster role or not
     .Values.proxy.metricsServer.nodes.enabled
     .Values.multiNamespaceMode.enabled
     .Values.coredns.plugin.enabled -}}
-{{- true -}}
+{{- .Values.suppressClusterRoleCreation | ternary "" "true" -}}
 {{- end -}}
 {{- end -}}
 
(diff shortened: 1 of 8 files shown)


FabianKramm:
Hey @db3f ! Thanks for creating this PR, we are currently in the process of refactoring the helm values and this will be addressed as part of that refactoring with version v0.20.0, so I'll leave this open for now.

mavogel:
Fixed in `v0.20.0-beta.10`. See 
- https://github.com/loft-sh/vcluster/blob/v0.20.0-beta.10/chart/templates/_rbac.tpl#L13 
- and https://github.com/loft-sh/vcluster/blob/v0.20.0-beta.10/chart/values.yaml#L692

Verified with the `k3s` distribution. Can be closed IMO

PR #1563, opened 2024-02-26
fix(helm): resources not embedded in initContainer for k0s

### LINKED ISSUE #1562: Helm-Chart does not allow setting resources for initContainer in StatefulSet

Kubernetes clusters can enforce that all containers and initContainers must specify resource requests and/or resource limits. With the most recent vcluster helm charts, this is not possible:

https://github.com/loft-sh/vcluster/blob/da4a60269d9e8b44ada0820bfc1e7deba93a2b0f/charts/k0s/templates/syncer.yaml#L119-L133

Allow setting the resources for the vcluster initContainer in the values.yaml.

Set default values for the vcluster initContainer

### Additional context

_No response_

cinimins:
/kind bugfix

resolves #1562 

k0s: setting resources for vcluster initContainer

### DIFF
--- a/charts/k0s/templates/syncer.yaml
+++ b/charts/k0s/templates/syncer.yaml
@@ -131,6 +131,8 @@ spec:
         volumeMounts:
           - name: binaries
             mountPath: /binaries
+        resources:
+{{ toYaml .Values.vcluster.resources | indent 10 }}
       containers:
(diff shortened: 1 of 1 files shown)


FabianKramm (approved):
@cinimins thanks for the PR! LGTM!

PR #1566, opened 2024-02-27
fix(helm): resources not embedded in initContainer for syncer in k8s …

### LINKED ISSUE #1562: Helm-Chart does not allow setting resources for initContainer in StatefulSet

cinimins:
…and eks

/kind bugfix

resolves #1562

k8s/eks: setting resources for syncer initContainer

### DIFF
--- a/charts/eks/templates/syncer.yaml
+++ b/charts/eks/templates/syncer.yaml
@@ -152,6 +152,8 @@ spec:
         {{- if .Values.syncer.imagePullPolicy }}
         imagePullPolicy: {{ .Values.syncer.imagePullPolicy }}
         {{- end }}
+        resources:
+{{ toYaml .Values.syncer.resources | indent 10 }}
       {{- if not .Values.controller.disabled }}
       - name: kube-controller-manager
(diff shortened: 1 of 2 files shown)


ThomasK33 (approved):

facchettos:
This should have been applied to all init containers @ThomasK33. Can you open a pr for that?

ThomasK33:
> This should have been applied to all init containers @ThomasK33. Can you open a pr for that?

It's applied to the vcluster init container in all charts, while plugins define their resources already.
What other init containers are you referring to @facchettos?

facchettos:
There is a series of init containers for EKS and k8s, it has only been applied to one of them here

PR #1567, opened 2024-02-28
added resources for eks and k8s

facchettos:
/kind enhancement
resolves eng-2926

we can now set limits for init containers 


ThomasK33 (approved):
LGTM, thanks for adding them.
I missed those during yesterday's review.

PR #1581, opened 2024-03-04
Improve binaries volume config

matskiv:
/kind enhancement

There are cases where an emptyDir volume of sufficient size can not be provided, so users should be allowed to override the /binaries volume with an alternative.
Additionally, current ephemeral storage requests are no sufficient for some distros:
```
k0s
# du -h /binaries/*
203.2M	/binaries/k0s

k8s
# du -h /binaries/*
108.3M	/binaries/kube-apiserver
100.9M	/binaries/kube-controller-manager
69.4M	/binaries/vcluster

eks
# du -h /binaries/*
105.8M	/binaries/kube-apiserver
98.4M	/binaries/kube-controller-manager
69.4M	/binaries/vcluster
``` 

Fixed an issue where vcluster would request too little ephemeral storage.
Added .syncer.storage.binariesVolume helm chart value to allow overriding /binaries volume.

### DIFF
--- a/charts/eks/templates/syncer.yaml
+++ b/charts/eks/templates/syncer.yaml
@@ -101,11 +101,7 @@ spec:
           emptyDir: {}
         - name: certs
           emptyDir: {}
-      {{- if .Values.volumes }}
-{{ toYaml .Values.volumes | indent 8 }}
-      {{- end }}
-        - name: binaries
-          emptyDir: {}
+{{ toYaml .Values.syncer.storage.binariesVolume | indent 8 }}
       {{- if .Values.syncer.volumes }}
 {{ toYaml .Values.syncer.volumes | indent 8 }}
       {{- end }}
--- a/charts/eks/values.yaml
+++ b/charts/eks/values.yaml
@@ -166,7 +166,7 @@ syncer:
       ephemeral-storage: 8Gi
       memory: 2Gi
     requests:
-      ephemeral-storage: 200Mi
+      ephemeral-storage: 300Mi
       cpu: 10m
       memory: 256Mi
   # Extra volumes
--- a/charts/eks/values.yaml
+++ b/charts/eks/values.yaml
@@ -196,14 +196,19 @@ syncer:
   serviceAnnotations: {}
   # Storage settings for the vcluster
   storage:
-   # If this is disabled, vcluster will use an emptyDir instead
-   # of a PersistentVolumeClaim
-   persistence: true
-   # Size of the persistent volume claim
-   size: 5Gi
-   # Optional StorageClass used for the pvc
-   # if empty default StorageClass defined in your host cluster will be used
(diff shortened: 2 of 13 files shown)


charts/eks/templates/syncer.yaml
```diff
@@ -106,8 +106,12 @@ spec:
       {{- if .Values.volumes }}
 {{ toYaml .Values.volumes | indent 8 }}
       {{- end }}
+      {{- if .Values.syncer.storage.binariesVolume }}
+{{ toYaml .Values.syncer.storage.binariesVolume | indent 8 }}
+      {{- else }}
         - name: binaries
           emptyDir: {}
+      {{- end }}
```
ThomasK33, line 114: I'd instead just set the default in the `.Values.syncer.storage.binariesVolume` and add a `toYaml` here instead of the if/else here, as it adds more lines and more complexity to the templating.

charts/eks/values.yaml
```diff
@@ -196,14 +196,18 @@ syncer:
...
+    persistence: true
+    # Size of the persistent volume claim
+    size: 5Gi
+    # Optional StorageClass used for the pvc
+    # if empty default StorageClass defined in your host cluster will be used
+    #className:
+    # Optional override for the "binaries" volume, in case the default emptyDir doesn't work for you
+    #binariesVolume:
+    # - name: binaries
+    #   emptyDir: {}
```
ThomasK33, line 210: Not sure if I'm a fan of this. Effectively, there will only be one binaries volume, which must be called binaries.

I'd rather rewrite this to something like below, and embed it using a `toYaml` below the `- name: binaries` lines in the syncer.yaml.

```suggestion
    binariesVolume:
      emptyDir: {}
```

ThomasK33 (dismissed):

facchettos:
What is the scenario for an `emptyDir` not being able to be used? I think we should probably hold off adding even more configuration options if we are not sure it's going to be useful

facchettos:
What is the scenario which would prevent us from using `emptyDir` ? I think we should hold off on adding more options if we are not sure they'd be useful

charts/eks/templates/syncer.yaml
matskiv: Yeah, sounds reasonable. I change it now.

matskiv:
> What is the scenario which would prevent us from using emptyDir ? I think we should hold off on adding more options if we are not sure they'd be useful

Some clusters have strict limits for the emptyDir size, and in such cases a different type of volume will have to be used.

As for the failing test, I'll figure it out tomorrow. But after a quick check - seems I need to update structs in the loft-sh/vcluster-values, but this is not documented as far as I can see(nothing in the CONTRIBUTING.md or the test case code).

FabianKramm (approved):

PR #1583, opened 2024-03-05
merge vCluster charts & new values.yaml

FabianKramm:
resolves ENG-2955


chart/templates/_rbac.tpl
```diff
@@ -0,0 +1,139 @@
...
+*/}}
+{{- define "vcluster.createClusterRole" -}}
+{{- if not .Values.rbac.clusterRole.disabled -}}
+{{- if or
+    .Values.rbac.clusterRole.overwriteRules
+    (not (empty (include "vcluster.rbac.clusterRoleExtraRules" . )))
+    (not (empty (include "vcluster.plugin.clusterRoleExtraRules" . )))
+    (not (empty (include "vcluster.generic.clusterRoleExtraRules" . )))
+    .Values.networking.replicateServices.fromHost
+    .Values.pro
```
ThomasK33, line 20: Didn't we want to refactor this, so that a pro one doesn't by default need a cluster role?

cmd/vclusterctl/cmd/app/create/pro.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/loft-sh/loftctl/v3/pkg/config"
 	"github.com/loft-sh/loftctl/v3/pkg/vcluster"
 	"github.com/loft-sh/log"
-	"github.com/loft-sh/vcluster-values/values"
+	config2 "github.com/loft-sh/vcluster/config"
```
ThomasK33, line 24: Wouldn't it make sense to have this called config and alias the loftctl config to "loftctlconfig"?

hack/compat-matrix/main.go
```diff
@@ -8,10 +8,10 @@ import (
 	"slices"
 	"strings"
 
+	config2 "github.com/loft-sh/vcluster/config"
```
ThomasK33, line 11:
```suggestion
	"github.com/loft-sh/vcluster/config"
```

chart/templates/_rbac.tpl
FabianKramm: This is legacy and won't be used for the new version anymore. We still need this for older loft versions though as they will set that field automatically in a controller.

cmd/vclusterctl/cmd/app/create/pro.go
FabianKramm: Yeah I was too lazy to do this :D

chart/templates/_backingstore.tpl
```diff
@@ -0,0 +1,22 @@
...
+*/}}
+{{- define "vcluster.externalEtcd.enabled" -}}
+{{- if and (eq (include "vcluster.distro" .) "k8s") (not .Values.controlPlane.backingStore.embeddedEtcd.enabled) -}}
+{{- true -}}
+{{- else if and (eq (include "vcluster.distro" .) "eks") (not .Values.controlPlane.backingStore.embeddedEtcd.enabled) -}}
+{{- true -}}
+{{- else if .Values.controlPlane.backingStore.externalEtcd.enabled -}}
+{{- true -}}
+{{- end -}}
+{{- end -}}
```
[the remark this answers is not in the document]
FabianKramm: Not sure if worth, most other projects also don't do it like this also because I believe this might render extra spaces. E.g. ingress-nginx is also not inlining else and ifs in there template files (https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/templates/_helpers.tpl)

pkg/config/config.go
```diff
@@ -1,155 +1,211 @@
...
-type Hooks struct {
-	// HostToVirtual is a hook that is executed before syncing from the host to the virtual cluster
-	HostToVirtual []*Hook `json:"hostToVirtual,omitempty" yaml:"hostToVirtual,omitempty"`
-
-	// VirtualToHost is a hook that is executed before syncing from the virtual to the host cluster
-	VirtualToHost []*Hook `json:"virtualToHost,omitempty" yaml:"virtualToHost,omitempty"`
+func (v VirtualClusterConfig) Distro() string {
+	if v.Config.ControlPlane.Distro.K3S.Enabled {
+		return config.K3SDistro
+	} else if v.Config.ControlPlane.Distro.K0S.Enabled {
```
facchettos, line 31: this is a little redundant no? without that `if` the function is the same

pkg/config/config.go
```diff
@@ -1,155 +1,211 @@
...
-	Version string `json:"version,omitempty" yaml:"version,omitempty"`
+// VirtualClusterConfig wraps the config and adds extra info such as name, serviceName and targetNamespace
+type VirtualClusterConfig struct {
+	// Holds the vCluster config
+	config.Config `json:",inline"`
 
-	// Exports syncs a resource from the virtual cluster to the host
-	Exports []*Export `json:"export,omitempty" yaml:"export,omitempty"`
+	// Name is the name of the vCluster
+	Name string `json:"name,omitempty"`
```
facchettos, line 19: shouldn't the name be mandatory?

pkg/config/validation.go
```diff
@@ -0,0 +1,347 @@
...
+	"github.com/loft-sh/vcluster/pkg/util/toleration"
+	"github.com/samber/lo"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+)
+
+var allowedPodSecurityStandards = map[string]bool{
+	"privileged": true,
+	"baseline":   true,
+	"restricted": true,
+}
```
facchettos, line 20: you can add enums with jsonschema using tags

pkg/config/validation.go
```diff
@@ -0,0 +1,347 @@
...
+
+var (
+	verbs = []string{"get", "list", "create", "update", "patch", "watch", "delete", "deletecollection"}
+)
```
facchettos, line 24: you can add enums with jsonschema using tags so that will even work within the helm chart

pkg/config/validation.go
```diff
@@ -0,0 +1,347 @@
...
+
+func ValidateConfig(config *VirtualClusterConfig) error {
```
facchettos, line 26: this does more than validating no?

pkg/config/validation.go
```diff
@@ -0,0 +1,347 @@
...
+			return fmt.Errorf("exports[%d] is required", idx)
+		}
+
+		if exp.Kind == "" {
+			return fmt.Errorf("exports[%d].kind is required", idx)
+		}
+
+		if exp.APIVersion == "" {
+			return fmt.Errorf("exports[%d].APIVersion is required", idx)
+		}
```
facchettos, line 124: can't we have that from the json schema directly?

pkg/config/validation.go
```diff
@@ -0,0 +1,347 @@
...
+			return fmt.Errorf("imports[%d] is required", idx)
+		}
+
+		if imp.Kind == "" {
+			return fmt.Errorf("imports[%d].kind is required", idx)
+		}
+
+		if imp.APIVersion == "" {
+			return fmt.Errorf("imports[%d].APIVersion is required", idx)
+		}
```
facchettos, line 157: can't we use the json schema to do that?

pkg/config/validation.go
```diff
@@ -0,0 +1,347 @@
...
+		return nil
+	default:
+		return fmt.Errorf("unsupported patch type %s", patch.Operation)
+	}
+}
+
+func validateVerb(verb string) error {
+	if !lo.Contains(verbs, verb) {
+		return fmt.Errorf("invalid verb \"%s\"; expected on of %q", verb, verbs)
+	}
```
facchettos, line 233: can't we use `slices.Contain` here?

facchettos (commented):
I left a few comments mainly on the config part, will continue tomorrow first thing

cmd/vcluster/cmd/start.go
```diff
@@ -40,52 +47,53 @@ func NewStartCommand() *cobra.Command {
...
 			// execute command
-			return ExecuteStart(cobraCmd.Context(), vClusterOptions)
+			return ExecuteStart(cobraCmd.Context(), vClusterConfig)
 		},
 	}
 
-	options.AddFlags(cmd.Flags(), vClusterOptions)
-	pro.AddProFlags(cmd.Flags(), vClusterOptions)
+	cmd.Flags().StringVar(&startOptions.Config, "config", "/var/vcluster/config.yaml", "The path where to find the vCluster config to load")
+	cmd.Flags().StringArrayVar(&startOptions.SetValues, "set", []string{}, "Set values for the config. E.g. --set 'exportKubeConfig.secret.name=my-name'")
```
facchettos, line 62: not sure I understand what this is for. Isn't the config file final once it's in the vcluster pod?

cmd/vclusterctl/cmd/create.go
facchettos, line 1: I think we should get rid of the `--set` flag in here, otherwise the validation isn't accurate. We should have a few flags for the things that are likely to be change very often like name, image and what not but everything else should be in the vcluster.yaml

pkg/config/config.go
```diff
@@ -1,155 +1,211 @@
...
+		}
+	case config.EKSDistro, config.K8SDistro:
+		distroConfig = config.VirtualClusterKubeConfig{
+			KubeConfig:          "/pki/admin.conf",
+			ServerCAKey:         "/pki/ca.key",
+			ServerCACert:        "/pki/ca.crt",
+			ClientCACert:        "/pki/ca.crt",
+			RequestHeaderCACert: "/pki/front-proxy-ca.crt",
+		}
+	}
```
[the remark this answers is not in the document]
FabianKramm: problem is that the certs are generated not always by us and are called different even if in the same directory. Its a good idea, but I would separate this change from this PR

pkg/config/config.go
```diff
@@ -1,155 +1,211 @@
...
+	if retConfig.ServerCAKey == "" {
+		retConfig.ServerCAKey = distroConfig.ServerCAKey
+	}
+	if retConfig.ServerCACert == "" {
+		retConfig.ServerCACert = distroConfig.ServerCACert
+	}
+	if retConfig.RequestHeaderCACert == "" {
+		retConfig.RequestHeaderCACert = distroConfig.RequestHeaderCACert
+	}
```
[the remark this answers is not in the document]
FabianKramm: I guess its not too bad to have this explicit for now since its just 5 fields and people don't know how mergo operates and we rely on another dependency for 10 lines of code

pkg/config/validation.go
```diff
@@ -0,0 +1,347 @@
...
+	if config.Config.ControlPlane.Distro.K8S.Enabled {
+		enabledDistros++
+	}
+	if config.Config.ControlPlane.Distro.EKS.Enabled {
+		enabledDistros++
+	}
+
+	if enabledDistros > 1 {
+		return fmt.Errorf("please only enable a single distribution")
```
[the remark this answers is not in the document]
FabianKramm: Fixed

pkg/config/config.go
FabianKramm: Thing is more that we want to have k3s is all are disabled, so we need that function

pkg/config/config.go
FabianKramm: True, fixed it

pkg/config/validation.go
FabianKramm: We need this within golang code as well

pkg/config/validation.go
FabianKramm: We need this within golang code as well, e.g. if people don't use our helm chart

pkg/config/validation.go
FabianKramm: Its also setting couple of defaults

pkg/config/validation.go
FabianKramm: Better to do valdiation here as well, also I haven't changed this

pkg/config/validation.go
FabianKramm: We shouldn't rely only on jsonschema to do validation for us, there are customers who have their own helm chart

pkg/config/validation.go
FabianKramm: Does that provide any benefit?

cmd/vcluster/cmd/start.go
FabianKramm: Yes exactly

cmd/vclusterctl/cmd/create.go
FabianKramm: `--set` is mostly for development with DevSpace, when you don't want to redeploy for every config change and instead try things out. We can hide the flag if that helps

pkg/config/validation.go
facchettos: so the name should reflect it imo

.golangci.yml
```diff
@@ -31,7 +31,6 @@ linters:
     - revive
     - staticcheck
     - stylecheck
-    - tagalign
```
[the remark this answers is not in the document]
FabianKramm: I added it agian

pkg/config/validation.go
facchettos: it's the standard library, instead of using third party one. What's the benefit of using `lo` is the real question

cmd/vcluster/cmd/start.go
facchettos: but why?

pkg/config/config.go
facchettos: yeah but it's not the point, if you lose 
```go
	if v.Config.ControlPlane.Distro.K3S.Enabled {
		return config.K3SDistro
	} else
``` 
the end result is the same still. because the default will be to return k3s

pkg/config/validation.go
FabianKramm: Ah makes sense, will change!

pkg/config/validation.go
FabianKramm: Changed it!

pkg/config/validation.go
facchettos: but now if they do only use our helm charts they don't have the validation
We can run jsonschema validator within go and get the errors, that way validation would be consistent across the board

facchettos: https://github.com/santhosh-tekuri/jsonschema/tree/v5.3.1 for example would allow us to validate the schema for everything that we can, and then rely on go code to only validate things that aren't possible with the schema
That also would allow us to be able to validate for more than 1 version when we introduce breaking changes

Issue #1584, opened 2024-03-05
Waiting for vcluster to come up


eumel8:
### What happened?

I have a vcluster instance created in a namespace. From another pod in the same namespace I want to connect with vcluster cli to the vcluster. The ServiceAccount has permissions to get/list pods. I can also execute `vcluster list` and can see the vcluster instance, but I can't connect execute `vcluster -n <namespace> connect  <instance> ...`  but the only what I get is "Waiting for vcluster to come up". InCluster kube-config is used. Which connections or resources are required to connect? There is also not additional output with debug.

### What did you expect to happen?

```
vcluster -n <namespace> connect <instance> -- kubectl get nodes
```

### How can we reproduce it (as minimally and precisely as possible)?

install vcluster instance and an additionally Pod with vcluster/kubectl cli

### Anything else we need to know?

```bash
% kubectl -n vc2 exec -it vc2-register-rancher-create-pzkmd -- sh
$ kubectl get pod
NAME                                READY   STATUS    RESTARTS   AGE
vc2-register-rancher-create-pzkmd   1/1     Running   0          10m
vc2-vcluster-0                      1/1     Running   0          10m

$ vcluster list
  
        NAME     | CLUSTER | NAMESPACE | STATUS  | VERSION | CONNECTED |            CREATED            |  AGE  | DISTRO  
  ---------------+---------+-----------+---------+---------+-----------+-------------------------------+-------+---------
    vc2-vcluster |         | vc2       | Running | 0.19.3  |           | 2024-03-05 23:11:11 +0000 UTC | 6m50s | OSS     
  
$ vcluster -n vc2 connect vc2-vcluster --debug -- kubectl get nodes
23:18:36 debug Error creating pro client: couldn't find vCluster.Pro config: please make sure to run 'vcluster login' to connect to an existing instance or 'vcluster pro start' to deploy a new instance
23:18:36 info Waiting for vcluster to come up...
```

### Host cluster Kubernetes version

<details>

```console
$ kubectl version
# 1.26.11
```

</details>

### Host cluster Kubernetes distribution

<details>

```
# Rancher RKE1
```

</details>

### vlcuster version

<details>

```console
$ vcluster --version
# 1.18.1
```

</details>

### Vcluster Kubernetes distribution(k3s(default)), k8s, k0s)

<details>

```
# k3s
```

</details>

### OS and Arch

<details>

```
OS: Ubuntu 22.04
Arch: amd64
```
</details>

[...]

PR #1593, opened 2024-03-15
[helm] add additional sidecar container

eumel8:
/kind enhancement

In https://github.com/loft-sh/vcluster/issues/1584 I described our use case to make backups from K3s state.db on behalf of a sidecar container with some logic. Maybe there are other, so I added the possibiliry to add a sidecar container to the vcluster statefulset.

[helm] add additional sidecar container

.


FabianKramm:
@eumel8 thanks for your PR! In the latest version we have merged all helm charts into a single one and refactored the values.yaml to make it easier to configure vCluster (we are still in the progress of adding comments to the values.yaml). Would you mind rebasing our changes to the latest vCluster version?

eumel8:
@FabianKramm ooh, something went wrong in Github and the PR was closed. I setup a new one.

PR #1598, opened 2024-03-15
Schema comments: change vCluster to virtual cluster when not referring to the product itself

aimeeu:
A few nits:
- "vCluster" has a copyright, so we need to use virtual cluster when not talking about the product/app. 
- a few changes to standardize terms and products

/kind enhancement

Note: Pascal showed me what to do to generate the partials needed for the docs, so I can do that after this PR merges.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -4,131 +4,130 @@ import (
 	"regexp"
 )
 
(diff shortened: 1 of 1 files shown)


FabianKramm (approved):
@aimeeu thanks a lot for the PR! LGTM!

PR #1612, opened 2024-03-19
added draft for interceptor plugin

facchettos:
/kind feature

part of  eng-3064

add new methods to the plugin lib so that we can intercept requests for complex uses cases 


pkg/plugin/v2/plugin.go
```diff
@@ -85,10 +103,13 @@ func (m *Manager) Start(
 		}
 	}
 
+	port := 13370
 	// after loading all plugins we start them
 	for _, vClusterPlugin := range m.Plugins {
 		// build the start request
-		initRequest, err := m.buildInitRequest(filepath.Dir(vClusterPlugin.Path), syncerConfig, vConfig)
+		initRequest, err := m.buildInitRequest(filepath.Dir(vClusterPlugin.Path), syncerConfig, vConfig, port)
+		port++
```
ThomasK33, line 111: Is it possible to check if the port is available?

pkg/plugin/v2/plugin.go
```diff
@@ -117,12 +138,116 @@ func (m *Manager) Start(
...
+		if ok {
+			return true, portHandlerName.port, portHandlerName.handlerName
+		}
+	}
+	if resourcesMap, ok := groups["*"]; ok {
+		portHandlerName, ok := portForResource(resourcesMap, resource, verb, resourceName)
+		if ok {
+			return true, portHandlerName.port, portHandlerName.handlerName
+		}
+	}
```
ThomasK33, line 167: Could this be shortened to a 
```suggestion
	resourcesMap, ok := groups[group]
	if !ok {
		resourcesMap, ok = groups["*"]
	}
	if !ok {
		return false, 0, ""
	}
	
	portHandlerName, ok := portForResource(resourcesMap, resource, verb, resourceName)
	if !ok {
		return false, 0, ""
	}
	
	return true, portHandlerName.port, portHandlerName.handlerName
```

ThomasK33: This would also apply to the functions below.

ThomasK33 (changes requested):

pkg/plugin/v2/plugin.go
facchettos: how would it react if the port isn't available? should we exit ? If so that's kinda equivalent to the current code. I am not sure if changing the port if there's already something listening on it is a good idea. That should be a reserved range for vcluster plugins and so failing is ok if something is listening on it

ThomasK33: > how would it react if the port isn't available? should we exit ?

We might exit, we might also omit any port that's in use, and try the next one.

> That should be a reserved range for vcluster plugins and so failing is ok if something is listening on it

Is this reserved port range documented somewhere?

pkg/plugin/v2/plugin.go
```diff
@@ -117,12 +138,116 @@ func (m *Manager) Start(
...
+		}
+
 		klog.FromContext(ctx).Info("Successfully loaded plugin", "plugin", vClusterPlugin.Path)
 	}
 
 	return nil
 }
 
+// interceptorPortForResource returns the port and handler name for the given group, resource and verb
+func (m *Manager) interceptorPortForResource(group, resource, verb, resourceName string) (ok bool, port int, handlerName string) {
```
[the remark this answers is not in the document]
ThomasK33: None of the returned values are the same type, and their meaning can be deducted from the function name and go doc comment. `// interceptorPortForResource returns the port and handler name [...].`
Updating that to accommodate the boolean indicating whether it was found shouldn't be an issue.

In addition, it's about consistency, as we're not using named returns in other places. We might face potential shadowing issues in the future, which will require us to be more careful.

pkg/plugin/v2/plugin.go
facchettos: not yet, but also even if we check that the port is not in use now, there's no guarantee that the port will be available when the plugin starts, so it's not really changing anything

facchettos: this can be done from the sdk though, and that will be reliable

pkg/plugin/v2/plugin.go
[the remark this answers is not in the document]
facchettos: ```, or if the meaning of a result isn’t clear from context,```
imho here it helps, I can change `ok` to `exists` if really needed but 🤷 

`as we're not using named returns in other places` so we should never use them? I have to disagree with the `we've never done it before so we will never do it` argument.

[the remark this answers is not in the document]
ThomasK33: Named returns don't provide value here.

The example you linked showcases 

```go
// Location returns f's latitude and longitude.
// Negative values mean south and west, respectively.
func (f *Foo) Location() (lat, long float64, err error)
```

which makes the signature easier to read, although debatable.

In this instance, though, there is no technical benefit or documentation benefit resulting from those named returns.
No repetitive return types could be mixed up here, as it's returning three distinct types.
Please remove the named returns.

FabianKramm (approved):
@facchettos great work, looks great! LGTM!

PR #1636, opened 2024-03-29
Docs: Update config.go comments

aimeeu:
Add content to comments that belongs with the field rather than separate content on the docs page.

/kind documentation

Part of ENG-3147


config/config.go
```diff
@@ -763,46 +763,47 @@ type HostPathMapper struct {
...
 	Service CoreDNSService `json:"service,omitempty"`
-
-	// Deployment holds extra options for the coredns deployment deployed within the virtual cluster
+	
+    // Deployment holds extra options for the CoreDNS deployment deployed within the virtual cluster. 
+	// Customize the CoreDNS Deployment spec, metadata.labels, and metadata.annotations.
 	Deployment CoreDNSDeployment `json:"deployment,omitempty"`
 
-	// OverwriteConfig can be used to overwrite the coredns config
+	// Overwrite default config. Path to a custom Corefile. See https://coredns.io/2017/07/23/corefile-explained/.
```
johannesfrey, line 779: Same here.

config/config.go
```diff
@@ -870,7 +871,7 @@ type ControlPlaneAdvanced struct {
 	// upload all required vCluster images to a single private repository and set this value. Workload images are not affected by this.
 	DefaultImageRegistry string `json:"defaultImageRegistry,omitempty"`
 
-	// VirtualScheduler defines if a scheduler should be used within the virtual cluster or the scheduling decision for workloads will be made by the host cluster.
+	// Defines if a scheduler should be used within the virtual cluster or the scheduling decision for workloads will be made by the host cluster.
```
johannesfrey, line 874: Same here.

aimeeu:
Thanks for your comments @johannesfrey 
For the docs, starting the comment with the name of the exported field/type/func is redundant.

For example:
image


I can work within Go comment style guidelines if if I have to. For user-facing text, this is a case of deciding whether to strictly follow Go comments style or present content that's more user-friendly but doesn't strictly follow Go comment style rules. We aren't generating a Go API guide or Go library docs, if you see my point.

Probably better to let this PR sit until engineering is done updating the YAML structure. Then we can decide how to structure the comments, and I can update in a single attempt (I don't like merging/rebasing lol).

johannesfrey:
> Thanks for your comments @johannesfrey For the docs, starting the comment with the name of the exported field/type/func is redundant.
> 
> For example: image
> 
> I can work within Go comment style guidelines if if I have to. For user-facing text, this is a case of deciding whether to strictly follow Go comments style or present content that's more user-friendly but doesn't strictly follow Go comment style rules. We aren't generating a Go API guide or Go library docs, if you see my point.
> 
> Probably better to let this PR sit until engineering is done updating the YAML structure. Then we can decide how to structure the comments, and I can update in a single attempt (I don't like merging/rebasing lol).

Your welcome 🙂 .

Yes I totally get your point. Indeed we should rather strive for user-friendliness instead of strictly following the go style guidelines. Especially as we are using them verbatim as user-facing docs, which I forgot while doing the review.

PR #1641, opened 2024-04-02
DO NOT MERGE feat(docs): add version dropdown to link to new docs and mark current as legacy

pascalbreuninger:


pascalbreuninger:
do not merge

PR #1646, opened 2024-04-03
feat: add migration function

FabianKramm:
/kind feature

Resolves ENG-3085

**What does it do?**
* This PR adds a migration function that is able to convert the old helm `values.yaml` into the new `vcluster.yaml` format.
* Changes the default value of `fallbackHostDns` back to `false`, which was a mistake introduced when changing the `values.yaml` format


heiko-braun:
Thanks @FabianKramm, what about the cases following line https://github.com/loft-sh/vcluster/blob/f783bcd2e3f425ad759499e63ff71a1d7f5badf1/pkg/config/legacyconfig/migrate.go#L747

Are these cases where we bail out and the user has to manually remove fields from the previous config?

PR #1652, opened 2024-04-05
Define global in subchart json schema

### LINKED ISSUE #1651: Helm chart: Add global property to schema to enable use as subchart

It is not possible to use the vcluster chart as a subchart because of this problem:

https://github.com/helm/helm/issues/10392

Employ the [suggested workaround](https://github.com/helm/helm/issues/10392#issuecomment-1513258871) that has been applied to the cert-manager and open-telemetry charts, to name a few:
* https://github.com/cert-manager/helm-tool/pull/30
* https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/values.schema.json#L30-L36

N/A

### Additional context

https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values

joebowbeer:
/kind bugfix

resolves #1651

Fixed an issue where vcluster helm chart could not be used as a subchart.

This applies the [suggested workaround](https://github.com/helm/helm/issues/10392#issuecomment-1513258871) in the simplest possible way, and enables my invocation of helm template to succeed.


heiko-braun:
Hi @joebowbeer thanks for the contribution. The schema.json is generated from [1] using [2].

Can you adjust the PR to get changes into [1]?

[1] https://github.com/loft-sh/vcluster/blob/main/config/config.go
[2] https://github.com/loft-sh/vcluster/blob/3b4b5fb22b395d27f221a46df4cb30f2fc900f12/Justfile#L75

joebowbeer:
@heiko-braun I'll give it a shot. I suspect that the cert-manager approach will work here.

FabianKramm:
@joebowbeer if global shows up in the values.yaml it needs to be in the config as well, as we pass the whole values.yaml to the syncer and that will strictly parse it. So I guess we need to add the following to the config.go:
```
Global interface{} `json:"global,omitempty"`
```

joebowbeer:
@FabianKramm updated, PTAL

FabianKramm (approved):

PR #1660, opened 2024-04-08
Do not encode distro name into Chart URL

johannesfrey:
/kind bugfix

Resolves ENG-3230

Fixed an issue where vcluster upgrade leads to a 404 when upgrading a vcluster that was created before the unified Helm Chart. 


FabianKramm (approved):
@johannesfrey thanks for the PR! LGTM!

PR #1668, opened 2024-04-10
Prevent distro and backingstore switch

johannesfrey:
/kind enhancement

Resolves ENG-3142

Adds more heuristics to decide if a user tries to switch distro or backing stores.
Those "brittle" checks might be sufficient in the first iteration but should definitely made more robust in a follow-up!?


pkg/setup/config.go
```diff
@@ -64,5 +71,75 @@ func InitConfig(vConfig *config.VirtualClusterConfig) error {
 		}
 	}
 
+	// check if previously we were using eks or k8s as distro
+	if vConfig.Distro() != vclusterconfig.K8SDistro && vConfig.Distro() != vclusterconfig.EKSDistro {
```
FabianKramm, line 75: These checks won't work unfortunately as the /binaries folder is an empty dir which is wiped if the pod is recreated

pkg/setup/config.go
```diff
@@ -48,20 +54,52 @@ func InitConfig(vConfig *config.VirtualClusterConfig) error {
...
+
+		if annotatedStore, ok := secret.Annotations[AnnotationStore]; ok {
+			if string(backingStoreType) != annotatedStore {
+				return fmt.Errorf("seems like you were using %s as a store before and now have switched to %s, please make sure to not switch between vCluster stores", annotatedStore, vConfig.BackingStoreType())
+			}
+		} else {
+			// Otherwise we're dealing with a fresh start, and we can just set the initial used store.
+			secret.Annotations[AnnotationStore] = string(backingStoreType)
+
+			if _, err := vConfig.ControlPlaneClient.CoreV1().Secrets(vConfig.ControlPlaneNamespace).Update(ctx, secret, metav1.UpdateOptions{}); err != nil {
```
FabianKramm, line 95: Have we verified this is not overriden when a helm upgrade is done? Does this still work when using `controlPlane.advanced.globalMetadata.annotations`?

FabianKramm, line 95: Isn't 2 update a bit extreme for changing 2 annotations, why not using a single update or patch here?

FabianKramm (changes requested):

FabianKramm:
@johannesfrey @ThomasK33 please see my comments

pkg/helm/secrets.go
```diff
@@ -170,6 +171,46 @@ func (secrets *Secrets) Update(ctx context.Context, secret *corev1.Secret) (*cor
...
+	for _, item := range list.Items {
+		cpy := item
+		rls, err := decodeRelease(&cpy, string(item.Data["release"]))
+		if err != nil {
+			klog.Infof("list: failed to decode release: %v", err)
+			continue
+		} else if rls.Chart == nil || rls.Chart.Metadata == nil || rls.Info == nil {
+			klog.Infof("list: metadata info is empty of release: %s", rls.Name)
+			continue
+		}
```
johannesfrey, line 205:
```suggestion
		}
		results = append(results, rls)
```

pkg/setup/config.go
```diff
@@ -48,19 +61,248 @@ func InitConfig(vConfig *config.VirtualClusterConfig) error {
...
+	ls := kblabels.Set{}
+	ls["name"] = name
+	releases, err := helm.NewSecrets(client).ListUnfiltered(ctx, ls.AsSelector(), namespace)
+	if err != nil || len(releases) == 0 {
+		return false, nil
+	}
+
+	// (ThomasK33): if there is only one revision, we're dealing with an initial installation
+	// at which point we can just exit
+	if len(releases) == 1 {
```
johannesfrey, line 120: Why is that? Couldn't it be that a user only installed one 0.19.x revision before?

pkg/helm/secrets.go
FabianKramm: Good find! Fixed!

pkg/setup/config.go
FabianKramm: No since v0.20 code is executed as this is part of the control plane image, so it must be v0.20 as we know latest release is v0.20.

FabianKramm (approved):

PR #1671, opened 2024-04-11
feat: add "migrate values" command to migrate to new values format

johannesfrey:
/kind enhancement

Resolves ENG-3274

Add "migrate values" command to vcluster in order to convert the old values format to the new format introduced with 0.20

Current caveat is that the passed in values actually match`--distro`. Otherwise the `MigrateLegacyConfig` currently will simply create new default values for the given distro in the new format. So I made `--distro` mandatory with a warning that it should exactly match the provided values.

### DIFF
--- a/cmd/vclusterctl/cmd/migrate/values.go
+++ b/cmd/vclusterctl/cmd/migrate/values.go
@@ -38,10 +38,9 @@ func migrateValues(globalFlags *flags.GlobalFlags) *cobra.Command {
 Migrates values for a vcluster to the new format
 
 Examples:
-vcluster migrate values -f /my/k8s/values.yaml
-vcluster migrate values --distro k3s -f /my/k3s/values.yaml
-vcluster migrate values --distro k0s < /my/k0s/values.yaml
-cat /my/k8s/values.yaml | vcluster migrate values --distro k8s
+vcluster migrate values --distro k8s -f /my/k8s/values.yaml
+vcluster migrate values --distro k3s < /my/k3s/values.yaml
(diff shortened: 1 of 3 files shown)


heiko-braun:
Maybe we should make `--distro` mandatory to prevent people accidentally create configs for wrong distro unintentionally?

cmd/vclusterctl/cmd/migrate/values.go
```diff
@@ -0,0 +1,113 @@
...
+		Use:   "values",
+		Short: "Migrates current cluster values",
+		Long: `
+#######################################################
+############### vcluster migrate values ###############
+#######################################################
+Migrates values for a vcluster to the new format
+
+Examples:
+vcluster migrate values -f /my/k8s/values.yaml
```
heiko-braun, line 41: @johannesfrey with the distro being mandatory, this would need to adapted too

cmd/vclusterctl/cmd/migrate/values.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	cobraCmd := &cobra.Command{
+		Use:   "values",
+		Short: "Migrates current cluster values",
+		Long: `
+#######################################################
+############### vcluster migrate values ###############
+#######################################################
+Migrates values for a vcluster to the new format
```
johannesfrey, line 38: Not sure if we should be explicit here?
```suggestion
Migrates values for a vcluster to the 0.20 format
```

heiko-braun: Yes, I think so.

heiko-braun (approved):
lgtm

PR #1672, opened 2024-04-11
now takes into account the parameters from the kubeconfig config part

facchettos:
/kind bugfix

resolves eng-3330

Fixed an issue where vcluster wouldn't take into account some parameters


pkg/k8s/k8s.go
```diff
@@ -152,19 +152,20 @@ func StartK8S(
...
 				args = append(args, "--node-monitor-grace-period=180s")
 				args = append(args, "--node-monitor-period=30s")
 				args = append(args, "--pvclaimbinder-sync-period=60s")
-				args = append(args, "--requestheader-client-ca-file=/data/pki/front-proxy-ca.crt")
-				args = append(args, "--root-ca-file=/data/pki/ca.crt")
+				args = append(args, "--requestheader-client-ca-file="+vConfig.VirtualClusterKubeConfig().RequestHeaderCACert)
+				args = append(args, "--root-ca-file="+vConfig.VirtualClusterKubeConfig().ServerCACert)
 				args = append(args, "--service-account-private-key-file=/data/pki/sa.key")
 				args = append(args, "--use-service-account-credentials=true")
+				args = append(args, "--v=8")
```
FabianKramm, line 168: I think this should be not there

facchettos: ah oops indeed

pkg/k8s/k8s.go
```diff
@@ -152,17 +152,17 @@ func StartK8S(
 				args = append(args, "--authentication-kubeconfig=/data/pki/controller-manager.conf")
 				args = append(args, "--authorization-kubeconfig=/data/pki/controller-manager.conf")
 				args = append(args, "--bind-address=127.0.0.1")
-				args = append(args, "--client-ca-file=/data/pki/ca.crt")
+				args = append(args, "--client-ca-file="+vConfig.VirtualClusterKubeConfig().ServerCACert)
```
FabianKramm, line 155: This should be `ClientCACert`

PR #1673, opened 2024-04-12
fix: add coredns version map with key equales to 1.26 item

LinPr:
/kind bugfix

resolves #

fix: add coredns version map with kubernetes version equales to 1.26 item

According to https://github.com/coredns/deployment/blob/master/kubernetes/CoreDNS-k8s_version.md
when instsall kubernetes version 1.26,  it is supposed to install coredns/coredns:1.9.3 respectively
so I add this bug fix


FabianKramm (approved):
@LinPr thanks a lot for the PR! Good catch, yes we were actually missing that! LGTM!

PR #1674, opened 2024-04-12
added check for k0s to make sure we don't use incompatible options

facchettos:
/kind enhancement

resolves eng-3330


heiko-braun (approved):
lgtm

PR #1675, opened 2024-04-12
chore: validate passed in values files before passing it to helm

johannesfrey:
/kind enhancement

Resolves ENG-3334

Fixed an issue where vcluster would pass the values file to `helm install` without validating them beforehand.

### DIFF
--- a/cmd/vclusterctl/cmd/create.go
+++ b/cmd/vclusterctl/cmd/create.go
@@ -225,9 +225,11 @@ func (cmd *CreateCmd) Run(ctx context.Context, args []string) error {
 		defer f.Close()
 
 		cfg := &config.Config{}
-		err = cfg.Validate(f)
+		err = cfg.DecodeYAML(f)
 		if err != nil {
-			cmd.log.Infof("Consider using %q to convert the old values format to the new 0.20 format", "vcluster migrate values")
+			if errors.Is(err, config.ErrInvalidFileFormat) {
+				cmd.log.Infof("If you are using the old values format, consider using %q to convert it to the new 0.20 format", "vcluster migrate values")
+			}
 			return err
 		}
 	}
--- a/cmd/vclusterctl/cmd/create.go
+++ b/cmd/vclusterctl/cmd/create.go
@@ -184,22 +184,6 @@ func (cmd *CreateCmd) Run(ctx context.Context, args []string) error {
 		return err
 	}
 
-	// find out kubernetes version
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -72,6 +76,28 @@ type Config struct {
 	Plugin map[string]Plugin `json:"plugin,omitempty"`
 }
 
+func (c *Config) Validate(r io.Reader) error {
```
ThomasK33, line 79: Wouldn't `Decode` or `Unmarshall` be more fitting?

cmd/vclusterctl/cmd/create.go
```diff
@@ -232,9 +216,41 @@ func (cmd *CreateCmd) Run(ctx context.Context, args []string) error {
...
+		f, err := os.Open(p)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+
+		cfg := &config.Config{}
+		err = cfg.Validate(f)
+		if err != nil {
+			cmd.log.Infof("Consider using %q to convert the old values format to the new 0.20 format", "vcluster migrate values")
```
ThomasK33, line 230: This might not always be the case, as we also return errors when unable to ReadAll or convert the YAML to JSON.

We might want to introduce a new, concrete error `var ErrInvalidFileFormat = errors.New("invalid file format")` and wrap this error and the decode error together, then check here using `errors.Is` if the actual error returned.

config/config.go
johannesfrey: Good point. I'm even considering `DecodeYAML` to signal that we want a reader that we can read yaml from?!

ThomasK33: Yeah, I think DecodeYAML would be even more fitting

config/config_test.go
```diff
@@ -0,0 +1,104 @@
...
+`)),
+			},
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := &Config{}
+			if err := c.DecodeYAML(tt.args.r); (err != nil) != tt.wantErr {
+				t.Errorf("Config.Validate() error = %v, wantErr %v", err, tt.wantErr)
```
ThomasK33, line 100:
```suggestion
				t.Errorf("Config.DecodeYAML() error = %v, wantErr %v", err, tt.wantErr)
```

ThomasK33 (approved):

PR #1677, opened 2024-04-12
feat: show a message if user attempts to use pro / platform features

lizardruss:
/kind enhancement

resolves ENG-3327

Log a message if user attempt to use a pro feature before supported

Currently logs a message any time *any* pro feature is enabled, but an additional way to disable the check may be desired.

Also adds missing pro attributes for schema partials generation.


heiko-braun:
Fabians latest changes in essence add an ENV VAR that allow to override the checks Russ has put into place:

```
func ShouldCheckForProFeatures() bool {
	return os.Getenv("FORCE_VCLUSTER_PRO") != "true"
}
```

Issue #1654, opened 2024-04-05
Unable to create a virtual cluster


Arpit-yb:
### What happened?

vcluster create command itself is failing with the following error - 
~ ❯ vcluster create vc-1                                                                                                                                                                  
17:51:35 info Creating namespace vcluster-vc-1
17:51:36 info Create vcluster vc-1...
17:51:36 info execute command: helm upgrade vc-1 /var/folders/86/q11k1v1j51x8nbhcr9vzd80r0000gn/T/vcluster-0.19.5.tgz-1664191840 --kubeconfig /var/folders/86/q11k1v1j51x8nbhcr9vzd80r0000gn/T/4044563248 --namespace vcluster-vc-1 --install --repository-config='' --values /var/folders/86/q11k1v1j51x8nbhcr9vzd80r0000gn/T/1873813573
17:51:42 fatal error executing helm upgrade vc-1 /var/folders/86/q11k1v1j51x8nbhcr9vzd80r0000gn/T/vcluster-0.19.5.tgz-1664191840 --kubeconfig /var/folders/86/q11k1v1j51x8nbhcr9vzd80r0000gn/T/4044563248 --namespace vcluster-vc-1 --install --repository-config='' --values /var/folders/86/q11k1v1j51x8nbhcr9vzd80r0000gn/T/1873813573: Release "vc-1" does not exist. Installing it now.
Error: parse error at (vcluster/templates/_plugin.tpl:8): function "continue" not defined

This was working earlier, I started hitting this issue after upgrading the version to 0.19.5. 

### What did you expect to happen?

virtual cluster should have been created successfully

### How can we reproduce it (as minimally and precisely as possible)?

1. Install vcluster version 0.19.5
2. run - vcluster create <VC_NAME>

### Host cluster Kubernetes version

<details>

```console
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:18:45Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #1680, opened 2024-04-15
chore: adjust helm version check to be more explicit

johannesfrey:
/kind enhancement

Related to https://github.com/loft-sh/vcluster/issues/1654

### DIFF
--- a/pkg/util/clihelper/clihelper.go
+++ b/pkg/util/clihelper/clihelper.go
@@ -2,6 +2,7 @@ package clihelper
 
 import (
 	"context"
+	"fmt"
 	"math/rand"
 	"net"
 	"sort"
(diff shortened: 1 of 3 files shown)


pkg/util/clihelper/clihelper.go
```diff
@@ -144,8 +146,9 @@ func HasPodProblem(pod *corev1.Pod) bool {
 }
 
 func CheckHelmVersion(output string) error {
-	if !(strings.Contains(output, "Version:\"v3.")) {
-		return errors.New("Please ensure that the \"helm\" binary in your PATH is valid. Only Helm v3 is supported")
+	min := "v3.10.0"
```
ThomasK33, line 149: Can we make this a public const?

johannesfrey: Sure, in this package here or rather in the helm package?

ThomasK33: I think it's fine to keep it here, given that it'll be used in this package.
Can always refactor it later on, if it makes more sense somewhere else

ThomasK33 (approved):

FabianKramm (approved):

PR #1697, opened 2024-04-16
Updating from `product:pro` to `tier:pro` in config.go

deniseschannon:
Fixes ENG-3269

Adding in more comments on which are pro features

### DIFF
--- a/chart/values.schema.json
+++ b/chart/values.schema.json
@@ -56,7 +56,7 @@
         },
         "hostPathMapper": {
           "$ref": "#/$defs/HostPathMapper",
-          "description": "HostPathMapper defines if vCluster should rewrite host paths.",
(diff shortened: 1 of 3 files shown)


heiko-braun:
@deniseschannon Looks good, but I wonder what other code places need to be changed that currently expect `product: pro`?

deniseschannon:
@heiko-braun nowhere else as it was only used in this file right before v0.20 beta was released. The “addProToJSONSchema” function was added very recently so that the json schema has the pro tag. Afterwards the json schema is used in the docs repo to generate the Config references which we refer to as “partials”.

I wanted to update how we update the json schema to use tier: pro instead of pro:true but it wasn’t a priority for me. It would require way more time on my end as I’m a bit rusty on my development skills to change it in here and then how it’s fed and used in the docs repo. When we work on the automation of the schema changes into the docs repo, we should fix it then.

zerbitx (approved):

johannesfrey (approved):

PR #1701, opened 2024-04-17
fix: NetworkPolicies of Control-Planes

leoncamel:
/kind bugfix

resolves #

This pull request will fix networkpolicies of control-planes. Since from v0.19.x, the control-plane will pack into single
pod. The issue is not obvious. 

We encounter a issue with v0.18.x + k8s. After some investigation from our team, we found
the rule to system's host-dns is missing.

And we look into several version, we think the networkpolicies should be fixed.

### DIFF
--- a/chart/templates/networkpolicy.yaml
+++ b/chart/templates/networkpolicy.yaml
@@ -30,7 +30,7 @@ spec:
     - ports:
         - port: 443
         - port: 8443
-      to:
+    - to:
         - podSelector:
             matchLabels:
               release: {{ .Release.Name }}
@@ -78,21 +78,25 @@ spec:
     matchLabels:
       release: {{ .Release.Name }}
   egress:
-    # Allows outgoing connections to all pods with
-    # port 443, 8443 or 6443. This is needed for host Kubernetes
-    # access
+    # Allows outgoing connections to all necessary pods with
+    # port 443, 8443 or 6443 or system dns. This is needed for
+    # host Kubernetes access.
     - ports:
         - port: 443
         - port: 8443
         - port: 6443
-    # Allows outgoing connections to all vcluster workloads
-    # or kube system dns server
+        - port: 53
+          protocol: UDP
+        - port: 53
+          protocol: TCP
     - to:
-        - podSelector: {}
+        - podSelector:
+            matchLabels:
+              release: {{ .Release.Name }}
         - namespaceSelector:
             matchLabels:
               kubernetes.io/metadata.name: 'kube-system'
-          podSelector:
+        - podSelector:
             matchLabels:
               k8s-app: kube-dns
   policyTypes:


chart/templates/networkpolicy.yaml
```diff
@@ -78,21 +78,25 @@ spec:
...
     - to:
-        - podSelector: {}
+        - podSelector:
+            matchLabels:
+              release: {{ .Release.Name }}
         - namespaceSelector:
             matchLabels:
               kubernetes.io/metadata.name: 'kube-system'
-          podSelector:
+        - podSelector:
```
FabianKramm, line 99: Is this necessary? Since vCluster should only communicate with pods in the namespace `kube-system` and the labels `k8s-app=kube-dns`, this would change the logic to an `OR` and then basically allow communication with all pods in the kube-system namespace.

chart/templates/networkpolicy.yaml
```diff
@@ -78,21 +78,25 @@ spec:
...
     - ports:
         - port: 443
         - port: 8443
         - port: 6443
-    # Allows outgoing connections to all vcluster workloads
-    # or kube system dns server
+        - port: 53
+          protocol: UDP
+        - port: 53
+          protocol: TCP
```
FabianKramm, line 91: Since we already have the rule below that selects DNS in the kube-system namespace, why is this still required?

chart/templates/networkpolicy.yaml
```diff
@@ -30,7 +30,7 @@ spec:
     - ports:
         - port: 443
         - port: 8443
-      to:
+    - to:
```
FabianKramm, line 33: This changes logic from an `and` to an `or`, why is this required? This basically allows communication with all pods on port 443 and 8443 as well as all ports on the control-plane

chart/templates/networkpolicy.yaml
leoncamel: OK. I misunderstand the grammar of `NetworkPolicyEgressRule`. Just revert back

chart/templates/networkpolicy.yaml
leoncamel: Same as previous one. I misunderstand the grammar of `NetworkPolicyEgressRule`. Just revert back

chart/templates/networkpolicy.yaml
leoncamel: The previous NetworkPolicy `vc-work-{{ .Release.Name }}` will work on **workload node** based on its podSelector:

```
  podSelector:
    matchLabels:
      vcluster.loft.sh/managed-by: {{ .Release.Name }}
```

And, the control-plane of a vcluster still need to communicate **host's kube-dns**. That is why we want add this rule. And it did crash without this rule.

leoncamel:
@FabianKramm I just revert two typos, and explain the main idea about adding udp/53 rule for control-plane.

leoncamel:
@FabianKramm Any comment on this?

cbalan:
Hello. Thank you for your contribution. Via #3262, networkPolicies configuration has been refactored to accommodate ingress rules as well. I believe the changes discussed here are already addressed there. 

Please let me know if the original use case is not covered. Closing this pull request for now.

PR #1723, opened 2024-04-29
fix for synced pvs and retain SC

### LINKED ISSUE #1464: PV with reclaim policy can never be claimed again in vclusters
### What happened?

With a Vcluster (with persistentVolumes enabled, nodes sync enabled, hoststorageClasses enabled, A volume created with Reclaim policy can never be reclaim. The PV never return to "Available" state.

We should be able to reclaim this PV inside a vcluster and attach it again.

How to reproduce :

1. In a vCluster Create a PVC **with reclaim policy set to retain**, a PV is created
2. Remove the pods using it
3. Delete the PVC: the PV go to state "Released"
4. Edit the PV to remove in the spec the "claimRef" part
   In a non-vcluster env, the PV goes to state "Available" and can be claimed again
   In a vcluster env, the PV stays in the state "Released" and a claimRef is added (it can't be removed). The claim ref is the claim information on "host" cluster

*No response*

```console
$ kubectl version
v1.28.0
```

</details>

<details>

```
# Write here
```

</details>

### vlcuster version

<details>

```console
$ vcluster --version
0.18.1
```

v1.28.2+k3s1

### OS and Arch

```
OS: Talos
Arch:
```

facchettos:
/kind bugfix

resolves #1464
resolves eng-2701

Fixed an issue where vcluster would not make persistent volumes available when `retain` policy was set on the storage class.

The PV syncing in the host->vcluster direction was not syncing at the right time because it was relying exclusively on claimrefs, but did not detect the claimref being deleted from the vcluster object, and thus was forcing it back onto it

### DIFF
--- a/pkg/controllers/resources/persistentvolumes/syncer.go
+++ b/pkg/controllers/resources/persistentvolumes/syncer.go
@@ -34,7 +34,10 @@ func NewSyncer(ctx *synccontext.RegisterContext) (syncertypes.Object, error) {
 	return &persistentVolumeSyncer{
 		Translator: translator.NewClusterTranslator(ctx, "persistentvolume", &corev1.PersistentVolume{}, NewPersistentVolumeTranslator(), HostClusterPersistentVolumeAnnotation),
 
-		virtualClient: ctx.VirtualManager.GetClient(),
+		virtualClient:            ctx.VirtualManager.GetClient(),
+		physicalClient:           ctx.PhysicalManager.GetClient(),
+		useFakePersistentVolumes: !ctx.Config.Sync.ToHost.PersistentVolumes.Enabled,
+		isMultiNS:                ctx.Config.Experimental.MultiNamespaceMode.Enabled,
 	}, nil
 }
 
@@ -66,7 +69,10 @@ func NewPersistentVolumeTranslator() translate.PhysicalNameTranslator {
 type persistentVolumeSyncer struct {
 	translator.Translator
 
-	virtualClient client.Client
+	virtualClient            client.Client
+	physicalClient           client.Client
+	useFakePersistentVolumes bool
+	isMultiNS                bool
 }
 
 var _ syncertypes.IndicesRegisterer = &persistentVolumeSyncer{}
@@ -142,7 +148,7 @@ func (s *persistentVolumeSyncer) Sync(ctx *synccontext.SyncContext, pObj client.
 	}
 
 	// check if the persistent volume should get synced
-	sync, vPvc, err := s.shouldSync(ctx.Context, pPersistentVolume)
+	sync, vPvc, pPvc, err := s.shouldSync(ctx.Context, pPersistentVolume)
 	if err != nil {
 		return ctrl.Result{}, err
 	} else if !sync {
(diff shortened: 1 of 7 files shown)


FabianKramm (changes requested):
we need to add tests for this before we can merge here, preferably multiple unit tests + an e2e test

rohantmp:
Can do a review here, could you write a brief description?

FabianKramm:
@facchettos your tests are failing, can you fix them?

facchettos:
@rohantmp @FabianKramm I know the tests are failing, didn't have time to do more before I left for my PTO :)

heiko-braun:
@rohantmp make sense. please review it.

pkg/controllers/resources/persistentvolumes/syncer.go
```diff
@@ -142,26 +148,32 @@ func (s *persistentVolumeSyncer) Sync(ctx *synccontext.SyncContext, pObj client.
...
 	// check if there is a corresponding virtual pvc
-	updatedObj := s.translateUpdateBackwards(vPersistentVolume, pPersistentVolume, vPvc)
-	if updatedObj != nil {
-		ctx.Log.Infof("update virtual persistent volume %s, because spec has changed", vPersistentVolume.Name)
-		translator.PrintChanges(vPersistentVolume, updatedObj, ctx.Log)
-		err = ctx.VirtualClient.Update(ctx.Context, updatedObj)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
+	if syncBack {
```
FabianKramm, line 164: Why make a variable for this, can we not just put the condition here?

pkg/controllers/resources/persistentvolumes/syncer.go
```diff
@@ -240,33 +252,46 @@ func (s *persistentVolumeSyncer) SyncToVirtual(ctx *synccontext.SyncContext, pOb
...
 		if err != nil {
-			return false, nil, nil
+			return false, nil, nil, nil
 		}
-		return pObj.Spec.ClaimRef.Namespace == namespace && pObj.Spec.PersistentVolumeReclaimPolicy == corev1.PersistentVolumeReclaimRetain, nil, nil
+		return pObj.Spec.ClaimRef.Namespace == namespace && pObj.Spec.PersistentVolumeReclaimPolicy == corev1.PersistentVolumeReclaimRetain, nil, nil, nil
+	}
+
+	pPvc := &corev1.PersistentVolumeClaim{}
+	err = s.physicalClient.Get(ctx, types.NamespacedName{Namespace: pObj.Spec.ClaimRef.Namespace, Name: pObj.Spec.ClaimRef.Name}, pPvc)
```
FabianKramm, line 287: Shouldn't we validate here that the ClaimRef.Namespace actually matches the namespace we are working in as vCluster?

FabianKramm: e.g. this could cause problematic errors if we cannot list in that namespace, which would make everything fail as a non not found error is thrown

pkg/controllers/resources/persistentvolumes/translate.go
```diff
@@ -113,6 +113,11 @@ func (s *persistentVolumeSyncer) translateUpdate(ctx context.Context, vPv *corev
 		updated.Spec.Capacity = vPv.Spec.Capacity
 	}
 
+	if !equality.Semantic.DeepEqual(pPv.Spec.ClaimRef, vPv.Spec.ClaimRef) {
+		updated = translator.NewIfNil(updated, pPv)
+		updated.Spec.ClaimRef = vPv.Spec.ClaimRef
```
FabianKramm, line 118: Isn't that the wrong reference here, the claimref is not translated in this context correctly, or am I mistaken?

pkg/controllers/resources/persistentvolumes/translate.go
```diff
@@ -124,6 +129,10 @@ func (s *persistentVolumeSyncer) translateUpdate(ctx context.Context, vPv *corev
 	}
 
 	translatedStorageClassName := translateStorageClass(vPv.Spec.StorageClassName)
+	if !s.useFakePersistentVolumes {
+		translatedStorageClassName = vPv.Spec.StorageClassName
```
FabianKramm, line 133: Why is this necessary?

FabianKramm (changes requested):

pkg/controllers/resources/persistentvolumes/syncer.go
facchettos: I thought it was more readable this way, but I can change it if really needed

pkg/controllers/resources/persistentvolumes/translate.go
facchettos: because if we use real persistent volumes, they need to reference the proper storage class, for some reasons the storage class was updated to something different than the one at creation time when the persistent volume was updated

pkg/controllers/resources/persistentvolumes/syncer.go
facchettos: but isn't that what we want though? if the pv has a claimref with a pvc that we don't have access to, we don't own the pv

pkg/controllers/resources/persistentvolumes/syncer.go
FabianKramm: Yeah please change that

PR #1725, opened 2024-04-29
run tests on all commits for main

facchettos:
/kind enhancement

resolves #

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,6 +1,9 @@
 name: E2E CI
 
 on:
+  push:
+    branches:
+      - main
   release:
     types: [created]
   pull_request:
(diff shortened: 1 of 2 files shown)


FabianKramm:
@facchettos why do we need this?

facchettos:
@FabianKramm because we don't use either a merge queue or ensure that the pr branch is up to date with main, so there is some possibility that main breaks because of changes that don't conflict in git but do in the logic

FabianKramm:
Then we should enforce that the PR branch is up to date with main, because nobody is looking at the tests after merge and they just count towards our minute limit

facchettos:
enforcing the pr's head to be up to date will actually be more expensive than running the tests on main, because all PRs will need a new ci run after each merge to main before they can be merged (so best case is as expensive, any other case is more expensive)

cbron:
Closing, we can fix this problem when we run into it. For now we need to watch our minutes.

PR #1727, opened 2024-04-30
Add DecodeYAML to legacy configs

johannesfrey:
/kind enhancement

Part of ENG-3573

Add strict decoding in order to check for legacy configs.


config/legacyconfig/config.go
```diff
@@ -123,6 +175,7 @@ type BaseHelm struct {
 	Init               InitValues       `json:"init,omitempty"`
 	MultiNamespaceMode EnabledSwitch    `json:"multiNamespaceMode,omitempty"`
 	Telemetry          TelemetryValues  `json:"telemetry,omitempty"`
+	ServiceCIDR        string           `json:"serviceCIDR,omitempty"`
```
johannesfrey, line 178: Had to add this in order for 0.19 configs (where `serviceCIDR` is set per default) to work with the strict decoding with `DisallowUnknownFields`.

config/legacyconfig/config.go
```diff
@@ -199,6 +252,7 @@ type SyncNodes struct {
 }
 
 type SyncGeneric struct {
+	RBACValues
```
johannesfrey, line 255: Had to add this in order for the unit test (`generic sync example`) for 0.19 configs to work with the strict decoding with `DisallowUnknownFields`.

config/legacyconfig/config.go
```diff
@@ -12,6 +20,28 @@ type LegacyK0sAndK3s struct {
...
+
+	dec := json.NewDecoder(bytes.NewReader(j))
+	dec.DisallowUnknownFields()
+
+	err = dec.Decode(c)
+	if err != nil {
+		return fmt.Errorf("%w: %w", config.ErrInvalidFileFormat, err)
+	}
+
+	return nil
```
ThomasK33, line 42: Given that this function exists three times (including the current config decoding), can we refactor this into a generic function and then call that generic function from those three?

config/legacyconfig/config.go
ThomasK33: You could implement a custom [Unmarshaller](https://pkg.go.dev/encoding/json#Unmarshaler), that handles this specific case of unmarshalling, while leaving the rest to a DisallowUnknownFields.

config/legacyconfig/config.go
ThomasK33: Same as above, potentially use a custom Unmarshaller.

config/legacyconfig/config.go
johannesfrey: Good hint! But I would prefer to actually align the actual allowed fields in legacyconfig before doing exceptions/implement a custom marshaller. But if there is a good reason to **not** set this field here, I could totally go this route 🙂

johannesfrey: Probably @FabianKramm has more context?!

FabianKramm: This field was just forgotten here, so your solution @johannesfrey is correct, I wouldn't change that

config/legacyconfig/config.go
FabianKramm: Same as above its fine as it is

config/config.go
```diff
@@ -182,6 +161,13 @@ func (c *Config) IsProFeatureEnabled() bool {
 	return false
 }
 
+func UnmarshalYAMLStrict(data []byte, i any) error {
```
FabianKramm, line 164: Instead of any, shouldn't we use *Config here as type?

config/config.go
```diff
@@ -1,24 +1,21 @@
...
 	"errors"
 	"fmt"
-	"io"
 	"reflect"
 	"regexp"
 	"strings"
 
-	"github.com/ghodss/yaml"
 	"github.com/invopop/jsonschema"
+	"sigs.k8s.io/yaml"
```
FabianKramm, line 12: Why do we use that instead of ghodss/yaml? This will cause more dependencies on the loft-sh/vcluster-config side, is that intended?

johannesfrey: The current version of `ghodss/yaml` we depend on does not provide a strict marshaller that errors on unknown fields. But you are right, it doesn't make sense to import both `ghodss/yaml` and `sigs.k8s.io/yaml`. I could replace `ghodss/yaml` with the k8s yaml in following files, so that we only have one yaml dep? WDYT?

- https://github.com/loft-sh/vcluster/blob/main/config/diff.go#L9
- https://github.com/loft-sh/vcluster/blob/main/config/legacyconfig/migrate.go#L9

config/legacyconfig/config.go
johannesfrey: As discussed. Idea was to inline the fields under rbac directly.

FabianKramm (approved):

FabianKramm (approved):

PR #1728, opened 2024-04-30
feat: add vcluster activate & apply platform secret

FabianKramm:
/kind feature

resolves ENG-3605

**Changes:**
* Updated loft dependencies
* Added support for `vcluster activate` and auto-activate in `vcluster create`
* Fixed version problem with helm unit tests
* Removed fatal messages when using platform commands


pkg/platform/secret.go
```diff
@@ -0,0 +1,125 @@
...
+		err = c.Save()
+		if err != nil {
+			return fmt.Errorf("save vCluster platform config: %w", err)
+		}
+	}
+
+	// build secret payload
+	payload := map[string][]byte{
+		"accessKey": []byte(c.Config().VirtualClusterAccessKey),
+		"host":      []byte(strings.TrimPrefix(c.Config().Host, "https://")),
```
ThomasK33, line 81: Do we also want to trim potential HTTP from the host for insecure setups?

pkg/platform/secret.go
```diff
@@ -0,0 +1,125 @@
...
+	"strings"
+	"time"
+
+	managementv1 "github.com/loft-sh/api/v3/pkg/apis/management/v1"
+	storagev1 "github.com/loft-sh/api/v3/pkg/apis/storage/v1"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	client2 "sigs.k8s.io/controller-runtime/pkg/client"
```
ThomasK33, line 17: Not a big fan of calling this client2

ThomasK33 (approved):

pkg/platform/secret.go
FabianKramm: Actually the config / cli package ensures only https is allowed since Kubernetes needs that anyways, so I guess its safe to only cut https here

PR #1738, opened 2024-05-02
chore(vclusterctl): check for pre 0.20 values

johannesfrey:
/kind enhancement

Resolves ENG-3573

Provide checks that abort "vcluster create" when using pre-v0.20 formatting in values.yaml.

### DIFF
--- a/pkg/cli/create_helm.go
+++ b/pkg/cli/create_helm.go
@@ -1,6 +1,8 @@
 package cli
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
 	"encoding/base64"
 	"errors"
@@ -13,7 +15,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/ghodss/yaml"
 	"github.com/loft-sh/log"
 	"github.com/loft-sh/log/survey"
 	"github.com/loft-sh/log/terminal"
@@ -41,6 +42,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"sigs.k8s.io/yaml"
 )
 
 // CreateOptions holds the create cmd options
@@ -81,7 +83,7 @@ type CreateOptions struct {
 
 var CreatedByVClusterAnnotation = "vcluster.loft.sh/created"
 
-var AllowedDistros = []string{"k8s", "k3s", "k0s", "eks"}
+var AllowedDistros = []string{config.K8SDistro, config.K3SDistro, config.K0SDistro, config.EKSDistro}
 
 type createHelm struct {
 	*flags.GlobalFlags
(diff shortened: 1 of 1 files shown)


pkg/cli/create_helm.go
```diff
@@ -148,17 +150,40 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
...
+				return err
+			}
+		} else {
+			currentValues, err = helmValuesYAML(release)
+			if err != nil {
+				return err
+			}
 		}
+
+		command := fmt.Sprintf("vcluster convert config --distro %s - <<EOF\n%s\nEOF", currentDistro, string(currentValues))
```
FabianKramm, line 182: This will print a big chunk, can we not do something like:
```
helm get values ... | vcluster convert config ...
```

pkg/cli/create_helm.go
```diff
@@ -247,30 +278,6 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 		return err
 	}
 
-	// check if vcluster already exists
-	if !cmd.Upgrade {
```
FabianKramm, line 251: Why is that removed?

FabianKramm (changes requested):
See my comments

pkg/cli/create_helm.go
johannesfrey: It looked like a duplicate of line 133-151.

pkg/cli/create_helm.go
```diff
@@ -215,17 +240,23 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
...
 			}
-			return err
+			// TODO Delete after vCluster 0.19.x resp. the old config format is out of support.
+			// We cannot discriminate between k0s/k3s and eks/k8s. So we cannot prompt the actual values to convert, as this would cause false positives,
+			// because users are free to e.g. pass a k0s values file to a currently running k3s virtual cluster.
+			if isLegacyConfig(data) {
+				return fmt.Errorf("it appears you are using a vCluster configuration using pre-v0.20 formatting. Please run %q to convert the values to the latest format", "vcluster convert config")
+			}
+			// TODO end
 		}
```
facchettos, line 253: I would maybe structure the if differently here 
```go
if errors.Is(err, config.ErrInvalidConfig) {
  if isLegacyConfig(data){
    return fmt.Errorf("it appears you are using a vCluster configuration using pre-v0.20 formatting. Please run %q to convert the values to the latest format", "vcluster convert config")
  }
} else if err != nil {
  return err
}
```
that is simpler to follow imho

pkg/cli/create_helm.go
johannesfrey: Sure. Just to be sure. `helm get values` will retrieve the latest revision right? I tested locally and it looks like it 🤔

pkg/cli/create_helm.go
```diff
@@ -320,35 +327,66 @@ func isVClusterDeployed(release *helm.Release) bool {
...
+	if err := yaml.Unmarshal(releaseRaw, &helmChart); err != nil {
+		return nil, err
+	}
+
+	values, err := yaml.Marshal(helmChart.Chart.Values)
+	if err != nil {
+		return nil, err
+	}
+
+	return values, nil
```
facchettos, line 388: here you can just do 
```go
return yaml.Marshal(helmChart.Chart.Values)
```

pkg/cli/create_helm.go
```diff
@@ -320,35 +327,66 @@ func isVClusterDeployed(release *helm.Release) bool {
...
+	helmChart := struct {
+		Chart struct {
+			Values map[string]interface{} `json:"values"`
+		} `json:"chart"`
+	}{}
 
-// TODO end
+	if err := yaml.Unmarshal(releaseRaw, &helmChart); err != nil {
+		return nil, err
+	}
```
facchettos, line 381: can we use an encoder/decoder here instead of reading everything into memory or not?

pkg/cli/create_helm.go
johannesfrey: When going with [Fabian's suggestion](https://github.com/loft-sh/vcluster/pull/1738#discussion_r1592351094) this func will be removed.

pkg/cli/create_helm.go
```diff
@@ -148,17 +147,31 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
...
-	// TODO Delete after vCluster 0.19.x resp. the old config format is out of support.
-	if isVClusterDeployed(release) {
-		migratedValues, err := migrateLegacyHelmValues(release)
-		if err != nil {
-			return err
+	// TODO Refactor after vCluster 0.19.x resp. the old config format is out of support.
+	// Early abort if a user runs a virtual cluster < v0.20 without providing any values files during an upgrade.
+	// We do this because we don't want to "automagically" convert the old config implicitly, without the user
+	// realizing that the virtual cluster is running with the old config format.
+	if isVClusterDeployed(release) && isLegacyVCluster(release.Chart.Metadata.Version) && len(cmd.Values) <= 0 {
```
facchettos, line 154: little nit but I don't believe Len can return something less than 0

pkg/cli/create_helm.go
johannesfrey: **Update**: The `ErrInvalidConfig` did not provide any benefit. It was actually a left-over from when the config unmarshaling looked different. So I changed it to a simple err check.

FabianKramm (approved):

PR #1741, opened 2024-05-06
fix: set import-name correctly

FabianKramm:
/kind bugfix

Fixed an issue where the vCluster name was not set correctly and therefore the platform secret was not named correctly

### DIFF
--- a/pkg/platform/secret.go
+++ b/pkg/platform/secret.go
@@ -17,7 +17,9 @@ import (
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/platform/secret.go
```diff
@@ -17,7 +17,9 @@ import (
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func (c *client) ApplyPlatformSecret(ctx context.Context, kubeClient kubernetes.Interface, name, namespace, project string) error {
+var DefaultPlatformSecretName = "vcluster-platform-api-key"
```
facchettos, line 20: can't this be a const?

FabianKramm: Changed that!

facchettos (approved):

johannesfrey (approved):

PR #1742, opened 2024-05-06
chore(config): deprecate fields instead of removal

johannesfrey:
/kind enhancement

part of #ENG-3573

Needed so we can do a strict unmarshal of the new config format. Otherwise this won't work because values obtained from 0.20.0-beta.1 virtual clusters cannot be unmarshaled because of missing keys.

### DIFF
--- a/chart/values.schema.json
+++ b/chart/values.schema.json
@@ -1985,6 +1985,11 @@
     },
     "Platform": {
       "properties": {
+        "apiKey": {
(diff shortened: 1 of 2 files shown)


FabianKramm:
@johannesfrey do we really want to do this? Seems like additional migration effort for in-between beta's. I think its fine that this is breaking so far.

johannesfrey:
As discussed. No need to provide migration between betas.

Issue #1718, opened 2024-04-25
CoreDNS PreemptionPolicy conflicts with host cluster global default Priority Class


ruifung:
### What happened?

When trying to deploy a vcluster to my homelab cluster,
I found out that the coredns pod fails to sync because it has a preemptionPolicy set that is different from the default PriorityClass on my cluster.

The syncer fails to sync the coredns pod with the following error

```
syncer 2024-04-25 04:11:12    ERROR    controller/controller.go:329    controller pod: namespace kube-system name coredns-74987d4979-k272p: reconcileID "a0a6be23-8e5e-45bb-886c-414074d250d9": Reconciler error pods "coredns-74987d4979-k272p-x-kube-system-x-vcluster-rf-ebceb13e69" is forbidden: the string value of PreemptionPolicy (PreemptLowerPriority) must not be provided in pod spec; priority admission controller computed Never from the given PriorityClass name    {"component": "vcluster-pro"}
```

### What did you expect to happen?

For it to successfully sync the pod.
Ideally by omitting the preemptionPolicy here. Or mapping it properly to the host's system-cluster-critical PriorityClass.

### How can we reproduce it (as minimally and precisely as possible)?

1. Have a PriorityClass with globalDefault set to true with a preemption policy set to Never
2. Deploy vcluster with default settings
3. Observe syncer failing to sync coredns pod

### Anything else we need to know?

After temporarily removing the globalDefault PriorityClass on the host cluster, the pod successfully syncs, and I've observed that the resulting pod has an empty priorityClassName, which causes the host cluster to apply the default priorityClass which conflicts with the pod that has the values of system-cluster-critical inside the vcluster, but not actually having that priorityClassName.

### Host cluster Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.29.2                                                                                        
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3                                                        
Server Version: v1.29.3    
```

</details>

### Host cluster Kubernetes distribution

<details>

```
k8s
```

</details>

### vlcuster version

<details>

```console
$ vcluster --version
vcluster version 0.20.0-beta.1
```

</details>

### Vcluster Kubernetes distribution(k3s(default)), k8s, k0s)

<details>

```
k8s
```

</details>

### OS and Arch

<details>

```
OS: Talos Linux 1.7.0
Arch: x64
```

</details>

narcis96:
Hello 

Thank you for the information regarding the issue you are facing.
[...]

PR #1769, opened 2024-05-14
allow users to define a default priorityClass

narcis96:
Ref https://github.com/loft-sh/vcluster/issues/1718
Ref ENG-3563

### DIFF
--- a/chart/templates/coredns-configmap.yaml
+++ b/chart/templates/coredns-configmap.yaml
@@ -152,7 +152,7 @@ data:
 {{ toYaml .Values.controlPlane.coredns.deployment.pods.labels | indent 12 }}
           {{- end }}
         spec:
-          priorityClassName: "system-cluster-critical"
+          priorityClassName: {{ .Values.controlPlane.coredns.priorityClassName | default "\"system-cluster-critical\"" }}
           serviceAccountName: coredns
           nodeSelector:
             kubernetes.io/os: linux
--- a/chart/templates/coredns-configmap.yaml
+++ b/chart/templates/coredns-configmap.yaml
@@ -152,7 +152,7 @@ data:
 {{ toYaml .Values.controlPlane.coredns.deployment.pods.labels | indent 12 }}
           {{- end }}
         spec:
-          priorityClassName: {{ .Values.controlPlane.coredns.priorityClassName | quote}}
+          priorityClassName: {{ .Values.controlPlane.coredns.priorityClassName | quote }}
           serviceAccountName: coredns
           nodeSelector:
             kubernetes.io/os: linux
--- a/chart/templates/coredns-configmap.yaml
+++ b/chart/templates/coredns-configmap.yaml
@@ -152,7 +152,7 @@ data:
 {{ toYaml .Values.controlPlane.coredns.deployment.pods.labels | indent 12 }}
           {{- end }}
         spec:
-          priorityClassName: {{ .Values.controlPlane.coredns.priorityClassName | default "\"system-cluster-critical\"" }}
+          priorityClassName: {{ .Values.controlPlane.coredns.priorityClassName | default "system-cluster-critical" | quote }}
           serviceAccountName: coredns
           nodeSelector:
             kubernetes.io/os: linux
--- a/chart/templates/coredns-configmap.yaml
+++ b/chart/templates/coredns-configmap.yaml
@@ -152,7 +152,7 @@ data:
 {{ toYaml .Values.controlPlane.coredns.deployment.pods.labels | indent 12 }}
           {{- end }}
         spec:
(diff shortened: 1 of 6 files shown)


chart/values.yaml
```diff
@@ -457,6 +457,8 @@ controlPlane:
     overwriteManifests: ""
     # OverwriteConfig can be used to overwrite the coredns config
     overwriteConfig: ""
+    
+    priorityClassName: null
```
johannesfrey, line 461: You have to add this field under [`config.go`](https://github.com/loft-sh/vcluster/blob/main/config/config.go#L937-L955) and [`values.yaml`](https://github.com/loft-sh/vcluster/blob/main/config/values.yaml#L255) and then run `go run ./hack/schema/main.go`

johannesfrey (changes requested):

chart/values.yaml
narcis96: fixed

chart/templates/coredns-configmap.yaml
```diff
@@ -152,7 +152,7 @@ data:
 {{ toYaml .Values.controlPlane.coredns.deployment.pods.labels | indent 12 }}
           {{- end }}
         spec:
-          priorityClassName: "system-cluster-critical"
+          priorityClassName: {{ .Values.controlPlane.coredns.priorityClassName | default "\"system-cluster-critical\"" }}
```
johannesfrey, line 155:
```suggestion
          priorityClassName: {{ .Values.controlPlane.coredns.priorityClassName | default "system-cluster-critical" | quote }}
```

config/config.go
```diff
@@ -952,6 +952,9 @@ type CoreDNS struct {
 
 	// OverwriteManifests can be used to overwrite the coredns manifests used to deploy coredns
 	OverwriteManifests string `json:"overwriteManifests,omitempty"`
+
+	// PriorityClassName specifies the priority class name for CoreDNS pods.
```
johannesfrey, line 956:
```suggestion
	// PriorityClassName specifies the priority class name for the CoreDNS pods.
```

chart/values.schema.json
```diff
@@ -535,6 +535,10 @@
         "overwriteManifests": {
           "type": "string",
           "description": "OverwriteManifests can be used to overwrite the coredns manifests used to deploy coredns"
+        },
+        "priorityClassName": {
```
johannesfrey, line 539: I guess this also must find it's way into the docs: https://github.com/loft-sh/vcluster-docs/tree/main/docs/_partials/config/controlPlane/coredns

There is [`gen-partials.sh`](https://github.com/loft-sh/vcluster-docs/blob/main/hack/gen-partials.sh) which updates this. Might be that this done for the next beta release anyway though?! (cc @aimeeu)

heiko-braun:
@narcis96 
> Seems like you forgot to run 'go run hack/schema/main.go' before committing your changes!

chart/values.schema.json
aimeeu: The process is manual and yes, we generate new partials for each release.

narcis96: so we should do it in another MR, right ?

FabianKramm:
@narcis96 we should not have `system-critical` as a default anymore, its fine that we can configure it now, but it shouldn't be set by default

chart/templates/coredns-configmap.yaml
```diff
@@ -152,7 +152,7 @@ data:
 {{ toYaml .Values.controlPlane.coredns.deployment.pods.labels | indent 12 }}
           {{- end }}
         spec:
-          priorityClassName: "system-cluster-critical"
+          priorityClassName: {{ .Values.controlPlane.coredns.priorityClassName | default "" | quote}}
```
FabianKramm, line 155: Just do `{{ .Values.controlPlane.coredns.priorityClassName | quote }}` here

FabianKramm (changes requested):

chart/templates/coredns-configmap.yaml
narcis96: done

chart/templates/coredns-configmap.yaml
```diff
@@ -152,7 +152,7 @@ data:
 {{ toYaml .Values.controlPlane.coredns.deployment.pods.labels | indent 12 }}
           {{- end }}
         spec:
-          priorityClassName: "system-cluster-critical"
+          priorityClassName: {{ .Values.controlPlane.coredns.priorityClassName | quote}}
```
johannesfrey, line 155:
```suggestion
          priorityClassName: {{ .Values.controlPlane.coredns.priorityClassName | quote }}
```

johannesfrey (approved):

FabianKramm (approved):

PR #1770, opened 2024-05-14
chore(config): also enable hpm when using deprecated flag

johannesfrey:
/kind bugfix

Part of ENG-3708

Fixed an issue where vcluster did not convert a legacy config with a deprecated HPM flag.


rohantmp (approved):
LGTM :)

PR #1785, opened 2024-05-22
fix: resource already exists

imkira:
/kind bugfix

resolves #

During syncing of virtual namespaces to physical, the default service account already exists, which leads to errors during syncing. In this fix we consume the error in case the physical resource already exists.

Fixed an issue where vcluster would not sync due to already existing default service accounts

This is part of a collaborative work at Woven by Toyota with my colleagues @ajithcnambiar and @tauxi

### DIFF
--- a/pkg/controllers/syncer/translator/namespaced_translator.go
+++ b/pkg/controllers/syncer/translator/namespaced_translator.go
@@ -69,6 +69,10 @@ func (n *namespacedTranslator) SyncToHostCreate(ctx *context.SyncContext, vObj, 
 			ctx.Log.Debugf("error syncing %s %s/%s to physical cluster: %v", n.name, vObj.GetNamespace(), vObj.GetName(), err)
 			return ctrl.Result{RequeueAfter: time.Second}, nil
 		}
+		if kerrors.IsAlreadyExists(err) {
+			ctx.Log.Debugf("ignoring syncing %s %s/%s to physical cluster as it already exists", n.name, vObj.GetNamespace(), vObj.GetName())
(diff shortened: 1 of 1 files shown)


imkira:
@heiko-braun there's also this one which I rebased. PTAL cheers.

pkg/controllers/syncer/translator/namespaced_translator.go
```diff
@@ -80,6 +80,10 @@ func (n *namespacedTranslator) SyncToHostCreate(ctx *context.SyncContext, vObj,
 			ctx.Log.Debugf("error syncing %s %s/%s to physical cluster: %v", n.name, vObj.GetNamespace(), vObj.GetName(), err)
 			return ctrl.Result{RequeueAfter: time.Second}, nil
 		}
+		if kerrors.IsAlreadyExists(err) {
+			ctx.Log.Debugf("ignoring syncing %s %s/%s to physical cluster as it already exists", n.name, vObj.GetNamespace(), vObj.GetName())
```
zerbitx, line 84: Beyond a need to update this PR again with the new generics code, I'll have to give this some thought.  It has essentially moved to [here](https://github.com/loft-sh/vcluster/blob/main/pkg/syncer/syncer.go#L546) but placing it in such a central place means that any resource trying to create on sync that already exists could take over the original object and start making changes to it.  It might be better to let it fail here as its already doing, or put checks on specific resources like the [service accounts](https://github.com/loft-sh/vcluster/blob/main/pkg/controllers/resources/serviceaccounts/syncer.go#L56).

PR #1789, opened 2024-05-22
Adds enableServiceLinks to the STS config for controlPlane and controlPlane.backingStore

### LINKED ISSUE #1708: Allow setting enableServiceLinks to false in etcd and syncer
### What happened?

We are experimenting using vcluster to shard a cluster with 10K+ services between multiple vclusters.

It has been working perfectly for 2 weeks but we hit the first issue today: the etcd statefulset and the syncer deployment died and failed to start up due to there being too many environment variables. This is a known issue in clusters with too many services, and the workaround we usually employ is to disable pod's service links, as each service ends up as another environment variable to a pod.

Currently, however, there is no way to disable service links in the k8s chart 
https://github.com/loft-sh/vcluster/blob/v0.19/charts/k8s/templates/etcd-statefulset.yaml

Edit to add: we have disabled service links manually in the etcd statefulset and syncer deployment and everything seems to be working.

Either:
- for enableServiceLinks to be set to false (if vcluster does not work without DNS service discovery in the host cluster, I don't think there is a reason to leave that to the default true value)
- a chart value allowing service links to be disabled

Create 10K+ services in a cluster.

```
k get svc -A | wc -l
15405
```

I could open a PR for this (or for the related #1622 which would also fix this issue for us). I just see the charts on the v0.19 branch, is that the branch a potential PR should target?

<details>

```console
$ kubectl version
Client Version: v1.28.0
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.25.16-eks-b9c9ed7
WARNING: version difference between client (1.28) and server (1.25) exceeds the supported minor version skew of +/-1

```

</details>

<details>

```
# Write here
```

</details>

### vlcuster version

<details>

```console
$ vcluster --version
vcluster version 0.19.5
```

</details>

<details>

```
k8s
```

</details>

### OS and Arch

<details>

```
OS: linux
Arch: amd64
```
</details>

zerbitx:
/kind enhancement
/kind test

resolves ENG-3500
resolves #1708 

Added the ability to disable service links for the etcd STS as they can cause problems in a cluster with thousands of services.

vcluster.yaml docs should be updated as well.

👇 Demo showing creating one vCluster for each of the new `enableServiceLinks` config options.  Then getting the STS yaml, and parsing out the value that should, and does, read false in both cases:
gh1708-demo

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -620,7 +620,7 @@ type ControlPlaneStatefulSet struct {
 	// Persistence defines options around persistence for the statefulSet.
 	Persistence ControlPlanePersistence `json:"persistence,omitempty"`
 
-	// EnableServiceLinks for the sts pod
+	// EnableServiceLinks for the StatefulSet pod
 	EnableServiceLinks *bool `json:"enableServiceLinks,omitempty"`
 
 	LabelsAndAnnotations `json:",inline"`
@@ -883,7 +883,7 @@ type EtcdDeployStatefulSet struct {
 	// Enabled defines if the statefulSet should be deployed
 	Enabled bool `json:"enabled,omitempty"`
 
-	// EnableServiceLinks for the sts pod
+	// EnableServiceLinks for the StatefulSet pod
 	EnableServiceLinks *bool `json:"enableServiceLinks,omitempty"`
 
 	// Image is the image to use for the external etcd statefulSet
--- a/config/config.go
+++ b/config/config.go
@@ -620,6 +620,9 @@ type ControlPlaneStatefulSet struct {
 	// Persistence defines options around persistence for the statefulSet.
 	Persistence ControlPlanePersistence `json:"persistence,omitempty"`
 
+	// EnableServiceLinks for the sts pod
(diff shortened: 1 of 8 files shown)


config/config.go
```diff
@@ -620,6 +620,9 @@ type ControlPlaneStatefulSet struct {
 	// Persistence defines options around persistence for the statefulSet.
 	Persistence ControlPlanePersistence `json:"persistence,omitempty"`
 
+	// EnableServiceLinks for the sts pod
```
FabianKramm, line 623: Since we use StatefulSet or statefulSet everywhere, we should do:
```
EnableServiceLinks for the StatefulSet pod
```

config/config.go
```diff
@@ -880,6 +883,9 @@ type EtcdDeployStatefulSet struct {
 	// Enabled defines if the statefulSet should be deployed
 	Enabled bool `json:"enabled,omitempty"`
 
+	// EnableServiceLinks for the sts pod
```
FabianKramm, line 886: Same as above

FabianKramm (approved):
LGTM

PR #1790, opened 2024-05-23
Platform secrets

facchettos:
/kind feature

resolves eng-3731
depends on #1779 

### DIFF
--- a/cmd/vclusterctl/cmd/platform/set/secret.go
+++ b/cmd/vclusterctl/cmd/platform/set/secret.go
@@ -30,9 +30,8 @@ var (
 type SecretType string
 
 const (
-	ProjectNamespacePrefix            = "loft-p-"
-	ProjectSecret          SecretType = "Project Secret"
-	SharedSecret           SecretType = "Shared Secret"
+	ProjectSecret SecretType = "Project Secret"
+	SharedSecret  SecretType = "Shared Secret"
 )
 
 // SecretCmd holds the flags


cmd/vclusterctl/cmd/platform/get/get.go
```diff
@@ -19,5 +19,6 @@ func NewGetCmd(globalFlags *flags.GlobalFlags, cfg *config.CLI) *cobra.Command {
 	}
 
 	cmd.AddCommand(newClusterCmd(globalFlags, cfg))
+	cmd.AddCommand(NewSecretCmd(globalFlags, defaults))
```
ThomasK33, line 22: If `NewSecretCmd` is public, shouldn't `newClusterCmd` also be public?

cmd/vclusterctl/cmd/platform/set/secret.go
```diff
@@ -0,0 +1,281 @@
...
+	if namespace == "" {
+		namespace = "loft"
+	}
+
+	return namespace, nil
+}
+
+func GetProjectSecretNamespace(project string) (string, error) {
+	return ProjectNamespacePrefix + project, nil
+}
```
ThomasK33, line 281: This function doesn't work with the latest platform version.
Project namespace prefixes must be fetched at runtime and included in the selves response.

ThomasK33 (changes requested):

cmd/vclusterctl/cmd/platform/get/get.go
facchettos: no, it should be private too, just copied over from the loftctl code

cmd/vclusterctl/cmd/platform/set/secret.go
facchettos: yeah good catch I missed a commit from loftctl

cmd/vclusterctl/cmd/platform/set/secret.go
```diff
@@ -0,0 +1,273 @@
...
+)
+
+var (
+	ErrNoSecret = errors.New("please specify a secret key to set")
+)
+
+type SecretType string
+
+const (
+	ProjectNamespacePrefix            = "loft-p-"
```
ThomasK33, line 33: This constant is unused

cmd/vclusterctl/cmd/root.go
```diff
@@ -87,6 +90,10 @@ func BuildRoot(log log.Logger) (*cobra.Command, error) {
 	persistentFlags := rootCmd.PersistentFlags()
 	globalFlags = flags.SetGlobalFlags(persistentFlags, log)
 	cfg := globalFlags.LoadedConfig(log)
+	defaults, err := pdefaults.NewFromPath(pdefaults.ConfigFolder, pdefaults.ConfigFile)
+	if err != nil {
+		log.Debugf("Error loading defaults: %v", err)
```
ThomasK33, line 95: Shouldn't we exit here if default loading fails?

ThomasK33 (changes requested):

ThomasK33 (approved):

PR #1805, opened 2024-05-28
fix: credits command not working

rohantmp:
/kind bugfix

part of ENG-3803

Fixed an issue where vcluster credits was not properly populated


ThomasK33 (changes requested):
I know this isn't part of this PR, but vCluster does not use `modernc.org/mathutil` or `github.com/mattn/go-localereader`.

Would you mind removing the license attributions for those libs?

rohantmp:
done :)

ThomasK33 (approved):

PR #1806, opened 2024-05-29
removed get service cidr

facchettos:

/kind feature
resolves eng-3728


cmd/vclusterctl/cmd/get/get.go
ThomasK33, line 1: The get command is still needed, so you cannot delete it here.

ThomasK33 (dismissed):

cmd/vclusterctl/cmd/get/get.go
facchettos: `vcluster platform get` is needed, not `vcluster get`, there's only one subcommand for `vcluster get` which is `vcluster get service-cidr`

ThomasK33 (approved):

PR #1808, opened 2024-05-29
fix: disable global owner for multins mode

rohantmp:
/kind bugfix

Earlier, it was not obvious that global ownerref was disable for multins mode, and changing the position of the function reenabled it in this mode. Fixed that

### DIFF
--- a/pkg/setup/config.go
+++ b/pkg/setup/config.go
@@ -77,6 +77,7 @@ func InitAndValidateConfig(ctx context.Context, vConfig *config.VirtualClusterCo
 	err = SetGlobalOwner(
 		ctx,
 		vConfig.ControlPlaneClient,
+		vConfig.Experimental.MultiNamespaceMode.Enabled,
 		vConfig.WorkloadNamespace,
 		vConfig.WorkloadTargetNamespace,
 		vConfig.Experimental.SyncSettings.SetOwner,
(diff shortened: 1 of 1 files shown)


pkg/setup/config.go
```diff
@@ -308,26 +309,30 @@ func updateSecretAnnotations(ctx context.Context, client kubernetes.Interface, n
...
 	}
 
+	service, err := currentNamespaceClient.CoreV1().Services(currentNamespace).Get(ctx, serviceName, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrap(err, "get vcluster service")
+	}
+	// client doesn't populate typemeta sometimes
+	service.TypeMeta.APIVersion = "v1"
+	service.TypeMeta.Kind = "Service"
+	translate.Owner = service
```
ThomasK33, line 335: Shouldn't this be wrapped in a `if setOwner { ... }`?

pkg/setup/config.go
```diff
@@ -308,26 +309,30 @@ func updateSecretAnnotations(ctx context.Context, client kubernetes.Interface, n
 
 // SetGlobalOwner fetches the owning service and populates in translate.Owner if: the vcluster is configured to setOwner is,
 // and if the currentNamespace == targetNamespace (because cross namespace owner refs don't work).
-func SetGlobalOwner(ctx context.Context, currentNamespaceClient kubernetes.Interface, currentNamespace, targetNamespace string, setOwner bool, serviceName string) error {
-	if currentNamespace != targetNamespace {
-		if setOwner {
-			klog.Warningf("Skip setting owner, because current namespace %s != target namespace %s", currentNamespace, targetNamespace)
-		}
+func SetGlobalOwner(ctx context.Context, currentNamespaceClient kubernetes.Interface, multins bool, currentNamespace, targetNamespace string, setOwner bool, serviceName string) error {
+	if !setOwner {
```
rohantmp, line 313: I exit the func here if no setOwner

pkg/setup/config.go
rohantmp: I exit the func earlier if !setWwner

PR #1812, opened 2024-05-30
fix: don't init selves before platoform start

rohantmp:
/kind bugfix
/kind enhancement
/kind feature
/kind documentation
/kind test

resolves #


pkg/platform/client.go
```diff
@@ -122,6 +122,13 @@ func NewClientFromConfig(ctx context.Context, config *config.CLI) (Client, error
 	return c, nil
 }
 
+func NewClientFromConfig(_ context.Context, config *config.CLI) (Client, error) {
+	c := &client{
+		config: config,
+	}
+	return c, nil
+}
+
```
johannesfrey, line 131:
```suggestion
func NewClientFromConfig(config *config.CLI) Client {
	return &client{
		config: config,
	}
}
```

Though, this requires the remaining callers to be adjusted as well :)

johannesfrey (changes requested):
Thx for spotting this!
Would you mind adding a comment above InitClientFromConfig so that users know when this call should be used vs. just NewClientFromConfig.

johannesfrey (dismissed):

rohantmp:
Making the changes :)

johannesfrey (approved):

pkg/platform/client.go
```diff
@@ -122,6 +124,15 @@ func NewClientFromConfig(ctx context.Context, config *config.CLI) (Client, error
 	return c, nil
 }
 
+// NewClientFromConfig returns a client without the client identity initialized through the selves api.
+// Use this only when performing actions that don't require a log in (like login itself).
+func NewClientFromConfig(config *config.CLI) (Client, error) {
```
ThomasK33, line 129: Why does the function signature contain an error, if the function never returns one?

pkg/platform/client.go
```diff
@@ -106,7 +106,9 @@ type VirtualClusterInstanceProject struct {
 	Project        *managementv1.Project
 }
 
-func NewClientFromConfig(ctx context.Context, config *config.CLI) (Client, error) {
+// InitClientFromConfig returns a client with the client identity initialized through the selves api.
+// Use this by default, unless performing actions that don't require a log in (like login itself).
+func InitClientFromConfig(ctx context.Context, config *config.CLI) (Client, error) {
 	c := &client{
 		config: config,
 	}
```
ThomasK33, line 114: Couldn't we call `NewClientFromConfig` here instead of manually initializing the struct?

ThomasK33 (changes requested):

pkg/platform/client.go
rohantmp: good catch :)

ThomasK33 (approved):

PR #1813, opened 2024-05-30
fix/refactor: remove need to pass cli.Config to subcommands

rohantmp:
Earlier, `vcluster platform add cluster` had an uninitialized config.

I think generating from GlobalFlags reduces the need to drill down another var

/kind bugfix

### DIFF
--- a/cmd/vclusterctl/cmd/platform/add/cluster.go
+++ b/cmd/vclusterctl/cmd/platform/add/cluster.go
@@ -16,7 +16,6 @@ import (
 
 	managementv1 "github.com/loft-sh/api/v4/pkg/apis/management/v1"
 	storagev1 "github.com/loft-sh/api/v4/pkg/apis/storage/v1"
-	"github.com/loft-sh/vcluster/pkg/cli/config"
 	"github.com/loft-sh/vcluster/pkg/cli/flags"
 	"github.com/loft-sh/vcluster/pkg/platform"
 	"github.com/loft-sh/vcluster/pkg/platform/clihelper"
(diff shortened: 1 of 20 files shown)


pkg/cli/flags/flags.go
```diff
@@ -15,10 +17,16 @@ type GlobalFlags struct {
 	Context   string
 	Namespace string
 	LogOutput string
+
+	readOnce  sync.Once
+	cliConfig *config.CLI
 }
 
 func (g *GlobalFlags) LoadedConfig(log log.Logger) *config.CLI {
```
rohantmp, line 25: made it a once-only operation here

rohantmp:
This will probably need a last minute rebase again even if I rebase now given the files it touches. Let me know when reviewing and I'll do it then :)

pkg/cli/flags/flags.go
johannesfrey: I don't think this is necessary, as we already do this in `config.Read`: https://github.com/loft-sh/vcluster/blob/main/pkg/cli/config/config.go#L53
So, LoadedConfig is actually merely an "alias" for config.Read. I guess this the whole config handling is a bit inconsistent right now. WDYT of cleaning this in a follow-up after all the in-flight porting PRs are merged?

rohantmp: Sure, can remove the once here.

> WDYT of cleaning this in a follow-up after all the in-flight porting PRs are merged? 🙂

Do you mean reducing the change in this PR to just ensuring the `platform add cluster` config is initialized?

johannesfrey: Sry, I expressed myself a bit misleading 😅. The change of this PR looks good. I would only drop the duplicate once handling

rohantmp: Done :)

johannesfrey (approved):

ThomasK33 (approved):

PR #1820, opened 2024-05-31
fix(vclusterctl): write switched manager to config

johannesfrey:
Previously this was not persistent. Furthermore, moved the valid manager check inside SwitchManager, as this func is also used internally.

/kind bugfix

resolves #


cmd/vclusterctl/cmd/use/manager.go
```diff
@@ -52,13 +48,19 @@ func (cmd *ManagerCmd) Run(ctx context.Context, args []string) error {
 }
 
 func SwitchManager(ctx context.Context, cfg *config.CLI, manager string, log log.Logger) error {
-	if cfg.Manager.Type == config.ManagerPlatform {
+	managerType := config.ManagerType(manager)
+	if managerType != config.ManagerHelm && managerType != config.ManagerPlatform {
+		return fmt.Errorf("invalid manager type: %q, only \"helm\" or \"platform\" are valid", managerType)
+	}
```
ThomasK33, line 54: Technically speaking (from a DDD angle), this should be a function returning the specialized type.

```go
func ParseManagerType(manager string) (config.ManagerType, error) {
  switch manager {
  case "helm": 
    return config.ManagerHelm, nil
  case "platform":
    return config.ManagerPlatform, nil
  default:
    return "", fmt.Errorf("invalid manager type: %q, only \"helm\" or \"platform\" are valid", managerType)
  }
}
```

and then one would call

```suggestion
	managerType, err := ParseManagerType(manager)
	if err != nil {
		return fmt.Errorf("parse manager type: %w", err)
	}
```

ThomasK33 (changes requested):

ThomasK33 (approved):

PR #1821, opened 2024-05-31
feat: Added Disconnect command for every vCluster-related context

ThomasK33:
/kind enhancement

resolves ENG-3727


cmd/vclusterctl/cmd/platform/connect/vcluster.go
```diff
@@ -1,121 +0,0 @@
-package connect
```
johannesfrey, line 1: Why is this removed? I thought `vcluster connect` is not the same as`vcluster platform connect vcluster`, because in `vcluster connect` you don't have those flags: https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/platform/connect/vcluster.go#L23-L27
Or should they rather be moved over to the `ConnectOptions`?

ThomasK33: Yes, those should be part of the `vcluster connect` command.

Having two commands to do the same thing only adds to confusion about when to use which command.
Given that the `vcluster create` commands directly creates an instance in the platform, the `vcluster connect` command should be the one to also connect to it, with all the flags necessary.

johannesfrey (approved):

PR #1838, opened 2024-06-04
Drop loftctl deps

johannesfrey:
/kind enhancement

resolves ENG-3854

Removed loftctl dependencies

Happy to squash the commits before merge.


pkg/util/portforward/portforward.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package portforward
```
johannesfrey, line 17: FYI: I overwrote this with the one from loftctl. I guess this is the more recent version, as it's using ctx etc?!

FabianKramm (approved):

PR #1840, opened 2024-06-04
fix(vclusterctl): also parse manager type from --manager flag

johannesfrey:
/kind bugfix

resolves ENG-3851

Fixed an issue where vcluster would not consider the --manager flag.

### DIFF
--- a/cmd/vclusterctl/cmd/connect.go
+++ b/cmd/vclusterctl/cmd/connect.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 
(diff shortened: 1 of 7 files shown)


zerbitx:
I was looking at this yesterday and thought similar flag > config situations might come up.  I thought about adding a function to `clihelper.go` that looks something like 

```go
// Coalesce return the first non-empty/non-zero from a set of same type values
func Coalesce[T comparable](values ...T) T {
	var zeroType T

	for _, v := range values {
		if v != zeroType {
			return v
		}
	}

	return zeroType
}

// usage 
managerType, err := config.ParseManagerType(Coalesce(cmd.Manager, string(cfg.Manager.Type)))
```

but naming things is hard and I don't think coalesce makes it readable enough.  If you can think of a better name and want to try that out.  If not, this all still looks good to me.

zerbitx (approved):

ThomasK33 (approved):

PR #1856, opened 2024-06-10
[backport] fix: disable global owner for isolated control plane mode

rohantmp:
/kind bugfix

resolves #

currently, owner is set in both control-plane and workload clusters, so this option should be disabled for isolated control plane.

backport of https://github.com/loft-sh/vcluster/pull/1855


lizardruss (approved):
LGTM

PR #1865, opened 2024-06-11
Omit OSS vCluster that were imported into pro vClusters.

zerbitx:

resolves #ENG-3822

Omits OSS clusters where the name/namespace matches pro clusters with the `loft.sh/imported-by-agent` annotation.


heiko-braun (approved):
lgtm

PR #1887, opened 2024-06-25
V0.20 improve binaries volume config

neogopher:
/kind enhancement

resolves ENG-3953

Fixed an issue where vcluster would request too little ephemeral storage.

### DIFF
--- a/chart/templates/statefulset.yaml
+++ b/chart/templates/statefulset.yaml
@@ -92,8 +92,7 @@ spec:
 {{- include "vcluster.plugins.volumes" . | indent 8 }}
         - name: helm-cache
           emptyDir: {}
-        - name: binaries
-          emptyDir: {}
+{{ toYaml .Values.controlPlane.statefulSet.persistence.binariesVolume | indent 8 }}
         - name: tmp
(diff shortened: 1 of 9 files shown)


chart/templates/statefulset.yaml
```diff
@@ -92,8 +92,7 @@ spec:
 {{- include "vcluster.plugins.volumes" . | indent 8 }}
         - name: helm-cache
           emptyDir: {}
-        - name: binaries
-          emptyDir: {}
+{{ toYaml .Values.controlPlane.statefulSet.persistence.binariesVolume | indent 8 }}
```
johannesfrey, line 95: I'm not entirely sure about this, especially as we do it like this for other values as well above. But what would happen if someone uses the vcluster.yaml from e.g. a previous beta release where `binariesVolume` does not exist in the vcluster.yaml?!

neogopher: It should use the ones defined in the current values.yaml file by default, if it wasn't specified previously. I will validate this too, just to be sure.

johannesfrey (approved):

PR #1888, opened 2024-06-25
feat: update IsProFeatureEnabled for external.platform features

lizardruss:
/kind enhancement

part of ENG-3882

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -223,6 +223,10 @@ func (c *Config) IsProFeatureEnabled() bool {
 		return true
 	}
 
+	if len(c.External["platform"]) > 0 {
+		return true
+	}
+
 	return false
 }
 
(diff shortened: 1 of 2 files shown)


johannesfrey (approved):

config/config.go
```diff
@@ -223,6 +223,10 @@ func (c *Config) IsProFeatureEnabled() bool {
 		return true
 	}
 
+	if len(c.External["platform"]) > 0 {
```
FabianKramm, line 226: What happens if c.External is nil?

FabianKramm (changes requested):
Check my comment above

config/config.go
johannesfrey: If a map is nil, it will be treated as an empty map for the lookup. And if the key does not exist Go returns the zero value which is nil for `platform` and `len(nil)` being 0.

facchettos: uh, I didn't know that, it's just writing that makes it panic

johannesfrey: Yeah, really had to google and even then check in the playground 😅 https://go.dev/play/p/Sb6k7ApVEOx

FabianKramm (approved):

PR #1902, opened 2024-07-03
No agent found with Sleep Mode configured: vCluster logs, CLI output

lizardruss:
/kind enhancement

resolves ENG-3890

### DIFF
--- a/pkg/cli/create_helm.go
+++ b/pkg/cli/create_helm.go
@@ -268,6 +268,14 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 		cmd.Connect = false
 	}
 
+	if isSleepModeConfigured(vClusterConfig) {
+		if agentDeployed, err := cmd.isLoftAgentDeployed(ctx); err != nil {
+			return fmt.Errorf("is agent deployed: %w", err)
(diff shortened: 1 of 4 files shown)


pkg/cli/create_helm.go
```diff
@@ -365,6 +373,24 @@ func (cmd *createHelm) addVCluster(ctx context.Context, vClusterConfig *config.C
...
 
+func (cmd *createHelm) isLoftAgentDeployed(ctx context.Context) (bool, error) {
+	podList, err := cmd.kubeClient.CoreV1().Pods("").List(ctx, metav1.ListOptions{
+		LabelSelector: "app=loft",
+	})
+	if err != nil && !kerrors.IsNotFound(err) {
+		return false, err
+	}
+
+	return len(podList.Items) > 0, nil
```
rmweir, line 384: Can we narrow this down by namespace?

rmweir (changes requested):

FabianKramm (approved):

pkg/cli/create_helm.go
FabianKramm: Since people can deploy loft in any namespace, I'm not sure if this would be a good idea

PR #1903, opened 2024-07-04
refactor: integrations & apiservice

FabianKramm:
/kind enhancement

Changes:
* Refactored import / export syncer
* Refactored metrics-server integration
* Renamed `observability.proxy.metrics` -> `integrations.metricsServer`


pkg/apiservice/generic.go
```diff
@@ -0,0 +1,231 @@
...
+
+	s := serializer.NewCodecFactory(scheme.Scheme)
+	server := &http.Server{
+		Addr: "localhost:" + strconv.Itoa(proxyPort),
+		Handler: http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+			// we only allow traffic to discovery paths
+			if !isAPIServiceProxyPathAllowed(request.Method, request.URL.Path) {
+				klog.FromContext(ctx).Info("Denied access to api service proxy at path", "path", request.URL.Path, "method", request.Method)
+				responsewriters.ErrorNegotiated(
+					kerrors.NewForbidden(metav1.SchemeGroupVersion.WithResource("proxy").GroupResource(), "proxy", fmt.Errorf("other paths than discovery paths are not allowed")),
```
zerbitx, line 158: nit:
```suggestion
					kerrors.NewForbidden(metav1.SchemeGroupVersion.WithResource("proxy").GroupResource(), "proxy", fmt.Errorf("paths other than discovery paths are not allowed")),
```

pkg/util/unstructuredhelper/write.go
```diff
@@ -0,0 +1,318 @@
...
+	lastKey := pathSplitted[len(pathSplitted)-1]
+	if !targetMap.Has(lastKey) {
+		return
+	}
+
+	// to map
+	toMap := to
+	for i := 0; i < len(pathSplitted)-1; i++ {
+		toMap = toMap.Map(pathSplitted[i])
+	}
```
zerbitx, line 89: nit: traverses pathSplitted only once.

```suggestion
	targetMap := from
	// to map
	toMap := to
	for i := range pathSplitted {
		targetMap = targetMap.Map(pathSplitted[i])
		toMap = toMap.Map(pathSplitted[i])
	}

	// check if map has latest path
	lastKey := pathSplitted[len(pathSplitted)-1]
	if !targetMap.Has(lastKey) {
		return
	}
```

zerbitx (approved):

pkg/util/unstructuredhelper/write.go
```diff
@@ -0,0 +1,318 @@
...
+
+	// check if map has latest path
+	lastKey := pathSplitted[len(pathSplitted)-1]
+
+	// to map
+	toMap := to
+	for i := 0; i < len(pathSplitted)-1; i++ {
+		toMap = toMap.Map(pathSplitted[i])
+	}
```
zerbitx, line 55: nit: avoids second traversal

```suggestion
	targetMap := from
	// to map
	toMap := to
	for i := 0; i < len(pathSplitted)-1; i++ {
		targetMap = targetMap.Map(pathSplitted[i])
		toMap = toMap.Map(pathSplitted[i])
	}

	// check if map has latest path
	lastKey := pathSplitted[len(pathSplitted)-1]
```

chart/values.yaml
```diff
@@ -116,6 +116,17 @@ sync:
         all: false
         labels: {}
 
+# Integrations holds config for vCluster integrations with other operators or tools
```
johannesfrey, line 119:
```suggestion
# Integrations holds config for vCluster integrations with other operators or tools running on the host cluster
```
Not sure if this suggestion narrows down the use case of `integrations` too much, but I could see that users might be confused how this is different to `external` then.

FabianKramm: Makes sense, external is basically configurations for tooling outside of vCluster and integrations are part of vCluster itself

PR #1904, opened 2024-07-04
Add <NAME> arg to help output

heiko-braun:
/kind enhancement

resolves #
ENG-4070

### DIFF
--- a/cmd/vclusterctl/cmd/platform/add/cluster.go
+++ b/cmd/vclusterctl/cmd/platform/add/cluster.go
@@ -50,7 +50,7 @@ func NewClusterCmd(globalFlags *flags.GlobalFlags) *cobra.Command {
 	}
 
 	c := &cobra.Command{
-		Use:   "cluster",
+		Use:   "cluster CLUSTER_NAME",
 		Short: "add current cluster to vCluster platform",
 		Long: `#######################################################
 ############ vcluster platform add cluster ############
--- a/cmd/vclusterctl/cmd/platform/add/vcluster.go
+++ b/cmd/vclusterctl/cmd/platform/add/vcluster.go
@@ -33,7 +33,7 @@ vcluster platform add vcluster my-vcluster --namespace vcluster-my-vcluster --pr
 	`
 
 	addCmd := &cobra.Command{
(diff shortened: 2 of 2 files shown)


cmd/vclusterctl/cmd/platform/add/vcluster.go
```diff
@@ -33,7 +33,7 @@ vcluster platform add vcluster my-vcluster --namespace vcluster-my-vcluster --pr
 	`
 
 	addCmd := &cobra.Command{
-		Use:   "vcluster",
+		Use:   "vcluster VCLUSTER_NAME",
```
johannesfrey, line 36: We could also use this util func that we already use in some places (probably not consistent over the whole code base): https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/create.go#L35

cmd/vclusterctl/cmd/platform/add/vcluster.go
```diff
@@ -33,7 +33,7 @@ vcluster platform add vcluster my-vcluster --namespace vcluster-my-vcluster --pr
...
 		Short: "Adds an existing vCluster to the vCluster platform",
 		Long:  description,
 		Args:  cobra.ExactArgs(1),
```
johannesfrey, line 39: Also this here: https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/create.go#L46

cmd/vclusterctl/cmd/platform/add/cluster.go
```diff
@@ -50,7 +50,7 @@ func NewClusterCmd(globalFlags *flags.GlobalFlags) *cobra.Command {
 	}
 
 	c := &cobra.Command{
-		Use:   "cluster",
+		Use:   "cluster CLUSTER_NAME",
```
johannesfrey, line 53: If we go the util route we might want to add a `Cluster` entry there as well!?

FabianKramm (changes requested):
We need to be careful here as `Use` is also used to build the command itself

heiko-braun:
Can you elaborate @FabianKramm ?

PR #1913, opened 2024-07-08
Default to background-proxy to false for connect

zerbitx:
/kind bugfix

resolves #ENG-4054


FabianKramm (changes requested):
I think this should be addressed already, if this is still an issue we should only fix this on the platform side as this should stay the default for the OSS vclusters

PR #1925, opened 2024-07-12
refactor resource host to virtual mappings

FabianKramm:
/kind enhancement

Refactors the way vCluster maps virtual name to host names. This allows us to implement more flexible ways of rewritting names:
* Introduces a mapper per GVK to map virtual name to host name and back
* Allows globally accessing these mappers so that syncer can rewrite names used by other mappers this way
* Refactors all controllers to use the mappers instead of the internal mapping functions
* This will require changes to the vCluster SDK & Pro


pkg/controllers/resources/persistentvolumes/translate.go
```diff
@@ -99,15 +99,15 @@ func (s *persistentVolumeSyncer) translateUpdateBackwards(vPv *corev1.Persistent
...
 		updated = translator.NewIfNil(updated, vPv)
 		updated.Spec.Capacity = translatedSpec.Capacity
 	}
 
-	return updated
+	return updated, nil
 }
 
-func (s *persistentVolumeSyncer) translateUpdate(ctx context.Context, vPv *corev1.PersistentVolume, pPv *corev1.PersistentVolume) *corev1.PersistentVolume {
+func (s *persistentVolumeSyncer) translateUpdate(ctx context.Context, vPv *corev1.PersistentVolume, pPv *corev1.PersistentVolume) (*corev1.PersistentVolume, error) {
```
ThomasK33, line 110: Why is this func returning an error now, when it never returns an error value?

pkg/mappings/types.go
```diff
@@ -0,0 +1,140 @@
...
+	return Default.ByGVK(corev1.SchemeGroupVersion.WithKind("PersistentVolumeClaim"))
+}
+
+func PriorityClasses() Mapper {
+	return Default.ByGVK(schedulingv1.SchemeGroupVersion.WithKind("PriorityClass"))
+}
+
+func VirtualToHostName(vName, vNamespace string, mapper Mapper) string {
+	return mapper.VirtualToHost(context.TODO(), types.NamespacedName{Name: vName, Namespace: vNamespace}, nil).Name
+}
```
ThomasK33, line 136: This should probably take a context as first argument.

pkg/mappings/types.go
```diff
@@ -0,0 +1,140 @@
...
+
+func VirtualToHost(vName, vNamespace string, mapper Mapper) types.NamespacedName {
+	return mapper.VirtualToHost(context.TODO(), types.NamespacedName{Name: vName, Namespace: vNamespace}, nil)
+}
```
ThomasK33, line 140: Same here regarding the context.

pkg/setup/managers.go
```diff
@@ -0,0 +1,56 @@
...
+	// start the virtual cluster manager
+	go func() {
+		err := ctx.VirtualManager.Start(ctx.Context)
+		if err != nil {
+			panic(err)
+		}
+	}()
+
+	// Wait for caches to be synced
+	klog.Infof("Starting local & virtual managers...")
```
ThomasK33, line 50:
```suggestion
	klog.FromContext(ctx.Context).Info("Starting local & virtual managers...")
```

pkg/setup/managers.go
```diff
@@ -0,0 +1,56 @@
...
+	ctx.PhysicalManager.GetCache().WaitForCacheSync(ctx.Context)
+	ctx.VirtualManager.GetCache().WaitForCacheSync(ctx.Context)
+	klog.Infof("Successfully started local & virtual manager")
```
ThomasK33, line 53:
```suggestion
	klog.FromContext(ctx.Context).Info("Successfully started local & virtual manager")
```

pkg/controllers/resources/events/syncer.go
```diff
@@ -97,7 +62,7 @@ func (s *eventSyncer) Sync(ctx *synccontext.SyncContext, pObj client.Object, vOb
 	vOldEvent := vEvent.DeepCopy()
 	vEvent, err := s.buildVirtualEvent(ctx.Context, pEvent)
 	if err != nil {
-		return ctrl.Result{}, IgnoreAcceptableErrors(err)
+		return ctrl.Result{}, resources.IgnoreAcceptableErrors(err)
```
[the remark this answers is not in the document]
FabianKramm: Why not? Its the same function as before just in a different package now to avoid import cycle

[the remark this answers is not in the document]
ThomasK33: The functions are referencing different error instances, and the above function does not return those errors.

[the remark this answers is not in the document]
FabianKramm: It does, it calls `resources.GetInvolvedObject(ctx, s.virtualClient, pEvent)` as first thing that returns those errors

ThomasK33 (approved):

PR #1927, opened 2024-07-12
fix: detect how the vcluster was put to sleep and guide the user to t…

lizardruss:
/kind enhancement

Part of ENG-4106

Fixed an issue where vcluster pause / resume can interfere with the platform sleep mode feature.

Mostly adds messaging that will guide the user to pause / resume using the same driver.


lizardruss:
Different approach will be used.

PR #1930, opened 2024-07-15
refactor: improve patcher & crd sync

FabianKramm:
/kind enhancement

* Prints what exactly has been patched in debug mode
* Do not run initializers in parallel as this is unexpected and can lead to race conditions in integration / plugins

### DIFF
--- a/pkg/controllers/register.go
+++ b/pkg/controllers/register.go
@@ -39,6 +39,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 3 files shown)


pkg/controllers/register.go
```diff
@@ -126,23 +126,19 @@ func ExecuteInitializers(controllerCtx *config.ControllerContext, syncers []sync
 	registerContext := util.ToRegisterContext(controllerCtx)
 
 	// execute in parallel because each one might be time-consuming
```
ThomasK33, line 128: This comment should probably be updated too

FabianKramm: Yep true

ThomasK33 (approved):

PR #1932, opened 2024-07-15
now uses the patcher logic for ingresses

facchettos:
/kind enhancement

part of eng-4141


pkg/controllers/resources/ingresses/syncer.go
```diff
@@ -29,43 +33,39 @@ func (s *ingressSyncer) SyncToHost(ctx *synccontext.SyncContext, vObj client.Obj
...
-		ctx.Log.Infof("update virtual ingress %s/%s, because status is out of sync", vIngress.Namespace, vIngress.Name)
-		translator.PrintChanges(vIngress, newIngress, ctx.Log)
-		err := ctx.VirtualClient.Status().Update(ctx.Context, newIngress)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-
-		// we will requeue anyways
-		return ctrl.Result{}, nil
+		vObj = newIngress
```
FabianKramm, line 56: Can we use `vIngress.Status = pIngress.Status` here?

pkg/controllers/resources/ingresses/syncer.go
```diff
@@ -29,43 +33,39 @@ func (s *ingressSyncer) SyncToHost(ctx *synccontext.SyncContext, vObj client.Obj
...
 	}
+	annotations := maps.Clone(vObj.GetAnnotations())
 
 	newIngress := s.translateUpdate(ctx.Context, pIngress, vIngress)
```
FabianKramm, line 60: Why do we need to return newIngress here? Can we not directly change vIngress / pIngress?

pkg/controllers/resources/ingresses/syncer.go
```diff
@@ -29,43 +33,39 @@ func (s *ingressSyncer) SyncToHost(ctx *synccontext.SyncContext, vObj client.Obj
...
+		if err := patch.Patch(ctx, pObj, vObj); err != nil {
+			retErr = utilerrors.NewAggregate([]error{retErr, err})
 		}
+	}()
 
-		// we will requeue anyways
-		return ctrl.Result{}, nil
+	pIngress, vIngress, _, _ := synccontext.Cast[*networkingv1.Ingress](ctx, pObj, vObj)
+	if vIngress.Spec.IngressClassName == nil && pIngress.Spec.IngressClassName != nil {
+		vIngress.Spec.IngressClassName = pIngress.Spec.IngressClassName
```
FabianKramm, line 50: This should be actually bidirectional

pkg/controllers/resources/ingresses/syncer.go
```diff
@@ -29,43 +33,39 @@ func (s *ingressSyncer) SyncToHost(ctx *synccontext.SyncContext, vObj client.Obj
...
 	}
 
 	if !equality.Semantic.DeepEqual(vIngress.Status, pIngress.Status) {
```
FabianKramm, line 53: We actually don't have to compare here and could just assign it

pkg/controllers/resources/ingresses/util/util.go
```diff
@@ -80,12 +83,13 @@ func ProcessAlbAnnotations(namespace string, k string, v string) (string, string
 }
 
 func UpdateAnnotations(vObj client.Object) client.Object {
-	annotations := vObj.GetAnnotations()
-	for k, v := range annotations {
-		delete(annotations, k)
+	// we need to do that otherwise the original object will be modified
+	ingress := *vObj.(*networkingv1.Ingress)
```
FabianKramm, line 87: I don't think this a good practice, if you want to clone the object, do `DeepCopy()`, also why is modifying the object now an issue?

pkg/controllers/resources/ingresses/translate.go
```diff
@@ -26,35 +25,18 @@ func (s *ingressSyncer) TranslateMetadataUpdate(ctx context.Context, vObj client
 	return s.NamespacedTranslator.TranslateMetadataUpdate(ctx, util.UpdateAnnotations(vObj), pObj)
 }
 
-func (s *ingressSyncer) translateUpdate(ctx context.Context, pObj, vObj *networkingv1.Ingress) *networkingv1.Ingress {
-	var updated *networkingv1.Ingress
-
+func (s *ingressSyncer) translateUpdate(ctx context.Context, pObj, vObj *networkingv1.Ingress) {
 	translatedSpec := *translateSpec(vObj.Namespace, &vObj.Spec)
```
FabianKramm, line 29: Just do `pObj.Spec = *translateSpec(vObj.Namespace, &vObj.Spec)`

FabianKramm (changes requested):

pkg/controllers/resources/ingresses/util/util.go
facchettos: because the underlying map gets modified, and we didn't care about it before because there was just the host object being modified at that point. With the new defer stuff, if the map is modified the virtual annotations also get modified

FabianKramm (approved):

PR #1935, opened 2024-07-16
feat(helm): Automatically create role & bindings for namespace containing platform secret

ThomasK33:
/kind enhancement

resolves ENG-4148

### DIFF
--- a/chart/templates/platform-rbac.yaml
+++ b/chart/templates/platform-rbac.yaml
@@ -1,0 +1,31 @@
+{{- if include "vcluster.rbac.createPlatformSecretRole" . }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "vcluster.rbac.platformRoleName" . }}
+  namespace: {{ include "vcluster.rbac.platformSecretNamespace" .}}
+rules:
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get"]
+  resourceNames:
+    - {{ include "vcluster.rbac.platformSecretName" . }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ include "vcluster.rbac.platformRoleBindingName" . }}
+  namespace: {{ include "vcluster.rbac.platformSecretNamespace" .}}
+subjects:
+  - kind: ServiceAccount
+    {{- if .Values.controlPlane.advanced.serviceAccount.name }}
+    name: {{ .Values.controlPlane.advanced.serviceAccount.name }}
+    {{- else }}
+    name: vc-{{ .Release.Name }}
+    {{- end }}
+    namespace: {{ include "vcluster.rbac.platformSecretNamespace" .}}
+roleRef:
+  kind: Role
+  name: {{ include "vcluster.rbac.platformRoleName" . }}
+  apiGroup: rbac.authorization.k8s.io
+{{- end }}
--- a/Justfile
+++ b/Justfile
@@ -74,6 +74,7 @@ generate-vcluster-images version="0.0.0":
 generate-cli-docs:
   go run -mod vendor -tags pro ./hack/docs/main.go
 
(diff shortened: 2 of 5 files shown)


zerbitx (approved):
LGTM

chart/templates/platform-rbac.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if include "vcluster.rbac.createPlatformSecretRole" . }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: vcluster-platform-secret-role
```
FabianKramm, line 5: This needs to be dynamic based on the vcluster name as this would otherwise collide if there are multiple vclusters

FabianKramm (changes requested):

chart/templates/platform-rbac.yaml
FabianKramm: Same goes for the role binding

chart/templates/platform-rbac.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if include "vcluster.rbac.createPlatformSecretRole" . }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: vc-{{ .Release.Name }}-platform-secret-role
```
FabianKramm, line 5: This also needs to include release namespace and a way to make sure this is not too long

chart/templates/platform-rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get", "watch", "patch"]
+  resourceNames:
+    - {{ include "vcluster.rbac.platformSecretName" . }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: vc-{{ .Release.Name }}-platform-secret-role-binding
```
FabianKramm, line 17: This also needs to include release namespace and a way to make sure this is not too long

FabianKramm (changes requested):

chart/templates/platform-rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - kind: ServiceAccount
+    {{- if .Values.controlPlane.advanced.serviceAccount.name }}
+    name: {{ .Values.controlPlane.advanced.serviceAccount.name }}
+    {{- else }}
+    name: vc-{{ .Release.Name }}
+    {{- end }}
+    namespace: {{ include "vcluster.rbac.platformSecretNamespace" .}}
+roleRef:
+  kind: Role
+  name: vc-{{ .Release.Name }}-platform-secret-role
```
FabianKramm, line 29: This needs to be the role name

chart/templates/platform-rbac.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if include "vcluster.rbac.createPlatformSecretRole" . }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "vcluster.rbac.platformRoleName" . }}
+  namespace: {{ include "vcluster.rbac.platformSecretNamespace" .}}
+rules:
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get", "watch", "patch"]
```
FabianKramm, line 10: Why patch here?

FabianKramm (changes requested):

chart/templates/platform-rbac.yaml
ThomasK33: Because of this Patch here: https://github.com/loft-sh/vcluster/blob/4a32dc7bc4619ca17239d596b17882cca5e048c9/pkg/platform/secret.go#L89

FabianKramm: But thats part of the cli package, so we don't need that "server side"

FabianKramm (approved):

PR #1940, opened 2024-07-16
Events patcher

facchettos:
/kind enhancement

part of eng-4141

### DIFF
--- a/pkg/controllers/resources/events/syncer.go
+++ b/pkg/controllers/resources/events/syncer.go
@@ -2,18 +2,19 @@ package events
 
 import (
 	"context"
-	"strings"
+	"errors"
+	"fmt"
 
 	synccontext "github.com/loft-sh/vcluster/pkg/controllers/syncer/context"
-	"github.com/loft-sh/vcluster/pkg/controllers/syncer/translator"
 	"github.com/loft-sh/vcluster/pkg/mappings"
 	"github.com/loft-sh/vcluster/pkg/mappings/resources"
+	"github.com/loft-sh/vcluster/pkg/patcher"
 	syncer "github.com/loft-sh/vcluster/pkg/types"
 	"github.com/loft-sh/vcluster/pkg/util/translate"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/controllers/resources/events/syncer.go
```diff
@@ -34,12 +35,15 @@ var AcceptedKinds = map[schema.GroupVersionKind]bool{
 
 func New(ctx *synccontext.RegisterContext) (syncer.Object, error) {
 	return &eventSyncer{
-		virtualClient: ctx.VirtualManager.GetClient(),
-		hostClient:    ctx.PhysicalManager.GetClient(),
+		NamespacedTranslator: translator.NewNamespacedTranslator(ctx, "event", &corev1.Event{}),
```
FabianKramm, line 38: Is that needed?

pkg/controllers/resources/events/syncer.go
```diff
@@ -124,7 +111,9 @@ var _ syncer.ToVirtualSyncer = &eventSyncer{}
 
 func (s *eventSyncer) SyncToVirtual(ctx *synccontext.SyncContext, pObj client.Object) (ctrl.Result, error) {
 	// build the virtual event
-	vObj, err := s.buildVirtualEvent(ctx.Context, pObj.(*corev1.Event))
+	vObj := pObj.DeepCopyObject().(*corev1.Event)
+	vObj.ResourceVersion = ""
```
FabianKramm, line 115: This should be `translate.ResetObjectMetadata(vObj)`

facchettos: so if we don't call it manually beforehand, the resource version will still be set because vObj contains the resource version of the physical object at that point. 
https://github.com/loft-sh/vcluster/pull/1940/files#diff-841ade67d794007be038e9be601ecb81f2e9b7b2e5b10ec260d13345b54294c1R43 this is where we would be overwriting it. To avoid doing that we should have a parameter to know whether we are creating the object or not. I believe it's slightly worse than the current solution

FabianKramm (approved):

PR #1946, opened 2024-07-18
bugfix(chart): enable fallthrough for cluster.local queries only when .Values.networking.advanced.fallbackHostCluster is true

neogopher:
/kind bugfix

resolves #ENG-4151

Fixed an issue where vcluster would fallback to hostDNS when a resolveDNS rule was specified


FabianKramm (approved):

FabianKramm:
@neogopher good catch, thanks a lot!

PR #1948, opened 2024-07-18
refactor: delete object if it was deleted on host

FabianKramm:
/kind enhancement

Changes within this PR:
* Adds detection for delete events within the syncers
* Deletes virtual objects if the host object was deleted
* Combines NamespacedTranslator and ClusterTranslator into GenericTranslator


pkg/util/translate/single_namespace.go
```diff
@@ -104,8 +112,6 @@ func (s *singleNamespace) IsManagedCluster(obj runtime.Object) bool {
 	metaAccessor, err := meta.Accessor(obj)
 	if err != nil {
 		return false
```
ThomasK33, line 114: Don't we need the nil check here?

pkg/util/translate/single_namespace.go
```diff
@@ -104,8 +112,6 @@ func (s *singleNamespace) IsManagedCluster(obj runtime.Object) bool {
...
-	} else if metaAccessor.GetLabels() == nil {
-		return false
 	}
```
ThomasK33, line 115: Don't we need a label nil check here?

FabianKramm: No, I didn't knew that as well, but this works:
```
var m map[string]bool = nil

fmt.Println(m["my-key"])
```

ThomasK33 (approved):

PR #1950, opened 2024-07-18
BACKPORT v0.19 bugfix(chart): enable fallthrough for cluster.local queries only when fallbackHostDns is true

neogopher:
/kind bugfix

resolves #ENG-4127

Fixed an issue where vcluster would fallback to hostDNS when coredns plugin is enabled

### DIFF
--- a/charts/k3s/templates/_coredns.tpl
+++ b/charts/k3s/templates/_coredns.tpl
@@ -18,7 +18,7 @@ Corefile: |-
           {{- end }}
           {{- end }}
           pods insecure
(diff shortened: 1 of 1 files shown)


FabianKramm (commented):
Are we sure this is necessary for v0.19? Since the vcluster plugin here comes after the check, I think this is still correct

neogopher:
I have moved the vcluster plugin up the chain as requested.

FabianKramm (approved):

PR #1959, opened 2024-07-19
refactor: use patcher for pod controller

FabianKramm:
/kind enhancement

Part of ENG-4168

Changes:
* Pod Controller
* Persistent Volume Claim Controller
* Volume Snapshots Controller

### DIFF
--- a/pkg/controllers/resources/persistentvolumeclaims/syncer_test.go
+++ b/pkg/controllers/resources/persistentvolumeclaims/syncer_test.go
@@ -95,15 +95,13 @@ func TestSync(t *testing.T) {
 			Name:      pObjectMeta.Name,
 			Namespace: pObjectMeta.Namespace,
 			Annotations: map[string]string{
-				translate.NameAnnotation:               vObjectMeta.Name,
-				translate.NamespaceAnnotation:          vObjectMeta.Namespace,
-				translate.UIDAnnotation:                "",
-				translate.KindAnnotation:               corev1.SchemeGroupVersion.WithKind("PersistentVolumeClaim").String(),
-				translate.ManagedAnnotationsAnnotation: "otherAnnotationKey",
-				bindCompletedAnnotation:                "testannotation",
-				boundByControllerAnnotation:            "testannotation2",
-				storageProvisionerAnnotation:           "testannotation3",
-				"otherAnnotationKey":                   "don't update this",
+				translate.NameAnnotation:      vObjectMeta.Name,
+				translate.NamespaceAnnotation: vObjectMeta.Namespace,
+				translate.UIDAnnotation:       "",
+				translate.KindAnnotation:      corev1.SchemeGroupVersion.WithKind("PersistentVolumeClaim").String(),
+				bindCompletedAnnotation:       "testannotation",
+				boundByControllerAnnotation:   "testannotation2",
+				storageProvisionerAnnotation:  "testannotation3",
 			},
 			Labels: pObjectMeta.Labels,
 		},
@@ -150,7 +148,7 @@ func TestSync(t *testing.T) {
 			},
 			Sync: func(ctx *synccontext.RegisterContext) {
 				syncCtx, syncer := generictesting.FakeStartSyncer(t, ctx, New)
-				_, err := syncer.(*persistentVolumeClaimSyncer).SyncToHost(syncCtx, basePvc)
+				_, err := syncer.(*persistentVolumeClaimSyncer).SyncToHost(syncCtx, basePvc.DeepCopy())
 				assert.NilError(t, err)
 			},
 		},
@@ -166,7 +164,7 @@ func TestSync(t *testing.T) {
 			},
 			Sync: func(ctx *synccontext.RegisterContext) {
 				syncCtx, syncer := generictesting.FakeStartSyncer(t, ctx, New)
-				_, err := syncer.(*persistentVolumeClaimSyncer).SyncToHost(syncCtx, deletePvc)
+				_, err := syncer.(*persistentVolumeClaimSyncer).SyncToHost(syncCtx, deletePvc.DeepCopy())
 				assert.NilError(t, err)
 			},
 		},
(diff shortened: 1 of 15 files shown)


pkg/controllers/resources/persistentvolumeclaims/syncer_test.go
```diff
@@ -95,15 +95,13 @@ func TestSync(t *testing.T) {
 			Name:      pObjectMeta.Name,
 			Namespace: pObjectMeta.Namespace,
 			Annotations: map[string]string{
-				translate.NameAnnotation:               vObjectMeta.Name,
-				translate.NamespaceAnnotation:          vObjectMeta.Namespace,
-				translate.UIDAnnotation:                "",
-				translate.KindAnnotation:               corev1.SchemeGroupVersion.WithKind("PersistentVolumeClaim").String(),
-				translate.ManagedAnnotationsAnnotation: "otherAnnotationKey",
```
facchettos, line 102: why is this removed?

pkg/controllers/resources/persistentvolumeclaims/syncer_test.go
```diff
@@ -95,15 +95,13 @@ func TestSync(t *testing.T) {
...
-				bindCompletedAnnotation:                "testannotation",
-				boundByControllerAnnotation:            "testannotation2",
-				storageProvisionerAnnotation:           "testannotation3",
-				"otherAnnotationKey":                   "don't update this",
```
facchettos, line 106: and this too

pkg/controllers/resources/persistentvolumeclaims/translate.go
```diff
@@ -99,61 +98,38 @@ func (s *persistentVolumeClaimSyncer) translateSelector(ctx *synccontext.SyncCon
...
 		}
-		updated.Spec.Resources.Requests["storage"] = vObj.Spec.Resources.Requests["storage"]
-	}
-
-	changed, updatedAnnotations, updatedLabels := s.TranslateMetadataUpdate(ctx, vObj, pObj)
-	if changed {
-		updated = translator.NewIfNil(updated, pObj)
-		updated.Annotations = updatedAnnotations
-		updated.Labels = updatedLabels
+		pObj.Spec.Resources.Requests["storage"] = vObj.Spec.Resources.Requests["storage"]
```
facchettos, line 107: couldn't this be 
```go
pObj.Spec.Resources.Requests = vObj.Spec.Resources.Requests
```
?

pkg/controllers/resources/persistentvolumeclaims/translate.go
```diff
@@ -99,61 +98,38 @@ func (s *persistentVolumeClaimSyncer) translateSelector(ctx *synccontext.SyncCon
...
-		if updated.Annotations[bindCompletedAnnotation] != pObj.Annotations[bindCompletedAnnotation] {
-			updated.Annotations[bindCompletedAnnotation] = pObj.Annotations[bindCompletedAnnotation]
-		}
-		if updated.Annotations[boundByControllerAnnotation] != pObj.Annotations[boundByControllerAnnotation] {
-			updated.Annotations[boundByControllerAnnotation] = pObj.Annotations[boundByControllerAnnotation]
-		}
-		if updated.Annotations[storageProvisionerAnnotation] != pObj.Annotations[storageProvisionerAnnotation] {
-			updated.Annotations[storageProvisionerAnnotation] = pObj.Annotations[storageProvisionerAnnotation]
+func (s *persistentVolumeClaimSyncer) translateUpdateBackwards(pObj, vObj *corev1.PersistentVolumeClaim) {
+	if vObj.Annotations[bindCompletedAnnotation] != pObj.Annotations[bindCompletedAnnotation] {
```
facchettos, line 117: don't think this is needed

pkg/controllers/resources/persistentvolumeclaims/translate.go
```diff
@@ -99,61 +98,38 @@ func (s *persistentVolumeClaimSyncer) translateSelector(ctx *synccontext.SyncCon
...
+		vObj.Annotations[bindCompletedAnnotation] = pObj.Annotations[bindCompletedAnnotation]
 	}
-
-	return updated
-}
-
-func translateUpdateNeeded(pAnnotations, vAnnotations map[string]string) bool {
-	if pAnnotations == nil {
-		pAnnotations = map[string]string{}
+	if vObj.Annotations[boundByControllerAnnotation] != pObj.Annotations[boundByControllerAnnotation] {
```
facchettos, line 123: all these checks aren't needed I guess

pkg/controllers/resources/persistentvolumeclaims/syncer_test.go
FabianKramm: The "otherAnnotationKey" and "translate.ManagedAnnotationsAnnotation" are removed because the object that was passed in doesn't have these annotations, so the syncer is correctly removing these. If we want to keep them we need to create a new object that has these as well, but I don't think this needs to be part of this testing logic. It worked before because the syncer was not updating metadata in this step and would require a second Sync call to remove that metadata

pkg/controllers/resources/persistentvolumeclaims/translate.go
FabianKramm: Oh yes true, good catch

pkg/controllers/resources/persistentvolumeclaims/syncer_test.go
facchettos: makes sense, no need to keep it then I agree

pkg/controllers/resources/persistentvolumeclaims/translate.go
FabianKramm: We need that or otherwise these annotations would be always set, in the beginning both pObj and vObj don't have them, but this would result in an empty annotation on the virtual side

facchettos (approved):

PR #1965, opened 2024-07-19
fix: order of volumesnapshot syncers

FabianKramm:
/kind bugfix

Moved CRD creation into the mapper as the index for the mappers are called before the Init function for controllers was called. To avoid confusion the Initializer function was removed as well.


pkg/mappings/resources/volumesnapshotcontents.go
```diff
@@ -28,3 +34,313 @@ func CreateVolumeSnapshotContentsMapper(ctx *synccontext.RegisterContext) (mappi
 		return vVSC.Annotations[constants.HostClusterVSCAnnotation]
 	})
 }
+
+var volumeSnapshotContentsCRD = `
```
facchettos, line 38: wouldn't that be a little nicer with embed?

pkg/patcher/patcher.go
```diff
@@ -46,6 +46,8 @@ func NewSyncerPatcher(ctx *synccontext.SyncContext, pObj, vObj client.Object, op
 		return nil, fmt.Errorf("create virtual patcher: %w", err)
 	}
 
+	vPatcher.direction = "virtual"
+	pPatcher.direction = "host"
```
facchettos, line 50: is this going to be required for most patchers that need logging? I'd put it in the constructor if that's the case

FabianKramm: Sure let me put it in the constructor

pkg/util/crds.go
```diff
@@ -17,6 +17,11 @@ import (
 )
 
 func EnsureCRD(ctx context.Context, config *rest.Config, manifest []byte, groupVersionKind schema.GroupVersionKind) error {
+	// TODO: find a better way to deal with this
+	if config.Host == "unit-test-client" {
+		return nil
+	}
+
```
facchettos, line 24: could we have a `func` variable inside the util package that we can replace at runtime with the fake manager? 

something like 
```go
var ensureCrd func(ctx context.Context, config *rest.Config, manifest []byte, groupVersionKind schema.GroupVersionKind) error {
      create/update the crd
}

func UnitTestSetup(){
   ensureCRD = func(...)error{
    return nil
   }
}
```

FabianKramm: Oh yeah good idea, let me do that

pkg/controllers/syncer/testing/manager.go
```diff
@@ -30,7 +30,7 @@ type fakeManager struct {
 
 func (f *fakeManager) SetFields(interface{}) error { return nil }
 
-func (f *fakeManager) GetConfig() *rest.Config { return &rest.Config{Host: "127.0.0.1"} }
+func (f *fakeManager) GetConfig() *rest.Config { return &rest.Config{Host: "unit-test-client"} }
```
facchettos, line 33: is it still used?

facchettos (approved):

pkg/controllers/syncer/testing/manager.go
FabianKramm: I think this actually is still a good idea as this shows up in tests a lot nicer than 127.0.0.1

PR #1967, opened 2024-07-21
refactor: move mappings into register context

FabianKramm:
/kind enhancement

Moves the mappings store into the controller / register / syncer context instead of exposing it globally which allows for more control and separate mapping stores


pkg/controllers/resources/persistentvolumes/fake_syncer_test.go
```diff
@@ -3,13 +3,13 @@ package persistentvolumes
 import (
 	"testing"
 
-	synccontext "github.com/loft-sh/vcluster/pkg/controllers/syncer/context"
+	"github.com/loft-sh/vcluster/pkg/syncer/synccontext"
+	testing2 "github.com/loft-sh/vcluster/pkg/syncer/testing"
```
facchettos, line 7: can't we give this a better name? something like syncertesting maybe?

FabianKramm: Yep will do

facchettos (approved):

PR #1970, opened 2024-07-22
refactor: labels & annotations translation

FabianKramm:
/kind enhancement

Changes:
* Refactored util/translate
* Renamed Physical to Host in util/translate
* Untangled metadata update
* Added VirtualAnnotations function


pkg/util/translate/translate.go
```diff
@@ -182,7 +265,28 @@ type ApplyMapsOptions struct {
...
+	for key, value := range toMap {
+		if exists(excludeKeys, key) {
+			if value != "" {
+				retMap[key] = value
+			}
+		}
+	}
+
+	return retMap
+}
```
facchettos, line 287:
```suggestion
func copyMaps(fromMap, toMap map[string]string, excludeKeys []string) map[string]string {
	retMap := maps.Clone(fromMap)
	for _, v := excludedKeys {
	  delete(retMap, v)
	  retMap[v] = toMap[v]
	}

	return retMap
}
```

PR #1971, opened 2024-07-22
Test cleanup

facchettos:
/kind enhancement

part of eng-4178

### DIFF
--- a/pkg/controllers/resources/nodes/syncer_test.go
+++ b/pkg/controllers/resources/nodes/syncer_test.go
@@ -3,13 +3,11 @@ package nodes
 import (
 	"testing"
 
-	"github.com/loft-sh/vcluster/pkg/syncer"
 	"github.com/loft-sh/vcluster/pkg/syncer/synccontext"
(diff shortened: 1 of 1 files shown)


pkg/controllers/resources/nodes/syncer_test.go
```diff
@@ -167,10 +176,104 @@ func TestSyncBothExist(t *testing.T) {
...
+			virtualNodeExists:  true,
+			withVirtualPod:     true,
+			clearImage:         true,
+			imagesPhysicalNode: []corev1.ContainerImage{{Names: []string{"ghcr.io/jetpack/calico"}}},
+		},
+		{
+			name:               "Don't Clear Nodes",
+			virtualNodeExists:  true,
+			withVirtualPod:     true,
+			clearImage:         false,
```
FabianKramm, line 204: Since the default is false, is this needed?

FabianKramm (approved):

pkg/controllers/resources/nodes/syncer_test.go
facchettos: just wanted to be explicit, but yeah it's not really needed

PR #1972, opened 2024-07-23
refactor: label translation

FabianKramm:
/kind enhancement

### DIFF
--- a/pkg/util/translate/translate.go
+++ b/pkg/util/translate/translate.go
@@ -531,9 +531,7 @@ func MergeLabelSelectors(elems ...*metav1.LabelSelector) *metav1.LabelSelector {
 				out.MatchLabels[k] = v
 			}
 		}
-		for _, expr := range selector.MatchExpressions {
-			out.MatchExpressions = append(out.MatchExpressions, expr)
-		}
+		out.MatchExpressions = append(out.MatchExpressions, selector.MatchExpressions...)
 	}
 	return out
 }
--- a/pkg/util/translate/translate.go
+++ b/pkg/util/translate/translate.go
@@ -73,6 +73,76 @@ func HostMetadata[T client.Object](ctx *synccontext.SyncContext, vObj T, name ty
 	return pObj
 }
 
+func HostLabelsMap(ctx *synccontext.SyncContext, vLabels, pLabels map[string]string, vNamespace string) map[string]string {
+	if vLabels == nil {
+		return nil
+	}
+
+	newLabels := map[string]string{}
+	for k, v := range vLabels {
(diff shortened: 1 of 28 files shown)


pkg/util/translate/translate.go
```diff
@@ -450,3 +519,25 @@ func KindExists(config *rest.Config, groupVersionKind schema.GroupVersionKind) (
...
+		for k, v := range selector.MatchLabels {
+			if out.MatchLabels == nil {
+				out.MatchLabels = map[string]string{}
+			}
+			out.MatchLabels[k] = v
+		}
+		for _, expr := range selector.MatchExpressions {
+			if out.MatchExpressions == nil {
+				out.MatchExpressions = []metav1.LabelSelectorRequirement{}
+			}
```
facchettos, line 538: is this needed? append should take care of it I think

pkg/util/translate/translate.go
```diff
@@ -450,3 +519,25 @@ func KindExists(config *rest.Config, groupVersionKind schema.GroupVersionKind) (
...
+func MergeLabelSelectors(elems ...*metav1.LabelSelector) *metav1.LabelSelector {
+	out := &metav1.LabelSelector{}
+	for _, selector := range elems {
+		if selector == nil {
+			continue
+		}
+		for k, v := range selector.MatchLabels {
+			if out.MatchLabels == nil {
+				out.MatchLabels = map[string]string{}
+			}
```
facchettos, line 532: I'd probably take this out of the loop and do 
```go
if len(selector.MatchLabels)>0{
    out.MatchLabels = make(map[string]string, len(selector.MatchLabels))
}

File: pkg/util/translate/translate.go (hunk 1 above)
Comment (FabianKramm) [reply pkg/util/translate/translate.go]: Removed it

File: pkg/util/translate/translate.go [hunk 3]
```diff
@@ -73,6 +73,79 @@ func HostMetadata[T client.Object](ctx *synccontext.SyncContext, vObj T, name ty
...
+
+	newLabelSelector := &metav1.LabelSelector{}
+	if labelSelector.MatchLabels != nil {
+		newLabelSelector.MatchLabels = map[string]string{}
+		for k, v := range labelSelector.MatchLabels {
+			newLabelSelector.MatchLabels[labelFunc(ctx, k)] = v
+		}
+	}
+	if len(labelSelector.MatchExpressions) > 0 {
+		newLabelSelector.MatchExpressions = []metav1.LabelSelectorRequirement{}
```
facchettos, line 136: this is not needed

FabianKramm: Yep true, I fixed that

PR #1981, opened 2024-07-24
refactor: make syncer generic

FabianKramm:
/kind enhancement

Restructures the syncer and makes the parameters generic so we don't have to cast objects anymore

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -44,6 +44,8 @@ func NewSyncController(ctx *synccontext.RegisterContext, syncer syncertypes.Sync
 	return &SyncController{
 		syncer: syncer,
 
+		genericSyncer: syncer.Syncer(),
+
 		config: ctx.Config,
 
 		mappings: ctx.Mappings,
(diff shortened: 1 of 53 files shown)


pkg/syncer/syncer.go
```diff
@@ -167,9 +169,24 @@ func (r *SyncController) Reconcile(ctx context.Context, origReq ctrl.Request) (_
...
-		return r.syncer.Sync(syncContext, pObj, vObj)
+		return r.genericSyncer.Sync(syncContext, &synccontext.SyncEvent[client.Object]{
+			Type:   syncEventType,
+			Source: syncEventSource,
+
+			Virtual:    vObj,
+			VirtualOld: nil,
+
+			Host:    pObj,
+			HostOld: nil,
```
facchettos, line 180: this is because it's not implemented yet right?

FabianKramm: Yes exactly, I will remove that

facchettos (approved):

PR #1987, opened 2024-07-25
fix: do not explicitly set zero values if key is not present

johannesfrey:
/kind bugfix

resolves ENG-4125

Fixed an issue where vcluster would overwrite explicit default values with the type's zero value.

There is an alternative PR here, which takes over the default values: https://github.com/loft-sh/vcluster/pull/1986
Not entirely sure which path we should follow?!


FabianKramm:
Closed in favour of #1991

PR #2007, opened 2024-08-01
[fix] add exception for the platform in network policy

facchettos:
/kind bugfix

resolves eng-4121

Fixed an issue where vcluster would not configure network policies correctly when trying to connect to the platform 


config/config.go
```diff
@@ -1438,7 +1438,8 @@ type LimitRange struct {
 
 type NetworkPolicy struct {
 	// Enabled defines if the network policy should be deployed by vCluster.
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled         bool `json:"enabled,omitempty"`
+	PlatformEnabled bool `json:"platformEnabled,omitempty"`
```
FabianKramm, line 1442: Platform enabled is a little bit misleading, can we move this into `outgoingConnections.platform: true`?

chart/templates/networkpolicy.yaml
```diff
@@ -95,6 +95,12 @@ spec:
           podSelector:
             matchLabels:
               k8s-app: kube-dns
+        {{- if .Values.policies.networkPolicy.platformEnabled }}
```
johannesfrey, line 98: We want to toggle this separately and this shouldn't be just hard-coded like we did with dns when `.Values.policies.networkPolicy.enabled` is set to true?!

config/config.go
facchettos: I followed what was in the ticket

config/config.go
```diff
@@ -1451,6 +1451,8 @@ type OutgoingConnections struct {
 	// to the pods matched by a NetworkPolicySpec's podSelector. The except entry describes CIDRs
 	// that should not be included within this rule.
 	IPBlock IPBlock `json:"ipBlock,omitempty"`
+
+	Platform bool
```
FabianKramm, line 1455: Please add a comment as well above the field

FabianKramm (approved):

PR #2011, opened 2024-08-01
Adding test for installing and deleting virtual cluster from kubectl

sowmyav27:
**What issue type does this pull request address?** (keep at least one, remove the others) 
/kind test

**What does this pull request do?**
Adds a test - install and delete vcluster using kubectl

**Please provide a short message that should be published in the vcluster release notes**
N/A

**What else do we need to know?**

PR #2013, opened 2024-08-01
PR for Adding tests for installing and deleting virtual cluster from kubectl

sowmyav27:
/kind test

- But this is for testing the GHA with some minor changes.
- Original PR: https://github.com/loft-sh/vcluster/pull/2011
- If this works, depending on how we want to proceed, I might close of these PRs
- fixed comments from here - https://github.com/loft-sh/vcluster/pull/2011#pullrequestreview-2213293663

**Details:**
- Implemented a module for running the install and delete commands for virtual cluster through helm and kubectl in the `e2e.yaml` GHA workflow file. 
- Added bash scripts for the actions - helm and kubectl

**To-do in next PR** 
- once this PR is merged - install and delete using argoCD and terraform

Adds a test - install and delete vcluster using kubectl and helm

N/A

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -139,6 +139,67 @@ jobs:
           name: e2e-binaries
           path: ./test/*/*.test
           retention-days: 7
+  
+  vcluster-install-delete:
+    name: Install and delete virtual cluster
+    needs:
+      - build-and-push-syncer-image
+      - build-vcluster-cli
+      
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - uses: azure/setup-helm@v4
+        name: Setup Helm
+        with:
+          version: "v3.11.0"
+
+      - name: Set up kind k8s cluster
+        uses: engineerd/setup-kind@v0.5.0
+        with:
+          version: "v0.20.0"
+          image: kindest/node:v1.30.0@sha256:047357ac0cfea04663786a612ba1eaba9702bef25227a794b52890dd8bcd692e
+
+      - name: Testing kind cluster set-up
+        run: |
+          set -x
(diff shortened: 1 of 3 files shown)


FabianKramm (changes requested):
Not sure if this is doing what we want, within our e2e test pipeline we want. The test itself looks good, but the problem is that the e2e pipeline itself is creating a vcluster as well via the regular method, it seems to me you only want to install & delete via kubectl and not want to have that other vcluster created as well. You either need to adjust the e2e pipeline itself for this (.github/workflows/e2e.yaml) or need to create a separate pipeline.

test/e2e_vcluster_install/deploy_and_delete_vcluster.go
```diff
@@ -0,0 +1,65 @@
...
+		err = kubectlCmd.Run()
+		framework.ExpectNoError(err)
+
+		gomega.Eventually(func() bool {
+
+			checkCmd := exec.Command("vcluster", "list")
+			output, err := checkCmd.CombinedOutput()
+			framework.ExpectNoError(err)
+			return err == nil && strings.Contains(string(output), vClusterName) && strings.Contains(string(output), "Running")
+		}, pollingDurationLong, pollingInterval).Should(gomega.BeTrue(), "Virtual cluster failed to come up Active")
```
zerbitx, line 50: nit: Recommend using the "With" methods like [these](https://github.com/loft-sh/vcluster/blob/main/test/e2e_plugin/plugin/plugin.go#L30-L31)

FabianKramm (approved):

FabianKramm (changes requested):
I tihnk whats weird is that we have all these different dimensions still for your new test. For what distros and dimensions do you want to run your test?

FabianKramm (approved):
@sowmyav27 LGTM! Thanks for the PR!

PR #2014, opened 2024-08-02
bugfix: do not output logs in vcluster connect if --print is set

hidalgopl:
/kind bugfix

resolves ENG-4270

Fix replaces the `Log` if `--print` flag is set. This new Log instance will log all non-error logs to `io.Discard`. Also, added two e2e tests:
1. First executes `vcluster connect -n vcluster --print vcluster` and writes it to temporary file, and then uses `kubectl --kubeconfig=<temporary-file> get pods -A` to check if outputed kubeconfig can be used to connect to virtual cluster.
2. Second tests checks if the `vcluster connect -n vcluster --print NOT_EXISTING_CLUSTER_NAME` still outputs error.

Fixed an issue where `vcluster connect --print` outputs invalid kubeconfig (it included log)

I initially tried to construct `cobra.Command` inside the e2e test (as it is done [here](https://github.com/loft-sh/vcluster/pull/2014/files#diff-e64a93b22baa2c465b8c347271dbde99882ebaf5ed3d8fb5d1f2d9d6d8850518L30), but I couldn't direct the output of the `cmd.Execute()` to the temporary kubeconfig file (even using `cmd.SetOut(kubeCfgFile)`). I believe reason for that is that we write to `os.Stdout` directly here: https://github.com/hidalgopl/vcluster/blob/fix-vcluster-connect-with-print/pkg/cli/connect_helm.go#L154 (so we don't take into account the `outWriter` from `cobra.Command`.

### DIFF
--- a/cmd/vclusterctl/cmd/connect.go
+++ b/cmd/vclusterctl/cmd/connect.go
@@ -4,6 +4,10 @@ import (
 	"cmp"
 	"context"
 	"fmt"
+	"io"
+	"os"
+
+	"github.com/sirupsen/logrus"
 
 	"github.com/loft-sh/log"
 	"github.com/loft-sh/vcluster/pkg/cli"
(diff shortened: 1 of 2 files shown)


FabianKramm (approved):
@hidalgopl thanks for the PR! Great work, LGTM!

PR #2029, opened 2024-08-06
Connect to vCluster Platform earlier

rmweir:
vCluster is now able to request a data source from vCluster Platform. vCluster is only able to use the data source if it attempts to connect to the vCluster Platform prior to deploying its own control plane. vCluster Platform is now connected before control plane is deployed but the platform controllers and tailscale server are started after because they require the control plane to be running.

/kind feature

Part of ENG-4279
Part of enabling using the same database server credential for multiple vcluster datastores by creating new database and user, then returning credentials for the created database and user.

### DIFF
--- a/cmd/vcluster/cmd/start.go
+++ b/cmd/vcluster/cmd/start.go
@@ -89,7 +89,12 @@ func ExecuteStart(ctx context.Context, options *StartOptions) error {
 	// set features for plugins to recognize
 	plugin.DefaultManager.SetProFeatures(pro.LicenseFeatures())
 
-	// check if we should create certs
+	// connect to vCluster platform if configured
+	startPlatformServersAndControllers, err := pro.ConnectToPlatform(ctx, vConfig)
+	if err != nil {
+		return fmt.Errorf("connect to platform: %w", err)
+	}
+
 	err = setup.Initialize(ctx, vConfig)
 	if err != nil {
 		return fmt.Errorf("initialize: %w", err)
@@ -99,6 +104,11 @@ func ExecuteStart(ctx context.Context, options *StartOptions) error {
 	controllerCtx, err := setup.NewControllerContext(ctx, vConfig)
 	if err != nil {
 		return fmt.Errorf("create controller context: %w", err)
+	}
+
+	err = startPlatformServersAndControllers(controllerCtx.VirtualManager)
+	if err != nil {
+		return fmt.Errorf("start platform controllers: %w", err)
 	}
 
 	// start integrations
@@ -132,11 +142,6 @@ func ExecuteStart(ctx context.Context, options *StartOptions) error {
 		if err != nil {
 			return fmt.Errorf("start integrated core dns: %w", err)
 		}
-	}
-
-	// connect to vCluster platform if configured
-	if err := pro.ConnectToPlatform(ctx, vConfig, controllerCtx.VirtualManager); err != nil {
-		return fmt.Errorf("connect to platform: %w", err)
 	}
 
 	// start leader election + controllers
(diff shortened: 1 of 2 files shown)


cmd/vcluster/cmd/start.go
```diff
@@ -89,7 +89,12 @@ func ExecuteStart(ctx context.Context, options *StartOptions) error {
 	// set features for plugins to recognize
 	plugin.DefaultManager.SetProFeatures(pro.LicenseFeatures())
 
-	// check if we should create certs
+	// connect to vCluster platform if configured
+	startPlatformServersAndControllers, err := pro.ConnectToPlatform(ctx, vConfig)
```
FabianKramm, line 93: Whats the benefit of returning a function here compared to having 2 separate pro functions? I think returning a function always makes the code a bit more complex and I don't really see why this is useful in this case, is this about the closure that is important?

pkg/pro/platform.go
```diff
@@ -7,6 +7,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
-var ConnectToPlatform = func(context.Context, *config.VirtualClusterConfig, manager.Manager) error {
-	return nil
+var ConnectToPlatform = func(context.Context, *config.VirtualClusterConfig) (func(mgr manager.Manager) error, error) {
+	return nil, nil
```
FabianKramm, line 11: Return a function that does nothing here to avoid lint and avoid the nil check above

cmd/vcluster/cmd/start.go
```diff
@@ -101,6 +106,13 @@ func ExecuteStart(ctx context.Context, options *StartOptions) error {
 		return fmt.Errorf("create controller context: %w", err)
 	}
 
+	if startPlatformServersAndControllers != nil {
+		err = startPlatformServersAndControllers(controllerCtx.VirtualManager)
+		if err != nil {
+			return fmt.Errorf("start platform controllers")
```
facchettos, line 112: you should probably wrap the error here

pkg/pro/platform.go
```diff
@@ -7,6 +7,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
-var ConnectToPlatform = func(context.Context, *config.VirtualClusterConfig, manager.Manager) error {
-	return nil
+var ConnectToPlatform = func(context.Context, *config.VirtualClusterConfig) (func(mgr manager.Manager) error, error) {
```
facchettos, line 10:
```suggestion
var ConnectToPlatform = func(context.Context, *config.VirtualClusterConfig) (func(_ manager.Manager) error, error) {
```

pkg/pro/platform.go
```diff
@@ -7,6 +7,6 @@ import (
...
+	return func(mgr manager.Manager) error { return nil }, nil
```
facchettos, line 11:
```suggestion
var ConnectToPlatform = func(context.Context, *config.VirtualClusterConfig) (func(manager.Manager) error, error) {
	return func(_ manager.Manager) error { return nil }, nil
```

cmd/vcluster/cmd/start.go
rmweir: @FabianKramm I was not so sure about this myself. The reason why I chose to do it is because the start controller logic is highly dependent on dependencies created above it. Decoupling them would mean returning a handful of variables that I do not think makes sense from a reader's perspective, nor are they useful to other functions. It would also mean introducing a new pro function that I do not think is necessary. There are other solutions but I'm not convinced they are nicer than this. The server/controllers are related to `ConnectToPlatform` at least. 

Another solution that I did consider was starting it in a go routine within `ConnectToPlatform` but blocking it with an empty channel, waitgroup, or some other sync mechanism, and then returning that. Do you think that would be better?

FabianKramm (approved):

PR #2030, opened 2024-08-06
refactor: only rewrite certain labels

FabianKramm:
/kind enhancement

resolves ENG-4289

Changes:
* Changes the way vCluster rewrites labels, instead of rewriting all labels it only rewrites a couple to avoid conflicts (app, release, marker, namespace label etc.)
* Removes the mapping store portion of rewriting labels
* Adds an additional unit test for checking if labels can be rewritten correctly
* Removes `experimental.syncSettings.syncLabels` as that's not needed anymore
* Migrating between an older vCluster version and this version should in almost all cases resolve itself. The only problematic cases can be:
  * Custom vCluster plugins that still do it the old way and somehow interact with other resources / require the old label format
  * Pods that reference their own labels via downwardAPI (rather uncommon as most use annotations), can be resolved by restarting / recreating the pod


lizardruss (approved):
LGTM

FWIW, Kurt has used this setting in a [few places in his Demo environments](https://github.com/search?q=org%3Aloft-demos%20syncLabels&type=code), but it doesn't look like it will be needed moving forward.

PR #2031, opened 2024-08-06
feature: add controlPlane.distro.k8s.version chart value for CAPI

hidalgopl:
… intergration

/kind enhancement

resolves ENG-2725
`controlPlane.distro.k8s.version` is a shortcut to set image tags for.

- controlPlane.distro.k8s.apiServer.image.tag
- controlPlane.distro.k8s.controllerManager.image.tag
- controlPlane.distro.k8s.schedule.image.tag

New behavior: `controlPlane.distro.k8s.version` will override image tags only if they're not specified by user (so they're different than current default tag, `v1.30.3`. This only affects `k8s` distro.

Add `controlPlane.distro.k8s.version` to ease setting up the cluster control plane components version.

It's a bit unfortunate that I had to add the logic to the helm chart, but I don't see other way around it (open to suggestions!). Added a bunch of unit tests.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -124,6 +124,17 @@ controlPlane:
     k8s:
       # Enabled specifies if the K8s distro should be enabled. Only one distro can be enabled at the same time.
       enabled: false
+      # Version specifies k8s components (scheduler, kube-controller-manager & apiserver) version.
+      # It is a shortcut for controlPlane.distro.k8s.apiServer.image.tag,
+      # controlPlane.distro.k8s.controllerManager.image.tag and
+      # controlPlane.distro.k8s.scheduler.image.tag
+      # If e.g. controlPlane.distro.k8s.version is set to v1.30.1 and
+      # controlPlane.distro.k8s.scheduler.image.tag
+      # (or controlPlane.distro.k8s.controllerManager.image.tag or controlPlane.distro.k8s.apiServer.image.tag)
+      # is set to v1.31.0,
+      # value from controlPlane.distro.k8s.<controlPlane-component>.image.tag will be used
+      # (where <controlPlane-component is apiServer, controllerManager and scheduler).
+      version: ""
       # APIServer holds configuration specific to starting the api server.
       apiServer:
         enabled: true
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -2,6 +2,8 @@ package config
 
 import (
 	_ "embed"
+	"fmt"
+	"os"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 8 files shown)


config/values.yaml
```diff
@@ -108,6 +108,7 @@ controlPlane:
         requests:
           cpu: 40m
           memory: 64Mi
+      version: ""
```
FabianKramm, line 111: I think we should move that up, so people see it sooner, just move it below enabled

FabianKramm:
Besides the small comment above, LGTM!

config/values.yaml
hidalgopl: Sounds good, I thought they're sorted alphabetically:) fixed!

chart/values.yaml
```diff
@@ -124,6 +124,17 @@ controlPlane:
...
+      # It is a shortcut for controlPlane.distro.k8s.apiServer.image.tag,
+      # controlPlane.distro.k8s.controllerManager.image.tag and
+      # controlPlane.distro.k8s.scheduler.image.tag
+      # If e.g. controlPlane.distro.k8s.version is set to v1.30.1 and
+      # controlPlane.distro.k8s.scheduler.image.tag
+      # (or controlPlane.distro.k8s.controllerManager.image.tag or controlPlane.distro.k8s.apiServer.image.tag)
+      # is set to v1.31.0,
+      # value from controlPlane.distro.k8s.<controlPlane-component>.image.tag will be used
+      # (where <controlPlane-component is apiServer, controllerManager and scheduler).
+      version: ""
```
johannesfrey, line 137: nit: Should we add a comment above the actual default tags below that if the tag is bumped the `$defaultTag` inside the _init-containers.tpl should also be bumped to the same value?!

hidalgopl: I can add a unit test in `config/config_test.go` that will check if `controlPlane.distro.<component>.image.tag` values are in sync with `$defaultTag` value in `_init-containers.tpl`. Then if anyone updates image tags without the defaultTag, this test will fail. WDYT?

johannesfrey: Sounds good, if it's not too much effort 🙂

hidalgopl: not at all, added :+1:

config/config_test.go
```diff
@@ -378,3 +380,23 @@ func TestConfig_IsProFeatureEnabled(t *testing.T) {
 		})
 	}
 }
+
+func TestIfDefaultImagesVersionsAreInSync(t *testing.T) {
+	defaultConfig, err := NewDefaultConfig()
+	assert.NilError(t, err)
+	initContainersTplFilePath := os.Getenv("REPO_ROOT_DIR") + "/chart/templates/_init-containers.tpl"
```
johannesfrey, line 387: Wouldn't this mean that the unit test will only work when run in CI (where this var is set?!)? I guess we also cannot work with `go:embed` because we are outside of the config dir...
As said. Totally fine if we only add a comment. Sorry for the additional effort because of my comment 😅

hidalgopl: I forgot to push changes in `./hack/test.sh`. How do you usually run the unit tests locally? If it's via `hack/test.sh` then it will work in locally as well as in CI. IMHO it would be more future proof to have a check in place for version sync, hence if it's not an issue from the developer experience POV, I'd keep this test.

johannesfrey: I sometimes just run `go test` directly (e.g. via `run package tests` in the IDE). Couldn't we just use `../chart/templates/_init-containers.tpl`, so we would not need the env var? Sure, if we move this unit test file or the template we have to adjust this as well.

hidalgopl: good point, gonna change it to a relative path

config/config_test.go
```diff
@@ -378,3 +380,24 @@ func TestConfig_IsProFeatureEnabled(t *testing.T) {
...
+	assert.NilError(t, err)
+	// this will fail when this test is moved or _init-containers.tpl is moved
+	initContainersTplFilePath := "../chart/templates/_init-containers.tpl"
+	tplBytes, err := os.ReadFile(initContainersTplFilePath)
+
+	assert.NilError(t, err)
+	assert.Equal(t, defaultConfig.ControlPlane.Distro.K8S.ControllerManager.Image.Tag, defaultConfig.ControlPlane.Distro.K8S.APIServer.Image.Tag)
+	assert.Equal(t, defaultConfig.ControlPlane.Distro.K8S.ControllerManager.Image.Tag, defaultConfig.ControlPlane.Distro.K8S.Scheduler.Image.Tag)
+	assert.Equal(t, defaultConfig.ControlPlane.Distro.K8S.APIServer.Image.Tag, defaultConfig.ControlPlane.Distro.K8S.Scheduler.Image.Tag)
+	expectedDefaultTag := fmt.Sprintf("{{- $defaultTag := \"%s\" -}}", defaultConfig.ControlPlane.Distro.K8S.ControllerManager.Image.Tag)
```
johannesfrey, line 395: nit:
```suggestion
	expectedDefaultTag := fmt.Sprintf("{{- $defaultTag := %q -}}", defaultConfig.ControlPlane.Distro.K8S.ControllerManager.Image.Tag)
```

johannesfrey (approved):

PR #2044, opened 2024-08-08
[feat] added priorityclass syncing from host

facchettos:
/kind feature

resolves eng-4179

### DIFF
--- a/pkg/controllers/resources/priorityclasses/translate.go
+++ b/pkg/controllers/resources/priorityclasses/translate.go
@@ -18,18 +18,36 @@ func (s *priorityClassSyncer) translate(ctx *synccontext.SyncContext, vObj clien
 	return priorityClass
 }
 
-func (s *priorityClassSyncer) translateUpdate(pObj, vObj, sourceObject, targetObject *schedulingv1.PriorityClass) {
+func (s *priorityClassSyncer) translateFromHost(ctx *synccontext.SyncContext, pObj client.Object) *schedulingv1.PriorityClass {
+	// translate the priority class
+	priorityClass := translate.VirtualMetadata(pObj.(*schedulingv1.PriorityClass), s.HostToVirtual(ctx, types.NamespacedName{Name: pObj.GetName(), Namespace: pObj.GetNamespace()}, pObj))
+	if priorityClass.Name == "" {
+		priorityClass.Name = pObj.GetName()
+	}
+	return priorityClass
+}
+
+func (s *priorityClassSyncer) translateUpdate(event *synccontext.SyncEvent[*schedulingv1.PriorityClass]) {
+	targetObject := event.TargetObject()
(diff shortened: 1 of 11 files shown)


pkg/controllers/resources/priorityclasses/translate.go
```diff
@@ -18,18 +18,42 @@ func (s *priorityClassSyncer) translate(ctx *synccontext.SyncContext, vObj clien
...
 }
 
-func (s *priorityClassSyncer) translateUpdate(pObj, vObj, sourceObject, targetObject *schedulingv1.PriorityClass) {
+func (s *priorityClassSyncer) translateFromHost(ctx *synccontext.SyncContext, pObj client.Object) *schedulingv1.PriorityClass {
+	// translate the priority class
+	priorityClass := translate.VirtualMetadata(pObj.(*schedulingv1.PriorityClass), s.HostToVirtual(ctx, types.NamespacedName{Name: pObj.GetName(), Namespace: pObj.GetNamespace()}, pObj))
+	priorityClass.GlobalDefault = false
+	if priorityClass.Value > 1000000000 {
+		priorityClass.Value = 1000000000
+	}
```
FabianKramm, line 27: Why are we doing these adjustments here, I know they are needed when going from virtual -> host, but in this case we could just leave them be, no?

pkg/controllers/resources/priorityclasses/translate.go
```diff
@@ -18,18 +18,38 @@ func (s *priorityClassSyncer) translate(ctx *synccontext.SyncContext, vObj clien
...
 
 	// check value
-	translatedValue := vObj.Value
+	translatedValue := sourceObject.Value
 	if translatedValue > 1000000000 {
 		translatedValue = 1000000000
 	}
-	pObj.Value = translatedValue
+	targetObject.Value = translatedValue
```
FabianKramm, line 54: I think we should only do this if syncing toHost, fromHost we don't mind the value

pkg/controllers/resources/priorityclasses/translate.go
```diff
@@ -18,18 +18,40 @@ func (s *priorityClassSyncer) translate(ctx *synccontext.SyncContext, vObj clien
...
 	// check value
-	translatedValue := vObj.Value
-	if translatedValue > 1000000000 {
-		translatedValue = 1000000000
+	if event.Source == synccontext.SyncEventSourceVirtual {
+		translatedValue := sourceObject.Value
+		if translatedValue > 1000000000 {
+			translatedValue = 1000000000
+		}
+		targetObject.Value = translatedValue
```
FabianKramm, line 55: Can you move that into the switch case and then exchange targetObject and sourceObject with the correct vObj and pObj?

FabianKramm (approved):

PR #2046, opened 2024-08-08
add --ca-cert flag to the vcluster platform add vcluster, so it can b…

hidalgopl:
…e used by passing .additionalCA from platform config

/kind enhancement

part of ENG-4224

adds --ca-cert flag and adds `certificate-authority-data` to the platform secret to enable secure connection between platform and vcluster. Overall behavior is described [in this PR](https://github.com/loft-sh/vcluster-pro/pull/384)

### DIFF
--- a/pkg/cli/create_helm.go
+++ b/pkg/cli/create_helm.go
@@ -380,7 +380,7 @@ func (cmd *createHelm) addVCluster(ctx context.Context, vClusterConfig *config.C
 		return nil
 	}
 
-	err = platform.ApplyPlatformSecret(ctx, cmd.LoadedConfig(cmd.log), cmd.kubeClient, "", cmd.Namespace, cmd.Project, "", "", false)
+	err = platform.ApplyPlatformSecret(ctx, cmd.LoadedConfig(cmd.log), cmd.kubeClient, "", cmd.Namespace, cmd.Project, "", "", false, []byte{})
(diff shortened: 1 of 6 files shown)


pkg/cli/create_helm.go
```diff
@@ -380,7 +380,7 @@ func (cmd *createHelm) addVCluster(ctx context.Context, vClusterConfig *config.C
 		return nil
 	}
 
-	err = platform.ApplyPlatformSecret(ctx, cmd.LoadedConfig(cmd.log), cmd.kubeClient, "", cmd.Namespace, cmd.Project, "", "", false)
+	err = platform.ApplyPlatformSecret(ctx, cmd.LoadedConfig(cmd.log), cmd.kubeClient, "", cmd.Namespace, cmd.Project, "", "", false, cmd.LoadedConfig(cmd.log).Platform.CertificateAuthorityData)
```
FabianKramm, line 383: Shouldn't this come from the flag? Why is this read from the config, while the other options are read from the flag?

hidalgopl: I'm not sure what's the preferred way to do this here.
I can change it for a flag, which will mean that `vcluster create vcluster` with driver `helm` will get a new flag. Do you think this should be a desired behavior?

FabianKramm (approved):

PR #2051, opened 2024-08-08
Adding in deploy changes tests for vcluster

sowmyav27:
Tests -- Adding in deploy changes tests for vcluster
- networkPolicy, limitRange and resourceQuota


test/deploy_changes/common/common_helpers.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func DeployChangesToVClusterUsingCLI(f *framework.Framework) {
+	gomega.Eventually(func() bool {
+		stdout := &bytes.Buffer{}
+		deployCmd := exec.Command("vcluster", "create", "--upgrade", f.VclusterName, "--namespace", f.VclusterNamespace, "--local-chart-dir", chartPath, "-f", filePath)
+		deployCmd.Stdout = stdout
+		err := deployCmd.Run()
+		framework.ExpectNoError(err)
+		return err == nil && strings.Contains(stdout.String(), "Switched active kube context to")
```
rmweir, line 38: Would an error be normal here, if so which error? We don't want to poll on an error if its not going to change.

test/deploy_changes/common/common_helpers.go
```diff
@@ -0,0 +1,57 @@
...
+		framework.ExpectNoError(err, "Error disconnecting from vCluster")
+	}
+}
+
+func VerifyClusterIsActive(f *framework.Framework) {
+	gomega.Eventually(func() bool {
+		checkCmd := exec.Command("vcluster", "list")
+		output, err := checkCmd.CombinedOutput()
+		framework.ExpectNoError(err)
+		return err == nil && strings.Contains(string(output), f.VclusterName) && strings.Contains(string(output), "Running")
```
rmweir, line 55: I think this check is too general. Here is the output of my `vcluster list`:
```
      NAME    |        NAMESPACE         | STATUS  |    VERSION     | CONNECTED |    AGE     
  ------------+--------------------------+---------+----------------+-----------+------------
    vcluster2 | loft-default-v-vcluster2 | Running | 0.20.0-beta.15 |           | 266h4m40s  
    vcluster  | vcluster                 | Running | uLmOpUH        |           | 238h42m5s  
```

Let's say instead `vcluster2` was not in `Running` and f.VclusterName was `vcluster2`. The output still contains `vcluster2` and `Running`. I think instead you may want to split by new line or, even better use a regex. I still would be worried about using contains with the former because you have to consider on a single line that f.VclusterName could be a substring of another vcluster name, like with `vcluster` and `vcluster2`.

test/deploy_changes/deploy_changes_networking/commonValues.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - from: test/nginx
+        to: nginx
+    fromHost:
+      - from: test/test
+        to: default/test
+      - from: test/nginx
+        to: default/nginx
+experimental:
+  syncSettings:
+    setOwner: true
```
rmweir, line 45: nit: add newline

test/deploy_changes/deploy_changes_sync/deploy_changes_sync.go
```diff
@@ -0,0 +1,67 @@
...
+	filePath              = "commonValues.yaml"
+	chartPath             = "../../../chart"
+	initialNsLabelKey     = "testing-ns-label"
+	initialNsLabelValue   = "testing-ns-label-value"
+	testingContainerName  = "nginx"
+	testingContainerImage = "nginxinc/nginx-unprivileged"
+)
+
+var _ = ginkgo.Describe("Deploy sync changes to vCluster", func() {
+	ginkgo.It("should deploy a vCluster using kubectl and verify sync changes ", func() {
```
rmweir, line 30: This says `deploy a vCluster` but it seems there is one already deployed. Redeploy or update would be more accurate.

rmweir (changes requested):

FabianKramm (changes requested):
This is a pretty big change as that moves all tests to a different folder and I'm not sure this is what we want as e2e tests are pretty hidden now. I would like to eventually have a consistent naming across vcluster, vcluster-pro and loft-enterprise. Whats the reasoning behind moving them to `functional_tests`? I find the naming a little odd as we have now `test` -> `functional_tests` and `test` -> `deploy_changes`, which seems a little inconsistent to me, as one has `tests` in it and the other not. When looking at Kubernetes, they also do `test` -> `e2e`, so I would like to stay with that pattern, but we could introduce `test` -> `deploy` or something similar for your new tests

PR #2052, opened 2024-08-09
[enhancement] now refuses to create multiple virtual clusters in same ns

facchettos:
/kind enhancement

resolves eng-4308

Fixed an issue where vcluster would let users create multiple virtual clusters in the same ns

### DIFF
--- a/pkg/cli/create_helm.go
+++ b/pkg/cli/create_helm.go
@@ -130,7 +130,7 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 	if err != nil {
 		return err
 	}
-	vclusters, err := find.ListVClusters(ctx, "", "", cmd.Namespace, log)
+	vclusters, err := find.ListVClusters(ctx, cmd.Context, "", cmd.Namespace, log)
 	if err != nil {
 		return err
 	}
--- a/pkg/cli/create_helm.go
+++ b/pkg/cli/create_helm.go
@@ -130,6 +130,15 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 	if err != nil {
 		return err
 	}
+	vclusters, err := find.ListVClusters(ctx, "", "", cmd.Namespace, log)
+	if err != nil {
+		return err
+	}
+	for _, v := range vclusters {
+		if v.Namespace == cmd.Namespace && v.Name != vClusterName {
+			return fmt.Errorf("there is already a virtual cluster in namespace %s", cmd.Namespace)
+		}
+	}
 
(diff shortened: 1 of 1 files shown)


pkg/cli/create_helm.go
```diff
@@ -130,6 +130,15 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 	if err != nil {
 		return err
 	}
+	vclusters, err := find.ListVClusters(ctx, "", "", cmd.Namespace, log)
+	if err != nil {
+		return err
+	}
+	for _, v := range vclusters {
+		if v.Namespace == cmd.Namespace && v.Name != vClusterName {
+			return fmt.Errorf("there is already a virtual cluster in the %s namespace", cmd.Namespace)
```
johannesfrey, line 139: nit:
```suggestion
			return fmt.Errorf("there is already a virtual cluster in namespace %s", cmd.Namespace)
```

johannesfrey (approved):

cmd/vclusterctl/cmd/create.go
```diff
@@ -77,6 +78,16 @@ func (cmd *CreateCmd) Run(ctx context.Context, args []string) error {
...
 
+	vclusters, err := find.ListVClusters(ctx, "", "", cmd.Namespace, cmd.log)
+	if err != nil {
+		return err
+	}
+	for _, v := range vclusters {
+		if v.Namespace == cmd.Namespace && v.Name != args[0] {
+			return fmt.Errorf("there is already a virtual cluster in namespace %s", cmd.Namespace)
+		}
+	}
```
FabianKramm, line 89: I think this should go into the helm driver only, as there can be many vClusters in the platform in different host clusters

FabianKramm (changes requested):
See comment above

cmd/vclusterctl/cmd/create.go
facchettos: I had it there before, but do we support creating multiple virtual clusters in the same ns with the platform?

facchettos: it seems that the `ListOSS` function should work with platform created virtual clusters too, since it's fetching the statefulset/deployment

pkg/cli/create_helm.go
```diff
@@ -130,6 +130,15 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 	if err != nil {
 		return err
 	}
+	vclusters, err := find.ListVClusters(ctx, "", "", cmd.Namespace, log)
```
FabianKramm, line 133: We should specify cmd.Context in here as well

FabianKramm (approved):

PR #2054, opened 2024-08-12
Add workflow to trigger doc PR in vcluster-docs

zerbitx:
/kind documentation

resolves #DOC-226

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -94,6 +94,26 @@ jobs:
             });
               console.log(response);
 
+      - name: Trigger cli-docs workflow
+        uses: actions/github-script@v7
+        continue-on-error: true
+        with:
+          github-token: ${{ secrets.GH_ACCESS_TOKEN }}
+          script: |
+            const version = '${{ steps.get_version.outputs.release_version }}';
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -94,6 +94,26 @@ jobs:
...
+              owner: 'loft-sh',
+              repo: 'vcluster-docs',
+              workflow_id: 'cluster-docs.yaml',
+              ref: 'main',
+              inputs: {
+                version,
+                ref,
+              }
+            });
+            console.log(response);
```
lizardruss, line 115: Should this use `core.setOutput()` and `core.setFailed()`, similar to [this action](https://github.com/loft-sh/devpod/blob/cee1fd74d5fd17ccd3160c0dd23e77c4b22db2e1/.github/workflows/release.yaml#L68-L80)?

lizardruss (approved):
Looks good, just a question about handling the response.

.github/workflows/release.yaml
zerbitx: I believe that's only for setting up outputs to be consumed in the same workflow, where this is trying to trigger one on another repository its passed in the post data of the dispatch.

PR #2074, opened 2024-08-16
fix(chart): allow not specifying resourceQuota quotas and limitRange resources

JuniorJPDJ:
/kind bugfix

resolves #

Allows not setting limits and resources in resourceQuota and limitRange. It's impossible to unset some of those values when the default values are specified. If using Umbrella Chart even `null` doesn't work. This makes it possible to specify only part of those values and set it to own.

Allows not setting some of limits and resources in resourceQuota and limitRange.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -740,21 +740,21 @@ policies:
     annotations: {}
     # Quota are the quota options
     quota:
-      requests.cpu: 10
-      requests.memory: 20Gi
-      requests.storage: "100Gi"
-      requests.ephemeral-storage: 60Gi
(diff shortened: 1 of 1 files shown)


FabianKramm:
@JuniorJPDJ thanks for the PR! Not sure if it makes sense to just comment these out, maybe we can introduce an option or new field that allows you to override all other values similar to what we have with roles?

JuniorJPDJ:
IMO this doesn't make sense as it's just config showcase / default value, which by the way is disabled by default.
Comment is there so everyone see what the keys can be.

PR #2075, opened 2024-08-16
feat: Ignore updates to Rancher managed annotations

neogopher:
/kind feature

resolves ENG-4177

vcluster will now ignore updates to Rancher managed annotations inside the virtual cluster when node syncing is enabled.

### DIFF
--- a/pkg/controllers/resources/nodes/syncer_test.go
+++ b/pkg/controllers/resources/nodes/syncer_test.go
@@ -142,22 +142,24 @@ func TestSyncBothExist(t *testing.T) {
 		},
 	}
 	testCases := []struct {
-		name                    string
-		withVirtualPod          bool
-		withPhysicalPod         bool
-		virtualNodeExists       bool
-		modifiedPhysical        bool
-		expectNoVNode           bool
-		syncFromHostLabel       map[string]string
-		hostLabel               map[string]string
-		virtualFinalLabels      map[string]string
-		virtualFinalAnnotations map[string]string
-		physicalAnnotations     map[string]string
-		imagesPhysicalNode      []corev1.ContainerImage
-		enforceTolerations      []string
-		clearImage              bool
-		physicalTaints          []corev1.Taint
-		expectedTaints          []corev1.Taint
+		name                      string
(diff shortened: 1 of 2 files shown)


FabianKramm (approved):

lalyos:
I had this issue and it caused the manager consume 100% cpu. 

I just describe my findings for reference, if someone else running into it.

My use case was:
- a k3s host cluster with nginx-ingress controller
- a vcluster with no ingress controller inside, but ingress sync switched on: `helm upgrade -i myvc loft/vcluster --version v0.19.7 -n vcluster-myvc --set sync.ingresses.enabled=true`

than I started to create ingresses inside the vcluster and they were nicely synced to the host cluster and everything worked fine.

- there was a separate k8s cluster with rancher, and the host cluster was imported.
- up to this point everything worked fine
- once I imported the vcluster into rancher a race condition started

The cattle-cluster-agent was deployed into the host cluster and also into the vcluster, and both tried to set the annotation on the ingress.
- This resulted and endless growing number of revisions in the etcd (kine) storage.
- In turn the sqlite db become several GB
- which further caused kine to be unable to run the `compact` command
- which made the api-server so slow that a simple `k get po` took 8-10 sec.

## Trasintional workaround

Before this patch was introduced, I was able to decrease the db size from **8GB to 30mb**
every night by doing this process:

- stop k3s via systemctl
- download a standalone kine from https://github.com/k3s-io/kine/releases/tag/v0.13.7
- start `kine --debug &` in /var/lib/rancher/k3s/server/
- wait for a successfull `COMPACT`
- stop kine
- manually trigger a sqlite compact: `sqlite3 /var/lib/rancher/k3s/server/db/state.db <<< "VACUUM"`

by decreasing the sqlite db under 100mb the cpu load was under 30%.
although the db increased again during the day, but repeating this process made it manageble.

## How to check if this hit you

If you see `ingresses` in the top 5 resources with most revision, usually over 50,000 than this patch will help you.
```
qlite3 /var/lib/rancher/k3s/server/db/state.db <<< "select count(name) as c, name from kine group by name order by c desc limit 5"
```

PR #2082, opened 2024-08-19
feat: add dnsPolicy, dnsConfig config options for the controlPlane

matskiv:
/kind feature

Add dnsPolicy, dnsConfig config options for the controlPlane ti be able to set these fields in the pod spec of the vcluster pod.

Added dnsPolicy and dnsConfig config options for the controlPlane field in vcluster.yaml


FabianKramm (changes requested):
@matskiv thanks a lot for the PR! Would be great if you could add 2 helm unit tests for each of the new options to make sure they work as expected

FabianKramm (approved):

PR #2103, opened 2024-09-02
sync values.schema.json only on the latest release

hidalgopl:
/kind enhancement

part of ENG-4185

add a separate workflow to sync `values.schema.json` only on the latest release

tested workflow triggers in the playground repo https://github.com/hidalgopl/gh-actions-playground

### DIFF
--- a/.github/workflows/sync-config-schema.yaml
+++ b/.github/workflows/sync-config-schema.yaml
@@ -1,0 +1,68 @@
+name: Sync Config Schema
+on:
+  release:
+    types:
+      - published # this will run whenever a release (or pre-release is created)
+      - released # this will run when a (not-pre-)release is created or was changed from pre-release to a release
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    outputs:
+      latest_release_tag: ${{ steps.release.outputs.latest_release }}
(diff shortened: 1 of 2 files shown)


.github/workflows/sync-config-schema.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+          # set git info
+          git config --global user.name "Loft Bot"
+          git config --global user.email 'loft-bot@users.noreply.github.com'
+          echo "Changes detected"
+
+          # commit changes
+          git commit -m "chore: sync vCluster config schema"
+          git push -u origin -f main
+          echo "Pushed commit to main branch"
```
FabianKramm, line 68: Not sure if I understand the logic here correctly. On release we commit to main branch? Shouldn't we also create a release in the vcluster-config repo?

hidalgopl: I implemented the logic described here: https://linear.app/loft/issue/ENG-4185/create-merged-schema-in-githubcomloft-shvcluster-config
I believe the idea is to commit to main in vcluster-config on vcluster / platform latest releases and then on vcluster-config run a merging job whenever `values.schema.json` or `platform.schema.json` got updated

FabianKramm (approved):

PR #2112, opened 2024-09-04
Consolidate create/switch/update context with --connect

johannesfrey:
/kind enhancement

resolves #


FabianKramm (commented):
@johannesfrey looks good to me, but I'm a little scared to remove the update-current flag, since it is referenced a lot. Can we basically just remove the functionality of it, but still have it as a hidden flag that is deprecated?

FabianKramm (approved):

pkg/cli/create_platform.go
```diff
@@ -164,7 +164,7 @@ func CreatePlatform(ctx context.Context, options *CreateOptions, globalFlags *fl
 	// check if we should connect to the vcluster or print the kubeconfig
 	if options.Connect || options.Print {
 		return ConnectPlatform(ctx, &ConnectOptions{
-			UpdateCurrent:         options.UpdateCurrent,
+			UpdateCurrent:         true,
```
rohantmp, line 167: I think ConnectPlatform and UpdateKubeconfig above duplicate some of the connection functionality

rohantmp (changes requested):

pkg/cli/create_platform.go
rohantmp: The `ConnectPlatform()` function ends with a `writkubeconfig()` function

pkg/cli/create_platform.go
```diff
@@ -145,9 +145,9 @@ func CreatePlatform(ctx context.Context, options *CreateOptions, globalFlags *fl
 	}
 	log.Donef("Successfully created the virtual cluster %s in project %s", virtualClusterName, options.Project)
 
-	if options.CreateContext {
+	if options.Connect {
```
rohantmp, line 148: I wonder if this block could just be removed

pkg/cli/create_platform.go
johannesfrey: Good point. I checked the code and it looks like `ConnectPlatform` does call `platform.CreateVirtualClusterInstanceOptions` internally, so essentially doing the same as the block L148-L162 you mentioned above. I went ahead and remove this block and checked `create` in the different combinations (`--connect=true`, `--connect=false`, `--connect=false --print=true` once without a platform and once with being logged into the platform). Looks good so far.

pkg/cli/create_platform.go
johannesfrey: I removed it. See the comment above.

rohantmp (approved):
LGTM :)

FabianKramm (approved):

rohantmp (approved):

PR #2113, opened 2024-09-04
add --all flag to vcluster platform add vcluster

hidalgopl:
/kind feature

resolves ENG-4109

Adds `--all` flag to `vcluster platform add vcluster`. This can be used to import all vClusters from host cluster to the platform.

### DIFF
--- a/cmd/vclusterctl/cmd/platform/add/vcluster.go
+++ b/cmd/vclusterctl/cmd/platform/add/vcluster.go
@@ -29,7 +29,12 @@ Adds a vCluster to the vCluster platform.
 
 Example:
 vcluster platform add vcluster my-vcluster --namespace vcluster-my-vcluster --project my-project --import-name my-vcluster
+
+Add all vClusters in the host cluster:
+vcluster platform add vcluster ignored --project my-project --all
+
 ###############################################
+
 	`
 
 	addCmd := &cobra.Command{
@@ -51,6 +56,7 @@ vcluster platform add vcluster my-vcluster --namespace vcluster-my-vcluster --pr
 	addCmd.Flags().BytesBase64Var(&cmd.CertificateAuthorityData, "ca-data", []byte{}, "additional, base64 encoded certificate authority data that will be passed to the platform secret")
 	// This is hidden until the platform side will be ready to use it
 	_ = addCmd.Flags().MarkHidden("ca-data")
+	addCmd.Flags().BoolVar(&cmd.All, "all", false, "all will try to add all vCluster found in all namespaces in the host cluster. If this flag is set, vcluster name argument is ignored")
 
 	return addCmd
 }
--- a/cmd/vclusterctl/cmd/platform/add/vcluster.go
+++ b/cmd/vclusterctl/cmd/platform/add/vcluster.go
@@ -30,7 +30,7 @@ Adds a vCluster to the vCluster platform.
 Example:
 vcluster platform add vcluster my-vcluster --namespace vcluster-my-vcluster --project my-project --import-name my-vcluster
 
-Add all vClusters in the host cluster:
+Add all vCluster instances in the host cluster:
 vcluster platform add vcluster --project my-project --all
 
 ###############################################
--- a/cmd/vclusterctl/cmd/platform/add/vcluster.go
+++ b/cmd/vclusterctl/cmd/platform/add/vcluster.go
@@ -31,7 +31,7 @@ Example:
 vcluster platform add vcluster my-vcluster --namespace vcluster-my-vcluster --project my-project --import-name my-vcluster
 
 Add all vClusters in the host cluster:
-vcluster platform add vcluster ignored --project my-project --all
+vcluster platform add vcluster --project my-project --all
 
 ###############################################
 
@@ -41,7 +41,7 @@ vcluster platform add vcluster ignored --project my-project --all
 		Use:   "vcluster",
 		Short: "Adds an existing vCluster to the vCluster platform",
 		Long:  description,
-		Args:  cobra.ExactArgs(1),
+		Args:  cobra.RangeArgs(0, 1),
 		RunE: func(cobraCmd *cobra.Command, args []string) error {
 			return cmd.Run(cobraCmd.Context(), args)
 		},
@@ -56,12 +56,12 @@ vcluster platform add vcluster ignored --project my-project --all
 	addCmd.Flags().BytesBase64Var(&cmd.CertificateAuthorityData, "ca-data", []byte{}, "additional, base64 encoded certificate authority data that will be passed to the platform secret")
 	// This is hidden until the platform side will be ready to use it
 	_ = addCmd.Flags().MarkHidden("ca-data")
-	addCmd.Flags().BoolVar(&cmd.All, "all", false, "all will try to add all vCluster found in all namespaces in the host cluster. If this flag is set, vcluster name argument is ignored")
+	addCmd.Flags().BoolVar(&cmd.All, "all", false, "all will try to add Virtual Cluster found in all namespaces in the host cluster. If this flag is set, any provided vCluster name argument is ignored")
 
 	return addCmd
 }
 
 // Run executes the functionality
 func (cmd *VClusterCmd) Run(ctx context.Context, args []string) error {
-	return cli.AddVClusterHelm(ctx, &cmd.AddVClusterOptions, cmd.GlobalFlags, args[0], cmd.Log)
+	return cli.AddVClusterHelm(ctx, &cmd.AddVClusterOptions, cmd.GlobalFlags, args, cmd.Log)
 }
--- a/pkg/cli/add_vcluster_helm.go
+++ b/pkg/cli/add_vcluster_helm.go
@@ -4,11 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"time"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/tools/clientcmd"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/loft-sh/log"
 	"github.com/loft-sh/log/survey"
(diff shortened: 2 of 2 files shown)


pkg/cli/add_vcluster_helm.go
```diff
@@ -33,22 +39,105 @@ func AddVClusterHelm(
...
+		if err != nil {
+			return err
+		}
+		namespaces, err := hostKubeClient.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+		log.Debugf("looking for vclusters in %d namespaces", len(namespaces.Items))
+		for _, ns := range namespaces.Items {
+			log.Infof("looking for a vcluster in %s namespace", ns.GetName())
```
rmweir, line 63:
```suggestion
			log.Infof("looking for a vclusters in %s namespace", ns.GetName())
```

pkg/cli/add_vcluster_helm.go
```diff
@@ -2,9 +2,14 @@ package cli
...
 import (
 	"context"
+	"errors"
 	"fmt"
+	"strings"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/clientcmd"
+
```
rmweir, line 12: Remove newline.

pkg/cli/add_vcluster_helm.go
```diff
@@ -33,22 +39,105 @@ func AddVClusterHelm(
...
+			log.Infof("looking for a vcluster in %s namespace", ns.GetName())
+			vClustersInNamespace, err := find.ListVClusters(ctx, globalFlags.Context, "", ns.GetName(), log)
+			if err != nil {
+				return err
+			}
+			if len(vClustersInNamespace) == 0 {
+				log.Infof("no vClusters found in context %s and namespace %s", globalFlags.Context, ns.GetName())
+				continue
+			}
+			vClusters = append(vClusters, vClustersInNamespace...)
```
rmweir, line 72: I don't think you need to list all the namespaces and then iterate over them. I see we [pass an empty ns ](https://github.com/loft-sh/vcluster/blob/cf9a1a48a2b38594df2c4139bbb1d62f23f9bf69/pkg/cli/list_helm.go#L96) to `ListVClusters` elsewhere. That should just give you the vclusters in all namespaces.

cmd/vclusterctl/cmd/platform/add/vcluster.go
```diff
@@ -29,7 +29,12 @@ Adds a vCluster to the vCluster platform.
 
 Example:
 vcluster platform add vcluster my-vcluster --namespace vcluster-my-vcluster --project my-project --import-name my-vcluster
+
+Add all vClusters in the host cluster:
+vcluster platform add vcluster ignored --project my-project --all
```
rmweir, line 34: Instead of forcing them to pass an ignored vcluster name, set [Args](https://github.com/loft-sh/vcluster/blob/cf9a1a48a2b38594df2c4139bbb1d62f23f9bf69/cmd/vclusterctl/cmd/platform/add/vcluster.go#L39) to `RangeArgs(0, 1)` then return an error if args is 0 and `--all` is false.

cmd/vclusterctl/cmd/platform/add/vcluster.go
```diff
@@ -51,6 +56,7 @@ vcluster platform add vcluster my-vcluster --namespace vcluster-my-vcluster --pr
 	addCmd.Flags().BytesBase64Var(&cmd.CertificateAuthorityData, "ca-data", []byte{}, "additional, base64 encoded certificate authority data that will be passed to the platform secret")
 	// This is hidden until the platform side will be ready to use it
 	_ = addCmd.Flags().MarkHidden("ca-data")
+	addCmd.Flags().BoolVar(&cmd.All, "all", false, "all will try to add all vCluster found in all namespaces in the host cluster. If this flag is set, vcluster name argument is ignored")
```
rmweir, line 59:
```suggestion
	addCmd.Flags().BoolVar(&cmd.All, "all", false, "all will try to add all vClusters found in all namespaces in the host cluster. If this flag is set, any provided vcluster name argument is ignored")
```

pkg/cli/add_vcluster_helm.go
```diff
@@ -33,22 +39,105 @@ func AddVClusterHelm(
...
+	}
+	return errors.New(errMsg.String())
+}
 
+func (vce *VClusterAddError) addErr(vClusterName string, err error) {
+	if err == nil {
+		return
+	}
+	vce.errs = append(vce.errs, fmt.Errorf("cannot add vcluster %s: %w", vClusterName, err))
+}
```
rmweir, line 130: For aggregating errors, use a pattern like [this](https://github.com/loft-sh/vcluster/blob/cf9a1a48a2b38594df2c4139bbb1d62f23f9bf69/pkg/controllers/resources/nodes/nodeservice/node_service.go#L83-L114) instead. TL;DR append them to a slice and then use `utilerrors.NewAggregate`.

rmweir (changes requested):

cmd/vclusterctl/cmd/platform/add/vcluster.go
FabianKramm: Yes, please change that

hidalgopl: good point!

pkg/cli/add_vcluster_helm.go
facchettos: because of the trademark, we should either use, `virtual cluster` or `vcluster instance` but not use `vcluster(s)` like that

cmd/vclusterctl/cmd/platform/add/vcluster.go
```diff
@@ -29,14 +29,19 @@ Adds a vCluster to the vCluster platform.
 
 Example:
 vcluster platform add vcluster my-vcluster --namespace vcluster-my-vcluster --project my-project --import-name my-vcluster
+
+Add all vClusters in the host cluster:
```
facchettos, line 33: see my comment about the trademark

pkg/cli/add_vcluster_helm.go
```diff
@@ -24,31 +27,79 @@ type AddVClusterOptions struct {
...
-	// check if vCluster exists
-	vCluster, err := find.GetVCluster(ctx, globalFlags.Context, vClusterName, globalFlags.Namespace, log)
-	if err != nil {
-		return err
+	var vClusters []find.VCluster
+	if len(args) == 0 && !options.All {
+		return errors.New("empty vCluster name but no --all flag set, please either set vCluster name to add one cluster or set --all flag to add all of them")
+	}
+	if options.All {
+		log.Debugf("add vcluster called with --all flag")
```
facchettos, line 45: is that really useful outside of development context?

pkg/cli/add_vcluster_helm.go
```diff
@@ -24,31 +27,79 @@ type AddVClusterOptions struct {
...
+		log.Info("looking for vClusters in all namespaces")
+		vClustersInNamespace, err := find.ListVClusters(ctx, globalFlags.Context, "", "", log)
+		if err != nil {
+			return err
+		}
+		if len(vClustersInNamespace) == 0 {
+			log.Infof("no vClusters found in context %s", globalFlags.Context)
```
facchettos, line 52: tm

pkg/cli/add_vcluster_helm.go
```diff
@@ -24,31 +27,79 @@ type AddVClusterOptions struct {
...
 		return err
 	}
+
+	// create kube client
 	kubeClient, err := kubernetes.NewForConfig(restConfig)
 	if err != nil {
 		return err
 	}
+	var addErrors []error
+	log.Debugf("trying to add %d vClusters to platform", len(vClusters))
```
facchettos, line 81: tm

pkg/cli/add_vcluster_helm.go
```diff
@@ -24,31 +27,79 @@ type AddVClusterOptions struct {
...
+	for _, vCluster := range vClusters {
+		vCluster := vCluster
+		log.Infof("adding %s vCluster to platform", vCluster.Name)
```
facchettos, line 84: tm

pkg/cli/add_vcluster_helm.go
```diff
@@ -24,31 +27,79 @@ type AddVClusterOptions struct {
...
+	args []string,
 	log log.Logger,
 ) error {
-	// check if vCluster exists
-	vCluster, err := find.GetVCluster(ctx, globalFlags.Context, vClusterName, globalFlags.Namespace, log)
-	if err != nil {
-		return err
+	var vClusters []find.VCluster
+	if len(args) == 0 && !options.All {
+		return errors.New("empty vCluster name but no --all flag set, please either set vCluster name to add one cluster or set --all flag to add all of them")
```
facchettos, line 42: tm

facchettos, line 42: I personally wouldn't exit with an error here. I'd rather try to display the default help message instead

hidalgopl: This is probably a question of the CLI style, but I feel we should error here (as it is an user input error). 
For example, I can imagine someone using this CLI command in the CI job, in a script with `set -e` and passing vcluster name dynamically (e.g. from the env var). In the case when name is empty and we don't error here (and exit with `0`), then the pipeline may continue and user may expect that vcluster got added to platform.

rmweir (approved):

pkg/cli/add_vcluster_helm.go
hidalgopl: changed

pkg/cli/add_vcluster_helm.go
hidalgopl: addressed

pkg/cli/add_vcluster_helm.go
hidalgopl: good catch - removed

PR #2120, opened 2024-09-05
Add db fields

rmweir:
/kind feature

Part of ENG-4279

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1090,7 +1090,17 @@ type Database struct {
 	Embedded DatabaseKine `json:"embedded,omitempty"`
 
 	// External defines that an external database should be used as the backend for the virtual cluster
-	External DatabaseKine `json:"external,omitempty"`
+	External ExternalDatabaseKine `json:"external,omitempty"`
+}
+
+type ExternalDatabaseKine struct {
+	DatabaseKine
+
+	// Connector specifies a secret located in a connected vCluster Platform that contains database server connection information
+	// to be used by Platform to create a database and database user for the vCluster.
+	// and non-privileged user. A kine endpoint should be created using the database and user on Platform registration.
+	// This is optional.
+	Connector string `json:"connector,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -1044,7 +1044,16 @@ type Database struct {
...
+	External ExternalDatabaseKine `json:"external,omitempty"`
+}
+
+type ExternalDatabaseKine struct {
+	DatabaseKine
+
+	// DBServerCredentials specifies database server connection information to be used by Platform to create a database
+	// and non-privileged user. A kine endpoint should be created using the database and user on Platform registration.
+	// This is optional.
+	DBServerCredentials string `json:"dbServerCredentials,omitempty"`
```
FabianKramm, line 1056: I'm not sure if this really makes it clear what happens here. This is platform only functionality right? Also is this a reference to database credentials or the credentials themselves? I think we should rather name this something like `platformDatabaseCredentialsRef`.

FabianKramm (changes requested):

config/config.go
```diff
@@ -1066,7 +1066,17 @@ type Database struct {
...
+}
+
+type ExternalDatabaseKine struct {
+	DatabaseKine
+
+	// Connector specifies a secret located in a connected vCluster Platform that contains database server connection information
+	// to be used by Platform to create a database and database user for the vCluster.
+	// and non-privileged user. A kine endpoint should be created using the database and user on Platform registration.
+	// This is optional.
+	Connector string `json:"connector,omitempty"`
```
FabianKramm, line 1079: I don't think Connector is a good name? What should connector mean?

FabianKramm (changes requested):

config/config.go
LukasGentele: This has been resolved in a discussion a few weeks ago

LukasGentele (approved):

FabianKramm (approved):

PR #2131, opened 2024-09-11
fix(cli): do not upgrade if latest stable is older(eg: beta)

rohantmp:
part of ENG-4354


FabianKramm (changes requested):
@rohantmp why do we not allow that?

FabianKramm (approved):

PR #2132, opened 2024-09-11
Fix: reset password namespace & start chart name

FabianKramm:
/kind bugfix

### DIFF
--- a/cmd/vclusterctl/cmd/platform/start.go
+++ b/cmd/vclusterctl/cmd/platform/start.go
@@ -3,7 +3,9 @@ package platform
 import (
 	"context"
 	"fmt"
+	"strings"
 
+	"github.com/blang/semver"
 	"github.com/loft-sh/log"
 	"github.com/loft-sh/log/survey"
 	"github.com/loft-sh/log/terminal"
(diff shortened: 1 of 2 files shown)


neogopher (approved):

cmd/vclusterctl/cmd/platform/start.go
```diff
@@ -83,6 +85,15 @@ func (cmd *StartCmd) Run(ctx context.Context) error {
 			cmd.Version = latestVersion
 		}
 	}
+
+	// if < v4.0.0 then use ChartName loft
+	parsedVersion, err := semver.Parse(strings.TrimPrefix(cmd.Version, "v"))
+	if err != nil {
+		return fmt.Errorf("parse provided version %s: %w", cmd.Version, err)
+	} else if parsedVersion.LT(semver.MustParse("4.0.0-alpha.0")) && cmd.ChartName == "vcluster-platform" {
```
deniseschannon, line 93: Why is this hard-coded? 

Upgraded platforms will still have `loft`, but only fresh installs will have `vcluster-platform`. 

Also, users could choose any namespace name that they want.

FabianKramm: This is not about the namespace its the chart that doesn't exist prior to v4.0.0-alpha.0

Issue #2105, opened 2024-09-02
Can't deploy helm chart when artefact file name is different than chart name


wim1988:
### What happened?

I wanted to deploy a helm chart into my vCluster, but vCluster was logging the error: 
ERROR    deploy/start.go:42    Error deploying manifests: couldn't find chart    {"component": "vcluster"}

### What did you expect to happen?

vCluster installs the specified helm chart.

### How can we reproduce it (as minimally and precisely as possible)?

Specify a helm release like this:
...
- chart:
  name: strimzi-kafka-operator
  repo: https://strimzi.io/charts
  version: "0.41.0"
...

vCluster will download and store the chart in its working directory as strimzi-kafka-operator-helm-3-chart-0.41.0.tgz.
Then it will fail with the error I mentioned. If I manually enter the vCluster pod and rename the file to strimzi-kafka-operator-0.41.0.tgz, it continues and succeeds.
This is because it only works, if the downloaded artefact name equals the specified helm chart name. I checked your code and this seems to be the issue.

### Host cluster Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.29.0-eks-5e0fdde
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.10
```

</details>

### vcluster version

<details>

```console
$ vcluster --version
0.20.0
(also tested with 0.20.0-beta.5 and beta.16)
```

</details>

### VCluster Config

<details>

```
values:
  sync:
    toHost:
      pods:
        rewriteHosts:
          initContainer:
            image: "library/alpine:3.13.1"
  controlPlane:
    distro:
      k3s:
        enabled: true
    statefulSet:
      persistence:
        volumeClaim:
          retentionPolicy: Retain
      scheduling:
        podManagementPolicy: OrderedReady
  experimental:
    deploy:
      vcluster:
        helm:
          - chart:
              name: strimzi-kafka-operator
              repo: https://strimzi.io/charts
              version: "0.41.0"
            release:
              name: strimzi-kafka-operator
              namespace: operators
```

</details>

PR #2133, opened 2024-09-11
fix: init manifests fail if chart artifact name differs from chart name

rohantmp:
part of ENG-4457

verified manually that bundled charts work with and without a provided name

*Notes for reviewer:*
- charts are now put in unique directories so that the artifact name (which cannot be controlled by `helm pull`) doesn't matter. The pulled chart will be the only file in that directory.

fixes: #2105


rohantmp:
ugh, tests passed, but there was a typo in commit message. Ready to review

PR #2136, opened 2024-09-12
fix(cli): improve UX around vcluster upgrade command

rohantmp:
part of ENG-4354

a continuation of #2131  with logging to give the user some guidance

screenshot:
image

image

### DIFF
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -76,10 +76,8 @@ func CheckForNewerVersion() (string, error) {
 			errLatestVersion = err
 			return
 		}
-
-		v := semver.MustParse(version)
 		// if latest not found or latest version found is less than or equal to the current version, do not upgrade
-		if !found || latest.Version.Compare(v) != 1 {
+		if !found {
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


rohantmp:
I don't think this requires a backport as it mainly affects unstable versions. Closed the backport PR

pkg/upgrade/upgrade.go
```diff
@@ -147,13 +145,27 @@ func Upgrade(flagVersion string, log log.Logger) error {
...
 	}
-	if newerVersion == "" {
+	if latestStable == "" {
+		log.Infof("Current binary is the latest version: %s", version)
+		return nil
+	}
+
+	version = GetVersion()
+	semverCurrentVersion := semver.MustParse(version)
+	semverLatestStableVersion := semver.MustParse(latestStable)
```
FabianKramm, line 159: Can this not include a `v` and fail?

rohantmp: No, neither of these are from user input.  `latestStable` orginates from a `semver.Version.String()`

FabianKramm (approved):

PR #2147, opened 2024-09-16
Changes fallback ns to vcluster-platform

zerbitx:
Adds function to look up loft deployment's namespace

/kind enhancement

resolves #ENG-4560

Tries to look up the platform install's deployment to get the namespace rather than assuming "loft" if the user doesn't specify and the flag default isn't vcluster-platform.

### DIFF
--- a/cmd/vclusterctl/cmd/platform/add/cluster.go
+++ b/cmd/vclusterctl/cmd/platform/add/cluster.go
@@ -121,13 +121,29 @@ func (cmd *ClusterCmd) Run(ctx context.Context, args []string) error {
 					User: user,
 					Team: team,
 				},
-				NetworkPeer: true,
-				Access:      getAccess(user, team),
+				NetworkPeer:         true,
+				ManagementNamespace: cmd.Namespace,
+				Access:              getAccess(user, team),
 			},
 		},
 	}, metav1.CreateOptions{})
 	if err != nil && !kerrors.IsAlreadyExists(err) {
 		return fmt.Errorf("create cluster: %w", err)
+	}
+
+	// get namespace to install if cluster already exists
+	if kerrors.IsAlreadyExists(err) {
+		cluster, err := managementClient.Loft().ManagementV1().Clusters().Get(ctx, clusterName, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("get cluster: %w", err)
+		}
+
+		cmd.Namespace = cluster.Spec.ManagementNamespace
+		if cmd.Namespace == "" {
+			cmd.Namespace = "loft" // since this is hardcoded in the platform at https://github.com/loft-sh/loft-enterprise/blob/b716f86a83d5f037ad993a0c3467b54393ef3b1f/pkg/util/agenthelper/helper.go#L9
+		}
+
+		cmd.Log.Infof("Using namespace %s because cluster already exists", cmd.Namespace)
 	}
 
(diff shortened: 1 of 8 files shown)


cmd/vclusterctl/cmd/platform/add/cluster.go
```diff
@@ -70,7 +70,7 @@ vcluster platform add cluster my-cluster
 		},
 	}
 
-	c.Flags().StringVar(&cmd.Namespace, "namespace", "loft", "The namespace to generate the service account in. The namespace will be created if it does not exist")
```
FabianKramm, line 73: Can we test that this works as expected, I'm worried this will not work as we hardcode loft in the platform backend I believe

cmd/vclusterctl/cmd/platform/backup/management.go
```diff
@@ -67,8 +68,10 @@ vcluster platform backup management
 		},
 	}
 
+	platformNamespace, _ := clihelper.VClusterPlatformInstallationNamespace(context.Background())
```
ThomasK33, line 71: Is it safe to ignore this error here?

zerbitx: Yeah, its a bit weird.  If this fails the namespace flag value will just be an empty string and [this prompt](https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/platform/backup/management.go#L96-L99) will allow them to proceed.  The backup will proceed, but with warnings and output like

```shell
09:52:23 info Backing up clusterroletemplates...
09:52:23 info Backing up clusteraccesses...
...
09:52:23 info Backing up runners...
09:52:23 info Backing up projects...
09:52:23 warn backup clusterrole templates: no matches for kind "ClusterRoleTemplate" in version "storage.loft.sh/v1"
09:52:23 warn backup cluster accesses: no matches for kind "ClusterAccess" in version "storage.loft.sh/v1"
09:52:23 warn backup users: no matches for kind "User" in version "storage.loft.sh/v1"
09:52:23 warn backup teams: no matches for kind "Team" in version "storage.loft.sh/v1"
...
09:52:23 warn backup runners: no matches for kind "Runner" in version "storage.loft.sh/v1"
09:52:23 warn backup projects: no matches for kind "Project" in version "storage.loft.sh/v1"
```

So in the interests of having it work like it did before, I just ignored the error, and I didn't fall back to `vcluster-platform` since that should have been found if it was a valid alternative.

cmd/vclusterctl/cmd/platform/add/cluster.go
zerbitx: Ok I went back and did a few more tests of this without specifying the namespace to get `vcluster-platform` as the default and tried out some random namespaces too.   Perhaps I'm misunderstanding something.  I didn't think the namespace here mattered much, it would just be the namespace of the deployment, to then make a management.loft.sh/v1 cluster object.  Can you point me to some places I should look in the platform to double check besides my just running the add command and creating vcluster instances in the external cluster(s)?

FabianKramm: I added the relevant logic

zerbitx:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|v0.20|[](https://github.com/loft-sh/vcluster/pull/2167)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport)

PR #2148, opened 2024-09-16
Captures helm error before trying to call WaitForReadyLoftPod

zerbitx:
/kind bugfix

resolves #ENG-4585

Fixed an issue where if the helm command to add a cluster failed, the error wouldn't be reported quickly because waiting for the pod began before the helm error was checked.


andyluak (approved):

ThomasK33 (commented):
If this is going to be sequential, we don't need the goroutine or the errChan.

FabianKramm:
Yeah why is that running in parallel at all?

FabianKramm (approved):

PR #2115, opened 2024-09-05
fix: use proper config location

rohantmp:
respect the --config flag and it's defaults

part of ENG-4398

PR #2149, opened 2024-09-17
[0.20] fix: create --connect fails because of improper config location, remove duplicate flags, drop duplicate functionality

rohantmp:
resolves ENG-4557

manual backport of https://github.com/loft-sh/vcluster/pull/2115 and https://github.com/loft-sh/vcluster/pull/2112

Hold off on merging without considering the changes in flags. I do think they should be backported as the combination doesn't make sense and both flags have to be disabled to set to false to not connect when creating.


rohantmp:
Hold off on merging without considering the changes in flags. I do think they should be backported as the combination doesn't make sense and both flags have to be disabled to set to false to not connect when creating.

FabianKramm (approved):

PR #2155, opened 2024-09-18
fix: update platform add command to reference the platform instead of loft

deniseschannon:
/kind bugfix

part of #eng-4516

Fixed an issue where vcluster platform add referenced `loft` instead of the new platform references. 

### DIFF
--- a/cmd/vclusterctl/cmd/platform/add/cluster.go
+++ b/cmd/vclusterctl/cmd/platform/add/cluster.go
@@ -106,7 +106,7 @@ func (cmd *ClusterCmd) Run(ctx context.Context, args []string) error {
 
 	loftVersion, err := platformClient.Version()
 	if err != nil {
-		return fmt.Errorf("get loft version: %w", err)
+		return fmt.Errorf("get platform version: %w", err)
 	}
 
 	// TODO(ThomasK33): Eventually change this into an Apply instead of a Create call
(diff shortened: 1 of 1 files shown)


cmd/vclusterctl/cmd/platform/add/cluster.go
```diff
@@ -70,7 +70,7 @@ vcluster platform add cluster my-cluster
 		},
 	}
 
-	c.Flags().StringVar(&cmd.Namespace, "namespace", "loft", "The namespace to generate the service account in. The namespace will be created if it does not exist")
+	c.Flags().StringVar(&cmd.Namespace, "namespace", "vcluster-platform", "The namespace to generate the service account in. The namespace will be created if it does not exist")
```
FabianKramm, line 73: This breaks a lot of things as the Cluster Object is not updated with the namespace, I think https://github.com/loft-sh/vcluster/pull/2147 does this better

FabianKramm (changes requested):

cmd/vclusterctl/cmd/platform/add/cluster.go
deniseschannon: I forgot about this when making this PR. I'll remove it from my PR and see if @zerbitx can help fix this one since I've found it.

FabianKramm (approved):

PR #2168, opened 2024-09-23
Make it possible to create virtual clusters with numbers only name

johannesfrey:
/kind bugfix

resolves ENG-4551

Make it possible to create virtual clusters with digits only name

In order to be able to create virtual clusters with a digits-only name we have to change the way how we create services based on the name. Previously we just used the name verbatim which in case of pure digits violates the DNS-1035 naming convention for service names used by Kubernetes. 
It turned out this change cannot only be made to our helm templates as there have been several places in the code base that implicitly rely on services being named like the virtual cluster itself. I hope I found all the places and adjusted them to rely on the service name directly.
The change should be backwards compatible as the `vc-` prefix is only prepended to the name when the vcluster name consists of digits only (thx @ThomasK33). So if users already have virtual clusters running with a regular name (e.g. "foo") the service's name will remain "foo" as well.

**TODO after merge**
- [ ] Bump vcluster dep in vcluster-pro

### DIFF
--- a/chart/templates/_helper.tpl
+++ b/chart/templates/_helper.tpl
@@ -15,3 +15,11 @@
 {{ .repository }}:{{ .tag }}
 {{- end -}}
 {{- end -}}
+
+{{- define "vcluster.name" -}}
+{{- if regexMatch "^[0-9]+$" .Release.Name -}}
+{{- printf "vc-%s" .Release.Name -}}
+{{- else -}}
+{{- printf "%s" .Release.Name }}
+{{- end -}}
+{{- end -}}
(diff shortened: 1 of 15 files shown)


chart/templates/_helper.tpl
```diff
@@ -15,3 +15,7 @@
 {{ .repository }}:{{ .tag }}
 {{- end -}}
 {{- end -}}
+
+{{- define "vcluster.name" -}}
+{{- printf "vc-%s" .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
```
ThomasK33, line 21: Can we make this conditional so it only prepends the `vc-` prefix if it starts with a number/non-alphabetic character?

Otherwise, helm might recreate the vCluster services on an upgrade.

PR #2169, opened 2024-09-23
fix: handle trailing slash in mount paths for hostpath mapper

neogopher:
/kind bugfix

resolves ENG-4427

Fixed an issue where bind mounts for hostpath mapping would not forward changes correctly when mountpath contained a trailing forward-slash

### DIFF
--- a/pkg/controllers/resources/pods/translate/hostpath.go
+++ b/pkg/controllers/resources/pods/translate/hostpath.go
@@ -25,6 +25,9 @@ const (
 func (t *translator) ensureMountPropagation(pPod *corev1.Pod) {
 	for i, container := range pPod.Spec.Containers {
 		for j, volumeMount := range container.VolumeMounts {
+			// handle scenarios where path ends with a /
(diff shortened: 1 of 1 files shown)


pkg/controllers/resources/pods/translate/hostpath.go
```diff
@@ -25,6 +25,9 @@ const (
 func (t *translator) ensureMountPropagation(pPod *corev1.Pod) {
 	for i, container := range pPod.Spec.Containers {
 		for j, volumeMount := range container.VolumeMounts {
+			// handle scenarios where path ends with a /
+			volumeMount.MountPath = strings.TrimSuffix(volumeMount.MountPath, "/")
```
hidalgopl, line 29: Could you add unit test case for it to the `TestVolumeTranslation`?

FabianKramm: I'll merge this for now, we can do a separate PR with the unit test

PR #2181, opened 2024-09-25
Create sub-directories in config for versions

rmweir:
Some vcluster config logic is version specific. Splitting the config package into version specific sub-directories helps maintain backwards compatibility for package consumers. The subdirectories allow us to use version specific code as needed without carrying it forward to newer versions, creating unnecessary technical debt. Separate go module versions were proposed as an alternative but were rejected in favor of sub-directories for now, as sub-directories are  simpler to implement and does not necessitate consumers to import multiple modules.

/kind enhancement

resolves #


rmweir:
Moving to different solution: removing config dependence entirely entirely.

PR #2198, opened 2024-10-04
chore: Update translate to patches

ThomasK33:
/kind enhancement

resolves ENG-4775

Renamed translate field to patches

### DIFF
--- a/chart/values.schema.json
+++ b/chart/values.schema.json
@@ -1016,18 +1016,18 @@
       "additionalProperties": false,
       "type": "object"
     },
-    "EnableSwitchWithTranslate": {
+    "EnableSwitchWithPatches": {
       "properties": {
         "enabled": {
           "type": "boolean",
           "description": "Enabled defines if this option should be enabled."
         },
-        "translate": {
+        "patches": {
           "items": {
             "$ref": "#/$defs/TranslatePatch"
           },
           "type": "array",
-          "description": "Translate the patch according to the given patches."
+          "description": "Patches patch the resource according to the provided specification."
         }
       },
       "additionalProperties": false,
@@ -2910,12 +2910,12 @@
           "type": "boolean",
           "description": "All defines if all resources of that type should get synced or only the necessary ones that are needed."
         },
-        "translate": {
+        "patches": {
           "items": {
             "$ref": "#/$defs/TranslatePatch"
           },
           "type": "array",
-          "description": "Translate the patch according to the given patches."
+          "description": "Patches patch the resource according to the provided specification."
         }
       },
       "additionalProperties": false,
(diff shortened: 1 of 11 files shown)


chart/values.schema.json
```diff
@@ -3161,12 +3161,12 @@
...
           "description": "Enabled defines if this option should be enabled."
         },
-        "translate": {
+        "patches": {
           "items": {
             "$ref": "#/$defs/TranslatePatch"
           },
           "type": "array",
-          "description": "Translate the patch according to the given patches."
+          "description": "Patches the resources according to the given patches."
```
rmweir, line 3169: This is ambiguous, maybe keep "translate" here?

config/config.go
```diff
@@ -494,8 +494,8 @@ type SyncToHostCustomResource struct {
 	// Enabled defines if this option should be enabled.
 	Enabled bool `json:"enabled,omitempty"`
 
-	// Translate the patch according to the given patches.
-	Translate []TranslatePatch `json:"translate,omitempty"`
+	// Patches the resources according to the given patches.
+	Patches []TranslatePatch `json:"patches,omitempty"`
 }
 
 type TranslatePatch struct {
```
rmweir, line 501: Do we want to rename this to just "Patch"?

chart/values.schema.json
```diff
@@ -3047,12 +3047,12 @@
...
           "description": "RewriteHosts is a special option needed to rewrite statefulset containers to allow the correct FQDN. virtual cluster will add\na small container to each stateful set pod that will initially rewrite the /etc/hosts file to match the FQDN expected by\nthe virtual cluster."
         },
-        "translate": {
+        "patches": {
           "items": {
             "$ref": "#/$defs/TranslatePatch"
           },
           "type": "array",
-          "description": "Translate the patch according to the given patches."
+          "description": "Patches the resource according to the given patches."
```
rmweir, line 3055: This is ambiguous, maybe keep "translate" here?

rmweir (changes requested):

config/config.go
deniseschannon: No, it's an array and multiple patches can be used.

chart/values.schema.json
deniseschannon: Patch is a common word used to do exactly what we're doing which is update/change a resource. 

https://kubernetes.io/docs/reference/kubectl/generated/kubectl_patch/

config/config.go
rmweir: I'm referring to the struct not the field name. There is also the struct ` EnableSwitchWithTranslate` which should probably be changed to `EnabledSwitchWithPatch`.

rmweir (commented):
1. I think everything prefixed with `Translate` can have that removed or swapped for `Patch`.
2. I think `EnableSwitchWithTranslate` especially need to be changed to `EnabledSwitchWithPatch`.
3. Patch can both be a noun and a verb but I don't think it should be both in a single sentence. I would prefer a description such as "Defines patches to be applied to resources." or "Defines patches for resources", as patches are _usually_ used for one thing- patching.

chart/values.schema.json
ThomasK33: How would I keep it here, given that this file gets generated off the Go struct?

config/config.go
ThomasK33: "TranslatePatch" as struct is fine, as internally we perform translation operations (see the translate package in pkg).

ThomasK33:
> Patch can both be a noun and a verb but I don't think it should be both in a single sentence. I would prefer a description such as "Defines patches to be applied to resources." or "Defines patches for resources", as patches are usually used for one thing- patching.

How does that align with Go doc conventions?

An example from the Go std lib: https://pkg.go.dev/io#Copy

FabianKramm (approved):

PR #2211, opened 2024-10-08
fix: change warnings to info, increase time between messages

zerbitx:
/kind enhancement

resolves #ENG-4817

### DIFF
--- a/pkg/platform/helper.go
+++ b/pkg/platform/helper.go
@@ -34,7 +34,7 @@ import (
 
 var errNoClusterAccess = errors.New("the user has no access to any cluster")
 
-var waitDuration = 20 * time.Second
+var defaultPollInterval = 40 * time.Second
 
 type VirtualClusterInstanceProject struct {
 	VirtualCluster *managementv1.VirtualClusterInstance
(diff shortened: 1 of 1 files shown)


pkg/platform/helper.go
```diff
@@ -1007,7 +1007,7 @@ func WaitForVCluster(ctx context.Context, client Client, clusterName, spaceName,
 	}
 
 	now := time.Now()
-	nextMessage := now.Add(waitDuration)
+	nextMessage := now.Add(defaultPollInterval)
```
lizardruss, line 1010: Is it intentional to use this instead of `pollInterval` here?

pkg/platform/helper.go
```diff
@@ -1020,7 +1020,7 @@ func WaitForVCluster(ctx context.Context, client Client, clusterName, spaceName,
 				log.Info("Waiting for virtual cluster to be available...")
 			}
 
-			nextMessage = time.Now().Add(waitDuration)
+			nextMessage = time.Now().Add(defaultPollInterval)
```
lizardruss, line 1023: Same question as above

lizardruss (changes requested):
Possible change to make the usage of `defaultPollInterval` consistent

pkg/platform/helper.go
zerbitx: Ah, no.   That function is actually not used anywhere. I'll remove it.

FabianKramm (approved):

PR #2216, opened 2024-10-09
feat: prompt for missing vcluster name in create command

zerbitx:
/kind enhancement

resolves #ENG-4851

### DIFF
--- a/cmd/vclusterctl/cmd/create.go
+++ b/cmd/vclusterctl/cmd/create.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"cmp"
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/loft-sh/log"
@@ -44,12 +45,18 @@ Example:
 vcluster create test --namespace test
 #######################################################
 	`,
-		Args: util.VClusterNameOnlyValidator,
 		RunE: func(cobraCmd *cobra.Command, args []string) error {
+			newArgs, err := util.PromptForArgs(cmd.log, args, "vcluster name")
+			if err != nil && errors.Is(err, util.ErrNonInteractive) {
+				if err := util.VClusterNameOnlyValidator(cobraCmd, args); err != nil {
+					return err
+				}
+			}
+
 			// Check for newer version
 			upgrade.PrintNewerVersionWarning()
 
-			return cmd.Run(cobraCmd.Context(), args)
+			return cmd.Run(cobraCmd.Context(), newArgs)
 		},
 	}
 
--- a/pkg/cli/util/args.go
+++ b/pkg/cli/util/args.go
@@ -1,9 +1,13 @@
 package util
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
+	"github.com/loft-sh/log"
+	"github.com/loft-sh/log/survey"
+	"github.com/loft-sh/log/terminal"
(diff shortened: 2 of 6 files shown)


ThomasK33 (approved):

pkg/cli/util/args.go
```diff
@@ -67,3 +69,24 @@ func NamedPositionalArgsValidator(failMissing, failExtra bool, expectedArgs ...s
...
 }
+
+// PromptForMissingArgs expects that the number of args, matched the number of argNames, in the order they should appear
+// and will prompt one by one for the missing args adding them to the args slice and returning a new set for a command to use
+func PromptForMissingArgs(l log.Logger, args []string, argNames ...string) ([]string, error) {
+	if len(args) == len(argNames) {
+		return args, nil
+	}
+
+	for i := range argNames[len(args):] {
```
FabianKramm, line 80: We should handle the case where this is not a interactive terminal as in a pipeline you would never know why this command has not succeeded as it just hangs for the question

FabianKramm (changes requested):
See comment above

zerbitx:
I also noticed warnings in CI about `--distro` being deprecated, so I made a few changes to add the distro to the values yaml's for the distro.

cmd/vclusterctl/cmd/create.go
```diff
@@ -44,12 +45,18 @@ Example:
 vcluster create test --namespace test
 #######################################################
 	`,
-		Args: util.VClusterNameOnlyValidator,
 		RunE: func(cobraCmd *cobra.Command, args []string) error {
+			newArgs, err := util.PromptForMissingArgsIfInteractive(cmd.log, args, "vcluster name")
```
ThomasK33, line 49: Nit: this function name seems excessive. 

Something like `PromptForMissingArgs` or even `PromptForArgs` should be acceptable, and the details about whether something is interactive are more of a doc comment than a name (plus what kind of errors it will return in which circumstances).

pkg/cli/util/args.go
```diff
@@ -14,6 +18,8 @@ var (
 	VClusterNameOnlyUseLine string
 
 	VClusterNameOnlyValidator cobra.PositionalArgs
+
+	ErrNonInteractive = errors.New("terminal is not interactive")
```
FabianKramm, line 22: I think this error is a little awkward when you forget to put an argument it will yield `terminal is not interactive` which is not really what you would want. It should say something like `please specify vcluster ... [VCLUSTER_NAME]` or something more understandable

FabianKramm (changes requested):

pkg/cli/util/args.go
FabianKramm: Nvm see you do the check in the caller

FabianKramm (approved):

ThomasK33 (approved):

pkg/cli/util/args.go
zerbitx: The way I'm calling it, it will report the same error it did before for none interactive usage since each of the three places its called it checks for `ErrNonInteractive` and then falls back to trying the validation that was there before.  I thought about adding a `cobra.PositionalArgs` to the arguments, so it could take care of that automatically, but thought it might be seen as overkill.  So like [this](https://github.com/zerbitx/vcluster/pull/1/files?w=1)

PR #2217, opened 2024-10-10
feat: require valid admin email for platform start

zerbitx:
/kind enhancement

resolves #ENG-4850


cmd/vclusterctl/cmd/platform/start.go
```diff
@@ -142,3 +149,31 @@ func (cmd *StartCmd) Run(ctx context.Context) error {
...
+	if ret.Disposable {
+		return fmt.Errorf(`disposable address "%s" not allowed`, email)
+	}
+
+	if !ret.HasMxRecords {
+		return fmt.Errorf(`address "%s" without MX record not allowed`, email)
+	}
+
+	return nil
+}
```
ThomasK33, line 179: This will perform SMTP or MX lookups from a CLI, potentially blocking folks in air-gapped or restricted networks from spinning up the platform.

I'd expect this logic in the platform code if connections can be established.
The CLI should only perform a quick sanity check using a regex or comparable mechanics.

ThomasK33 (changes requested):

FabianKramm (changes requested):
I'm heavily against adding 132k lines of code for verfying an email address

PR #2221, opened 2024-10-10
ENG-4867: Fixes platform create output so it prints internal vcluster create command instead of external

cbron:
/kind bugfix

resolves ENG-4867

Fixed an issue where `vcluster platform start` command printed the wrong type of vcluster create command


lizardruss (approved):
:tada:

PR #2224, opened 2024-10-11
feat: require valid admin email for platform start

zerbitx:
/kind enhancement

resolves #ENG-3850

> 💡 The go:geneate portion can be removed if its seen as "overkill", I threw it in to make updating the list easier in the future.

## Requirements from the issue:
If in interactive shell, then prompt for email address. With the email address:

Check if Valid Email - use a parser

Don't allow any disposable emails: Use this list to verify if it's a disposable one. This is in the existing email-verifier library. https://raw.githubusercontent.com/disposable/disposable-email-domains/master/domains.json 

Check if mx_record is found

Connect to see if it's valid with a 10 second timeout. If it doesn't come back in 10 seconds, it's considered invalid. 

If the validation step doesn't succeed, then keep asking the question. 

```vcluster platform start
By providing your email, you accept our Terms of Service and Privacy Statement: 
Terms of Service: https://www.loft.sh/legal/terms
Privacy Statement: https://www.loft.sh/legal/privacy
```


FabianKramm (changes requested):
This is a huge change as well, I would like to see how this impacts binary size and why such a huge dependency is needed

deniseschannon:
@FabianKramm It looks like most of this code changes are from packaging the disposable domains. That file alone is 169,412 lines of code.

deniseschannon:
@zerbitx - Let's do this. Let's replace the list with this much smaller one: https://github.com/disposable-email-domains/disposable-email-domains/blob/main/disposable_email_blocklist.conf

That will significantly decrease the size of the PR.

pkg/cli/email/validate.go
```diff
@@ -0,0 +1,154 @@
...
+	localPart, domainPart := parts[0], parts[1]
+
+	// Check if both the local part and domain part are non-empty
+	if len(localPart) == 0 || len(domainPart) == 0 {
+		return errors.New("missing parts")
+	}
+
+	// Validate the domain part contains a dot and has text on both sides
+	if strings.Count(domainPart, ".") == 0 || strings.HasPrefix(domainPart, ".") || strings.HasSuffix(domainPart, ".") {
+		return errors.New(`missing "."`)
```
johannesfrey, line 73: Not sure if I understood this correctly, but it seems that the comment does not match the implementation.
If it's about the dots should it be something like this:

```suggestion
  if strings.Count(domainPart, ".") == 0 {
      return errors.New(`missing "."`)
  }
  if strings.HasPrefix(domainPart, ".") || strings.HasSuffix(domainPart, ".") {
      return errors.New(`"." cannot be at the start or end of domain`)
  }
```

pkg/cli/email/validate.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+
+	// Check if the local part doesn't start or end with a dot
+	if strings.HasPrefix(localPart, ".") || strings.HasSuffix(localPart, ".") {
+		return fmt.Errorf(`cannot have "." at the beginning or end of "%s"`, emailAddress)
+	}
+
+	// Check that the local part doesn't contain consecutive dots
+	if strings.Contains(localPart, "..") {
```
johannesfrey, line 90: Wondering if this could also be used for the double dot check for the `domainPart`? :thinking:

pkg/cli/email/validate.go
```diff
@@ -0,0 +1,154 @@
...
+	defer cancel()
+	errc := make(chan error)
+
+	// Fire off a go routine for each record and SMTP port
+	// The first to return a nil error will return success
+	for _, record := range records {
+		for _, port := range smtpPorts {
+			go func() {
+				_, connectError := smtp.Dial(fmt.Sprintf("%s:%d", record.Host, port))
+				errc <- connectError
```
johannesfrey, line 139: Couldn't this lead to the other go routines being blocked by trying to send to errc, which isn't read anymore in case of a canceled context or the first nil err received?

zerbitx: Good catch, updated.

pkg/cli/email/validate.go
zerbitx: The RFC check catches this.  I just made it default instead of trying to do the smaller checks in case its not checked.

zerbitx:
@deniseschannon @FabianKramm I changed the list to be a zip file to make it smaller as an embed, using the condensed list suggested.

johannesfrey (approved):
lgtm!

lizardruss (approved):
lgtm

FabianKramm (approved):

PR #2228, opened 2024-10-15
fix: Added custom resources to host to pro check

ThomasK33:
/kind bugfix

Add custom resources to host syncing to pro check


hidalgopl (changes requested):
could you add a test case to https://github.com/loft-sh/vcluster/blob/main/config/config_test.go#L144 as well? Thanks!

johannesfrey (approved):

hidalgopl (approved):

PR #2229, opened 2024-10-15
fix: Separate volumesnapshot mappers into individual configurations

ThomasK33:
/kind bugfix
/kind enhancement
/kind feature
/kind documentation
/kind test

resolves #

### DIFF
--- a/pkg/mappings/resources/volumesnapshotclasses.go
+++ b/pkg/mappings/resources/volumesnapshotclasses.go
@@ -13,7 +13,7 @@ import (
 var volumeSnapshotClassesCRD string
 
 func CreateVolumeSnapshotClassesMapper(ctx *synccontext.RegisterContext) (synccontext.Mapper, error) {
-	if !ctx.Config.Sync.ToHost.VolumeSnapshots.Enabled {
+	if !ctx.Config.Sync.FromHost.VolumeSnapshotClasses.Enabled {
 		return generic.NewMirrorMapper(&volumesnapshotv1.VolumeSnapshotClass{})
 	}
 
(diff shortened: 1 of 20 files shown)


rmweir (approved):
LGTM

pkg/mappings/resources/volumesnapshotclasses.go
```diff
@@ -13,7 +13,7 @@ import (
 var volumeSnapshotClassesCRD string
 
 func CreateVolumeSnapshotClassesMapper(ctx *synccontext.RegisterContext) (synccontext.Mapper, error) {
-	if !ctx.Config.Sync.ToHost.VolumeSnapshots.Enabled {
+	if !ctx.Config.Sync.FromHost.VolumeSnapshotClasses.Enabled {
```
zerbitx, line 16: I'm not understanding why we check this?   Since it tries to return a NewMirrorMapper whether its enabled or not, if the EnsureCRD check is successful.  Is it to get the error from EnsureCRD instead of from `apiutil.GVKForObject(obj, scheme.Scheme)` used in NewMirrorMapper?

zerbitx (approved):
had a non-blocking question, otherwise lgtm

pkg/mappings/resources/volumesnapshotclasses.go
ThomasK33: The `sync.toHost.volumesnapShot` toggle was split into three different toggles in a previous pull request.

This change is mostly for consistency's sake so that the volumesnapshotclass mapper is controlled by its toggle.

PR #2233, opened 2024-10-16
fix: Only get PartialObjectMetadata in mappings GC

ThomasK33:
/kind bugfix

resolves #

Fixed an issue where vCluster mappings garbage collection would hang and cause the entire syncer process to deadlock.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -103,7 +103,7 @@ jobs:
       - id: set-paths-matrix
         run: |
           set -x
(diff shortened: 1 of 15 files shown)


FabianKramm (approved):

cbron:
Some stuff in here we'll need to address in future (ie: https://github.com/loft-sh/vcluster/pull/2233/files#diff-201bae62fde93bdc51a8720dd2f82e08a531df9a8d332a95090d14cf1de1091aR91), but merging now to cut a beta

PR #2241, opened 2024-10-22
Prevent recreating virtual cluster resource during delete

lizardruss:
/kind bugfix

resolves ENG-4924

Fixed an issue where the vcluster syncer would mistakenly re-create a resource when reconciling the update event that adds the `deletionTimestamp`

## Scenario
The syncer incorrectly re-creates a resource while it's being deleted. The sequence of events observed:
- A replicationcontroller with a `terminationGracePeriod` == 0 is deleted and its pods are garbage collected
- An update event that sets the deletionTimestamp is queued
- An delete event for the same resource is queued
- Syncer reconciles the update event but is unable to load the resource because it's already deleted. This triggers a call to `SyncToVirtual()`, which mistakenly re-creates the resource.

This scenario occurs more frequently with pods, but the solution has been made generic in case it occurs with different resource kinds.

## Solution
Add a `PendingDelete` event type to communicate that an update event occurred to a virtual resource being deleted. If the `SyncToVirtual` flow is reached by an event where the source == `Virtual` and the type == `PendingDelete`, then skip calling SyncToVirtual.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -169,9 +169,7 @@ build-dev-image tag="":
   docker build -t vcluster:dev-{{tag}} -f Dockerfile.release --build-arg TARGETARCH=$(uname -m) --build-arg TARGETOS=linux .
   rm ./vcluster
 
-run-conformance k8s_version="1.31.1" mode="conformance-lite" tag="conf": (build-dev-image tag)
-  minikube start --kubernetes-version {{k8s_version}} --nodes=2
-  minikube addons enable metrics-server
+run-conformance k8s_version="1.31.1" mode="conformance-lite" tag="conf": (create-conformance k8s_version) (build-dev-image tag)
   minikube image load vcluster:dev-{{tag}}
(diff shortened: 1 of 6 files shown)


Justfile
```diff
@@ -183,3 +181,15 @@ conformance-status:
...
+
+dev-conformance *ARGS:
+  devspace dev --profile test-conformance --namespace vcluster {{ARGS}}
+
+create-conformance k8s_version="1.31.1":
+  minikube start --kubernetes-version {{k8s_version}} --nodes=2
+  minikube addons enable metrics-server
+
+delete-conformance:
+  minikube delete || true
```
ThomasK33, line 193: Just can [ignore errors](https://github.com/casey/just#ignoring-errors) by prefixing the line with a dash

```suggestion
  -minikube delete
```

ThomasK33 (approved):

PR #2246, opened 2024-10-25
reduce concurrent reconciles of the same virtual resource

lizardruss:
/kind bugfix

resolves ENG-4924

Fixed an issue where vcluster incorrectly re-created pods during garbage collection

The previous method of passing through the `syncEventType` would potentially enqueue the same object 4 different ways:
- virtual object: `ns/obj`
- delete virtual object: `delete#ns/obj`
- physical object: `host#ns/obj`
- delete physical object: `delete#host#ns/obj`

Usually this would not be a concern, but the following scenario occurs frequently enough to cause a conformance test to fail:
1. virtual object is deleted
1. a update event for the virtual object is received that adds the `deletionTimestamp`
1. a delete event for the virtual object is received
1. these events would enqueue the object as objects with the same name in different namespaces from the controller runtime's perspective
1. the two objects are reconciled concurrently by the controller-runtime
1. while the `Reconcile` function is processing the virtual object update the virtual object is deleted, this leads to the `SyncToVirtual` implementation incorrectly re-creating the virtual object

This PR attempts to solve this by:
- Reducing the amount of concurrent reconciles for the same virtual object by dropping the `delete#` prefix
- Using a different method of passing the `syncEventType` through to `Reconcile` (and the various sync event types). Event types are recorded by virtual object UID, only the last event is saved, and a delete event for a UID cannot be overwritten.

### DIFF
--- a/pkg/syncer/handler.go
+++ b/pkg/syncer/handler.go
@@ -3,6 +3,7 @@ package syncer
 import (
 	"context"
 
+	"github.com/loft-sh/vcluster/pkg/syncer/synccontext"
 	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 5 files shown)


FabianKramm:
Closing this since we went for another solution

PR #2258, opened 2024-11-01
refactor: use object cache instead of event info

FabianKramm:
/kind bugfix
/kind enhancement

resolves ENG-4997

### This PR does the following things:
* Removes the host event & delete event within syncer
* Introduces an optional object cache syncers can use to save the before state of objects, which helps to determine what exactly has changed on the other side
* Moves the patch & apply from vCluster pro into oss
* Refactors all syncers that were using the "old" bidirectional syncing approach to use the new one


pkg/syncer/syncer.go
```diff
@@ -444,11 +362,14 @@ func (r *SyncController) enqueueVirtual(_ context.Context, obj client.Object, q
 	})
 }
 
-func (r *SyncController) enqueuePhysical(ctx context.Context, obj client.Object, q workqueue.TypedRateLimitingInterface[ctrl.Request], isDelete, _ bool) {
+func (r *SyncController) enqueuePhysical(ctx context.Context, obj client.Object, q workqueue.TypedRateLimitingInterface[ctrl.Request], isDelete bool) {
 	if obj == nil {
 		return
 	}
 
+	// add object to cache if it's not there yet
```
lizardruss, line 370: Comment is misleading now, and this line could be moved down next to `vReq`

zerbitx (approved):

pkg/controllers/resources/services/syncer.go
```diff
@@ -117,40 +118,93 @@ func (s *serviceSyncer) Sync(ctx *synccontext.SyncContext, event *synccontext.Sy
...
 	// remove the ServiceBlockDeletion annotation if it's not needed
 	if event.Virtual.Spec.ClusterIP == event.Host.Spec.ClusterIP {
 		delete(event.Host.Annotations, ServiceBlockDeletion)
 	}
 
 	// translate selector
-	if event.Source == synccontext.SyncEventSourceHost {
-		event.Virtual.Spec.Selector = translate.VirtualLabelsMap(event.Host.Spec.Selector, event.Virtual.Spec.Selector)
-	} else {
+	// TODO: ryan - convert to bidirectional
```
zerbitx, line 203: We can remove this todo

pkg/controllers/resources/priorityclasses/syncer_test.go
```diff
@@ -88,9 +88,10 @@ func TestSyncToHost(t *testing.T) {
 
 			syncCtx, syncer := syncertesting.FakeStartSyncer(t, registerContext, New)
 			event := synccontext.NewSyncToHostEvent(vObj.DeepCopy())
-			if tC.isDelete {
-				event.Type = synccontext.SyncEventTypeDelete
-			}
+			// TODO: fix this
```
lizardruss, line 91: Do we need to address these TODOs or can they be removed?

lizardruss (commented):
Looks like this was merged already, but some small comments

PR #2259, opened 2024-11-01
Exclude rancher managed annotation from ingress sync

neogopher:
/kind feature

resolves ENG-4976

vcluster will now ignore updates to Rancher managed annotations when syncing ingress to & from the virtual cluster.


lizardruss (approved):
lgtm

PR #2260, opened 2024-11-01
Fix unit tests for sync-refactoring

lizardruss:
/kind test

resolves ENG-5033

Updating unit tests after the syncer refactoring

### DIFF
--- a/pkg/controllers/resources/services/translate.go
+++ b/pkg/controllers/resources/services/translate.go
@@ -50,3 +50,73 @@ func StripNodePorts(vObj *corev1.Service) {
 		vObj.Spec.Ports[i].NodePort = 0
 	}
 }
+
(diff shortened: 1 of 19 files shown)


pkg/controllers/resources/services/translate.go
```diff
@@ -50,3 +50,73 @@ func StripNodePorts(vObj *corev1.Service) {
 		vObj.Spec.Ports[i].NodePort = 0
 	}
 }
+
+// AlignSpecWithServiceType removes any fields that are invalid for the specific service type
+func AlignSpecWithServiceType(svc *corev1.Service) {
```
zerbitx, line 55: I added this to fix a unit test that expected the externalName to be there, but it seems to be breaking conformance tests, so it should probably be removed and just adjust the test expectation instead.

PR #2263, opened 2024-11-04
fix: lost host event information

FabianKramm:
/kind bugfix

Fixes ENG-5083

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/loft-sh/vcluster/pkg/config"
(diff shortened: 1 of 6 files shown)


pkg/syncer/syncer.go
```diff
@@ -108,7 +114,15 @@ func (r *SyncController) newSyncContext(ctx context.Context, logName string) *sy
 	}
 }
 
-func (r *SyncController) Reconcile(ctx context.Context, vReq ctrl.Request) (_ ctrl.Result, retErr error) {
+func (r *SyncController) Reconcile(ctx context.Context, vReq reconcile.Request) (_ ctrl.Result, retErr error) {
+	// extract request
+	r.hostNameRequestLookupLock.Lock()
+	pReq, ok := r.hostNameRequestLookup[vReq]
+	if ok {
+		delete(r.hostNameRequestLookup, vReq)
```
zerbitx, line 122: Should we wait to delete until after we have a successful reconciliation? 

```go
	pReq, ok := r.hostNameRequestLookup[vReq]

	// create sync context
	syncContext := r.newSyncContext(ctx, vReq.Name)
	defer func(founcHostEvent bool) {
		if err := syncContext.Close(); err != nil {
			retErr = errors.Join(retErr, err)
		}

		if retErr == nil && founcHostEvent {
			delete(r.hostNameRequestLookup, vReq)
		}
		r.hostNameRequestLookupLock.Unlock()
	}(ok)
```

FabianKramm: Oh yeah good point

PR #2268, opened 2024-11-08
fix: Skip trying to dial an MX record's domain

zerbitx:
/kind bugfix

resolves #ENG-5104

Some ISPs block the SMTP ports so trying to dial will always fail even though the email is valid.  This changes it to accept the email address as long as the domain returns at least one MX record.


lizardruss (approved):
lgtm

PR #2273, opened 2024-11-10
Apply exportKubeconfig server setting only to additional kubeconfig secret

### LINKED ISSUE #2234: Setting server in exportKubeConfig overwrites default secret
### What happened?

When using `exportKubeConfig` to create an additional kubeconfig secret with a new server, it seems that both the new secret and the default (`vc-NAME`) secret get updated with the new server.

Unless I'm misunderstanding the docs, it seems like the `exportKubeConfig` configuration should only update the additional secret?
> vCluster always creates a kubeconfig in a secret called vc-NAME in the namespace where you deployed vCluster. Configure exportKubeConfig so vCluster creates an additional secret with the given configuration.

from: https://www.vcluster.com/docs/vcluster/configure/vcluster-yaml/export-kube-config

I would expect the default `vc-NAME` kubeconfig secret to still be configured with the default localhost:8443 server.

Create vcluster with following vcluster.yaml:
```
$ vcluster create cluster-a --namespace team-a --values vcluster.yaml
```
```
exportKubeConfig:
  server: https://cluster-a
  secret:
    name: vc-cluster-a-new
```

New kubeconfig secret (`vc-cluster-a-new`) is created as expected:
```
$ kubectl --namespace team-a get secret vc-cluster-a-new --template={{.data.config}} | base64 -d
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: ...
    server: https://cluster-a
  name: kubernetes
...
```

However the default secret's (`vc-cluster-a`) localhost:8443 seems to also have been overridden:
```
$ kubectl --namespace team-a get secret vc-cluster-a --template={{.data.config}} | base64 -d
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: ...
    server: https://cluster-a
  name: kubernetes
...
```

For context, I'm trying to have two versions of the kubeconfig, one with localhost that can be used to connect to the cluster via port forwarding and the other that can be used from within the host cluster for something like Flux to make API calls to the vcluster's apiserver. I could manually create the second secret, but would be nice of `exportKubeConfig` could be used.

<details>

```console
$ kubectl version
Client Version: v1.31.1
Kustomize Version: v5.4.2
Server Version: v1.31.0
```

</details>

### vcluster version

<details>

```console
$ vcluster --version
vcluster version 0.20.1
```

</details>

### VCluster Config

<details>

```
exportKubeConfig:
  server: https://cluster-a
  secret:
    name: vc-cluster-a-new
```

</details>

RokibulHasan7:
/kind bugfix

resolves #2234

Fixed issue where custom server setting in `exportKubeConfig` was overwriting the default kubeconfig secret.

### DIFF
--- a/pkg/setup/controllers.go
+++ b/pkg/setup/controllers.go
@@ -210,24 +210,23 @@ func WriteKubeConfigToSecret(ctx context.Context, virtualConfig *rest.Config, cu
 		return err
 	}
 
-	// should use special server?
+	var customSyncerConfig *clientcmdapi.Config
 	if options.ExportKubeConfig.Server != "" {
-		// exchange kube config server & resolve certificate
-		for key, cluster := range syncerConfig.Clusters {
-			if cluster == nil {
-				continue
-			}
-
-			syncerConfig.Clusters[key] = &clientcmdapi.Cluster{
-				Server:                   options.ExportKubeConfig.Server,
-				Extensions:               make(map[string]runtime.Object),
-				CertificateAuthorityData: cluster.CertificateAuthorityData,
-				InsecureSkipTLSVerify:    options.ExportKubeConfig.Insecure,
+		// Create a deep copy of syncerConfig to modify the server for the additional secret
+		customSyncerConfig = syncerConfig.DeepCopy()
+		for key, cluster := range customSyncerConfig.Clusters {
+			if cluster != nil {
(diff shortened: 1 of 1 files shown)


FabianKramm (approved):
@RokibulHasan7 thanks for creating this PR! LGTM!

PR #2289, opened 2024-11-21
feat: Render control plane ingress spec as template in helm chart

### LINKED ISSUE #1760: Consider making charts more dynamic via use of tpl

Not a problem but a need and having seen more productivity in other contexts.
Bitnami charts make heavy use of the "tpl" function to allow many/most/all evaluations to be dynamic.
This would also reduce the need to resort to other tools (namely gitops or other ways of evaluating values at deploy time).

Convert the new v0.20.x chart (or on a follow up version) to use tpl when evaluating variables in the various templates.
the bitnami.common chart already includes a "common.tplvalues.render" template/function.

Using combinations of gitops, external tools (ansible, bash), or variables in tools like vcluster or other TACOS. All of them increase complexity, even for simple/trivial scenarios.

### Additional context

Examples on no particular order:
* https://github.com/bitnami/charts/blob/main/bitnami/zookeeper/templates/statefulset.yaml#L15
* https://github.com/bitnami/charts/blob/main/bitnami/keycloak/templates/admin-ingress.yaml#L31
* https://github.com/kubernetes/dashboard/blob/main/charts/kubernetes-dashboard/templates/security/pdb.yaml#L22
* https://github.com/seaweedfs/seaweedfs/blob/master/k8s/charts/seaweedfs/templates/master-statefulset.yaml#L226

garryod:
/kind enhancement

partially resolves #1760 by rendering the `.Values.controlPlane.ingress.spec` as a helm template when substituting it into the ingress template

Allows use of templates in chart `.Values.controlPlane.ingress.spec`


FabianKramm (commented):
@garryod thanks a lot for the PR! LGTM, but would you mind adding a helm unit test for this so we got that new functionality covered? Also seems some unit tests are failing, would be nice if you could take a look

chart/templates/ingress.yaml
```diff
@@ -18,8 +18,8 @@ metadata:
 {{ toYaml .Values.controlPlane.ingress.labels | indent 4 }}
   {{- end }}
 spec:
-  {{- if .Values.controlPlane.ingress.spec }}
-{{ toYaml .Values.controlPlane.ingress.spec | indent 2 }}
+  {{- with .Values.controlPlane.ingress.spec }}
+  {{ tpl (toYaml .) $ | indent 2 }}
```
FabianKramm, line 22: Looks like changing the indent here causes issues

garryod:
> @garryod thanks a lot for the PR! LGTM, but would you mind adding a helm unit test for this so we got that new functionality covered? Also seems some unit tests are failing, would be nice if you could take a look

Sure thing, should all be resolved now

FabianKramm (approved):

FabianKramm:
@garryod thanks, LGTM!

PR #2297, opened 2024-11-25
bugfix: set correct etcdEndpoints for non embedded etcd based on vcluster config

neogopher:
/kind bugfix

resolves ENG-5249

Fixed an issue where vcluster would always set etcd endpoints to the regular etcd service instead of based on config values

When non-embedded etcd is used for the backing store, we have both the regular as well as a headless service pointing to the etcd statefulset, for use by default. However, if one wishes to use just the headless service and disable the regular one, it is not possible, as the regular service name is hardcoded and referred-to by the etcd clients in the code. 

With this PR, we are setting the etcdEndpoints in the code appropriately after checking the fields in the vcluster values. Thus, the etcd clients will now refer to - 
- the regular service if enabled (via .Values.controlPlane.backingStore.etcd.deploy.service.enabled) or else 
- the headless service as a fallback/default option.

### DIFF
--- a/chart/templates/etcd-headless-service.yaml
+++ b/chart/templates/etcd-headless-service.yaml
@@ -1,6 +1,5 @@
 {{- if not .Values.experimental.isolatedControlPlane.headless }}
 {{- if or .Values.controlPlane.backingStore.etcd.deploy.enabled (include "vcluster.etcd.embedded.migrate" .) }}
-{{- if .Values.controlPlane.backingStore.etcd.deploy.headlessService.enabled }}
 apiVersion: v1
 kind: Service
 metadata:
@@ -31,6 +30,5 @@ spec:
   selector:
     app: vcluster-etcd
     release: "{{ .Release.Name }}"
-{{- end }}
 {{- end }}
 {{- end }}
--- a/pkg/etcd/client.go
+++ b/pkg/etcd/client.go
@@ -59,7 +59,7 @@ func NewFromConfig(ctx context.Context, vConfig *config.VirtualClusterConfig) (C
 			etcdEndpoints = "https://127.0.0.1:2379"
 		} else if vConfig.ControlPlane.BackingStore.Etcd.Deploy.Service.Enabled {
 			etcdEndpoints = "https://" + vConfig.Name + "-etcd:2379"
-		} else if vConfig.ControlPlane.BackingStore.Etcd.Deploy.HeadlessService.Enabled {
+		} else {
 			etcdEndpoints = "https://" + vConfig.Name + "-etcd-headless:2379"
 		}
 	} else if vConfig.Distro() == vconfig.K8SDistro {
--- a/pkg/etcd/client.go
+++ b/pkg/etcd/client.go
@@ -57,8 +57,10 @@ func NewFromConfig(ctx context.Context, vConfig *config.VirtualClusterConfig) (C
 
 		if vConfig.ControlPlane.BackingStore.Etcd.Embedded.Enabled {
 			etcdEndpoints = "https://127.0.0.1:2379"
-		} else {
+		} else if vConfig.ControlPlane.BackingStore.Etcd.Deploy.Service.Enabled {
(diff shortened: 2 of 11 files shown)


pkg/etcd/client.go
```diff
@@ -57,8 +57,10 @@ func NewFromConfig(ctx context.Context, vConfig *config.VirtualClusterConfig) (C
 
 		if vConfig.ControlPlane.BackingStore.Etcd.Embedded.Enabled {
 			etcdEndpoints = "https://127.0.0.1:2379"
-		} else {
+		} else if vConfig.ControlPlane.BackingStore.Etcd.Deploy.Service.Enabled {
 			etcdEndpoints = "https://" + vConfig.Name + "-etcd:2379"
+		} else if vConfig.ControlPlane.BackingStore.Etcd.Deploy.HeadlessService.Enabled {
```
ThomasK33, line 62: Can a scenario exist in which `Deploy.Enable` is true, but neither `Service.Deploy` nor `HeadlessService.Enabled` is true?

pkg/k3s/k3s.go
```diff
@@ -51,17 +51,24 @@ func StartK3S(ctx context.Context, vConfig *config.VirtualClusterConfig, service
 			args = append(args, "--kube-apiserver-arg=endpoint-reconciler-type=none")
 		}
 		if vConfig.ControlPlane.BackingStore.Etcd.Deploy.Enabled {
+			var etcdEndpoints string
```
ThomasK33, line 54: NIT: Given that we only have a single endpoint here, this might become confusing.

```suggestion
			var etcdEndpoint string
```

pkg/etcd/client.go
neogopher: Yes, It is possible to reach this situation when both the above fields are set to disabled, and subsequently this would lead to a critical failure. 

On a related note, I have another thought. Given that the headless service is the [governing service](https://github.com/loft-sh/vcluster/blob/94e59b63d9e3bc82c0c55b2ffaed35ea4881f9d4/chart/templates/etcd-statefulset.yaml#L26) for the etcd statefulset and is required to be present whenever the statefulset exists, does it then make more sense to remove the `headlessservice.enabled` field and have its creation defined primarily by `Deploy.Enable`. This will prevent it from being manually disabled by someone like above. Also, this will further simplify all the checks I added in this PR, where we will have the headless version as the last/fallback option under the else clause.

pkg/k3s/k3s.go
neogopher: I just tried to match it with the name of the function parameter where it is used and also other [places](https://github.com/neogopher/vcluster/blob/ce4030a9679d4880f73a1884e1abf0857429a489/pkg/etcd/client.go#L66-L67) where it is exists as of today, but I agree it is largely singular in our code.

Have updated just the ones I added.

chart/templates/etcd-headless-service.yaml
```diff
@@ -1,6 +1,5 @@
 {{- if not .Values.experimental.isolatedControlPlane.headless }}
 {{- if or .Values.controlPlane.backingStore.etcd.deploy.enabled (include "vcluster.etcd.embedded.migrate" .) }}
-{{- if .Values.controlPlane.backingStore.etcd.deploy.headlessService.enabled }}
```
FabianKramm, line 3: Whats the reasoning behind removing this option? Sounds like that was not interfering with what you are trying to change.

FabianKramm (approved):

chart/templates/etcd-headless-service.yaml
neogopher: This was removed to prevent someone from manually disabling it and thus causing the headless service to not be created and subsequently hampering the etcd statefulset's working. We discussed it [here](https://github.com/loft-sh/vcluster/pull/2297#discussion_r1858750161).

pkg/etcd/client.go
neogopher: Changes are done now.

PR #2317, opened 2024-12-05
fix(cli): add --force-remove-finalizers to platform destroy

rohantmp:
/kind enhancement

resolves ENG-5245
resolves ENG-5264

Adds prompt for this flag, also changes logging to mention that externally deployed vclusters are not cleaned up.

### DIFF
--- a/cmd/vclusterctl/cmd/platform/destroy.go
+++ b/cmd/vclusterctl/cmd/platform/destroy.go
@@ -107,7 +107,7 @@ func (cmd *DestroyCmd) Run(ctx context.Context) error {
 	if !cmd.NonInteractive {
 		deleteOpt := "delete"
 		out, err := cmd.Log.Question(&survey.QuestionOptions{
-			Question: fmt.Sprintf("IMPORTANT! You are destroying the vCluster Platform installation in the namespace %q.\nThis may result in data loss. Please ensure your kube-context is pointed at the right cluster.\n Please type %q to continue:", cmd.Namespace, deleteOpt),
+			Question: fmt.Sprintf("IMPORTANT! You are destroying the vCluster Platform installation in the namespace %q.\n  This may result in data loss.\n  Externally deployed virtual clusters depending on an external database connection will be irrecoverable after the platform is destroyed.\n  Please ensure your kube-context is pointed at the right cluster.\nPlease type %q to continue:", cmd.Namespace, deleteOpt),
 		})
 		if err != nil {
 			return fmt.Errorf("failed to prompt for confirmation: %w", err)
--- a/cmd/vclusterctl/cmd/platform/destroy.go
+++ b/cmd/vclusterctl/cmd/platform/destroy.go
@@ -49,7 +49,7 @@ before running this command:
 2. Helm v3 must be installed
 
 
-VirtualClusterInstances managed with driver helm will be deleted, but the underlying virtual cluster will not be uninstalled
+VirtualClusterInstances managed with driver helm will be deleted, but the underlying virtual cluster will not be uninstalled.
 
 ########################################################
 	`,
@@ -65,7 +65,8 @@ VirtualClusterInstances managed with driver helm will be deleted, but the underl
 	destroyCmd.Flags().BoolVar(&cmd.IgnoreNotFound, "ignore-not-found", false, "Exit successfully if platform installation is not found")
 	destroyCmd.Flags().BoolVar(&cmd.Force, "force", false, "Try uninstalling even if the platform is not installed. '--namespace' is required if true")
 	destroyCmd.Flags().BoolVar(&cmd.NonInteractive, "non-interactive", false, "Will not prompt for confirmation")
-	destroyCmd.Flags().IntVar(&cmd.TimeoutMinutes, "timeout-minutes", 5, "How long to try deleting the platform before giving up")
+	destroyCmd.Flags().IntVar(&cmd.TimeoutMinutes, "timeout-minutes", 5, "How long to try deleting the platform before giving up. May increase when removing finalizers if --remove-finalizers is used")
+	destroyCmd.Flags().BoolVar(&cmd.ForceRemoveFinalizers, "force-remove-finalizers", false, "IMPORTANT! Removing finalizers may cause unintended behaviours like leaving resources behind, but will ensure the platform is uninstalled.")
 
 	return destroyCmd
 }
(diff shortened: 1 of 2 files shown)


rmweir (changes requested):
I have additional thoughts about cleanup of "external" virtual clusters that I am asking about and will finish reviewing once those are resolved.

pkg/cli/destroy/destroy.go
```diff
@@ -240,38 +256,101 @@ func Destroy(ctx context.Context, opts DeleteOptions) error {
...
+						log.Infof("deleting %v: %q", resource, namespacedName)
+					} else {
+						log.Infof("deleting externally deployed %v, the virtual cluster itself will remain: %q", resource, namespacedName)
+					}
+					err := resourceClient.Namespace(object.GetNamespace()).Delete(ctx, object.GetName(), metav1.DeleteOptions{})
+					if kerrors.IsNotFound(err) {
+						continue
+					} else if err != nil {
+						return false, err
+					}
```
rmweir, line 322: Spoke to Lukas regarding external cluster that still may be irreparably affected by deletion. At the moment this only is cluster that use db connector and it will not be common for new features to have a similar effect. So we can make another exception here and provide a prompt. You can see if a vcluster is using a db connector by checking if there are any secrets with the `loft.sh/database-vcluster` set to `virtualClusterInstance.Status.ServiceUID`. If a vcluster is external _and_ has one of these secrets, provide a prompt with wording similar to "Warning: `<vclustername>` is an external cluster but has a datastore managed by platform. Deleting this will destroy your virtual cluster's datastore and it will no longer be usable as the vcluster pod will enter a bad state."

rmweir (changes requested):

rohantmp:
With changes: 
image

pkg/cli/destroy/destroy.go
rohantmp: Added the changes!

rohantmp: However, I parsed the object's config instead of searching for secrets

cmd/vclusterctl/cmd/platform/destroy.go
```diff
@@ -115,6 +116,19 @@ func (cmd *DestroyCmd) Run(ctx context.Context) error {
 			cmd.Log.Info("destroy cancelled")
 			return nil
 		}
+		if cmd.ForceRemoveFinalizers {
+			forceRemoveOpt := "force-remove"
+			out, err := cmd.Log.Question(&survey.QuestionOptions{
+				Question: fmt.Sprintf("IMPORTANT! You have selected the --force-remove-finalizers option. Please make sure you know what you're doing. Removing finalizers may cause unintended behaviours like leaving resources behind, but will ensure the platform is uninstalled. To confirm, please type %q", forceRemoveOpt),
```
cbron, line 122:
```suggestion
				Question: fmt.Sprintf("IMPORTANT! You have selected the --force-remove-finalizers option. Please ensure you understand the consequences. Removing finalizers may cause unintended behaviours like leaving resources behind, but will ensure the platform is uninstalled. To confirm, please type %q", forceRemoveOpt),
```

cmd/vclusterctl/cmd/platform/destroy.go
```diff
@@ -65,7 +65,8 @@ VirtualClusterInstances managed with driver helm will be deleted, but the underl
 	destroyCmd.Flags().BoolVar(&cmd.IgnoreNotFound, "ignore-not-found", false, "Exit successfully if platform installation is not found")
 	destroyCmd.Flags().BoolVar(&cmd.Force, "force", false, "Try uninstalling even if the platform is not installed. '--namespace' is required if true")
 	destroyCmd.Flags().BoolVar(&cmd.NonInteractive, "non-interactive", false, "Will not prompt for confirmation")
-	destroyCmd.Flags().IntVar(&cmd.TimeoutMinutes, "timeout-minutes", 5, "How long to try deleting the platform before giving up")
+	destroyCmd.Flags().IntVar(&cmd.TimeoutMinutes, "timeout-minutes", 5, "How long to try deleting the platform before giving up. May increase when removing finalizers if --remove-finalizers is used")
+	destroyCmd.Flags().BoolVar(&cmd.ForceRemoveFinalizers, "force-remove-finalizers", false, "IMPORTANT! Removing finalizers may cause unintended behaviours like leaving resources behind, but will ensure the platform is uninstalled.")
```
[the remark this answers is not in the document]
rohantmp: Hmm, if you use `--force`, it wouldn't skip that prompt as it is now. You have to be more specific with non-interactive.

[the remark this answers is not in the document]
rohantmp: IMO, the finalizers part is important here because of the nuance between force removing, and force removing finalizers. Removing finalizers is sometimes the opposite of a true `--force`.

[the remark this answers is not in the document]
cbron: Was that last comment supposed to have something besides the quote ?

I didn't realize there was a `non-interactive` option, so that would be used in CI/OPS environments, which is what I was mostly thinking of.  And I suppose Ricardo's feature does use a finalizer, even though it's a separate feature.

pkg/cli/destroy/destroy.go
```diff
@@ -240,38 +258,127 @@ func Destroy(ctx context.Context, opts DeleteOptions) error {
...
+					isExternalVCluster = virtualClusterInstance.Spec.External
+				}
+				if isExternalVCluster && virtualClusterInstance.Status.VirtualCluster != nil {
+					vConfig := &config.Config{}
+					err = config.UnmarshalYAMLStrict([]byte(virtualClusterInstance.Status.VirtualCluster.HelmRelease.Values), vConfig)
+					if err != nil {
+						return false, fmt.Errorf("failed to unmarshal virtual cluster config for %v %q: %w", resource, namespacedName, err)
+					}
+					if !nonInteractive {
+						if vConfig.ControlPlane.BackingStore.Database.External.Enabled && vConfig.ControlPlane.BackingStore.Database.External.Connector != "" {
```
rmweir, line 319: Sorry one mroe thing, you should still check this and report it if its non-interactive. Just without the "Do you want to continue?"

rmweir (changes requested):

pkg/cli/destroy/destroy.go
rmweir: Also, it look like if its not non interactive you and not using external connector, you don't show any warning. I think you want something like this:
```
if vConfig.ControlPlane.BackingStore.Database.External.Enabled && vConfig.ControlPlane.BackingStore.Database.External.Connector != "" {
	if !nonInteractive {
		yesOpt := "yes"
		noOpt := "no"
		out, err := log.Question(&survey.QuestionOptions{
			Options:  []string{yesOpt, noOpt},
			Question: fmt.Sprintf("IMPORTANT! You are removing an externally deployed virtual cluster %q from the platform.\n It will not be destroyed as the deployment is managed externally, but its connection to its database will be removed.\nDo you want to continue?", namespacedName),
		})
		if err != nil {
			return false, fmt.Errorf("failed to prompt for confirmation: %w", err)
		}
		if out != yesOpt {
			return false, fmt.Errorf("destroy cancelled during prompt")
		}
	}
} else {
	log.Warnf("removing an externally deployed virtual cluster %q from the platform. It will not be destroyed as the deployment is managed externally, but its connection to its database will be removed.", namespacedName)
}
```

cbron: Fixed in latest commit

rmweir (approved):
LGTM

PR #2329, opened 2024-12-13
vcluster config convert statefulSet.affinity bugfix

hidalgopl:
/kind bugfix

resolves ENG-5301

Fixed an issue where vcluster convert config skipped statefulset affinity rules

### DIFF
--- a/config/diff.go
+++ b/config/diff.go
@@ -12,6 +12,14 @@ import (
 // ErrUnsupportedType is returned if the type is not implemented
 var ErrUnsupportedType = errors.New("unsupported type")
 
+// includeNilValuesForKeys prevents removing certain keys with null values.
+// This is useful when default value for a given field is not nil, user set this field for null in oldConfig,
+// and we want to keep this field as null in the newConfig.
+var includeNilValuesForKeys = map[string]struct{}{
+	"ephemeral-storage":          {},
+	"requests.ephemeral-storage": {},
+}
+
 func Diff(fromConfig *Config, toConfig *Config) (string, error) {
 	// convert to map[string]interface{}
 	fromRaw := map[string]interface{}{}
@@ -112,7 +120,8 @@ func prune(in interface{}) interface{} {
 
 		for k, v := range inType {
 			inType[k] = prune(v)
-			if inType[k] == nil {
+			_, ok := includeNilValuesForKeys[k]
+			if inType[k] == nil && !ok {
 				delete(inType, k)
 			}
 		}
--- a/config/diff.go
+++ b/config/diff.go
@@ -12,14 +12,6 @@ import (
 // ErrUnsupportedType is returned if the type is not implemented
 var ErrUnsupportedType = errors.New("unsupported type")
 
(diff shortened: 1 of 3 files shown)


config/diff.go
```diff
@@ -12,6 +12,14 @@ import (
 // ErrUnsupportedType is returned if the type is not implemented
 var ErrUnsupportedType = errors.New("unsupported type")
 
+// includeNilValuesForKeys prevents removing certain keys with null values.
+// This is useful when default value for a given field is not nil, user set this field for null in oldConfig,
+// and we want to keep this field as null in the newConfig.
+var includeNilValuesForKeys = map[string]struct{}{
```
FabianKramm, line 18: This looks not correct to me. Why is this only a problem for ephemeral-storage and not for cpu and memory? Also I don't think this is the right way to do it, since it seems this is more a logic problem that values that are nil before are now not there anymore, which should be solved through a generic solution

config/diff.go
```diff
@@ -112,7 +120,8 @@ func prune(in interface{}) interface{} {
 
 		for k, v := range inType {
 			inType[k] = prune(v)
-			if inType[k] == nil {
+			_, ok := includeNilValuesForKeys[k]
```
FabianKramm, line 123: Same as above, I don't think we should check for specific keys in a generic function, it seems more like we need to solve this in a generic way

FabianKramm (changes requested):

config/diff.go
hidalgopl: ok, generic solution would be removing this block:
```
if inType[k] == nil && !ok {
				delete(inType, k)
			}
```. Then, we would output `null` for options that are set to nil. PTAL, changed it and adjusted unit tests.

File: config/legacyconfig/migrate_test.go [hunk 3]
```diff
@@ -89,6 +104,9 @@ plugin:
   statefulSet:
     scheduling:
       podManagementPolicy: OrderedReady
+external:
+  platform:
+    apiKey: null
```
FabianKramm, line 109: That also doesn't look correct to me, why is this now always null? Is that expected?

config/diff.go
```diff
@@ -112,11 +112,7 @@ func prune(in interface{}) interface{} {
 
 		for k, v := range inType {
 			inType[k] = prune(v)
-			if inType[k] == nil {
-				delete(inType, k)
-			}
```
FabianKramm, line 117: Whats the reasoning behind removing this? This leads to a lot more fields showing up as "null" which is not what we want.

FabianKramm (changes requested):
See comments above

config/diff.go
hidalgopl: This prunes null values, but there are cases when we want to keep them (e.g. originally reported issue when user wants to set ephemeral storage limit to nil).
I agree that we should prune some of the keys - e.g. `external` key if it the value is an empty map.

Reverted to keep deleting keys when the value is e.g. empty map, but to keep them if the value before pruning is nil. Please take another look

FabianKramm (approved):

PR #2341, opened 2024-12-18
fix: avoid unnecessary proxy container creation for already connected vCluster

RokibulHasan7:
/kind bugfix

resolves #

This PR adds a check to prevent the creation of a background proxy container when already connected to the specified vCluster. For example, if the command `vcluster connect vcluster-test` is run while already connected to `vcluster-test`, the unnecessary proxy container will no longer be created.

### DIFF
--- a/pkg/cli/connect_helm.go
+++ b/pkg/cli/connect_helm.go
@@ -93,7 +93,8 @@ func (cmd *connectHelm) connect(ctx context.Context, vCluster *find.VCluster, co
 		return err
 	}
 	if currentContext == find.VClusterContextName(vCluster.Name, vCluster.Namespace, vCluster.Context) {
-		return fmt.Errorf("already connected to vcluster %q", vCluster.Name)
+		cmd.Log.Infof("already connected to vcluster %q", vCluster.Name)
+		return nil
 	}
 
 	// prepare clients and find vcluster
--- a/pkg/cli/connect_helm.go
+++ b/pkg/cli/connect_helm.go
@@ -88,17 +88,13 @@ func ConnectHelm(ctx context.Context, options *ConnectOptions, globalFlags *flag
 }
 
 func (cmd *connectHelm) connect(ctx context.Context, vCluster *find.VCluster, command []string) error {
-	currentContext, _, err := find.CurrentContext()
-	if err != nil {
-		return err
-	}
-	if currentContext == find.VClusterContextName(vCluster.Name, vCluster.Namespace, vCluster.Context) {
+	if connected, _ := checkIfAlreadyConnected(ctx, vCluster); connected {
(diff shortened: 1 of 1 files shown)


pkg/cli/connect_helm.go
```diff
@@ -88,8 +88,16 @@ func ConnectHelm(ctx context.Context, options *ConnectOptions, globalFlags *flag
 }
 
 func (cmd *connectHelm) connect(ctx context.Context, vCluster *find.VCluster, command []string) error {
+	currentContext, _, err := find.CurrentContext()
+	if err != nil {
+		return err
+	}
+	if currentContext == find.VClusterContextName(vCluster.Name, vCluster.Namespace, vCluster.Context) {
+		return fmt.Errorf("already connected to vcluster %q", vCluster.Name)
```
FabianKramm, line 96: I'm not sure if printing an error here is the right choice as that would prevent this correct:
```
vcluster connect vcluster => works
vcluster connect vcluster => error
```
I think it's fine to not do certain things on the second connect but I think the command should suceed as otherwise this leads to a worse UX for the user

FabianKramm (changes requested):

pkg/cli/connect_helm.go
```diff
@@ -88,8 +88,17 @@ func ConnectHelm(ctx context.Context, options *ConnectOptions, globalFlags *flag
 }
 
 func (cmd *connectHelm) connect(ctx context.Context, vCluster *find.VCluster, command []string) error {
+	currentContext, _, err := find.CurrentContext()
+	if err != nil {
+		return err
+	}
+	if currentContext == find.VClusterContextName(vCluster.Name, vCluster.Namespace, vCluster.Context) {
+		cmd.Log.Infof("already connected to vcluster %q", vCluster.Name)
```
FabianKramm, line 96: Can we somehow check if the vCluster is reachable here like we do below so that we reconnect if we cannot reach the vCluster?

FabianKramm (changes requested):

pkg/cli/connect_helm.go
```diff
@@ -693,3 +698,28 @@ func (cmd *connectHelm) waitForVCluster(ctx context.Context, vKubeConfig clientc
...
+		if err != nil {
+			return false, err
+		}
+
+		vKubeClient, err := kubernetes.NewForConfig(kubeConfig)
+		if err != nil {
+			return false, err
+		}
+
+		_, err = vKubeClient.CoreV1().ServiceAccounts("default").Get(ctx, "default", metav1.GetOptions{})
```
FabianKramm, line 718: Just a small nit and then I think we can merge this, but I guess we want to have a timeout context here, because otherwise this might hang forever

FabianKramm (changes requested):

FabianKramm (approved):

PR #2348, opened 2024-12-18
Validate comments in config/config.go before generating jsonschema

hidalgopl:
/kind enhancement

- add checking for invalid characters that will break .mdx generation  in docs to hack/schema/main.go;
- fix sync-config-schema.yaml vcluster-config syncing

Enhancement: do not allow invalid characters in config/config.go comments

### DIFF
--- a/.github/workflows/sync-config-schema.yaml
+++ b/.github/workflows/sync-config-schema.yaml
@@ -135,8 +135,8 @@ jobs:
           GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
           RELEASE_TAG: ${{ steps.release.outputs.release_tag }}
         run: |
+          rm -rf vcluster-config/ || true
           git clone --single-branch https://github.com/loft-sh/vcluster-config.git
-
           # copy generated schema from vcluster chart values to vcluster-config
           cp chart/values.schema.json vcluster-config/values.schema.json
           cp -R config/. vcluster-config/config/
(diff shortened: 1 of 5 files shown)


hack/schema/main.go
```diff
@@ -239,6 +239,12 @@ func getReflector() (*jsonschema.Reflector, error) {
 	}
 	r.CommentMap = commentMap
 
+	for k, comment := range commentMap {
+		if strings.Contains(comment, "<") || strings.Contains(comment, ">") {
```
cbron, line 243: Nit: Long term might be better to check for more consistent strings. I think there is a Constant in golang core library, though this looks like special case.

.github/workflows/sync-config-schema.yaml
```diff
@@ -135,8 +135,8 @@ jobs:
           GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
           RELEASE_TAG: ${{ steps.release.outputs.release_tag }}
         run: |
+          rm -rf vcluster-config/ || true
```
cbron, line 138: Was this on purpose ?

hidalgopl: yes, the CI failed because in some cases `vcluster-config` is already cloned in the previous step. We want to make sure that we commit to vcluster-config only changes made in this step (that's why we remove and clone again)

hack/schema/main.go
hidalgopl: yeah, so if anyone puts a `<` or `>` in the comments [here](https://github.com/loft-sh/vcluster/blob/main/config/config.go), it breaks `.mdx` rendering (as we use comments as a source of content for field description)

cbron (approved):

FabianKramm (approved):

PR #2349, opened 2024-12-18
bugfix: node allocatable resources correction

neogopher:
/kind bugfix

resolves ENG-5440

Fixed an issue where vcluster would calculate allocatable resources inside the virtual cluster incorrectly when virtual cluster scheduler was enabled

With this PR, we are updating the code to use the same utility function that kubectl uses for the resource calculation.

### DIFF
--- a/pkg/controllers/resources/nodes/translate.go
+++ b/pkg/controllers/resources/nodes/translate.go
@@ -14,6 +14,7 @@ import (
 	"github.com/loft-sh/vcluster/pkg/util/translate"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
+	resourceutil "k8s.io/kubectl/pkg/util/resource"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


hidalgopl (approved):

pkg/controllers/resources/nodes/translate.go
```diff
@@ -187,39 +190,33 @@ func (s *nodeSyncer) translateUpdateStatus(ctx *synccontext.SyncContext, pNode *
...
-						cpu -= container.Resources.Requests.Cpu().MilliValue()
-						memory -= container.Resources.Requests.Memory().Value()
-						storageEphemeral -= container.Resources.Requests.StorageEphemeral().Value()
-					}
-					for _, container := range pod.Spec.Containers {
-						cpu -= container.Resources.Requests.Cpu().MilliValue()
-						memory -= container.Resources.Requests.Memory().Value()
-						storageEphemeral -= container.Resources.Requests.StorageEphemeral().Value()
+
+					reqs, _ := resourceutil.PodRequestsAndLimits(&pod)
```
hidalgopl, line 194: so it seems that we were summing up init containers instead of picking the one with the biggest resource request + not accounting for `pod.spec.overhead`, right?

neogopher: yes exactly. This lead our calculations to be incorrect as summing them up caused them to go beyond the maximum allocatable size of the node, and yes pod.spec.overhead was never taken into account.

johannesfrey (approved):

PR #2368, opened 2025-01-02
fix: resync labels initially

FabianKramm:
/kind bugfix

Fixed an issue where vcluster would not resync the labels when upgrading from pre-0.21 to 0.21 or above


rmweir (approved):
LGTM

PR #2369, opened 2025-01-02
Fix coredns

rmweir:
Prior, if the coredns deployment was using the old label, "kube-dns",
it would not be updated during the coredns manifests apply. This was
because label selectors are immutable. Now, if the label is outdated
on the deployment, the deployment is deleted prior to the coredns
manifests apply.

/kind bugfix

Part of ENG-5232

Fixed an issue where vcluster dns would stop working on upgrade.

### DIFF
--- a/pkg/setup/controllers.go
+++ b/pkg/setup/controllers.go
@@ -175,7 +175,23 @@ func StartControllers(controllerContext *synccontext.ControllerContext, syncers 
 
 func ApplyCoreDNS(controllerContext *synccontext.ControllerContext) {
 	_ = wait.ExponentialBackoffWithContext(controllerContext.Context, wait.Backoff{Duration: time.Second, Factor: 1.5, Cap: time.Minute, Steps: math.MaxInt32}, func(ctx context.Context) (bool, error) {
-		err := coredns.ApplyManifest(ctx, controllerContext.Config.ControlPlane.Advanced.DefaultImageRegistry, controllerContext.VirtualManager.GetConfig(), controllerContext.VirtualClusterVersion)
+		dnsDeployment := &appsv1.Deployment{}
+		err := controllerContext.VirtualManager.GetClient().Get(controllerContext.Context, types.NamespacedName{Namespace: "kube-system", Name: "coredns"}, dnsDeployment)
+		if err != nil && !kerrors.IsNotFound(err) {
+			return false, err
+		}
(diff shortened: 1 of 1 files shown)


pkg/setup/controllers.go
```diff
@@ -175,7 +175,25 @@ func StartControllers(controllerContext *synccontext.ControllerContext, syncers
 
 func ApplyCoreDNS(controllerContext *synccontext.ControllerContext) {
 	_ = wait.ExponentialBackoffWithContext(controllerContext.Context, wait.Backoff{Duration: time.Second, Factor: 1.5, Cap: time.Minute, Steps: math.MaxInt32}, func(ctx context.Context) (bool, error) {
-		err := coredns.ApplyManifest(ctx, controllerContext.Config.ControlPlane.Advanced.DefaultImageRegistry, controllerContext.VirtualManager.GetConfig(), controllerContext.VirtualClusterVersion)
+		var dnsDeployment appsv1.Deployment
+		err := controllerContext.VirtualManager.GetClient().Get(controllerContext.Context, types.NamespacedName{Namespace: "kube-system", Name: "coredns"}, &dnsDeployment)
+		if err != nil {
+			if !kerrors.IsNotFound(err) {
```
FabianKramm, line 181: I think this should be like below `if err != nil && !kerrors.IsNotFound(err)`

pkg/setup/controllers.go
```diff
@@ -175,7 +175,25 @@ func StartControllers(controllerContext *synccontext.ControllerContext, syncers
 
 func ApplyCoreDNS(controllerContext *synccontext.ControllerContext) {
 	_ = wait.ExponentialBackoffWithContext(controllerContext.Context, wait.Backoff{Duration: time.Second, Factor: 1.5, Cap: time.Minute, Steps: math.MaxInt32}, func(ctx context.Context) (bool, error) {
-		err := coredns.ApplyManifest(ctx, controllerContext.Config.ControlPlane.Advanced.DefaultImageRegistry, controllerContext.VirtualManager.GetConfig(), controllerContext.VirtualClusterVersion)
+		var dnsDeployment appsv1.Deployment
```
FabianKramm, line 178: Make this a pointer then it is easier to check below as well

FabianKramm (dismissed):

FabianKramm (approved):

PR #2378, opened 2025-01-06
Support labelSelectors to filter out services before syncing to the host cluster

### LINKED ISSUE #1622: Support filtering exported services or other resources

Currently, exporting `services` from virtual to host can only be toggled on/off. The [docs](https://www.vcluster.com/docs/syncer/config#enable-or-disable-synced-resources) understandably warn against disabling exporting `services`.

Sync configuration should allow using `metav1.LabelSelector`, in particular `MatchExpressions` to allow negative expressions (i.e `label/export-service NotIn ["false"]`

None that I can think of.

### Additional context

`endpointslice` with unexported parent service should probably not be exported as well. This could be checked by the syncer when exporting `endpoints` or config validation could force the `endpoint` configuration to match `service`. (is there a use case where one should be exported but not the other?)

ApsTomar:
/kind enhancement

resolves #1622 

Added a feature enhancement to support filtering out services using label selectors, in particular MatchExpressions (for eg. label/export-service NotIn ["false"]

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -509,7 +509,10 @@ type EnableSwitchWithPatches struct {
 }
 
 type SyncSelector struct {
-	MatchLabels      map[string]string                 `json:"matchLabels,omitempty"`
+	// MatchLabels are the labels to select the object from.
+	MatchLabels map[string]string `json:"matchLabels,omitempty"`
+
+	// MatchExpressions is a list of label selector requirements which supports operators such as In, NotIn, Exists and DoesNotExist.
 	MatchExpressions []metav1.LabelSelectorRequirement `json:"matchExpressions,omitempty"`
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -5,13 +5,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"reflect"
 	"regexp"
 	"strings"
 	"time"
 
 	"github.com/invopop/jsonschema"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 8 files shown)


config/config.go
```diff
@@ -500,10 +501,21 @@ type EnableSwitchWithPatches struct {
 	// Enabled defines if this option should be enabled.
 	Enabled bool `json:"enabled,omitempty"`
 
+	// Selector defines the labelSelector based filtering for syncing the resources to the host cluster.
+	Selector *SyncSelector `json:"selector,omitempty"`
```
FabianKramm, line 505: Adding this here would also apply to other resources that use EnableSwitchWithPatches, is that wanted?

FabianKramm (changes requested):
I'm not sure if we actually want to do this, whats the use case for allowing something like this? Services wouldn't work at all if they are not synced, so it seems to me this is not something we want to support

config/config.go
ApsTomar: We can have a separate object specifically for this purpose. But I've added this in EnableSwitchWithPatches, keeping in mind that it would provide us an option to enable such selector based filtering for other resources as well. And the user can user let it be empty if they don't require any filtering.

ApsTomar:
>  it seems to me this is not something we want to support
@FabianKramm So, is it safe to assume that this feature enhancement which is requested is not a desirable one and hence the issue can be closed?

ApsTomar:
>  it seems to me this is not something we want to support
@FabianKramm If I understood this correctly then this feature enhancement which is requested is not a desirable one and hence the issue can be closed?
In that case, let me know if there is anything in the PR which is worth adding to vcluster, if not then I'll close this PR once the issue is closed.

cbron:
I believe this feature is no longer being implemented

PR #2385, opened 2025-01-08
fix: set resource version to 1 on old

FabianKramm:
/kind bugfix

resolves ENG-5478


zerbitx (approved):
lgtm

PR #2406, opened 2025-01-16
fix: Only error when we verify they're using a disposable domain.

zerbitx:
This change ignores any errors in trying to get the disposable domains list and only returns an error if the domain being used is found in the list.

I tested this by providing a disposable domain to get the error and by using a list in an invalid json format to get the logging, and see it ignore the error.

/kind bugfix

resolves #ENG-5542

### DIFF
--- a/pkg/cli/email/validate.go
+++ b/pkg/cli/email/validate.go
@@ -68,25 +68,25 @@ func Validate(emailAddress string, options ...Option) error {
 func checkDisposableDomains(domain string) error {
 	fb, err := domainsZip.ReadFile("disposable_domains.zip")
 	if err != nil {
-		return err
+		return nil
 	}
 
 	zipReader, err := zip.NewReader(bytes.NewReader(fb), int64(len(fb)))
 	if err != nil {
-		return fmt.Errorf("failed to read disposable domains list for verification: %w", err)
+		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


cbron (dismissed):
lgtm, though you should wait for @lizardruss as he would know logging patterns better than I. Debug feels fine since you would almost never want to know or care that this failed. If anything it's a bad thing as it gives a clue to make circumvention easier.

pkg/cli/email/validate.go
```diff
@@ -66,27 +70,36 @@ func Validate(emailAddress string, options ...Option) error {
 }
 
 func checkDisposableDomains(domain string) error {
+	logger := log.GetInstance()
+	if os.Getenv("DEBUG") == "true" {
+		logger.SetLevel(logrus.DebugLevel)
+	}
```
lizardruss, line 76: I *think* a logger is already initialized with the correct level based on the `--debug` flag. Did `log.GetInstance()` not return a logger with that level already?

lizardruss (changes requested):
Question regarding the logger initialization.

pkg/cli/email/validate.go
zerbitx: I just removed it, since like you said there would be no work around anyway.

lizardruss (approved):

PR #2411, opened 2025-01-17
Exclude backingStore.etcd's image from getting populated when disabled

### LINKED ISSUE #2398: Change logic to not include controlPlane.backingStore.etcd 's image when it's not enabled on the vCluster
# Problem & Impact

On a default cluster that's created using 0.20 vCluster - vcluster create <name>, generate a yaml using helm get values VCLUSTER_NAME -n VCLUSTER_NAMESPACE command, and you will notice controlPlane.backingStore.etcd 's image is populated even though its disabled by default (The host cluster is on 1.29 in this case)

```
controlPlane:
  backingStore:
    etcd:
      deploy:
        statefulSet:
          image:
            tag: 3.5.10-0
  distro:
    k0s:
      image:
        tag: v1.29.6-k0s.0
    k3s:
      image:
        tag: v1.29.6-k3s1
    k8s:
      apiServer:
        image:
          tag: v1.29.6
      controllerManager:
        image:
          tag: v1.29.6
      scheduler:
        image:
          tag: v1.29.6
  service:
    spec:
      type: NodePort
sync:
  fromHost:
    nodes:
      enabled: true
telemetry:
  instanceCreator: vclusterctl
  machineID: 4f52a059183efb60979cbe81e105eac94519bb3be9910fbc141fe3dfc38a5833
```

# Goals

Change logic to not include controlPlane.backingStore.etcd 's image when it's not enabled on the vCluster

ApsTomar:
/kind bugfix

resolves ENG-5526 (#2398) 

Fixed an issue in vcluster v0.20 where the controlPlane.backingStore.etcd's image is getting populated with default values even when it has been disabled.

While reproducing the issue, I came to a conclusion that this issue happens in vcluster version 0.20 when k8s version < 1.30, a quick fix for the user would be to either upgrade vcluster to latest version OR upgrade k8s to a version >= 1.30

Now, for the users on vcluster 0.20 and having k8s version < 1.30, this PR would resolve the issue. Please refer to the artifact attached below. With this fix, controlPlane.backingStore.etcd's image will not get populated if disabled.

One thing to note here- similar to etcd image, there are distro images as well present in the output as mentioned in the issue. These can also be fixed in the similar fashion as this PR. Let me know if we want to resolve that along with this etcd image fix.

Screenshot from 2025-01-17 21-22-43

### DIFF
--- a/config/default_extra_values.go
+++ b/config/default_extra_values.go
@@ -173,18 +173,9 @@ func applyK8SExtraValues(vConfig *Config, options *ExtraValuesOptions) error {
 		return err
 	}
 
-	// get etcd image
-	etcdImage, err := getImageByVersion(options.KubernetesVersion, K8SEtcdVersionMap)
-	if err != nil {
-		return err
-	}
-
 	// build values
 	if apiImage != "" {
 		vConfig.ControlPlane.Distro.K8S.Version = parseImage(apiImage).Tag
-	}
-	if etcdImage != "" {
-		vConfig.ControlPlane.BackingStore.Etcd.Deploy.StatefulSet.Image = parseImage(etcdImage)
 	}
 
 	return nil


config/default_extra_values.go
```diff
@@ -173,19 +173,10 @@ func applyK8SExtraValues(vConfig *Config, options *ExtraValuesOptions) error {
...
-	etcdImage, err := getImageByVersion(options.KubernetesVersion, K8SEtcdVersionMap)
-	if err != nil {
-		return err
-	}
-
 	// build values
 	if apiImage != "" {
 		vConfig.ControlPlane.Distro.K8S.Version = parseImage(apiImage).Tag
 	}
-	if etcdImage != "" {
```
lizardruss, line 186: I think you still need to set the image when the etcd deployment is enabled.

lizardruss (changes requested):
Some logic updates needed.

config/default_extra_values.go
ApsTomar: With these changes, the behaviour is now consistent with the latest vcluster (v0.22) when `etcd.Deploy` is enabled. But I understand the concern, I will check again to be sure and update with more details.

ApsTomar: @lizardruss I have checked and verified about the scenario that you mentioned- `when etcd deployment is enabled` - in that case it's going to take the default image value present in [default values.yaml](https://github.com/loft-sh/vcluster/blob/v0.20/config/values.yaml#L180), as expected. 

The reason being, we first get the default helm values (as mentioned above), then we read the values.yaml file provided by the user and merge override only those values which are supplied by the user to create the vcluster config. So, `etcd.Deploy` will be set to true and the image value will be used from default values.

Below are the test artifacts supporting the above-mentioned facts-

- Here is the values.yaml file (vc20.yaml) which sets `etcd.Deploy.Enabled = true` and the etcd pod with the expected image version `3.5.13`:
image

- Here is the result of `helm get values` command:
image

PR #2388, opened 2025-01-09
chore: bump k8s version to 1.32

rohantmp:
**What issue type does this pull request address?** (keep at least one, remove the others) 
/kind bugfix
/kind enhancement
/kind feature
/kind documentation
/kind test

**What does this pull request do? Which issues does it resolve?** (use `resolves #<issue_number>` if possible) 
resolves ENG-5387
resolves ENG-5390
resolves ENG-5428

**Please provide a short message that should be published in the vcluster release notes**
Update k8s client dependencies to 1.32

(depends on pending controller-runtime release)

**What else do we need to know?**

PR #2427, opened 2025-01-25
Conformance test fix

kale-amruta:
/kind bugfix
/kind enhancement
/kind feature
/kind documentation
/kind test

resolves #


ApsTomar:
There are two more tests which are failing. These tests were recently added in conformance test suite (a couple of months back). Below are the tests and the RCA:
1. should run through the lifecycle of a node
2. should run through the lifecycle of a csinode

**Why are these tests failing?**

These tests perform the lifecycle operations (CRUD) for nodes and CSI nodes on vcluster respectively but since vcluster doesn't inherently support the node creation on the virtual cluster, so `syncToHost` method simply deletes the node that was created during the conformance test. Hence, the test can't perform any further action such as get or update and the test fails.

**Probable Solutions:**

Since this is a case of _a feature not supported_ so ideally it should be skipped but that might impact the certification failure as mentioned in conformance test instructions.
In that case, I tried some workarounds which certainly work but they have some downsides as well. I've detailed them out, please take a look:

1. The conformance test creates the nodes which have a certain prefix in the name and are unschedulable in nature. This can be used to identify the test nodes and we can skip their deletion during sync operation. Later the test clean-up will delete that node after completion. The caveat is that- a user can misuse this to create a similar node which vcluster won't be able to delete during the regular sync operation.

2. We can add some `gracePeriod` to the node deletion in `syncToHost` method which will keep the node active till the tests are passed. The caveat is that- it'll impact all the node deletion process and the users might not like that the nodes are taking more time to get deleted now. Also, what should be the ideal grace period is also uncertain and might make the test flaky.

3. Better solution might be a combination of both of the above points: we identify the node created during the test (similar to point #1) and then mark it for deletion with a decent grace period (30 to 60 sec). This will not only make sure that the code change won't impact vcluster's natural behaviour but also won't allow any user to misuse it because it'll eventually get deleted after the grace period runs out.

Let me know your opinions and based on that I will create the PR to fix these remaining 2 conformance tests.

cbron:
Original PR was here, I will close: https://github.com/loft-sh/vcluster/pull/2388

pkg/controllers/resources/pods/syncer.go
```diff
@@ -274,6 +275,23 @@ func (s *podSyncer) Sync(ctx *synccontext.SyncContext, event *synccontext.SyncEv
...
+		// results in dangling pods on vcluster
+		if !equality.Semantic.DeepEqual(event.Virtual.Status, event.Host.Status) {
+			updated := event.Virtual.DeepCopy()
+			updated.Status = *event.Host.Status.DeepCopy()
+			ctx.Log.Infof("update virtual pod %s, because status has changed", event.Virtual.Name)
+			err := ctx.VirtualClient.Status().Update(ctx, updated)
+			if err != nil && !kerrors.IsNotFound(err) {
+				return ctrl.Result{}, err
+			}
+		}
```
lizardruss, line 293: I wonder if a straight status update is desired, or the patch + diff logic [down here](https://github.com/loft-sh/vcluster/pull/2427/files#diff-31528d4b6188bc9038b3a499cb5ff53a3b7d184f62fb9ccef03c9e5cfb2c78aeR356-R375) would also fix the issue. In this logic branch that update logic is skipped.

It may be as simple as removing the [return here](https://github.com/loft-sh/vcluster/pull/2427/files#diff-31528d4b6188bc9038b3a499cb5ff53a3b7d184f62fb9ccef03c9e5cfb2c78aeR295), but could have other side effects.

lizardruss (commented):
Left some questions and suggestions to try.

pkg/controllers/resources/pods/syncer.go
kale-amruta: @lizardruss we can remove the `return` statement, but it will unnecessarily patch all the other updates as well to the pod spec. Question is do we want to patch other updates to spec/metdata if the host pod is being deleted? AFAIU the check of `if event.Host.DeletionTimestamp != nil` is to identify if we should delete the virtual pod OR process the other updates to virtual pod/host pod.

lizardruss: Maybe. One example might be patching to remove finalizers, but seems like we would have come across that case before.

kale-amruta: I tried this approach of removing finalizers , but that broke other conformance tests, hence discarded that approach.

kale-amruta: Okay, this is what I did in the latest commit - https://github.com/loft-sh/vcluster/pull/2427/commits/00fd0bd956beb1a9bff43d5de0943d7adab93f29
- Patch just the status in podTranslator.Diff when host.DeletionTimestamp is not nil
- skip all the other updates apart from status when host.DeletionTimeStamp is not nil

Let me know if you think this is good.

pkg/controllers/resources/pods/translate/diff.go
```diff
@@ -9,10 +9,26 @@ import (
...
+	event.Host.Status.QOSClass = event.VirtualOld.Status.QOSClass
+
+	if event.Host.DeletionTimestamp != nil {
+		// Copy status updates from host to virtual if there is a status difference
+		if !equality.Semantic.DeepEqual(event.Virtual.Status, event.Host.Status) {
+			event.Virtual.Status = *event.Host.Status.DeepCopy()
+		}
+		// return since host is being deleted, no updates required
+		return nil
+	}
```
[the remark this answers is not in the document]
kale-amruta: Yes, since [this](https://github.com/loft-sh/vcluster/pull/2427/files/00fd0bd956beb1a9bff43d5de0943d7adab93f29#diff-15faba7aa47de3def60ba11ad95c92113e2c10ff395a2ebc453c8a3e64182b8eR33) piece of code does a birectional update, we would not want to copy the virtual status to host pod status in case of host pod is being deleted.

Yes all the other updates after [this](https://github.com/loft-sh/vcluster/pull/2427/files#diff-15faba7aa47de3def60ba11ad95c92113e2c10ff395a2ebc453c8a3e64182b8eR48) should be skipped and are not required when host is being deleted. 

I think I will revert to original solution then.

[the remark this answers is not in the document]
kale-amruta: I have reverted it to original solution

lizardruss (approved):

PR #2459, opened 2025-02-07
Secrets From Host Sync

hidalgopl:
/kind feature
/kind test

resolves ENG-5724 ENG-5739 ENG-5736

I moved common logic for from host syncing to `pkg/controllers/generic/from_host_syncer.go` & `pkg/syncer/translator/from_config_translator.go`. 
As a result, implementation for particular kinds (ConfigMaps, Secrets) requires minimum kind-specific code.
Added e2e test for secrets too.

Secretss From Host Sync


johannesfrey (commented):
First round. Mostly nits.

pkg/controllers/generic/from_host_syncer.go
```diff
@@ -0,0 +1,319 @@
...
+	// sync.toHost.(configMaps/secrets).enabled = true
+	// sync.toHost.(configMaps/secrets).all = true
+	// sync.fromHost.(configMaps/secrets).enabled = true
+	// sync.fromHost.(configMaps/secrets).mappings: (mapping from vCluster host namespace to NAMESPACE_A)
+	// experimental.deploy.vCluster.manifests: (manifest for (config map / secret) in NAMESPACE_A)
+	// then, fromHost syncer will enqueue (config map/ secret) coming from experimental.deploy.vCluster.manifests
+	// syncer will try to call SyncToHost, as physical object does not exist
+	// fromHostSyncer.SyncToHost will delete virtual object, which is not wanted and expected by user.
+	// Therefore, we need to exclude objects from experimental.deploy.vCluster.(manifests/manifestsTemplate)
+	// in the genericFromHostSyncer.
```
johannesfrey, line 235: I'm wondering if we could either rather use a PredicateFunc in `WatchesRawSource` that filters out any ConfigMaps/Secrets coming from the experimental.deploy or check its presence in SyncToHost and don't delete in case it has the marker.
But I guess we would need some markers (label/annotation) for that. Either on the host object (PredicateFunc) or on the virtual object (SyncToHost).
But by looking at https://github.com/loft-sh/vcluster/blob/main/pkg/controllers/deploy/deploy.go it doesn't look like its applying any marker, except the `vcluster-deploy` ConfigMap in kube-system, which doesn't really help in that case.

Regardless of that, IMHO I would also challenge if this experimental feature is "worth it" to add additional complexity here. But I guess this decision would require some additional opinions (cc @deniseschannon @cbron )

test/e2e/syncer/fromhost/from_host_secrets.go
```diff
@@ -0,0 +1,159 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = ginkgo.Describe("Secrets are synced to host and can be used in Pods", ginkgo.Ordered, func() {
+	var (
+		f                   *framework.Framework
+		configMap1          *corev1.Secret
+		configMap2          *corev1.Secret
+		cm1Name             = "dummy"
```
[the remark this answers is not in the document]
hidalgopl: renamed

chart/values.yaml
```diff
@@ -153,6 +153,30 @@ sync:
     priorityClasses:
       # Enabled defines if this option should be enabled.
       enabled: false
+    # Secrets defines if config maps in the host should get synced to the virtual cluster.
```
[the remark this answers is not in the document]
hidalgopl: good catch, changed

pkg/controllers/generic/from_host_syncer.go
hidalgopl: yes, there are no markers, we create them "as-is". Keep in mind that this happens only in very specific combinations of fields in config.
I talked with @matskiv about the `experimental.deploy.vcluster` and it seems that it isn't widely used.

I'd prefer to exclude these objects from `experimental.deploy` in the `ExcludeVirtual` (as is currently),  as it is a part of `syncertypes.ObjectExcluder` interface, which seems to be defined exactly for cases like this :))

pkg/config/validation_test.go
```diff
@@ -236,7 +236,104 @@ func TestValidateFromHostSyncMappings(t *testing.T) {
...
 					Mappings: map[string]string{
-						"default/my-cm": "barfoo",
+						"default/*": "barfoo",
+					},
+				},
+			},
+			expectErr: expectErr,
+		},
+		{
+			name: "invalid config 5",
```
zerbitx, line 246: nit: Would be nice if the names described what was invalid like "mappings has invalid key"

pkg/controllers/generic/from_host_syncer.go
cbron: From my POV, it's totally fine to have this feature not be compatible with the experimental feature to bring down complexity and speed things up. As long as it's documented in the code or elsewhere. Then we can fix it if we get a request for it.

pkg/controllers/generic/from_host_syncer.go
```diff
@@ -0,0 +1,318 @@
...
+			if found {
+				virtualConfigMapsToSkip[configMapKey] = true
+			}
+		}
+	} else {
+		configMapKey, found := s.processTemplate(deploy.VCluster.ManifestsTemplate, &ctx.Config.Config, ctx.Config.Name, ctx.Config.WorkloadTargetNamespace)
+		if found {
+			virtualConfigMapsToSkip[configMapKey] = true
+		}
+	}
```
zerbitx, line 266: You shouldn't need the `strings.Contains` check.  If it doesn't contain "---" the split will return a single element slice, so it could be more like 

```go
	for _, manifest := range strings.Split(deploy.VCluster.Manifests, "---") {
		configMapKey, found := s.processManifest(manifest)
		if found {
			virtualConfigMapsToSkip[configMapKey] = true
		}
	}

	for _, manifest := range strings.Split(deploy.VCluster.ManifestsTemplate, "---") {
		configMapKey, found := s.processTemplate(manifest, &ctx.Config.Config, ctx.Config.Name, ctx.Config.WorkloadTargetNamespace)
		if found {
			virtualConfigMapsToSkip[configMapKey] = true
		}
	}
```

pkg/syncer/translator/from_config_translator.go
```diff
@@ -0,0 +1,181 @@
...
+	translatorName string
+	skipFuncs      []ShouldSkipHostObjectFunc
+}
+
+// ShouldSkipHostObjectFunc takes object's host name and namespace
+// and returns true if object should be skipped in the from host sync.
+type ShouldSkipHostObjectFunc func(hostName, hostNamespace string) bool
+
+func NewFromHostTranslatorForGVK(ctx *synccontext.RegisterContext, gvk schema.GroupVersionKind, mappings map[string]string, skipFuncs ...ShouldSkipHostObjectFunc) (syncer.FromConfigTranslator, error) {
+	hostToVirtual := mappings
```
zerbitx, line 32:
```suggestion
func NewFromHostTranslatorForGVK(ctx *synccontext.RegisterContext, gvk schema.GroupVersionKind, hostToVirtual map[string]string, skipFuncs ...ShouldSkipHostObjectFunc) (syncer.FromConfigTranslator, error) {
```

pkg/syncer/translator/from_config_translator.go
```diff
@@ -0,0 +1,181 @@
...
+	// last chance, if user specified "": <namespace>/*
+	if virtual, ok := resourceMappings[constants.VClusterNamespaceInHostMappingSpecialCharacter]; ok {
+		if vClusterHostNamespace == hostNamespace {
+			virtualParts := strings.Split(virtual, "/")
+			if len(virtualParts) == 2 {
+				return types.NamespacedName{Namespace: virtualParts[0], Name: hostName}, true
+			} else if !strings.Contains(virtual, "/") {
+				// then the mapping is "": "virtual-namespace" where "" means vCluster host namespace
+				// in this case, we want to return virtual-namespace/hostName
+				return types.NamespacedName{Namespace: virtual, Name: hostName}, true
```
zerbitx, line 144:
```suggestion
			return types.NamespacedName{Namespace: virtualParts[0], Name: hostName}, true
```

pkg/syncer/translator/from_config_translator.go
```diff
@@ -0,0 +1,181 @@
...
+	// first, let's try matching by namespace/name
+	if virtual, ok := resourceMappings[key]; ok {
+		virtualParts := strings.Split(virtual, "/")
+		if len(virtualParts) == 2 {
+			ns := virtualParts[0]
+			name := virtualParts[1]
+			if name == "*" {
+				name = hostName
+			}
+			return types.NamespacedName{Namespace: ns, Name: name}, true
```
zerbitx, line 122: I just learned from Jan's [PR](https://github.com/loft-sh/loft-enterprise/pull/3706/files?w=1#diff-e7b4913cc69eb9882b7b87493cdd22c0b1286b0529a03aa26078c234c8abec2fR57) that `strings.Cut` exists. Its a nit suggestion, but this and other splits could be changed to something like. 
```suggestion
		ns, name, _ := strings.Cut(virtual, "/")
		if name == "*" {
			name = hostName
		}
		return types.NamespacedName{Namespace: ns, Name: name}, true
```

pkg/controllers/generic/from_host_syncer.go
hidalgopl: good point, changed

pkg/syncer/translator/from_config_translator.go
hidalgopl: good call :+1:

pkg/syncer/translator/from_config_translator.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func (c *fromHostTranslate) Resource() client.Object {
+	switch c.gvk.Kind {
+	case "ConfigMap":
+		return &corev1.ConfigMap{}
+	case "Secret":
+		return &corev1.Secret{}
+	default:
+		return nil
```
zerbitx, line 59: Asking because I don't know.  Would it ever be ok to return a nil here?  Its rarely the case that a panic is the right thing to do, but this _**might**_ be one case.  Thoughts?

pkg/controllers/generic/from_host_syncer.go
```diff
@@ -0,0 +1,306 @@
...
+
+	vObj := translate.VirtualMetadata(event.Host, s.HostToVirtual(ctx, types.NamespacedName{Name: event.Host.GetName(), Namespace: event.Host.GetNamespace()}, event.Host))
+
+	err := pro.ApplyPatchesVirtualObject(ctx, nil, vObj, event.Host, s.GetProPatches(ctx), false)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// make sure namespace exists
```
zerbitx, line 112: nit: I think the namespace check can be moved before the `vObj` variable and `ApplyPatchesVirtualObject` call.

pkg/syncer/translator/from_config_translator.go
hidalgopl: It's only for now. Once this is merged, I'll open another PR to re-use this translator for CRDs - and I already changed this default case there.

pkg/config/validation_test.go
hidalgopl: changed :+1:

pkg/config/validation.go
```diff
@@ -599,6 +605,9 @@ func validateFromHostSyncMappings(s config.EnableSwitchWithResourcesMappings, re
 		return fmt.Errorf("config.sync.fromHost.%s.mappings are empty", resourceNamePlural)
 	}
 	for key, value := range s.Selector.Mappings {
+		if key == "*" {
```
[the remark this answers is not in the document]
hidalgopl: this is to ensure that if someone puts something like this to mapping:
```yaml
mappings:
  "*": <any-value>
  ```
 he will get a descriptive error message saying that `"*"` is not a valid key

File: pkg/controllers/generic/from_host_syncer.go (hunk 10 above)
Comment (hidalgopl) [reply pkg/controllers/generic/from_host_syncer.go:107]: :+1: moved

File: pkg/config/validation.go (hunk 11 above)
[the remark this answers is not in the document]
Comment (johannesfrey) [reply pkg/config/validation.go:609]: Might be that I'm not looking correctly but won't the `*` be also captured by `if !strings.Contains(key, "/") && key != constants.VClusterNamespaceInHostMappingSpecialCharacter`, which has an even more descriptive error message? 🤔

**Review by johannesfrey — APPROVED** [review]
Except last nit question.

**Review by johannesfrey — APPROVED** [review]

File: test/e2e/values.yaml [hunk 12]
```diff
@@ -40,4 +40,10 @@ sync:
         mappings:
           "from-host-sync-test/*": "barfoo/*"
           "default/my-cm": "barfoo/cm-my"
-          "": "my-new-ns"
+          "": "default"
```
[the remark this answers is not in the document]
hidalgopl: Having it set to default implicitly tests an edge case that I've described here: https://github.com/loft-sh/vcluster/pull/2459/files/aa7e6f693c814acdb18dd0742f4d3e3fbe383c03#diff-f209cd83c0ee845d41f0f79172bf9b41fee25febd64ab42493927f8ba1c980afR223-R233 (as the `experimental.deploy.vcluster` config maps are created in default namespace too). Feel free to add another `key: value` in your PR to the `values.yaml` that will cover the case you're testing there

PR #2481, opened 2025-02-14
From host sync additional config validation

hidalgopl:
/kind bugfix
/kind enhancement
/kind test

resolves ENG-5739

Fixed an issue where vcluster was allowing invalid values in from host sync mapping entries

this contains config validation that got lost in the revert + better organization of functions in `pkg/syncer/from_host_syncer.go`, so we can use them in other repo.

### DIFF
--- a/pkg/config/parse.go
+++ b/pkg/config/parse.go
@@ -3,6 +3,10 @@ package config
 import (
 	"fmt"
 	"os"
+	"strings"
+
+	"github.com/loft-sh/vcluster/pkg/constants"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 
 	"github.com/loft-sh/vcluster/config"
 	"github.com/loft-sh/vcluster/pkg/strvals"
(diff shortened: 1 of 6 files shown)


FabianKramm (approved):

hidalgopl:
> Can we add some E2E test for it 🙏

This would be quite an effort, given that our e2e CI works in the following way:
1. Each test suite in `test/` directory has it's own `values.yaml` (vcluster config)
2. We compile each test suite first
3. Then, we fan out in CI, so each test suite gets it's own CI job, creates kind cluster, deploys vCluster with the `values.yaml` from test suite directory, then executes test suite binary against it

So, to have an e2e test that checks invalid vCluster config values, we would need to have a snowflake test suite that we treat differently in the CI (we won't deploy vCluster in this case, because the test suite itself should do that multiple times with various `vcluster.yaml` combinations and check whether the `vcluster create ...` failed as expected.

I think for now having good unit test coverage for config validation functions should work fine.

PR #2483, opened 2025-02-14
feat: add sync.toHost.pods.priorityClassName

neogopher:
/kind feature

resolves ENG-5095

With this PR we are:
- Adding sync.toHost.pods.priorityClassName config field which will add the specified priorityClassName to all workloads pods being synced to the host
- Fixing a bug that prevented the priorityClassName to be applied correctly when sync.fromHost.priorityClasses.enabled is true.

### DIFF
--- a/pkg/controllers/resources/pods/translate/translator.go
+++ b/pkg/controllers/resources/pods/translate/translator.go
@@ -102,10 +102,11 @@ func NewTranslator(ctx *synccontext.RegisterContext, eventRecorder record.EventR
 		overrideHostsImage:     ctx.Config.Sync.ToHost.Pods.RewriteHosts.InitContainer.Image,
 		overrideHostsResources: resourceRequirements,
 
-		serviceAccountsEnabled: ctx.Config.Sync.ToHost.ServiceAccounts.Enabled,
-		priorityClassesEnabled: ctx.Config.Sync.ToHost.PriorityClasses.Enabled,
-		enableScheduler:        ctx.Config.ControlPlane.Advanced.VirtualScheduler.Enabled,
-		fakeKubeletIPs:         ctx.Config.Networking.Advanced.ProxyKubelets.ByIP,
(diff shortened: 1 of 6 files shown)


pkg/controllers/resources/pods/translate/translator.go
```diff
@@ -172,15 +174,26 @@ func (t *translator) Translate(ctx *synccontext.SyncContext, vPod *corev1.Pod, s
...
-	// check if priority classes are enabled
-	if !t.priorityClassesEnabled {
-		pPod.Spec.PriorityClassName = ""
-		pPod.Spec.Priority = nil
-	} else if pPod.Spec.PriorityClassName != "" {
+	// by default the priority class would be synced as-is to the host
+	// this will readily work when config.Sync.FromHost.PriorityClasses.Enabled is true
+	// however, in case when config.Sync.ToHost.PriorityClasses.Enabled is true, we
+	// need to translate the priorityClassName before syncing to host
+	if t.priorityClassesSyncEnabled && pPod.Spec.PriorityClassName != "" {
```
zerbitx, line 181: I was trying to think of a little refactor so that we're not checking `t.priorityClassesSyncEnabled` in both the if and else if, but it does read better to check that both are disabled.  Very nit suggestion, but maybe naming the conditionals in variables, like `allSyncingDisabled := !t.priorityClassesSyncEnabled && !t.hostPriorityClassesSyncEnabled`.

zerbitx (approved):

cbron:
Going to merge this to fit into RC in time. Feel free to make follow up commits to fix nits ^. Thx!

pkg/controllers/resources/pods/translate/translator.go
neogopher: @zerbitx I have created a separate PR for this [here](https://github.com/loft-sh/vcluster/pull/2498)

PR #2485, opened 2025-02-14
Deprecate creation of multiple vclusters in same namespace

ApsTomar:
/kind bugfix

resolves ENG-5363

- Deprecate the `reuse-namespace` flag.
- Add a flag `reuse-namespace` in vcluster config with immediate deprecation warning.

**Testing**:

- CLI: here's the deprecation warning coming up if someone uses flag `reuse-namespace`:

Screenshot from 2025-02-14 19-35-25

- Using helm, multiple vcluster creation is allowed but with a note about `reuse-namespace` config and a deprecation warning like this:

image

### DIFF
--- a/cmd/vcluster/cmd/start.go
+++ b/cmd/vcluster/cmd/start.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"runtime/debug"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/loft-sh/vcluster/pkg/config"
 	"github.com/loft-sh/vcluster/pkg/constants"
 	"github.com/loft-sh/vcluster/pkg/integrations"
@@ -18,6 +20,10 @@ import (
 	"github.com/loft-sh/vcluster/pkg/telemetry"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
@@ -83,6 +89,21 @@ func ExecuteStart(ctx context.Context, options *StartOptions) error {
 	// initialize feature gate from environment
 	if err := pro.LicenseInit(ctx, vConfig); err != nil {
 		return fmt.Errorf("license init: %w", err)
+	}
+
+	// check if there are existing vclusters in the current namespace. This can be achieved by listing the vcluster services
+	vclusterServices, err := listVclusterServices(ctx, vConfig.ControlPlaneClient, vConfig.ControlPlaneNamespace)
+	if err != nil {
+		return err
+	}
+
+	// add a deprecation warning if reuseNamespace flag is being set to true
+	if !vConfig.ReuseNamespace {
+		if len(vclusterServices) > 0 {
+			return fmt.Errorf("there is already a virtual cluster in the current namespace %s", vConfig.ControlPlaneNamespace)
+		}
+	} else {
+		logrus.Warnf("Creation of multiple vclusters within the same namespace will be deprecated soon.")
 	}
 
 	err = setup.Initialize(ctx, vConfig)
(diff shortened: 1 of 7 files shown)


cmd/vcluster/cmd/start.go
```diff
@@ -85,6 +91,21 @@ func ExecuteStart(ctx context.Context, options *StartOptions) error {
...
+	// check if there are existing vclusters in the current namespace. This can be achieved by listing the vcluster services
+	vclusterServices, err := listVclusterServices(ctx, vConfig.ControlPlaneClient, vConfig.ControlPlaneNamespace)
+	if err != nil {
+		return err
+	}
+
+	// add a deprecation warning if reuseNamespace flag is being set to true
+	if !vConfig.ReuseNamespace {
+		if len(vclusterServices) > 0 {
+			return fmt.Errorf("there is already a virtual cluster in the current namespace %s", vConfig.ControlPlaneNamespace)
```
cbron, line 103: This shouldn't error out yet, just log a warning. We'll start erroring in 0.24. I believe this is why all the tests are failing.

cmd/vcluster/cmd/start.go
```diff
@@ -85,6 +91,21 @@ func ExecuteStart(ctx context.Context, options *StartOptions) error {
...
+		}
+	} else {
+		logrus.Warnf("Creation of multiple vclusters within the same namespace will be deprecated soon.")
+	}
```
cbron, line 107: Won't this error for everyone all the time ? This should be wrapped by the length check so it only occurs when multiple are actually being used.

cbron (changes requested):

cmd/vcluster/cmd/start.go
cbron: Also we need to tell them that that the config.yaml setting is required to run this in 0.24. Which is why we are introducing it now.

cmd/vcluster/cmd/start.go
ApsTomar: I've made the above change.
But just for my understanding, so this means that there'll be a difference in vcluster creation behaviour via helm vs. via CLI. 
For helm if the user creates another vcluster **even without using the reuse-ns config** i.e. `reuse-ns=false`, we'll allow it with a warning. 
But for CLI, if the user does the same then we'll error out because for CLI this is not a new flag or feature. Am I correct?

cmd/vcluster/cmd/start.go
ApsTomar: This error is removed altogether since it's irrelevant now, given that we're only giving warning in v0.23, also I've added the note about setting the config with the deprecation warning. Please review.

zerbitx (approved):
lgtm

cmd/vcluster/cmd/start.go
```diff
@@ -85,6 +91,19 @@ func ExecuteStart(ctx context.Context, options *StartOptions) error {
...
 
+	// check if there are existing vclusters in the current namespace. This can be achieved by listing the vcluster services
+	vclusterServices, err := listVclusterServices(ctx, vConfig.ControlPlaneClient, vConfig.ControlPlaneNamespace)
+	if err != nil {
+		return err
+	}
+
+	// add a note for setting reuse-namespace config in v0.24 and a deprecation warning for multiple vcluster creation scenario
+	if len(vclusterServices) > 0 {
+		logrus.Warnf("Please note that in next release i.e. v0.24, for creating multiple vclusters within the " +
```
johannesfrey, line 102: nit: Can we use `log.Logger` like we already do in other commands?! E.g. like in "pkg/cli/create_helm.go" below.

cmd/vcluster/cmd/start.go
```diff
@@ -85,6 +91,19 @@ func ExecuteStart(ctx context.Context, options *StartOptions) error {
...
+			"same namespace, it'll be mandatory to set 'reuse-namespace=true' in vcluster config. " +
+			"This config and the scenario of creating multiple vclusters in the same namespace will be deprecated soon.")
```
cbron, line 104:
```suggestion
			"This config and the scenario of creating multiple vclusters in the same namespace is deprecated and will be removed soon.")
```

pkg/cli/create_helm.go
```diff
@@ -139,6 +139,8 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 				return fmt.Errorf("there is already a virtual cluster in namespace %s", cmd.Namespace)
 			}
 		}
+	} else {
+		cmd.log.Warn("Creation of multiple vclusters within the same namespace will be deprecated soon.")
```
cbron, line 143:
```suggestion
		cmd.log.Warn("Creation of multiple vClusters within the same namespace is deprecated and will be removed soon.")
```

cbron (approved):

PR #2522, opened 2025-02-21
fix: service label sync

FabianKramm:
/kind bugfix

Fixed an issue where vcluster upgrades wouldn't rewrite the service labels correctly anymore (regression)


zerbitx (approved):
lgtm

PR #2528, opened 2025-02-24
refactor: cache authorizer & short local k8s timeout

FabianKramm:
/kind enhancement

Changes:
* Caches the subjectaccessreview calls
* Reduces the timeout to wait for local k8s connection before falling back to background proxy

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"reflect"
 	"regexp"
 	"strings"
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -333,6 +334,10 @@ func ValidateStoreAndDistroChanges(currentStoreType, previousStoreType StoreType
 }
 
 func (c *Config) IsProFeatureEnabled() bool {
+	if os.Getenv("SKIP_VALIDATE_PRO_FEATURES") == "true" {
```
hidalgopl, line 337: where is this one set and why is it needed?

FabianKramm: its mostly for development to skip the pro checks when the cli is not authenticated / logged in

cbron:
backports: 
* 23: https://github.com/loft-sh/vcluster/pull/2534
* 22: https://github.com/loft-sh/vcluster/pull/2532

PR #2529, opened 2025-02-24
[bugfix- deprecate reuse-ns]: change approach for checking existing vclusters in a namespace

ApsTomar:
/kind bugfix

resolves ENG-5363

Fixed an issue where multiple vcluster creation inside same namespace was providing warning incorrectly. 

Since fetching services is not a fool-proof method of knowing whether we have vclusters existing in a namespace so now I'm reusing an existing method to retrieve existing vclusters.

### DIFF
--- a/cmd/vcluster/cmd/start.go
+++ b/cmd/vcluster/cmd/start.go
@@ -7,6 +7,7 @@ import (
 	"runtime/debug"
 
 	"github.com/loft-sh/log"
+	"github.com/loft-sh/vcluster/pkg/cli/find"
 	"github.com/loft-sh/vcluster/pkg/config"
 	"github.com/loft-sh/vcluster/pkg/constants"
 	"github.com/loft-sh/vcluster/pkg/integrations"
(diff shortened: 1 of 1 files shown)


cmd/vcluster/cmd/start.go
```diff
@@ -90,15 +88,21 @@ func ExecuteStart(ctx context.Context, options *StartOptions) error {
...
 	if err != nil {
 		return err
 	}
-
-	logger := log.GetInstance()
-	// add a note for setting reuse-namespace config in v0.24 and a deprecation warning for multiple vcluster creation scenario
-	if len(vclusterServices) > 0 {
+	var vClusterExists bool
+	for _, v := range vClusters {
+		if v.Namespace == vConfig.ControlPlaneNamespace && v.Name != vClusterName {
```
neogopher, line 99: I see that we are already supplying vConfig.ControlPlaneNamespace to the find.ListVClusters function on line no. 93, so that it will limit the search to this namespace only. This makes the check for `v.Namespace == vConfig.ControlPlaneNamespace` redundant no?

ApsTomar: Yes I thought so too. There is a similar usage in CLI functionality ([reference](https://github.com/loft-sh/vcluster/blob/main/pkg/cli/create_helm.go#L137)). Since we're close to release, I leaned on the side of extra caution and chose to reuse it without any alteration (though it's bugging me as well!). 

Second reason is that, I'm working on the same part of code for next task as well so I'll test this part out with and without that namespace check and then remove it once I'm sure that it doesn't impact anything.

PR #2544, opened 2025-02-26
[v0.22] Update etcd to 3.5.17

ApsTomar:
/kind bugfix

resolves ENG-5330

Update etcd to 3.5.17


config/default_extra_values.go
```diff
@@ -68,6 +68,7 @@ var K8SSchedulerVersionMap = map[string]string{
 
 // K8SEtcdVersionMap holds the supported etcd
 var K8SEtcdVersionMap = map[string]string{
+	"1.32": "registry.k8s.io/etcd:3.5.17-0",
```
cbron, line 71: I don't believe we should be adding 1.32 support to 0.22, but I'm not positive.

ApsTomar: yes, we don't have 1.32 in v0.22, so does that mean we'll be changing the mapping for 1.31 for etcd from 3.5.15 to 3.5.17?
```"1.31": "registry.k8s.io/etcd:3.5.17-0",```

cbron: @hidalgopl or @lizardruss do you know if we should leave this as-is or upgrade it ? Ideally new clusters would start using 3.5.17, as the earlier versions had a consistency issue.

cbron: Looked into this a bit and I don't think it really makes a difference. But for now can you not add 1.32 and instead change the 1.31 version

ApsTomar: done, please take a look

cbron (approved):

PR #2563, opened 2025-03-06
feat(e2e): snapshot restore e2e tests

adriankabala:
/kind test

resolves #


test/e2e/snapshot/snapshot.go
```diff
@@ -117,15 +199,36 @@ var _ = ginkgo.Describe("Snapshot VCluster", func() {
...
-		_, err = f.VClusterClient.CoreV1().Services("default").Create(f.Context, &corev1.Service{
+	})
+
+	ginkgo.AfterAll(func() {
+		// delete the snapshot pvc
+		err := f.HostClient.CoreV1().PersistentVolumeClaims(pvc.Namespace).Delete(f.Context, pvc.Name, metav1.DeleteOptions{})
+		framework.ExpectNoError(err)
+	})
+
+	ginkgo.It("Restore should overwrite data", func() {
```
sowmyav27, line 222: Can we have the testcase say "Restore should restore resources in the vCluster" instead?

test/e2e/snapshot/snapshot.go
```diff
@@ -117,15 +199,36 @@ var _ = ginkgo.Describe("Snapshot VCluster", func() {
...
+		// now delete the serviceToRestore
+		err := f.VClusterClient.CoreV1().Services(defaultNamespace).Delete(f.Context, "snapshot-restore", metav1.DeleteOptions{})
```
sowmyav27, line 224: Can we have two testcases and not make one so complicated with creation/deletion/etc ?

**Testcase#1 --> ginkgo.It("Verify if only the resources in snapshot are available in vCluster after restore", func()**
1. Create vCluster
2. Create service, configmap, deployment, secret
3. Take snapshot
4. Create more service, configmap, deployment, secret
5. Restore, verify only resources from Step#2 are available

**Testcase#2 --> ginkgo.It("verify if deleted resources are recreated in vCluster after restore", func()**
1. Create vCluster
2. Create service, configmap, deployment, secret
3. Take snapshot
4. Delete all resources created in Step#2
5. Restore, verify all resources in Step#2 are recreated.

And also This testcase --> `ginkgo.It("Check if expected resources are available after restore", func() {` Should all be part of the two testcases, as that will make it complete e2e test.

sowmyav27 (changes requested):

sowmyav27 (changes requested):

sowmyav27 (approved):
Can you squash your commits pls. Otherwise lgtm

PR #2573, opened 2025-03-10
fix(bump): Bump oauth2

ulyssessouza:
/kind bugfix

Bump on oauth2 version


go.mod
```diff
@@ -170,7 +170,7 @@ require (
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/protobuf v1.5.4 // indirect
 	github.com/google/btree v1.1.3 // indirect
-	github.com/google/go-cmp v0.6.0 // indirect
+	github.com/google/go-cmp v0.6.0
```
nprokopic, line 173: Newer 0.7.0 is also available https://github.com/google/go-cmp/releases/tag/v0.7.0

nprokopic: Probably worth bumping to the latest, while you're at it 😬

ulyssessouza: This is a funny one... I actually didn't change it manually. 
It comes from a `go mod tidy`. and it's not a bump, it's just the `//indirect` comment that was removed.

nprokopic: Funny one indeed 🤔 In that case IMO just leave it as it :)

Issue #2591, opened 2025-03-14
Improve timeout handling for setup.Initialize


mayooot:
### Is your feature request related to a problem?

We are currently using vcluster in our production environment, and we encountered an occasional problem where the `<vcluster-id>-certs` secret was not created after the vcluster was created, which ultimately resulted in apiserver, controller-manager, and etcd not being Running.

Our guess is that it is a k8s cluster network issue that causes the syncer container to fail when creating the secret.

<img width="1470" alt="Image" src="https://github.com/user-attachments/assets/a04deb8f-f211-4cc4-a3ad-9a7ab827193d" />

### Which solution do you suggest?

We modified the Initialize method in `pkg/setup/initialize.go` to set the timeout to 5 minutes, which I think alleviates the problem.

And we created a 30s context in the initialize method, mainly to solve the problem of calling k8s resources due to network problems that cause the request to get stuck, and the `PollUntilContextTimeout` can't go to the next loop.

zerbitx:
The polling has been removed, so the timeout(s) shouldn't be needed.  Feel free to re-open if you're still seeing this issue.   See the comment on the [PR](https://github.com/loft-sh/vcluster/pull/2592?notification_referrer_id=NT_kwDOAAGIm7IxNTMwNDk1NTMwNjoxMDA1MDc&notifications_query=reason%3Amention#issuecomment-3254812632) as well.

PR #2592, opened 2025-03-14
fix(setup): improve timeout handling for certificate generation

mayooot:

/kind feature

https://github.com/loft-sh/vcluster/issues/2591

- Increase overall timeout in Initialize function from 2 to 5 minutes
- Add 30-second timeout contexts for initialize

### DIFF
--- a/pkg/setup/initialize.go
+++ b/pkg/setup/initialize.go
@@ -29,10 +29,12 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const initializeTimeout = 5 * time.Minute
+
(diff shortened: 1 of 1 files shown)


mayooot:
@zerbitx @FabianKramm Could you please review  this?

zerbitx:
@mayooot Thanks for the contribution!  Some recent changes have removed the polling and gone with a strategy of allowing it to fail on the first error, rather than polling.  The timeouts shouldn't be needed anymore.  I'm going to close this PR and the issue, but please reopen and tag me if you find it still not working as you would expect.

PR #2593, opened 2025-03-14
feat: warn of experimental feature changes

zerbitx:
/kind bugfix
/kind enhancement

resolves #ENG-6136

New Behavior:

 - If the current configuration as sleepMode under experimental and they're upgrading to 0.24+
   - Show advice on how to update their config
   - Prompt if they'd like to proceed, warning that the experimental feature will be lost
   - Fail if they choose no
   - Proceed if they agree
   
### Note, only prompts in interactive terminals.  
In non-interactive terminals it warns, but proceeds otherwise.  Without proceeding, if it just returned the error they would be incapable of upgrading since their already deployed values won't validate against a new schema.   

Example Usage:

### DIFF
--- a/config/experimental_helper.go
+++ b/config/experimental_helper.go
@@ -1,0 +1,66 @@
+package config
+
+import (
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/loft-sh/log"
+	"github.com/loft-sh/vcluster/pkg/upgrade"
+	"golang.org/x/mod/semver"
+)
+
+type (
+	ExperimentalConfig struct {
+		Experimental map[string]any
+	}
+)
+
+var advisors = map[string]func() (warning string){
+	"sleepMode": SleepModeWarning,
+}
+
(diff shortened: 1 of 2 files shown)


config/experimental_helper.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+type (
+	ExperimentalConfig struct {
+		Experimental map[string]any
+	}
+)
+
+var advisors = map[string]func(exp *ExperimentalConfig) (warning string){
+	"sleepMode": SleepModeWaring,
```
lizardruss, line 19:
```suggestion
	"sleepMode": SleepModeWarning,
```

lizardruss (commented):
Tentative approval of the code (with one small correction), but I didn't see any comments on the linear issue if this is the solution we want to move forward with.

lizardruss (dismissed):
LGTM

lizardruss (approved):

PR #2596, opened 2025-03-17
[ENG-5979] refactor(assets): Refactor assets listing

ulyssessouza:
/kind enhancement

resolves #ENG-5979

- Unifies the image assets listing executables.
- Add a `--latest` option to generate a minimal list for just the latest version of the distro matrix.
- Add a `--optional` option to get the images not listed in `--latest`.
- Fixes the problem os not enumerating all kube versions by linking the `Justfile` with the versions map in the code.


cbron (commented):
Looks like right track, I'll review once ready, would be good to get someone else to as well.

Justfile
```diff
@@ -51,19 +51,30 @@ copy-assets:
...
+generate-vcluster-optional-images version="0.0.0":
+  go run -mod vendor ./hack/assets/cmd/main.go --optional {{ version }} > ./release/vcluster-images-optional.txt
 
 # Generate versioned vCluster image files for multiple versions and distros
 [private]
 generate-matrix-specific-images version="0.0.0":
   #!/usr/bin/env bash
 
-  distros=("k8s" "k3s" "k0s")
-  versions=("1.30" "1.29" "1.28")
```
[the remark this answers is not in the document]
ulyssessouza: Yes! So the idea here is to manage that from the code and the CI will follow.

Justfile
```diff
@@ -51,19 +51,30 @@ copy-assets:
 # Generate the vcluster images file
 [private]
 generate-vcluster-images version="0.0.0":
-  go run -mod vendor ./hack/assets/main.go {{ version }} > ./release/vcluster-images.txt
+  go run -mod vendor ./hack/assets/cmd/main.go {{ version }} > ./release/vcluster-images.txt
+
+# Generate the vcluster latest/minimal images file
+[private]
+generate-vcluster-latest-images version="0.0.0":
+  go run -mod vendor ./hack/assets/cmd/main.go --latest {{ version }} > ./release/vcluster-images-latest.txt
```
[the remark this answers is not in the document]
ulyssessouza: Yes. That's also built in the release.
What I imagined is that the user would use this one when having a simple deployment that uses the latest images, while the above (the full one) when the user can make some tries. To be able to change/downgrade the versions in case of a bug in the latest version.

I just don't know how realistic this scenario is 🫤

Justfile
```diff
@@ -51,19 +51,30 @@ copy-assets:
 # Generate the vcluster images file
 [private]
 generate-vcluster-images version="0.0.0":
-  go run -mod vendor ./hack/assets/main.go {{ version }} > ./release/vcluster-images.txt
+  go run -mod vendor ./hack/assets/cmd/main.go {{ version }} > ./release/vcluster-images.txt
```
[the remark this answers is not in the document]
cbron: Correct, thanks

hack/assets/assets.go
```diff
@@ -0,0 +1,214 @@
...
+			selectedImages = append(selectedImages, sortedImages[1:]...)
+			continue
+		}
+		selectedImages = append(selectedImages, sortedImages[:]...)
+	}
+	return selectedImages
+}
+
+// GetVclusterProImages returns a list of vcluster images
+func GetVclusterProImages(cleanTag string) []string {
```
cbron, line 154: > The -pro image should stay for vcluster, while the vcluster one should be removed and the vcluster-oss one should be included on optional one

I oddly don't see the normal one here ? But it's getting generated somewhere. We also need the `-oss` one.

Justfile
```diff
@@ -48,22 +50,26 @@ copy-assets:
...
+[private]
+generate-vcluster-latest-images version="0.0.0":
+  {{ASSETS_RUN}} --latest {{ version }} > ./release/images.txt
+
+# Generate the vcluster optional images file
 [private]
-generate-vcluster-images version="0.0.0":
-  go run -mod vendor ./hack/assets/main.go {{ version }} > ./release/vcluster-images.txt
+generate-vcluster-optional-images version="0.0.0":
+  {{ASSETS_RUN}} --optional {{ version }} > ./release/images-optional.txt
```
cbron, line 61: I was originally thinking optional would have latest+optional, but I think this way is good too.

Justfile
```diff
@@ -9,6 +9,8 @@ GOBIN := env("GOBIN", `go env GOPATH`+"/bin")
 DIST_FOLDER := if GOARCH == "amd64" { "dist/vcluster_linux_amd64_v1" } else if GOARCH == "arm64" { "dist/vcluster_linux_arm64_v8.0" } else { "unknown" }
```
cbron, line 9: All the places you'll need to change the filename and docs to match this. [This search lists 4](https://github.com/search?q=repo%3Aloft-sh%2Fvcluster%20vcluster-images.txt&type=code), but I also see one in contributing.md.

cbron (commented):
Mostly lgtm. Few more things added.

hack/assets/assets.go
ulyssessouza: Done

Justfile
ulyssessouza: Added that

cbron (approved):
lgtm

johannesfrey (approved):

PR #2600, opened 2025-03-17
fix: Fix justfile for build-dev-image

ulyssessouza:
/kind bugfix

Fixed the `just` build target `build-dev-image` when building on a darwin/arm64 (MacbookPro M4)

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -150,7 +150,7 @@ gen-license-report:
 build-dev-image tag="":
   TELEMETRY_PRIVATE_KEY="" goreleaser build --snapshot --clean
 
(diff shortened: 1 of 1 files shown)


cbron:
Can you try `just build-snapshot` as well and see if that works for you. On my M3 the dev-image command works (without this fix) but build-snapshot does not.

ulyssessouza:
@cbron Hmmm... `just snapshot` works for me. Also from what I see it doesn't have any `sed` that is missing the last part. That can be another thing. 
Can you provide the error so I can debug it?

cbron (approved):
Nvm. Goreleaser seems to have changed its folder from arm64 to arm64_v8, once I upgraded those ones worked.

cbron (approved):

PR #2618, opened 2025-03-21
fix: respect platform config insecure value on login

zerbitx:
/kind bugfix

resolves #ENG-6079

Fixed an issue where vcluster would always login with insecure when starting the platform rather than abiding by the platform config


johannesfrey (approved):
lgtm

PR #2620, opened 2025-03-24
Delete previously replicated services after disabling from host syncing

nprokopic:
/kind bugfix

towards #ENG-5567

Fixed an issue where vcluster does not delete services that were synced from host to virtual, but should not be synced anymore, because from host syncing for services has been disabled.

How it's tested:
- Added unit tests
- Added e2e test
- Tested manually with the following config that enables syncing services from host:
```
networking:
  replicateServices:
    fromHost:
    - from: host-hello/my-host-service
      to: virtual-hello/my-virtual-service
```

After removing this config the synced service has been deleted from the virtual cluster.

### DIFF
--- a/pkg/setup/cleanup.go
+++ b/pkg/setup/cleanup.go
@@ -3,29 +3,28 @@ package setup
 import (
 	"fmt"
 
+	"github.com/loft-sh/vcluster/config"
 	"github.com/loft-sh/vcluster/pkg/syncer/synccontext"
 	"github.com/loft-sh/vcluster/pkg/util/translate"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // deletePreviouslySyncedResources deletes resources that were synced from host to virtual, but
 // should not be synced anymore, because from host syncing has been disabled.
 func deletePreviouslySyncedResources(ctx *synccontext.ControllerContext) error {
-	err := deletePreviouslySyncedServices(ctx)
+	err := deletePreviouslyReplicatedServices(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to delete previously synced services: %w", err)
 	}
 	return nil
 }
 
-// deletePreviouslySyncedServices deletes services that were synced from host to virtual, but
+// deletePreviouslyReplicatedServices deletes services that were synced from host to virtual, but
 // should not be synced anymore, because from host syncing for services has been disabled.
-func deletePreviouslySyncedServices(ctx *synccontext.ControllerContext) error {
-	if len(ctx.Config.Networking.ReplicateServices.FromHost) > 0 {
-		return nil
-	}
+func deletePreviouslyReplicatedServices(ctx *synccontext.ControllerContext) error {
 	virtualClient := ctx.VirtualManager.GetClient()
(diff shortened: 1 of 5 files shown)


nprokopic:
Converted back to draft, as a more proper solution (for all resources) is in progress.

pkg/setup/cleanup.go
```diff
@@ -0,0 +1,56 @@
...
+		return nil
+	}
+	virtualClient := ctx.VirtualManager.GetClient()
+	listOptions := client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(labels.Set{
+			translate.ControllerLabel: "vcluster",
+		}),
+	}
+	previouslySyncedServices := &corev1.ServiceList{}
+	err := virtualClient.List(ctx, previouslySyncedServices, &listOptions)
```
hidalgopl, line 36: I believe this entire block can be simplified massively if you use:
```go
err := virtualClient.DeleteAllOf(controllerContext, client.MatchingLabels{translate.ControllerLabel: "vcluster"})
```

hidalgopl (commented):
Could we test it also in E2E? I think we could create a service(s) with such labels in `.github/workflows/e2e.yaml` (and `Justfile` e2e target) and then in one of the existing e2e tests we can check if it got deleted

pkg/setup/cleanup.go
nprokopic: I did that first :) It failed because it did not have permissions to delete a list of services, and expanding the role to be able to delete a list sounded like a recipe for potential issues 🙈 So I deleted them one by one.

nprokopic:
> Could we test it also in E2E? I think we could create a service(s) with such labels in `.github/workflows/e2e.yaml` (and `Justfile` e2e target) and then in one of the existing e2e tests we can check if it got deleted

Yeah, definitely. I'm back to working on an actual proper fix for cleaning up all resources, where I will also include e2e tests for this. This PR is now more of a backup fix (which is why I converted it back to draft), in case we need to release this more quickly. If we need to merge this one, I'll add tests here as well.

pkg/setup/cleanup.go
hidalgopl: ok, could you then not return an error if one of the delete calls fails? I'd instead collect possible errors in the slice `var errs []error` and then return `errors.Join(errs)`

This way we will do the best-effort cleanup

nprokopic: ✅

nprokopic:
> Could we test it also in E2E? I think we could create a service(s) with such labels in `.github/workflows/e2e.yaml` (and `Justfile` e2e target) and then in one of the existing e2e tests we can check if it got deleted

@hidalgopl if I would like to create some services (for which I want to test that they are cleaned up), can I add service manifests here? https://github.com/loft-sh/vcluster/blob/14580942e1a2188ccd1cf57302d440061a3dd69b/test/e2e/values.yaml#L4

nprokopic:
@hidalgopl adding service here worked https://github.com/loft-sh/vcluster/blob/ff9f8140ab1af4a225bc1c7a90a84feb4ed95396/test/e2e/values.yaml#L13

I have added a service that looks like it has been replicated (it has `vcluster.loft.sh/controlled-by: vcluster` label). There is no replication config for that service, so it gets cleaned up.

I checked the e2e tests logs, the (fake replicated) service is created initially and then deleted by the new cleanup functionality that is added in this PR.

FabianKramm (approved):

johannesfrey (approved):

PR #2623, opened 2025-03-25
feat: Try secure login first, and only overwrite platform config if --insecure is specified

zerbitx:
/kind enhancement

resolves #ENG-6079

### The intended behavior:

  * If cert works, only write insecure if explicitly provided via --insecure flag
  * If cert does not work,
    * and platform config has insecure configured, write insecure in file
    * and platform config has nothing configured, fail and tell user that --insecure flag is required to be used

### The changes
 Write to the config whatever was specified on the command line, defaulting to false and always makes an attempt at a secure login, even if it ultimately falls back to insecure and/or writes `insecure: true` to the platform config allowing insecure logins to virtual clusters as intended by a user.

⚠️ Currently insecure defaults to true with platform login and a user must specify `--insecure=false` this flips that.


johannesfrey (approved):
lgtm

PR #2625, opened 2025-03-26
Align host to virtual metadata sync across all resources

nprokopic:
/kind enhancement

Towards #ENG-5567

Aligned how metadata is synced from host to virtual resources. Now all virtual resources that are synced from host have `vcluster.loft.sh/managed-by` label.

This PR basically doesn't add anything new, but it just (mostly) aligns how metadata is synced from host to virtual for different resource types. ~15% of are code changes, while ~85% are test additions/changes.

Here are the additions and changes:

🎯 The main change - now all virtual resources that are synced from host have `vcluster.loft.sh/managed-by` label. This enables us to then look up resources that are synced from host when we have to clean them up (after from host syncing has been disabled).

➕ New func `translate.SyncHostMetadataToVirtual(...)` has been added to `translate` package. Almost all resources now use this function to sync metadata from host to virtual. It is essentially extended version of the existing `ApplyMetadata()` func, which was already used for nodes metadata syncing (now with few more annotations and labels set).

➕ As a consequence of almost all FromHost syncers reusing the existing `ApplyMetadata(...)` func (through the new `SyncHostMetadataToVirtual`), now all virtual resources that are synced from host also have these annotations (like `Nodes` until now):
- `vcluster.loft.sh/managed-annotations` which shows which annotations have been synced from host resource
- `vcluster.loft.sh/managed-labels` which shows which annotations have been synced from host resource

ℹ️ `Nodes` FromHost syncer has minor refactoring changes, where `vcluster.loft.sh/managed-by` is not set in the syncer code, since it's now set in the `translate.SyncHostMetadataToVirtual(...)` func.

➕ The following FromHost syncers (and their tests) have been updated to use the new `translate.SyncHostMetadataToVirtual(...)` func and therefore set `vcluster.loft.sh/managed-by` label and `vcluster.loft.sh/managed-annotations` and `vcluster.loft.sh/managed-labels` annotations in the virtual resource after syncing it from host:
  - `CSIDrivers`
  - `CSINodes`
  - `IngressClasses`
  - `StorageClasses`
  - `Events`
  - generic `FromHostSyncer` implementation (used for `ConfigMaps` and `Secrets`)
  - **TODO**: Custom resources

🧪 ⚠️ `CSIStorageCapacities` FromHost syncer only has tests updated, because it currently has metadata syncing implementation which is different than `Nodes` (and almost all other resources), and changing it would be a breaking change. Since the current implementation looks suspicious (it syncs metadata from host to virtual with functions that are used for syncing the other way around - from virtual to host), I will check if this should be fixed.

🧪 New tests have been added for the following from host syncers (as they were missing before):
  - `StorageClasses`
    - Test syncing new host resource to virtual
    - Test syncing host resource changes to virtual
    - Test syncing deleted host resource to virtual
  - generic `FromHostSyncer` implementation (used for `ConfigMaps` and `Secrets`)
    - Test syncing new host resource to virtual
    - Test syncing host resource changes to virtual
    - Test syncing deleted host resource to virtual

🧪 Previously, the `vcluster.loft.sh/managed-by` label was being ignored in tests (it was stripped down before comparing existing to expected resources). Now the `vcluster.loft.sh/managed-by` label is not stripped down and it is checked in tests, since this label is functionally important (it was used also before this PR in some places to determine if the resource was synced from host to virtual). The syncer controller tests and the following syncer tests have been updated to include the label in their test code:
- ConfigMap
- Endpoints
- Nodes
- PriorityClasses
- RuntimeClasses
- Secrets
- ServiceAccounts
- StorageClasses
- VolumeSnapshotContents

### DIFF
--- a/pkg/controllers/resources/nodes/syncer.go
+++ b/pkg/controllers/resources/nodes/syncer.go
@@ -307,14 +307,6 @@ func (s *nodeSyncer) Sync(ctx *synccontext.SyncContext, event *synccontext.SyncE
 
 	s.translateUpdateBackwards(event.Host, event.Virtual)
 
-	// Set the marker of managed-by vcluster so that
-	// we skip deleting the nodes which are not managed
-	// by vcluster in `SyncToHost` function
-	if len(event.Virtual.Labels) == 0 {
-		event.Virtual.Labels = map[string]string{}
-	}
-	event.Virtual.Labels[translate.MarkerLabel] = translate.VClusterName
-
 	return ctrl.Result{}, nil
 }
 
--- a/pkg/controllers/resources/nodes/translate.go
+++ b/pkg/controllers/resources/nodes/translate.go
@@ -29,7 +29,7 @@ func (s *nodeSyncer) translateUpdateBackwards(pNode *corev1.Node, vNode *corev1.
 	applyMetadataOptions := translate.ApplyMetadataOptions{
 		ExcludeAnnotations: []string{TaintsAnnotation, RancherAgentPodRequestsAnnotation, RancherAgentPodLimitsAnnotation},
 	}
-	translate.ApplyVirtualMetadata(pNode, vNode, applyMetadataOptions)
+	translate.SyncHostMetadataToVirtual(pNode, vNode, applyMetadataOptions)
 
 	// merge taints together
 	oldPhysical := []string{}
--- a/pkg/controllers/resources/nodes/translate.go
+++ b/pkg/controllers/resources/nodes/translate.go
@@ -26,8 +26,10 @@ var (
 func (s *nodeSyncer) translateUpdateBackwards(pNode *corev1.Node, vNode *corev1.Node) {
 	// merge labels & taints
 	translatedSpec := pNode.Spec.DeepCopy()
-	excludeAnnotations := []string{TaintsAnnotation, RancherAgentPodRequestsAnnotation, RancherAgentPodLimitsAnnotation}
-	labels, annotations := translate.ApplyMetadata(pNode.Annotations, vNode.Annotations, pNode.Labels, vNode.Labels, excludeAnnotations...)
+	applyMetadataOptions := translate.ApplyMetadataOptions{
+		ExcludeAnnotations: []string{TaintsAnnotation, RancherAgentPodRequestsAnnotation, RancherAgentPodLimitsAnnotation},
+	}
+	translate.ApplyVirtualMetadata(pNode, vNode, applyMetadataOptions)
 
 	// merge taints together
 	oldPhysical := []string{}
@@ -90,13 +92,10 @@ func (s *nodeSyncer) translateUpdateBackwards(pNode *corev1.Node, vNode *corev1.
 	if err != nil {
 		klog.Errorf("error encoding taints: %v", err)
 	} else {
-		if annotations == nil {
-			annotations = map[string]string{}
-		}
 		if len(physical) > 0 {
-			annotations[TaintsAnnotation] = string(out)
+			vNode.Annotations[TaintsAnnotation] = string(out)
 		} else {
-			delete(annotations, TaintsAnnotation)
+			delete(vNode.Annotations, TaintsAnnotation)
 		}
 	}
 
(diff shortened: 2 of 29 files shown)


pkg/controllers/resources/nodes/syncer.go
```diff
@@ -307,14 +307,6 @@ func (s *nodeSyncer) Sync(ctx *synccontext.SyncContext, event *synccontext.SyncE
 
 	s.translateUpdateBackwards(event.Host, event.Virtual)
 
-	// Set the marker of managed-by vcluster so that
```
nprokopic, line 310: This is now done in the `translate.SyncHostMetadataToVirtual(...)` func (that the nodes syncer is calling).

pkg/controllers/resources/nodes/translate.go
```diff
@@ -26,8 +26,10 @@ var (
 func (s *nodeSyncer) translateUpdateBackwards(pNode *corev1.Node, vNode *corev1.Node) {
 	// merge labels & taints
 	translatedSpec := pNode.Spec.DeepCopy()
-	excludeAnnotations := []string{TaintsAnnotation, RancherAgentPodRequestsAnnotation, RancherAgentPodLimitsAnnotation}
-	labels, annotations := translate.ApplyMetadata(pNode.Annotations, vNode.Annotations, pNode.Labels, vNode.Labels, excludeAnnotations...)
+	applyMetadataOptions := translate.ApplyMetadataOptions{
+		ExcludeAnnotations: []string{TaintsAnnotation, RancherAgentPodRequestsAnnotation, RancherAgentPodLimitsAnnotation},
+	}
+	translate.SyncHostMetadataToVirtual(pNode, vNode, applyMetadataOptions)
```
nprokopic, line 32: `translate.SyncHostMetadataToVirtual(...)` calls `translate.ApplyMetadata(...)` and also sets annotations and labels, which is why the code on the left side has been removed.

pkg/controllers/resources/storageclasses/syncer_test.go
```diff
@@ -32,7 +32,7 @@ func TestSync(t *testing.T) {
 			Name:            translate.Default.HostNameCluster(vObjectMeta.Name),
 			ResourceVersion: syncertesting.FakeClientResourceVersion,
 			Labels: map[string]string{
-				translate.MarkerLabel: translate.VClusterName,
+				translate.MarkerLabel: translate.SafeConcatName(testingutil.DefaultTestTargetNamespace, "x", translate.VClusterName),
```
nprokopic, line 35: Previous check (`translate.MarkerLabel: translate.VClusterName`) was incorrect, but since the tests were stripping down and not checking the `vcluster.loft.sh/managed-by` label, the was not catching the error in the test here.

pkg/syncer/testing/testing.go
```diff
@@ -225,14 +225,6 @@ func stripObject(obj runtime.Object) runtime.Object {
 		accessor.SetAnnotations(nil)
 	}
 
-	l := accessor.GetLabels()
-	if l != nil {
-		delete(l, "vcluster.loft.sh/managed-by")
```
nprokopic, line 230: All the tests have been updated to check for `vcluster.loft.sh/managed-by` label.

pkg/util/translate/translate.go
```diff
@@ -252,6 +254,44 @@ func ResetObjectMetadata(obj metav1.Object) {
...
+	labels[MarkerLabel] = VClusterName
+	if options.SetHostNameAndNamespaceAnnotations {
+		annotations[HostNameAnnotation] = pObj.GetName()
+		if pObj.GetNamespace() != "" {
+			annotations[HostNamespaceAnnotation] = pObj.GetNamespace()
+		}
+	}
+	vObj.SetAnnotations(annotations)
+	vObj.SetLabels(labels)
+}
```
FabianKramm, line 293: Unfortunately we cannot use this approach as we cannot rely on labels or annotations within the vCluster as the end user can freely change this

FabianKramm (changes requested):
See my comment above

nprokopic:

PR #2630, opened 2025-03-30
fix call errors.Wrap with nil value error

alingse:
/kind bugfix

resolves #


johannesfrey (approved):
Thx!

PR #2641, opened 2025-04-03
Add basic ServiceSyncer unit tests

nprokopic:
/kind test

towards #ENG-5567

Add basic unit tests for ServiceSyncer.

This PR is adding basic unit tests that would serve as a basic safety net when making changes needed to fix #ENG-5567. It's not trying to cover all possible service replication scenarios and there are definitely more tests that should be added.

Work on #ENG-5567 got stuck (pending decision on how to proceed forward), but these tests I wrote while doing that work add value on their own, so opening a PR here to add them, regardless of how we proceed with work on #ENG-5567.

### DIFF
--- a/pkg/controllers/servicesync/servicesync_test.go
+++ b/pkg/controllers/servicesync/servicesync_test.go
@@ -7,7 +7,11 @@ import (
 	syncertesting "github.com/loft-sh/vcluster/pkg/syncer/testing"
 	"github.com/loft-sh/vcluster/pkg/util/loghelper"
 	testingutil "github.com/loft-sh/vcluster/pkg/util/testing"
+	"github.com/loft-sh/vcluster/pkg/util/translate"
 	"gotest.tools/v3/assert"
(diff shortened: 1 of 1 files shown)


pkg/controllers/servicesync/servicesync_test.go
```diff
@@ -0,0 +1,188 @@
...
+				compareErr := syncertesting.CompareObjs(
+					fakeContext,
+					t,
+					testCase.Name+" virtual state",
+					fakeContext.VirtualManager.GetClient(),
+					serviceGVK,
+					scheme.Scheme,
+					testCase.ExpectedVirtualServices,
+					nil)
+				if err != nil {
```
zerbitx, line 182: Should be checking `compareErr`

nprokopic: 🙈 fixed it

nprokopic:
Please squash & merge if all is good :)

PR #2652, opened 2025-04-07
bugfix: fix the incorrect calculation of node resources when sidecar containers are being injected during sync to host with virtual scheduler enabled

neogopher:
/kind bugfix

resolves ENG-6278


cbron:
@neogopher can this be closed ?

PR #2654, opened 2025-04-07
feat: auto-migrate k3s to k8s

FabianKramm:
/kind feature


pkg/setup/initialize.go
```diff
@@ -31,8 +31,14 @@ import (
 
 // Initialize creates the required secrets and configmaps for the control plane to start
 func Initialize(ctx context.Context, options *config.VirtualClusterConfig) error {
+	// migrate k3s to k8s if needed
+	err := k8s.MigrateK3sToK8s(ctx, options.ControlPlaneClient, options.ControlPlaneNamespace, options)
```
hidalgopl, line 35: should we run this if embedded etcd is enabled? I think we already have a data migration logic for sqlite - > embedded etcd in the pro.

FabianKramm: yeah this is about migrating the distro not the backing store, so this wouldn't cause any issues

PR #2657, opened 2025-04-08
when CLI creates platform credentials secret, set vCluster service as…

hidalgopl:
… an owner, so it gets cleaned up when virtual cluster is deleted

/kind bugfix

part of ENG-6304

Fixed an issue where vcluster platform secret is not deleted when virtual cluster no longer exists

### DIFF
--- a/pkg/platform/secret.go
+++ b/pkg/platform/secret.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/aws/smithy-go/ptr"
 	"github.com/blang/semver"
 	managementv1 "github.com/loft-sh/api/v4/pkg/apis/management/v1"
 	storagev1 "github.com/loft-sh/api/v4/pkg/apis/storage/v1"
@@ -23,6 +22,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/ptr"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -127,8 +127,8 @@ func ApplyPlatformSecret(
 					Kind:               "Service",
 					Name:               vClusterService.Name,
 					UID:                vClusterService.UID,
-					Controller:         ptr.Bool(true),
-					BlockOwnerDeletion: ptr.Bool(false),
+					Controller:         ptr.To(true),
+					BlockOwnerDeletion: ptr.To(false),
 				},
 			}
 		}
--- a/pkg/platform/secret.go
+++ b/pkg/platform/secret.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/smithy-go/ptr"
 	"github.com/blang/semver"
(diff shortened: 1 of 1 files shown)


pkg/platform/secret.go
```diff
@@ -105,7 +106,9 @@ func ApplyPlatformSecret(
 	if err != nil && !kerrors.IsNotFound(err) {
 		return fmt.Errorf("error getting platform secret %s/%s: %w", namespace, DefaultPlatformSecretName, err)
 	} else if kerrors.IsNotFound(err) {
-		_, err = kubeClient.CoreV1().Secrets(namespace).Create(ctx, &corev1.Secret{
+		// try to find vCluster service
+		vClusterService, svcErr := kubeClient.CoreV1().Services(namespace).Get(ctx, name, metav1.GetOptions{})
+		secret := &corev1.Secret{
```
jjaferson, line 111: nit: maybe it would be better to first instantiate the secret struct and then retrieve the service. It would make the code a bit more readable and it's also a good practise handling error right after they occur.

pkg/platform/secret.go
```diff
@@ -114,7 +117,23 @@ func ApplyPlatformSecret(
 				},
 			},
 			Data: secretPayload,
-		}, metav1.CreateOptions{})
+		}
+		if svcErr == nil {
+			// set virtual cluster service as an owner for platform secret, so it can be cleaned up by kubernetes GC controller
+			// once virtual cluster is uninstalled
+			secret.OwnerReferences = []metav1.OwnerReference{
```
jjaferson, line 124: nit: we could also use SetOwnerReference from the [controllerutil library ](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.20.4/pkg/controller/controllerutil/controllerutil.go#L110)

jjaferson (commented):
just left two nit but the change lgmt

pkg/platform/secret.go
```diff
@@ -114,7 +117,23 @@ func ApplyPlatformSecret(
 				},
 			},
 			Data: secretPayload,
-		}, metav1.CreateOptions{})
+		}
+		if svcErr == nil {
```
zerbitx, line 121: Would we not want to return an error if the service can't be found?  Also there's a [SetOwnerReference](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetOwnerReference) method which would make this look more like 

```go
if err := controllerutil.SetOwnerReference(vClusterService, secret, scheme.Scheme); err != nil {
  return err
}
```

pkg/platform/secret.go
hidalgopl: I initially used it, but then decided I want to set `BlockOwnerDeletion` to `false`, and in `SetOwnerReference` it's always set to true

pkg/platform/secret.go
hidalgopl: I initially used it, but then decided I want to set `BlockOwnerDeletion` to `false`, and in `SetOwnerReference` it's always set to true

pkg/platform/secret.go
hidalgopl: good point, moved

pkg/platform/secret.go
zerbitx: Not worth changing at this point, but you can also pass an option for that as `controllerutil.WithBlockOwnerDeletion(false)` after the scheme

zerbitx (approved):
lgtm

PR #2658, opened 2025-04-08
Add replicated services to mappings store

nprokopic:
> [!WARNING]
> Do not merge!

**What issue type does this pull request address?** (keep at least one, remove the others) 
/kind enhancement

**What does this pull request do? Which issues does it resolve?** (use `resolves #<issue_number>` if possible) 
towards #ENG-5567

**Please provide a short message that should be published in the vcluster release notes**
Add replicated services to mappings store.

**What else do we need to know?** 
This PR builds on top of #2641 and ATM the purpose of this PR is just to review the potential solutions for #ENG-5567.

PR #2659, opened 2025-04-08
Add option to skip host namespace verification when saving mappings

nprokopic:
> [!warning]
> Do not merge!

/kind enhancement

towards #ENG-5567

Added option to skip host namespace verification when saving mappings for replicated services.

Services can be replicated from any namespace on host/virtual cluster to any namespace on virtual/host cluster, so the existing host resource namespace verification does not apply in this case.

This PR builds on top of https://github.com/loft-sh/vcluster/pull/2658 and ATM the purpose of this PR is just to review the potential solutions for #ENG-5567.


nprokopic:

PR #2660, opened 2025-04-08
Add sync direction to mappings in store (only for replicated services)

nprokopic:
> [!warning]
> Do not merge!

/kind enhancement

towards #ENG-5567

Added sync direction to mappings in store (for replicated services).

When replication/sync config is removed fro vCluster config, we need to know which service to delete. By saving sync direction in the mappings store, we can know which service, host or virtual, is the original, and which one is the replica.

This PR builds on top of https://github.com/loft-sh/vcluster/pull/2659 and ATM the purpose of this PR is just to review the potential solutions for #ENG-5567.


nprokopic:

Issue #2545, opened 2025-02-26
Helm version check fails with Fedora packaged Helm


jawsper:
### What happened?

Running Fedora 41, with Fedora package `helm` installed (version `3.17.0-1.fc41`).

`helm version --client --template '{{.Version}}'` outputs `3.17.0`

Running `vcluster create ...` or `vcluster delete ...` fails with the error message:
```
fatal please ensure that the "helm" binary in your PATH is valid. Currently only Helm >= v3.10.0 is supported
```

### What did you expect to happen?

No error about `helm` not being valid.

### How can we reproduce it (as minimally and precisely as possible)?

On Fedora 41, install helm:
```
sudo dnf install helm
```
And check that the version is missing the `v` prefix:
```
helm version --client --template '{{.Version}}'
```

### Anything else we need to know?

From looking at the code, it seems that the `semver` package requires the semver to start with a `v` which is not the case for the helm I have installed.

### Host cluster Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.30.9-dispatcher
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.30.9-gke.1046000

```

</details>

### vcluster version

<details>

```console
$ vcluster --version
vcluster version 0.22.4
```

</details>

### VCluster Config

<details>

```
# My vcluster.yaml / values.yaml here
controlPlane:
  distro:
    k3s:
      enabled: true
  statefulSet:
    resources:
      limits:
        memory: 4Gi
    scheduling:
      podManagementPolicy: OrderedReady
    persistence:
      volumeClaim:
        size: 10Gi
sync:
  fromHost:
    ingressClasses:
      enabled: true
  toHost:
    ingresses:
      enabled: true
    serviceAccounts:
      enabled: true

```

</details>

cbron:
Looks like that is coming from here: https://github.com/loft-sh/vcluster/blob/main/pkg/util/clihelper/clihelper.go#L119-L125
And the custom version is throwing off the check. Feel free to submit a PR, otherwise I'll throw this in our backlog to fix.

PR #2663, opened 2025-04-09
Helm version check fails with Fedora packaged Helm

mayooot:
/kind bugfix

resolves https://github.com/loft-sh/vcluster/issues/2545

CheckHelmVersion function adapts version that do not begin with 'v'

nothing

### DIFF
--- a/pkg/util/clihelper/clihelper.go
+++ b/pkg/util/clihelper/clihelper.go
@@ -117,6 +117,10 @@ func HasPodProblem(pod *corev1.Pod) bool {
 }
 
 func CheckHelmVersion(output string) error {
+	if !strings.HasPrefix(output, "v") {
+		output = "v" + output
+	}
+
 	if semver.Compare(output, MinHelmVersion) == -1 {
 		return fmt.Errorf("please ensure that the \"helm\" binary in your PATH is valid. Currently only Helm >= %s is supported", MinHelmVersion)
 	}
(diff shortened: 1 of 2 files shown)


pkg/util/clihelper/clihelper.go
```diff
@@ -117,6 +117,10 @@ func HasPodProblem(pod *corev1.Pod) bool {
 }
 
 func CheckHelmVersion(output string) error {
+	if !strings.HasPrefix(output, "v") {
+		output = "v" + output
+	}
+
 	if semver.Compare(output, MinHelmVersion) == -1 {
```
cbron, line 124: Hello, thanks for sending a PR. Could you add some unit tests for this ? Would like to track the types of version that we are supporting.

pkg/util/clihelper/clihelper_test.go
```diff
@@ -0,0 +1,53 @@
...
+			wantErr: true,
+		},
+		{
+			name:    "equal to minimum version - with v prefix",
+			version: MinHelmVersion,
+			wantErr: false,
+		},
+		{
+			name:    "equal to minimum version - without v prefix",
+			version: MinHelmVersion[1:], // remove 'v' prefix
```
ulyssessouza, line 40: nit: Breaks if `MinHelmVersion` changes to something without 'v'. But that's small enough to be changed/removed by the future change.

ulyssessouza (approved):
LGTM

pkg/util/clihelper/clihelper_test.go
mayooot: You're right, but I think the current changes are adequate.

PR #2664, opened 2025-04-09
fix: document bi-directionality of secret stores

johannesfrey:
/kind documentation

part of ENG-6260

Fixed an issue where vcluster docs where unclear about SecretStore sync bi-directionality


chart/values.schema.json
```diff
@@ -1897,7 +1897,7 @@
         },
         "stores": {
           "$ref": "#/$defs/EnableSwitch",
-          "description": "Stores defines if secret stores should get synced from the virtual cluster to the host cluster."
+          "description": "Stores defines if secret stores should get synced bi-directionally after an initial sync from the virtual cluster to the host cluster."
```
cbron, line 1900: This phrasing is slightly confusing, it sounds like you are defining the bi-directional part only. How about: 
```suggestion
          "description": "Stores defines if secret stores should get synced from the virtual cluster to the host cluster and then bi-directionally."
```

cbron (approved):

Issue #2546, opened 2025-02-26
vcluster requires List namespace privilege


antoinetran:
### What happened?

With version 0.20.0-beta.1, vcluster deployed without creating the namespace.
With version >= 0.20.0, vcluster adds a hardcoded `--create-namespace`, which requires `oc` to list namespace.

In environment where this privilege is not given, because only one namespace is allowed, this ends in error:
```
85 [debug] getting history for release my-vcluster
Release "my-vcluster" does not exist. Installing it now.
install.go:225: 2025-02-26 17:34:45.504054107 +0000 UTC m=+0.303771631 [debug] Original chart version: ""
install.go:242: 2025-02-26 17:34:45.504142182 +0000 UTC m=+0.303859705 [debug] CHART PATH: /tmp/vcluster-0.23.0.tgz-4163262726

client.go:142: 2025-02-26 17:34:46.887421897 +0000 UTC m=+1.687139405 [debug] creating 1 resource(s)
Error: 1 error occurred:
        * namespaces is forbidden: User "XX" cannot create resource "namespaces" in API group "" at the cluster scope

                                                                                                                                                                                                                                                        helm.go:86: 2025-02-26 17:34:46.932141242 +0000 UTC m=+1.731858762 [debug] 1 error occurred:
        * namespaces is forbidden: User "XX" cannot create resource "namespaces" in API group "" at the cluster scope
```


### What did you expect to happen?

The `--create-namespace` is not set if vcluster does not set it.

### How can we reproduce it (as minimally and precisely as possible)?

1. deploy Kubernetes cluster (with KIND?)
2. creates namespace
3. creates a RBAC to restrict list namespace and a serviceAccounts
4. deploy vcluster using that serviceAccounts

### Anything else we need to know?

Environment is OKD.

### Host cluster Kubernetes version

<details>

```console
$ oc version
Client Version: v4.2.0-alpha.0-2103-g48dcf59
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Kubernetes Version: v1.28.4
```

</details>


### vcluster version

0.20.0-beta.1 => OK (no `--create-namespace`)
0.20.0 and 0.23.0 => KO (`--create-namespace`)

### VCluster Config

<details>

nothing relevant

</details>

antoinetran:
[...]

PR #2672, opened 2025-04-09
feat: Add `create-namespace` flag to `vcluster platform add cluster` command

mayooot:
/kind feature

resolves https://github.com/loft-sh/vcluster/issues/2546

### DIFF
--- a/cmd/vclusterctl/cmd/platform/add/cluster.go
+++ b/cmd/vclusterctl/cmd/platform/add/cluster.go
@@ -173,10 +173,12 @@ func (cmd *ClusterCmd) Run(ctx context.Context, args []string) error {
 		helmArgs = []string{
 			"upgrade", "--install", "loft", cmp.Or(os.Getenv("DEVELOPMENT_CHART_DIR"), "./chart"),
 			"--namespace", namespace,
-			"--create-namespace",
 			"--set", "agentOnly=true",
 			"--set", "image=" + cmp.Or(os.Getenv("DEVELOPMENT_IMAGE"), "ghcr.io/loft-sh/enterprise:release-test"),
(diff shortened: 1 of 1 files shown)


cbron (approved):

cmd/vclusterctl/cmd/platform/add/cluster.go
```diff
@@ -167,10 +169,13 @@ func (cmd *ClusterCmd) Run(ctx context.Context, args []string) error {
...
 	}
 
+	if cmd.CreateNamespace {
+		helmArgs = append(helmArgs, "--create-namespace")
+	}
+
 	if os.Getenv("DEVELOPMENT") == "true" {
 		helmArgs = []string{
 			"upgrade", "--install", "loft", cmp.Or(os.Getenv("DEVELOPMENT_CHART_DIR"), "./chart"),
-			"--create-namespace",
```
cbron, line 173: Actually this one needs to stay, because we are overwriting helmArgs here.

cbron (changes requested):

cmd/vclusterctl/cmd/platform/add/cluster.go
cbron: It doesn't need to be forced, but how the code was before the helmArgs isn't used inside DEV conditional, meaning you could not have turned it on. You can make it optional in this block, but needs to use the flags.

mayooot: ok, how about I add a `--dev-create-namespace` flag?

cbron: I think you misunderstand. My point is that this part: 
```
if cmd.CreateNamespace {
			helmArgs = append(helmArgs, "--create-namespace")
		}
```
Should take effect in both the development section above (around line 172) and the non-development section below (which it already does). The flag should work whether you are in dev or not.

mayooot: Ok, I understand that in development mode the `helmArgs` string slice is reinitialized.

cbron (approved):

Issue #2662, opened 2025-04-08
vcluster 0.24 upgrade of k3s cluster generates fatal error


mlbiam:
### What happened?

Deployed a k3s vcluster:

```
$ vcluster --version
vcluster version 0.24.1
$ vcluster create myvcluster --distro k3s -n tenant1
.
.
.
.
$ kubectl get pod myvcluster-0 -n tenant1 -o yaml | grep image | grep rancher
    image: rancher/k3s:v1.32.1-k3s1
    image: docker.io/rancher/k3s:v1.32.1-k3s1
    imageID: docker.io/rancher/k3s@sha256:2ce69284b7f28bd1bde9f8503f0eb7fe77943cdbe3a88564b73ed6aef8b180d3
```

Then run an upgrade with the following YAML:

```yaml
controlPlane:
  distro:
    k3s:
      enabled: true
  statefulSet:
    scheduling:
      podManagementPolicy: OrderedReady
sync:
  fromHost:
    ingressClasses:
      enabled: true
  toHost:
    ingresses:
      enabled: true
```

with the command:

```
$ vcluster create myvcluster --upgrade -f ./vcluster-values.yaml -n tenant1
18:30:29 fatal seems like you were using k8s as a distro before and now have switched to k3s, please make sure to not switch between vCluster distros
```

### What did you expect to happen?

Complete the upgrade

### How can we reproduce it (as minimally and precisely as possible)?

All steps provided above

### Anything else we need to know?

running on a single node:

```
Client Version: v1.31.0
Kustomize Version: v5.4.2
Server Version: v1.30.0
```

### Host cluster Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.31.0
Kustomize Version: v5.4.2
Server Version: v1.30.0
```

</details>

### vcluster version

<details>

```console
$ vcluster --version
vcluster version 0.24.1
```

</details>

### VCluster Config

<details>

see issue description
```

</details>

**Comment by cbron** [comment]
Thanks for filing. I can reproduce, and checking helm values k3s did get set correctly: 
```
helm get values myvcluster2 -n tenant2
USER-SUPPLIED VALUES:
controlPlane:
  distro:
    k3s:
      enabled: true
```
[...]

PR #2679, opened 2025-04-11
fix: fatal error when upgrading a different distro

jjaferson:
Fixes a fatal error when updating a vCluster with an an specific distro other than the default one

/kind bugfix

resolves ENG-6422

Fixed an issue where updates of vclusters with different distro would cause an fatal error 

fix https://github.com/loft-sh/vcluster/issues/2662


stevenyslins:
Need to fix this issue because when I run `vcluster create --upgrade ha-vcluster --namespace vcluster --connect=false -f values.yaml`, I encounter the following problem: `fatal seems like you were using embedded-database as a store before and now have switched to external-etcd, please make sure to not switch between vCluster stores`

@jjaferson Thank you for providing the fix.

pkg/cli/create_helm.go
```diff
@@ -322,9 +322,15 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 	if err != nil {
 		return err
 	}
+
 	verb := "created"
 	if isVClusterDeployed(release) {
 		verb = "upgraded"
+		currentVClusterConfig, err := getConfigfileFromSecret(ctx, vClusterName, cmd.Namespace)
```
zerbitx, line 329: The `:=` is shadowing both `currentVClusterConfig` and `err`.  Recommend using `=` instead.

jjaferson: Good catch, both of them have been declared already. I'll fix it ;)

zerbitx (approved):
lgtm

PR #2683, opened 2025-04-12
Update golangci lint

nprokopic:
/kind enhancement

Update golangci-lint to the new major version v2 and remove custom linter.

More details about changes in this PR:
- Migrated golangci-lint config to the new v2 version. This is done automatically with `golangci-lint migrate` command.
- Addded back (most of) config comments, since the above migrate command does not preserve them.
- Re-formatted the newly generated config a bit, in order to try to minimize the diff after the migration.
- Newer linter versions showed new errors:
  - I fixed linter errors where it was only required to rename function parameters, so there were no changes to the code logic.
  - I added exclusions for new `staticcheck` errors where, in order to fix them, it would be required to make changes to code logic, e.g. use switch instead of if/else, call different funcs with different params, change boolean logic, etc. We can make these changes in separate PRs.
- Removed custom nilaway linter as discussed in Slack.


johannesfrey (dismissed):
Thx for doing this! lgtm

.golangci.yml
```diff
@@ -32,86 +31,89 @@ linters:
     - revive
     - sloglint
     - staticcheck
-    - stylecheck
-    - typecheck
```
jjaferson, line 36: are we removing these two or are they covered by other linters?

nprokopic:
Looking into this failed test https://github.com/loft-sh/vcluster/actions/runs/14423431427/job/40448950455?pr=2683

nprokopic:
> Looking into this failed test https://github.com/loft-sh/vcluster/actions/runs/14423431427/job/40448950455?pr=2683

The test failure is unrelated to the change from this PR, but trying to improve it nonetheless, so it's more reliable https://github.com/loft-sh/vcluster/pull/2687.

johannesfrey (approved):

nprokopic:
Please squash & merge the PR if there aren't any other concerns here 🙏

PR #2695, opened 2025-04-17
add istio config

hidalgopl:
/kind feature

resolves #

### DIFF
--- a/chart/templates/role.yaml
+++ b/chart/templates/role.yaml
@@ -120,10 +120,10 @@ rules:
   {{- end }}
   {{- if .Values.integrations.istio.enabled }}
   - apiGroups: [ "networking.istio.io" ]
-    resources: [ "*" ]
+    resources: [ "destinationrules", "gateways", "virtualservices" ]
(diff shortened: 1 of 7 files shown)


chart/templates/role.yaml
```diff
@@ -118,6 +118,14 @@ rules:
     verbs: ["create", "delete", "patch", "update", "get", "list", "watch"]
   {{- end }}
   {{- end }}
+  {{- if .Values.integrations.istio.enabled }}
+  - apiGroups: [ "networking.istio.io" ]
+    resources: [ "destinationrules", "gateways", "virtualservices" ]
```
zerbitx, line 123: Can we add `serviceentries` here for sleep mode even though we're not syncing them yet?

zerbitx (dismissed):
lgtm.  Had one comment, but don't want you to be blocked.

chart/templates/role.yaml
hidalgopl: added

zerbitx (approved):

PR #2697, opened 2025-04-17
Fix vcluster delete CLI when "Prevent deletion" is enabled via platform

nprokopic:
/kind bugfix

fixes ENG-6256

Fixed an issue where `vcluster delete` command would delete virtual clusters with `loft.sh/non-deletable: true`.

This fix has certain limitations. See the following scenarios where those limitations are described.

**1\. You are logged into the host cluster (where the platform runs).**

✅ In this case `vcluster delete` can always see if "Prevent deletion" is enabled for any cluster, so command fails correctly for protected clusters.

In this scenario it also doesn't matter if you are logged into the platform or not, and platform version does not matter, because `vcluster delete` directly checks `VirtualClusterInstance` CR.

**2\. You are logged into the connected host cluster (the platform is running on a different host cluster).**

Here we have two different sub-cases:

- ✅ You are logged into the platform. In this case `vcluster delete` fails correctly for any cluster that has "Prevent deletion" enabled, because it first tries to delete the virtual cluster via platform, which fails as expected.

- ⚠️ When you are NOT logged into the platform, there are few additional conditions for the check to work.
    - ✅ Virtual cluster is created with platform version 4.3.0 or newer (or had its config updated with platform version 4.3.0 or newer). `vcluster delete` fails correctly, as it checks the StatefulSet/Deployment annotation that is added in platform version 4.3.0.
   - :x: Virtual cluster is created with platform version older than 4.3.0 (and didn't have its config updated with platform version 4.3.0 or newer). `vcluster delete` still incorrectly deletes the cluster, because it doesn't see neither VirtualClusterInstance CR nor the annotation that the platform 4.3.0 sets.

### DIFF
--- a/pkg/platform/kube/client.go
+++ b/pkg/platform/kube/client.go
@@ -13,7 +13,7 @@ import (
 
 type Interface interface {
 	kubernetes.Interface
-	ApiExtensions() apiextensions.Interface
+	APIExtensions() apiextensions.Interface
 	Loft() loftclient.Interface
 	Agent() agentloftclient.Interface
 }
@@ -54,7 +54,7 @@ type client struct {
 	agentLoftClient     agentloftclient.Interface
 }
 
-func (c *client) ApiExtensions() apiextensions.Interface {
+func (c *client) APIExtensions() apiextensions.Interface {
 	return c.apiExtensionsClient
 }
 
--- a/pkg/platform/kube/client.go
+++ b/pkg/platform/kube/client.go
@@ -6,14 +6,12 @@ import (
 
 	"github.com/pkg/errors"
 
-	apiextensions "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
 type Interface interface {
 	kubernetes.Interface
-	APIExtensions() apiextensions.Interface
 	Loft() loftclient.Interface
 	Agent() agentloftclient.Interface
 }
@@ -22,11 +20,6 @@ func NewForConfig(c *rest.Config) (Interface, error) {
 	kubeClient, err := kubernetes.NewForConfig(c)
 	if err != nil {
 		return nil, errors.Wrap(err, "create kube client")
-	}
-
-	apiExtensionsClient, err := apiextensions.NewForConfig(c)
-	if err != nil {
-		return nil, errors.Wrap(err, "create api extensions client")
 	}
 
 	loftClient, err := loftclient.NewForConfig(c)
@@ -40,22 +33,16 @@ func NewForConfig(c *rest.Config) (Interface, error) {
 	}
 
 	return &client{
-		Interface:           kubeClient,
-		apiExtensionsClient: apiExtensionsClient,
-		loftClient:          loftClient,
-		agentLoftClient:     agentLoftClient,
+		Interface:       kubeClient,
+		loftClient:      loftClient,
+		agentLoftClient: agentLoftClient,
 	}, nil
 }
 
 type client struct {
 	kubernetes.Interface
-	apiExtensionsClient apiextensions.Interface
-	loftClient          loftclient.Interface
-	agentLoftClient     agentloftclient.Interface
-}
-
-func (c *client) APIExtensions() apiextensions.Interface {
-	return c.apiExtensionsClient
+	loftClient      loftclient.Interface
+	agentLoftClient agentloftclient.Interface
 }
 
 func (c *client) Loft() loftclient.Interface {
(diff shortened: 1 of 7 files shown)


nprokopic:
I've hit an issue while testing - in this PR I have added a check of `virtualclusterinstances.management.loft.sh` CRD, but when testing now that check failed (so protected cluster got deleted). Then I saw that there is a `virtualclusterinstances.storage.loft.sh` CRD, while I expected `virtualclusterinstances.management.loft.sh`. Looking into this now.

johannesfrey (dismissed):
lgtm

pkg/platform/kube/client.go
```diff
@@ -6,12 +6,14 @@ import (
...
 	"github.com/pkg/errors"
 
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
 type Interface interface {
 	kubernetes.Interface
+	APIExtensions() apiextensions.Interface
```
hidalgopl, line 16: this change slightly worries me:
this is a platform client that we use in the Pro version too. we can create a new client for `apiextensions` here, but it may be not usable in most cases (because vCluster does not have RBAC for `apiextensions` usually (https://github.com/loft-sh/vcluster/blob/main/chart/templates/clusterrole.yaml#L122-L125).
This doesn't mean that the client creation will fail, but any calls made via this client will.
given that we only need `apiextensionsv1` clientset in the CLI, maybe just create an `apiextensions` client directly there instead of adding it here?

hidalgopl (changes requested):

pkg/platform/kube/client.go
nprokopic: Isn't this already the case for `kubernetes.Interface` that we already have here, which bundles 50+ GroupVersions (most of APIs from the Kubernetes project), while virtual cluster Role can access only a subset of those (depending on the config)?

Ideally a piece of code, some component, which needs some client, would use a more specific client, e.g. an instance of `corev1.CoreV1Interface`, or `storagev1.StorageV1Interface`, or `apiextensions.Interface`, and it would also take care of the permissions (it would be a larger refactoring effort to make the CLI code work like this).

hidalgopl: > Ideally a piece of code, some component, which needs some client, would use a more specific client, e.g. an instance of corev1.CoreV1Interface, or storagev1.StorageV1Interface, or apiextensions.Interface, and it would also take care of the permissions (it would be a larger refactoring effort to make the CLI code work like this).

yeah, I agree.

> Isn't this already the case for kubernetes.Interface that we already have here, which bundles 50+ GroupVersions (most of APIs from the Kubernetes project), while virtual cluster Role can access only a subset of those (depending on the config)?

It is, but I still think we shouldn't add on top of that.

How about a slightly different approach:
Given that `apiextensions` client is only used to check for VirtualClusterInstance CRD existence, can we replace it with:
1. not adding `APIExtensions` to the platform client
2. instead, doing the following to check VirtualClusterInstance CRD existence with what's already in the platform client:
```go
resourceList, err := kubeClient.Discovery().ServerResourcesForGroupVersion("storage.loft.sh/v1")

if err != nil {
    log.Debug("virtualclusterinstances.storage.loft.sh CustomResourceDefinition not found, will not check virtual cluster on the platform")
    return vClusters, nil
}
found := false

for _, resource := range resourceList.APIResources {
    if resource.Name == "VirtualClusterInstance" {
        found = true
        break
    }
}

// if found you can list virtual cluster instances

```
(we can move this to separate function). This is similar to how we check for feature CRD existance in vcluster-pro [ref](https://github.com/loft-sh/vcluster-pro/blob/main/pkg/license/loader/loader.go#L204-L212)

nprokopic: > It is, but I still think we shouldn't add on top of that.

True, especially if we would like to transition towards using more specific clients.

> How about a slightly different approach:

The above example looks good 👌 let me try it and then I'll update the PR

hidalgopl (dismissed):

nprokopic:
> [nprokopic](https://github.com/nprokopic) dismissed [johannesfrey](https://github.com/johannesfrey)’s [stale review](https://github.com/loft-sh/vcluster/pull/2697#pullrequestreview-2799126360) via [597e630](https://github.com/loft-sh/vcluster/commit/597e6301f23288724ff563a31731bae37e0c99dc)

I think that GitHub does this automatically when new commits are pushed,IIRC related to the repo settings.

johannesfrey (dismissed):

cbron (approved):

PR #2701, opened 2025-04-22
[v0.24] Platform should start with last released stable version, by default

ApsTomar:
/kind bugfix

resolves ENG-6469

Fixed an issue where vcluster platform start command installs the platform with latest version even if that is an alpha or rc version. Ideally, it should start with last stable release version, for eg. v4.2.3

### DIFF
--- a/pkg/platform/version.go
+++ b/pkg/platform/version.go
@@ -15,7 +15,7 @@ import (
 )
 
 var (
-	MinimumVersionTag = "v4.0.0-alpha.18"
+	MinimumVersionTag = "v4.2.3"
 	MinimumVersion    = semver.MustParse(strings.TrimPrefix(MinimumVersionTag, "v"))
 )
 
--- a/pkg/platform/version.go
+++ b/pkg/platform/version.go
@@ -50,7 +50,7 @@ func LatestCompatibleVersion(ctx context.Context) (string, error) {
 		return MinimumVersionTag, nil
 	}
 
-	eligibleReleases := lo.FilterMap(releases, func(release *github.RepositoryRelease, _ int) (semver.Version, bool) {
+	eligibleStableReleases := lo.FilterMap(releases, func(release *github.RepositoryRelease, _ int) (semver.Version, bool) {
 		tagName := release.GetTagName()
 		if tagName == "" {
 			return semver.Version{}, false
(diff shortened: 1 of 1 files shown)


pkg/platform/version.go
```diff
@@ -50,7 +50,7 @@ func LatestCompatibleVersion(ctx context.Context) (string, error) {
 		return MinimumVersionTag, nil
```
cbron, line 50: Can you change the actual min version too, in pkg/platform/version.go

ApsTomar: yes, it's done as per our slack discussion.

pkg/platform/version.go
```diff
@@ -61,15 +61,19 @@ func LatestCompatibleVersion(ctx context.Context) (string, error) {
 			return semver.Version{}, false
 		}
 
+		// skip the pre-releases (alpha, rc etc.), in semVer stable releases don't contain "-"
+		if strings.Contains(ghVersion.String(), "-") {
```
cbron, line 65: Would be better to use the library: 
```
	v, _ := semver.Make("4.2.1-alpha.0+other")
	if len(v.Pre) > 0 {
	    fmt.Println(len(v.Pre))
	    fmt.Println(v.Pre)
	}
```
output
```
2
[alpha 0]
```

ApsTomar: Good suggestion, thanks! `semver.Make()` is an alias of `semver.Parse()` which is already being used one above so I utilised that.

pkg/platform/version.go
```diff
@@ -61,15 +61,19 @@ func LatestCompatibleVersion(ctx context.Context) (string, error) {
 			return semver.Version{}, false
 		}
 
+		// skip the pre-releases (alpha, rc, etc.)
+		if len(ghVersion.Pre) != 0 {
+			return semver.Version{}, false
```
cbron, line 66: Why return here instead of continue ? Won't that stop finding latest release as soon as it finds a pre release.

ApsTomar: No, actually this is a callback function which runs on every item of the `releases` array provided in the argument of `FilterMap()` method. If the `ghVersion` is not eligible, we return false as a second argument which basically tells the `FilterMap` method to filter it out i.e. reject it. Here is the [reference code](https://github.com/samber/lo/blob/master/slice.go#L57) for FilterMap.

zerbitx (approved):
lgtm

cbron (approved):

cbron:
Will merge this and other PR monday

PR #2711, opened 2025-04-24
fix(patches): Avoid same path patches

ulyssessouza:
/kind bugfix
/kind enhancement

resolves #ENG-6431

Avoids having 2 patches applying in the same path.
Please note that this rule applies to all the patches on all the base paths.

### DIFF
--- a/pkg/config/validation_test.go
+++ b/pkg/config/validation_test.go
@@ -668,3 +668,92 @@ func TestValidateExportKubeConfig(t *testing.T) {
 		})
 	}
 }
+
+const patchesPath = "spec.containers[*].name"
+
+func TestValidateAllSyncPatches(t *testing.T) {
+	cases := []struct {
+		name          string
+		configSync    config.Sync
+		expectedError error
+	}{
+		{
+			name: "Simple config sync with one patch",
+			configSync: config.Sync{
+				ToHost: config.SyncToHost{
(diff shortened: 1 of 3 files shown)


pkg/config/validation_test.go
```diff
@@ -668,3 +668,75 @@ func TestValidateExportKubeConfig(t *testing.T) {
...
+					},
+					{
+						Path:              patchesPath,
+						ReverseExpression: "value.slice(\"my-prefix\".length)",
+					},
+				},
+			},
+		},
+	}
+	err := ValidateAllSyncPatches(sync)
```
hidalgopl, line 738: minor: Have you thought about consolidating this 3 `Test*` function into one and using `t.Run("<test-name>", func()` to run them as subtests? I think we use this approach in most of the unit tests we have (example: https://github.com/loft-sh/vcluster/blob/main/pkg/config/validation_test.go#L141-L145).
 I guess in this case, the test case struct could contain fields `config.Sync` as an input and `expectedErr error` which would be used in assertion

ulyssessouza: Sure! I'll do that!

ulyssessouza: PTAL

hidalgopl (approved):

PR #2716, opened 2025-04-25
[ENG-6170] refactor(versions): Remove older versions of images

ulyssessouza:
/kind enhancement

resolves #ENG-6170

Basically remove versions of k8s under 1.30

### DIFF
--- a/pkg/constants/coredns.go
+++ b/pkg/constants/coredns.go
@@ -1,19 +1,9 @@
 package constants
 
+// Please refer to https://github.com/coredns/deployment/blob/master/kubernetes/CoreDNS-k8s_version.md
+
 var CoreDNSVersionMap = map[string]string{
+	"1.32": "coredns/coredns:1.11.3",
+	"1.31": "coredns/coredns:1.11.3",
 	"1.30": "coredns/coredns:1.11.3",
-	"1.29": "coredns/coredns:1.11.1",
-	"1.28": "coredns/coredns:1.10.1",
-	"1.27": "coredns/coredns:1.10.1",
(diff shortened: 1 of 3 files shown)


pkg/constants/coredns.go
```diff
@@ -2,18 +2,4 @@ package constants
...
-	"1.25": "coredns/coredns:1.9.3",
-	"1.24": "coredns/coredns:1.8.7",
-	"1.23": "coredns/coredns:1.8.6",
-	"1.22": "coredns/coredns:1.8.4",
-	"1.21": "coredns/coredns:1.8.3",
-	"1.20": "coredns/coredns:1.8.0",
-	"1.19": "coredns/coredns:1.6.9",
-	"1.18": "coredns/coredns:1.6.9",
-	"1.17": "coredns/coredns:1.6.9",
-	"1.16": "coredns/coredns:1.6.3",
```
ulyssessouza, line 18: @cbron I'm not very sure about this deletions here.
Could you please confirm?
Please note that the `images-optional.txt` will not mention `coredns` since there is only one version available.

hidalgopl (dismissed):

pkg/constants/coredns.go
ulyssessouza: Ok. I added `1.31` and `1.32` versions. But just notice that according to https://github.com/coredns/deployment/blob/master/kubernetes/CoreDNS-k8s_version.md the correct version for `1.30` is `coredns/coredns:1.11.1` and not `coredns/coredns:1.11.3`. I didn't change because I fear there is a hidden reason for that.

hidalgopl (approved):

Issue #1926, opened 2024-07-12
K8s Service cluster-ip is none when map host cluster service to vCluster service


mayooot:
### What happened?

When I create a vcluster and map the host cluster's service to vcluster, use the flowing file:

~~~shell
$ cat values.yaml
mapServices:
  fromHost:
  - from: web-namespace/nginx-service
    to: aps/nginx-service
~~~

Everything was working fine, but I find the vcluster service cluster-ip is none, so I tried to exec a pod and use `nslookup` to get the actual service ip. 

~~~shell
$ vcluster connect my-vcluster-k8s -n team-x  -- kubectl get svc -A
NAMESPACE     NAME            TYPE        CLUSTER-IP        EXTERNAL-IP   PORT(S)                  AGE
default       kubernetes      ClusterIP   192.168.194.152   <none>        443/TCP                  28h
kube-system   kube-dns        ClusterIP   192.168.194.236   <none>        53/UDP,53/TCP,9153/TCP   28h
aps           nginx-service   ClusterIP   None              <none>        80/TCP                   28h

$ vcluster connect my-vcluster-k8s -n team-x  -- kubectl exec -it curl-pod -n aps -- bash

nslookup nginx-service.aps.svc.cluster.local
Server:   192.168.194.236
Address:  192.168.194.236#53

Name: nginx-service.aps.svc.cluster.local
Address: 192.168.194.242
~~~

Obviously, `192.168.194.242` is the real k8s service cluster ip.
~~~ shell
k get svc -A | grep nginx-service
default                    nginx-service                              ClusterIP   192.168.194.186   <none>        80/TCP                          2d5h
web-namespace              nginx-service                              ClusterIP   192.168.194.242   <none>        80/TCP                          34h
~~~

So my question is, why not show the cluster-ip in the service of vcluster? Is a bug or is it something else?

### What did you expect to happen?

vcluster can show real service cluster-ip when mapping the service

### How can we reproduce it (as minimally and precisely as possible)?

be like: https://www.vcluster.com/docs/v0.19/networking/internal_traffic/host_to_vcluster

### Host cluster Kubernetes version

<details>

```console
$ vcluster connect my-vcluster-k8s -n team-x  -- kubectl version
Client Version: v1.30.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.29.0
```

</details>

### vcluster version

<details>

```console
$ vcluster version
vcluster version 0.20.0-beta.1
```

[...]

PR #2717, opened 2025-04-25
fix(servicesync): correctly set service clusterIP

mayooot:
/kind bugfix

https://github.com/loft-sh/vcluster/issues/1926

When I use syncer `--map-host-service` args to sync the serivce from the physical k8s cluster inside vcluster, the ClusterIP is None, which is present in every version of vcluster.

So I modified `servicesync.go` and built the image locally and tested it on a production vcluster and it works, synchronizing both the headless service and the clusterIP serivce.

Note that `ClusterIP: fromService.Spec.ClusterIP` is sufficient, there is no need to determine if `fromService.Spec.ClusterIP` is `corev1.ClusterIPNone`.


mayooot:
@FabianKramm Please review this pr, thx.

PR #2720, opened 2025-04-27
Pin nginx-unprivileged image in e2e tests

matskiv:
/kind bugfix

I suspect that [this issue](https://github.com/nginx/docker-nginx-unprivileged/issues/302) in the nginx image we use in e2e is gonna break our suite. The image needs to be pinned to previous tag.

Here is failure with a dummy commit before pinning the tag - https://github.com/loft-sh/vcluster/actions/runs/14696414303/job/41238649749?pr=2720


matskiv:
This last test fail seems to be a flake - https://github.com/loft-sh/vcluster/actions/runs/14696629961/job/41239176723

zerbitx (approved):
lgtm

PR #2722, opened 2025-04-28
istio config validation + helm chart fix

hidalgopl:
/kind bugfix
/kind enhancement

resolves #

- creates cluster role if istio integration is enabled
- adds vcluster.yaml validation to reject istio CRDs in `sync.toHost.customResources` section if istio integraiton is enabled

Fixed an issue where vcluster was not creating cluster role when istio integration is enabled


zerbitx (approved):
pending checks, lgtm

PR #2725, opened 2025-04-29
[fix] Ignore permission error when listing VirtualClusterInstances

nprokopic:
/kind bugfix

Fixed an issue where printing test logs failed due to permission error introduced in #2697.

### DIFF
--- a/pkg/cli/find/find.go
+++ b/pkg/cli/find/find.go
@@ -281,7 +281,10 @@ func ListVClusters(ctx context.Context, context, name, namespace string, log log
 	}
 	// Find virtual cluster instances, so we can pair them with OSS virtual clusters.
 	virtualClusterInstancesList, err := kubeClient.Loft().StorageV1().VirtualClusterInstances("").List(ctx, listOptions)
-	if err != nil {
+	if kerrors.IsForbidden(err) {
+		log.Debug("user does not have permission to list VirtualClusterInstances")
+		return vClusters, nil
+	} else if err != nil {
 		return nil, fmt.Errorf("failed to list virtual cluster instances: %w", err)
 	}
 	virtualClusterInstances := map[string]*storagev1.VirtualClusterInstance{}


nprokopic:
e2e test failed with this old known flaky error which is still happening unfortunately:

```
• [FAILED] [40.104 seconds]
AdmissionWebhook [It] should be able to deny attaching pod
/home/runner/work/vcluster/vcluster/test/e2e/webhook/webhook.go:102

  [FAILED] unexpected 'kubectl attach' error message. expected to contain "attaching to pod 'to-be-attached-pod' is not allowed", got "timed out waiting for command /usr/bin/kubectl --kubeconfig=/tmp/vcluster_e2e_kubeconfig_2261075586 --namespace=e2e-webhook-1-vedve attach --namespace=e2e-webhook-1-vedve to-be-attached-pod -i -c=container1:\nCommand stdout:\n\nstderr:\nIf you don't see a command prompt, try pressing enter.\n"
  In [It] at: /home/runner/work/vcluster/vcluster/test/framework/helper.go:62 @ 04/29/25 23:01:03.074
```

pkg/cli/find/find.go
```diff
@@ -281,7 +281,10 @@ func ListVClusters(ctx context.Context, context, name, namespace string, log log
 	}
 	// Find virtual cluster instances, so we can pair them with OSS virtual clusters.
 	virtualClusterInstancesList, err := kubeClient.Loft().StorageV1().VirtualClusterInstances("").List(ctx, listOptions)
-	if err != nil {
+	if kerrors.IsForbidden(err) {
```
nprokopic, line 284: Pairing VirtualClusterInstances to found OSS virtual clusters is done on a best effort basis, as VirtualClusterInstance CRD maybe does not even exist on the cluster (e.g. in case of OSS vcluster, or connected host cluster).

So here we have another scenario where this pairing will not work - when the user that is running the CLI does not have permissions to list VirtualClusterInstances. In this case the func just return the OSS virtual clusters that it found so far (like few lines above where it cannot find VirtualClusterInstance CRD).

johannesfrey (approved):

hidalgopl (approved):

PR #2738, opened 2025-05-01
docs: documents that controlplane.distro.k8s.version is deprecated

jjaferson:
/kind documentation

resolves ENG-6742

Documents that controlplane.distro.k8s.version is deprecated

### DIFF
--- a/chart/values.schema.json
+++ b/chart/values.schema.json
@@ -1025,7 +1025,7 @@
         },
         "version": {
           "type": "string",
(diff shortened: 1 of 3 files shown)


cbron (approved):

deniseschannon:
FYI, when we deprecate a field, we should search the docs for other places that it should be replaced. 

I did this one as part of a different PR that I noticed it. 

cc: @joaniekube @Piotr1215

PR #2739, opened 2025-05-01
chore: adds deprecation notice to k0s and k3s distros

jjaferson:
/kind documentation

resolves #

Adds deprecation notice to k0s and k3s distributions in the documentation.

### DIFF
--- a/chart/values.schema.json
+++ b/chart/values.schema.json
@@ -861,11 +861,11 @@
         },
         "k3s": {
           "$ref": "#/$defs/DistroK3s",
-          "description": "K3S holds K3s relevant configuration."
+          "description": "[Deprecated] K3S holds K3s relevant configuration."
         },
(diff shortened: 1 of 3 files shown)


cbron (approved):

jjaferson:
> Does this mean k3s will not be supported in future vcluster versions?

No, we will continue supporting k3s. While both k0s and k3s were deprecated as part of v0.25.x, support for k0s will be removed in v0.26.0. Support for k3s will remain to allow users time to migrate over to k8s.

deniseschannon:
We haven't made any decisions on when to remove k3s, but wanted to give users a heads up that we plan to remove it in the future. Since k3s was the initial default, we wanted to give more time for migrations. In regards to external database support, we are debating on whether or not to open source that feature when we finally remove k3s.

PR #2742, opened 2025-05-05
sync istio patches in pod controller .Sync

hidalgopl:
/kind bugfix

resolves ENG-6760

Fixed an issue where vcluster was applying istio patches only on host object creation

otherwise the pods won't be labeled if the namespace was label after the pod creation


zerbitx (approved):
lgtm

PR #2749, opened 2025-05-06
chore: removes support for k0s

jjaferson:
/kind enhancement

resolves # ENG-6628

Removed support for k0s


pkg/k8s/k8s.go
```diff
@@ -327,6 +332,75 @@ func StartBackingStore(ctx context.Context, vConfig *config.VirtualClusterConfig
...
+		return nil
+	}
+	err = cmd.Run()
+
+	// make sure we wait for scanner to be done
+	writer.CloseAndWait(ctx, err)
+	return fmt.Errorf("error running command %s: %w", command[0], err)
+}
+
+func ExecTemplate(templateContents string, name, namespace string, values *vclusterconfig.Config) ([]byte, error) {
```
jjaferson, line 372: I moved this function to the k8s package as the k0s package was removed and it's required to  sync manifest template to the virtual cluster.

https://github.com/loft-sh/vcluster/blob/main/test/e2e/manifests/init.go#L28

pkg/k8s/k8s.go
```diff
@@ -327,6 +332,75 @@ func StartBackingStore(ctx context.Context, vConfig *config.VirtualClusterConfig
 	return etcdEndpoints, etcdCertificates, nil
 }
 
+func RunCommand(ctx context.Context, command []string, component string) error {
```
FabianKramm, line 335: This is now part of the util/command package I don't think we need this here as well

FabianKramm (changes requested):
See comment above and please remove the docs pages

pkg/k8s/k8s.go
jjaferson: done

FabianKramm (approved):

reneleonhardt:
I can't find ENG-6628, why is removing k0s an enhancement?

PR #2751, opened 2025-05-06
refactor: change k8s image logic

FabianKramm:
/kind enhancement

resolves ENG-6779
resolves ENG-6743

**Basically changes the logic how we determine what Kubernetes version to use by doing the following:**
* If user sets `controlPlane.distro.k8s.version` we use that
* Otherwise we use `controlPlane.distro.k8s.image.tag` which has a default of v1.32.1 (if nothing is defined)

Previously we were using the host cluster k8s version if nothing was defined which caused issues and unexpected behavior


chart/templates/_init-containers.tpl
```diff
@@ -8,28 +8,12 @@
...
-{{- regexFind "^v[0-9]+\\.[0-9]+\\.[0-9]+" .Capabilities.KubeVersion.Version -}}
-{{- end -}}
-{{- end -}}
-
 {{/* Bump $defaultTag value whenever k8s version is bumped */}}
 {{- define "vcluster.k8s.image.tag" -}}
-{{- $defaultTag := "v1.32.1" -}}
-{{- if and (not (empty .Values.controlPlane.distro.k8s.version)) (eq .Values.controlPlane.distro.k8s.image.tag $defaultTag) -}}
+{{- if not (empty .Values.controlPlane.distro.k8s.version) -}}
 {{ .Values.controlPlane.distro.k8s.version }}
```
deniseschannon, line 14: With the deprecation of `version` in this release, should we have it try `tag` first and then `version`?

FabianKramm: Problem is we still need to compare it then to some default as we don't know if it would be set as tag is always set and we cannot determine if the user has set it

PR #2756, opened 2025-05-07
ENG-6785 | Namespace toHost syncing

janekbaraniewski:
/kind feature

part of #ENG-6785

Adds initial implementation for namespace toHost syncing


pkg/setup/config.go
```diff
@@ -38,8 +39,13 @@ func InitClients(vConfig *config.VirtualClusterConfig) error {
...
 	// get workload target namespace
-	if vConfig.Experimental.MultiNamespaceMode.Enabled {
-		translate.Default = translate.NewMultiNamespaceTranslator(vConfig.WorkloadNamespace)
+	if vConfig.Sync.ToHost.Namespaces.Enabled {
+		translate.Default, err = pro.GetWithSyncedNamespacesTranslator(vConfig.WorkloadNamespace, vConfig.Sync.ToHost.Namespaces.Mappings)
+		if err != nil {
+			return err
+		}
+		// if vConfig.Experimental.MultiNamespaceMode.Enabled {
+		// 	translate.Default = translate.NewMultiNamespaceTranslator(vConfig.WorkloadNamespace)
```
FabianKramm, line 48: That looks kinda weird

janekbaraniewski: 🤦  thx for catching this!

janekbaraniewski: cleaned up 👍

chart/templates/clusterrole.yaml
```diff
@@ -104,9 +104,9 @@ rules:
     resources: ["services", "endpoints"]
     verbs: ["get", "watch", "list"]
   {{- end }}
-  {{- if .Values.experimental.multiNamespaceMode.enabled }}
+  {{- if .Values.sync.toHost.namespaces.enabled }}
   - apiGroups: [""]
-    resources: ["namespaces", "serviceaccounts"]
+    resources: ["namespaces", "serviceaccounts", "services", "pods", "secrets", "configmaps", "persistentvolumeclaims", "endpoints", "events"]
```
FabianKramm, line 109: Why is that needed, we already transform the Role in chart/templates/role to a clusterrole that should cover this

config/config.go
```diff
@@ -679,6 +682,19 @@ type SyncFromHost struct {
 	Secrets EnableSwitchWithResourcesMappings `json:"secrets,omitempty"`
 }
 
+type SyncToHostNamespaces struct {
+	Enabled bool `json:"enabled,omitempty" jsonschema:"required"`
```
FabianKramm, line 686: Add a comment above please as we want every field described

config/config.go
```diff
@@ -2090,9 +2106,6 @@ type Experimental struct {
 	// GenericSync holds options to generically sync resources from virtual cluster to host.
 	GenericSync ExperimentalGenericSync `json:"genericSync,omitempty"`
 
-	// MultiNamespaceMode tells virtual cluster to sync to multiple namespaces instead of a single one. This will map each virtual cluster namespace to a single namespace in the host cluster.
-	MultiNamespaceMode ExperimentalMultiNamespaceMode `json:"multiNamespaceMode,omitempty"`
```
FabianKramm, line 2094: Looks like we havent deleted struct ExperimentalMultiNamespaceMode, can you double check?

pkg/controllers/resources/namespaces/syncer.go
```diff
@@ -67,15 +71,19 @@ func (s *namespaceSyncer) Syncer() syncertypes.Sync[client.Object] {
 }
 
 func (s *namespaceSyncer) SyncToHost(ctx *synccontext.SyncContext, event *synccontext.SyncToHostEvent[*corev1.Namespace]) (ctrl.Result, error) {
+	if event.HostOld != nil || event.Virtual.DeletionTimestamp != nil {
```
FabianKramm, line 74: You use HostOld here but haven't enabled ObjectCaching, this seems like a mistake to me

pkg/controllers/resources/namespaces/syncer.go
```diff
@@ -96,7 +104,28 @@ func (s *namespaceSyncer) Sync(ctx *synccontext.SyncContext, event *synccontext.
 
 func (s *namespaceSyncer) SyncToVirtual(ctx *synccontext.SyncContext, event *synccontext.SyncToVirtualEvent[*corev1.Namespace]) (_ ctrl.Result, retErr error) {
 	// virtual object is not here anymore, so we delete
-	return patcher.DeleteHostObject(ctx, event.Host, nil, "virtual object was deleted")
+	if event.VirtualOld != nil || translate.ShouldDeleteHostObject(event.Host) {
```
FabianKramm, line 107: Same as above, seems like you should enable objectcaching

pkg/controllers/resources/namespaces/syncer.go
```diff
@@ -96,7 +104,28 @@ func (s *namespaceSyncer) Sync(ctx *synccontext.SyncContext, event *synccontext.
...
+		return patcher.DeleteHostObject(ctx, event.Host, event.VirtualOld, "virtual object was deleted")
+	}
+
+	virtualName := ""
+	for vname, hname := range ctx.Config.Sync.ToHost.Namespaces.Mappings.ByName {
+		if hname == event.Host.Name {
```
FabianKramm, line 113: Isn't this a pattern, does it make sense to do a direct comparison here, what happens if I use "*" in my mapping

pkg/specialservices/resolver.go
```diff
@@ -40,6 +41,10 @@ type NameserverFinder struct {
 }
 
 func (f *NameserverFinder) DNSNamespace(ctx *synccontext.SyncContext) (client.Client, string) {
+	if ctx.Config.Sync.ToHost.Namespaces.Enabled {
+		// when namespaces are synced, we rely on translator logic
```
FabianKramm, line 45: But why? Please make this comment more specific and tell us why exactly this is needed here

test/e2e_generic/values.yaml
```diff
@@ -1,6 +1,4 @@
 experimental:
-  multiNamespaceMode:
-    enabled: true
```
FabianKramm, line 3: This will cause the test to fail I believe, why don't we use the new thing here?

FabianKramm (changes requested):

pkg/setup/config.go
```diff
@@ -38,8 +39,11 @@ func InitClients(vConfig *config.VirtualClusterConfig) error {
 	}
 
 	// get workload target namespace
-	if vConfig.Experimental.MultiNamespaceMode.Enabled {
-		translate.Default = translate.NewMultiNamespaceTranslator(vConfig.WorkloadNamespace)
+	if vConfig.Sync.ToHost.Namespaces.Enabled {
+		translate.Default, err = pro.GetWithSyncedNamespacesTranslator(vConfig.WorkloadNamespace, vConfig.Sync.ToHost.Namespaces.Mappings)
```
FabianKramm, line 43: I don't understand this, does that mean we don't use translate.NewMultiNamespaceTranslator anymore at all?

.github/workflows/e2e.yaml
```diff
@@ -257,7 +257,7 @@ jobs:
       matrix:
         distribution: ["k3s", "k8s"]
         test-suite-path: ${{fromJson(needs.get-testsuites-dir.outputs.matrix)}}
-        multinamespace-mode: ["false", "true"]
+        multinamespace-mode: ["false"]
```
FabianKramm, line 260: Oh btw if we get rid of this then completely delete that

pkg/controllers/resources/namespaces/syncer.go
janekbaraniewski: yup, thx! added `OptionsProvider` implementation 👍

pkg/controllers/resources/namespaces/syncer.go
janekbaraniewski: enabled 👍

config/config.go
janekbaraniewski: cleaned up!

config/config.go
janekbaraniewski: added 👍

chart/templates/clusterrole.yaml
janekbaraniewski: yeah this is a leftover from when I started developing this, no longer needed. Thx! cleaning up

pkg/setup/config.go
janekbaraniewski: yeah exactly. I didn't clean it up yet but current behaviour for namespace syncing is:
- use only new namespace translator
- allow only operations on namespaces defined in mappings
- for special cases fallback to singleNamespace logic

.github/workflows/e2e.yaml
janekbaraniewski: cleaned up 👍

pkg/controllers/resources/namespaces/syncer.go
janekbaraniewski: yeah thx for catching this, clearly forgot to update logic here to work with new translator. Updated 👍

pkg/specialservices/resolver.go
janekbaraniewski: Ok found the root cause - new namespace mapper needed a fallback that would handle those cases, that's why we were getting empty namespace. Fixed and cleaned this up

FabianKramm (approved):

PR #2760, opened 2025-05-07
feat: Hybrid Scheduling config and check if enabled

nprokopic:
/kind feature

towards ENG-6600

Added Hybrid Scheduling config and syncer error if Hybrid Scheduling is enabled, since it is a Pro feature.

Hybrid scheduling check should be also added in few places in the Helm chart, but I will add that in another PR, as Hybrid Scheduling does not work in the vcluster OSS and this PR is towards the first alpha release for testing.

### DIFF
--- a/pkg/controllers/resources/pods/scheduling/config.go
+++ b/pkg/controllers/resources/pods/scheduling/config.go
@@ -1,0 +1,34 @@
+package scheduling
+
+import (
+	"github.com/loft-sh/admin-apis/pkg/licenseapi"
+	"github.com/loft-sh/vcluster/pkg/pro"
+)
+
+type Config struct {
+	virtualSchedulerEnabled bool
+	hybridSchedulingEnabled bool
+	hostSchedulers          []string
+}
+
+// NewConfig creates a new scheduling config with specified vCluster scheduling options. In the case of vcluster OSS
+// when Hybrid Scheduling is enabled, this func returns an error, because Hybrid Scheduling is a Pro-only feature.
+var NewConfig = func(virtualSchedulerEnabled, hybridSchedulingEnabled bool, _ []string) (Config, error) {
+	if hybridSchedulingEnabled {
+		return Config{}, pro.NewFeatureError(string(licenseapi.HybridScheduling))
+	}
+
+	return Config{
+		virtualSchedulerEnabled: virtualSchedulerEnabled,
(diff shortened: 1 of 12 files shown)


nprokopic:
Similar changes might be needed here https://github.com/loft-sh/vcluster/blob/2e6d12556f3a6f00cc05daef710618e0a1296e35/pkg/controllers/resources/nodes/syncer.go#L113

johannesfrey:
> Similar changes might be needed here
> 
> https://github.com/loft-sh/vcluster/blob/2e6d12556f3a6f00cc05daef710618e0a1296e35/pkg/controllers/resources/nodes/syncer.go#L113

Will this be part of another PR?

nprokopic:
> > Similar changes might be needed here
> > https://github.com/loft-sh/vcluster/blob/2e6d12556f3a6f00cc05daef710618e0a1296e35/pkg/controllers/resources/nodes/syncer.go#L113
> 
> Will this be part of another PR?

Nope, it should go in this PR, just wasn't sure initially if it was needed. Just updated it.

Will convert the PR to draft, just to add few more checks I missed.

nprokopic:
Now really ready for review 😬

pkg/controllers/resources/pods/scheduling/config.go
```diff
@@ -0,0 +1,34 @@
...
+
+type Config struct {
+	VirtualSchedulerEnabled bool
+	HybridSchedulingEnabled bool
+	HostSchedulers          []string
+}
+
+// NewConfig creates a new scheduling config with specified vCluster scheduling options. In the case of vcluster OSS
+// when Hybrid Scheduling is enabled, this func returns an error, because Hybrid Scheduling is a Pro-only feature.
+var NewConfig = func(virtualSchedulerEnabled, hybridSchedulingEnabled bool, _ []string) (Config, error) {
```
johannesfrey, line 16: Just out of interest. The reason for making this and the one below function literals instead of regular function declarations is to be able to overwrite them with vcluster-pro!?

johannesfrey (approved):
Just one question. Besides that lgtm!

pkg/controllers/resources/pods/scheduling/config.go
nprokopic: Yes, exactly :)

PR #2761, opened 2025-05-07
refactor: remove docs

FabianKramm:
Remove the legacy docs from the oss repo and move them into https://github.com/loft-sh/legacy-docs-vcluster

Resolves DOC-529


Piotr1215 (approved):
There is also the docs hack that generates partials in the vcluster-docs repository. I think it should still work, just mentioning for context: https://github.com/loft-sh/vcluster/blob/main/hack/docs/main.go

PR #2774, opened 2025-05-09
fix: vCluster connecting with background proxy

jjaferson:
/kind bugfix

resolves ENG-6873

fixed vCluster connect background proxy is not port-forwarding to the localhost when host server address is 0.0.0.0


zerbitx (approved):
lgtm

PR #2785, opened 2025-05-13
ci: build `head` docker images on changes in main

sydorovdmytro:
Add Docker build step for main branch commits

- Automatically builds a Docker image on every commit to the `main` branch.
- Tags each Docker image with `head-<architecture>`, e.g., `vcluster-oss:head-arm64v8` or `vcluster-oss:head-amd64`.

/kind enhancement

References OPS-148


Piotr1215 (commented):
You can use default branch variable here `${{ github.event.repository.default_branch }}`

Piotr1215 (changes requested):
Left some comments and suggestions. Also we should at least run unit and helm tests before triggering the workflow, something like in the release workflow:

```yaml
- name: Execute unit tests
  run: ./hack/test.sh

- name: Run Helm Unit Tests
  run: |
    helm plugin install https://github.com/helm-unittest/helm-unittest --version v0.4.4
    helm unittest chart
```

sydorovdmytro:
Good idea! Added unit tests and addressed other comments

Piotr1215 (approved):

.goreleaser.head_images.yaml
```diff
@@ -0,0 +1,158 @@
+version: 2
```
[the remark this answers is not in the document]
sydorovdmytro: The existing `.goreleaser` file is designed to be used in the snapshot mode, which blocks all publish events like `docker push`.

.github/workflows/push-head-images.yaml
```diff
@@ -0,0 +1,84 @@
+name: Build and push head images
+
+on:
+  push:
+    branches:
+      - main
```
cbron, line 6: If this is too frequent, we could do a nightly build as well. We typically get between 0-6 merges a day to main, avg around 3. But as team grows that can go up. Scaling this back to once a day will lower cost of producing it. Or if we want to be crazy with throw a random generator in here, build if `rand.Intn(3) ==0`  😄

cbron (approved):
lgtm, thanks for handling!

.github/workflows/push-head-images.yaml
sydorovdmytro: Sure! Let's gather some feedback from the team and change if needed.

Piotr1215: I love the random number :D. Let's monitor the cost and see where it lands.

PR #2786, opened 2025-05-13
feat: add ability to force sync an event into default namespace

matskiv:
/kind enhancement

resolves ENG-6860

Added ability to force sync Events that don't have an `involvedObject` that is synced from vCluster. These events will be synced into the default namespace.

### DIFF
--- a/pkg/mappings/resources/events.go
+++ b/pkg/mappings/resources/events.go
@@ -17,6 +17,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const ForceSyncedEventNamespace = "default"
+
 var AcceptedKinds = map[schema.GroupVersionKind]bool{
 	corev1.SchemeGroupVersion.WithKind("Pod"):       true,
 	corev1.SchemeGroupVersion.WithKind("Service"):   true,
@@ -52,7 +54,7 @@ func (s *eventMapper) HostToVirtual(ctx *synccontext.SyncContext, req types.Name
 			klog.Infof("Error retrieving involved object for %s/%s: %v", req.Namespace, req.Name, err)
 		} else if pObj.GetAnnotations()[constants.SyncResourceAnnotation] == "true" {
 			return types.NamespacedName{
-				Namespace: "default",
+				Namespace: ForceSyncedEventNamespace,
 				Name:      pObj.GetName(),
 			}
 		}
--- a/pkg/mappings/resources/events.go
+++ b/pkg/mappings/resources/events.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/loft-sh/vcluster/pkg/constants"
 	"github.com/loft-sh/vcluster/pkg/syncer/synccontext"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


pkg/mappings/resources/events.go
```diff
@@ -49,6 +50,11 @@ func (s *eventMapper) HostToVirtual(ctx *synccontext.SyncContext, req types.Name
 		err = IgnoreAcceptableErrors(err)
 		if err != nil {
 			klog.Infof("Error retrieving involved object for %s/%s: %v", req.Namespace, req.Name, err)
+		} else if pObj.GetAnnotations()[constants.SyncResourceAnnotation] == "true" {
+			return types.NamespacedName{
+				Namespace: "default",
```
jjaferson, line 55: maybe move this into a variable or a constant? 

Just a question, why are we hardcoding it to the default ns?

matskiv: > maybe move this into a variable or a constant?

Ok, will do.
> Just a question, why are we hardcoding it to the default ns?

If the event is not related to any specific synced resource then we don't have anything to go by to make the ns decision. Thus "default" seemed like a reasonable fallback. I've also considered adding an annotation for the virtual namespace, but default is enough for the initial use case.

jjaferson: I see, that makes sense.

hidalgopl (commented):
1. Just to ensure, this will only work on events from `vcluster` host namespace, right? Because the syncer watches only events from `vcluster` host namespace.
2. Were you able to test it? I think we could extend one of the existing e2e tests to annotate existing event and check if it gets synced to the virtual properly.

cbron:
@matskiv ^

matskiv:
@hidalgopl to answer your questions:

> Just to ensure, this will only work on events from vcluster host namespace, right? Because the syncer watches only events from vcluster host namespace.

Correct. That is an expected limitation.

> Were you able to test it? I think we could extend one of the existing e2e tests to annotate existing event and check if it gets synced to the virtual properly.

I did test this by applying a new Event manifest, with the annotation.
All the existing events would typically be synced anyway. Should I create a new e2e test that will create a new event and verify it is synced?

pkg/mappings/resources/events.go
matskiv: I added a constant as requested.
I would prefer it in the `pkg/controllers/resources/events` package, but then there would be a circular dependency, and it seemed like overkill to add a nested package just for this one const.

test/e2e/syncer/events/events.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+			if !reflect.DeepEqual(virtual1.Message, event1.Message) {
+				f.Log.Errorf("expected %#v in virtual.Message got %#v", event1.Message, virtual1.Message)
+				return false
+			}
+			return true
+		}).
+			WithPolling(time.Second).
+			WithTimeout(framework.PollTimeout / 4).
+			Should(gomega.BeTrue())
```
lizardruss: nit: imo this reads a bit better and would report the failed assertion vs reporting "false != true"
```golang
                gomega.Eventually(func(g gomega.Gomega) {
			virtual1, err := f.VClusterClient.CoreV1().Events("default").Get(f.Context, event1Name, metav1.GetOptions{})
			g.Expect(err).NotTo(gomega.HaveOccurred())
			g.Expect(virtual1.Message).To(gomega.Equal(event1.Message))
		}).
			WithPolling(time.Second).
			WithTimeout(framework.PollTimeout / 4).
			Should(gomega.Succeed())
```

lizardruss (approved):
Code looks good, I added a suggestion for the e2e tests, but approving 👍

The reason the customer wants this seems a little specific to build into the core codebase, but I don't think there are other extension options that we're actively supporting.

test/e2e/syncer/events/events.go
matskiv: Good suggestion. I updated my commit.

lizardruss (approved):
lgtm!

PR #2793, opened 2025-05-15
Only set embedded if external datasource is empty

rmweir:
This is necessary to maintain compatibility. Pre-v0.25.0 external datasource was used regardless of whether the enabled flag was used.

/kind bugfix

Fixes #ENG-6900

Fixed an issue where vcluster where using a connector without the flag "controlPlane.backingStore.database.external.enabled" would no longer work. This would cause the backingstore to change to an embedded sqlite database after upgrading from v0.24.x to v0.25.0. This would would make the vcluster appear as thought it was reset.

If anyone happened to use a connector without enabled in v0.25.0 then they we will face the opposite issue after the fix. They will have been using sqlite and now will actually use the connector.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -342,7 +342,8 @@ func (c *Config) BackingStoreType() StoreType {
 }
 
 func (c *Config) EmbeddedDatabase() bool {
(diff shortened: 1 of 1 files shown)


rmweir:
Closing this for now, evaluating different options with @FabianKramm.

PR #2808, opened 2025-05-21
pref: Increase kubernetes config QPS and Burst

mayooot:

/kind enhancement

none

Increased QPS and Burst for kubernetes config from 40 to 1000 for QPS and 80 to 2000 fo Burst.


pkg/pro/remote.go
```diff
@@ -16,8 +16,8 @@ import (
 
 var GetRemoteClient = func(vConfig *config.VirtualClusterConfig) (*rest.Config, string, string, *rest.Config, string, string, error) {
 	inClusterConfig := ctrl.GetConfigOrDie()
-	inClusterConfig.QPS = 40
-	inClusterConfig.Burst = 80
+	inClusterConfig.QPS = 1000
+	inClusterConfig.Burst = 2000
```
matskiv, line 20: This is pretty steep increase. It probably would be okay if it was for the client targeting the vCluster API server, but here it is targeting the host cluster, and it makes sense to be more conservative there.

matskiv:
> What does this pull request do? Which issues does it resolve?
> 
> none

Can you provide some reasoning why this change should be made? Ideally with some numbers...

pkg/pro/remote.go
matskiv: I discussed this with maintainers, keeping the current values as defaults and adding possibility to override it via env vars is the preferred solution. Would you be able to update your PR accordingly?

mayooot: ok, I'll finish it later.

matskiv (approved):
Looks good. Thank you!

mayooot:
> Looks good. Thank you!

@matskiv  Please rerun the tests.

pkg/pro/remote.go
```diff
@@ -16,9 +19,30 @@ import (
 
 var GetRemoteClient = func(vConfig *config.VirtualClusterConfig) (*rest.Config, string, string, *rest.Config, string, string, error) {
 	inClusterConfig := ctrl.GetConfigOrDie()
-	inClusterConfig.QPS = 40
-	inClusterConfig.Burst = 80
-	inClusterConfig.Timeout = 0
+
+	// Get QPS from environment variable or default to 40
+	qpsStr := os.Getenv("K8S_CLIENT_QPS")
```
lizardruss: We may want to use a different name. The `K8S` prefix is not used anywhere in the code base. `KUBERNETES` could collide with environment variables used for other purposes. This is also specific to the physical host client, so we might consider `VCLUSTER_PHYSICAL_CLIENT_QPS` (to use the more recent terminology) and `VCLUSTER_VIRTUAL_CLIENT_QPS` could be used [here](https://github.com/loft-sh/vcluster/blob/main/pkg/setup/controller_context.go#L170-L172) to configure the virtual client.

pkg/pro/remote.go
```diff
@@ -16,9 +19,30 @@ import (
...
+	qps, err := strconv.ParseFloat(qpsStr, 32)
+	if err != nil || qpsStr == "" {
+		qps = 40
+	}
+	inClusterConfig.QPS = float32(qps)
+
+	// Get Burst from environment variable or default to 80
+	burstStr := os.Getenv("K8S_CLIENT_BURST")
```
lizardruss: similar comment as above

pkg/pro/remote.go
```diff
@@ -16,9 +19,30 @@ import (
...
+	burst, err := strconv.Atoi(burstStr)
+	if err != nil || burstStr == "" {
+		burst = 80
+	}
+	inClusterConfig.Burst = burst
+
+	// Get Timeout from environment variable or default to 0
+	timeoutStr := os.Getenv("K8S_CLIENT_TIMEOUT")
```
lizardruss: similar comment as above

lizardruss (commented):
Naming is hard. I've added my feedback, but will also seek additional maintainer feedback.

pkg/pro/remote.go
mayooot: > We may want to use a different name. The `K8S` prefix is not used anywhere in the code base. `KUBERNETES` could collide with environment variables used for other purposes. This is also specific to the physical host client, so we might consider `VCLUSTER_PHYSICAL_CLIENT_QPS` (to use the more recent terminology) and `VCLUSTER_VIRTUAL_CLIENT_QPS` could be used [here](https://github.com/loft-sh/vcluster/blob/main/pkg/setup/controller_context.go#L170-L172) to configure the virtual client.

Ok, I think using `VCLUSTER_PHYSICAL_CLIENT_` prefix is better.

mayooot:
> Naming is hard. I've added my feedback, but will also seek additional maintainer feedback.



> Naming is hard. I've added my feedback, but will also seek additional maintainer feedback.

Ok, I'm done, and you can check it.

lizardruss (approved):
lgtm

cbron:
@mayooot if you can resolve conflicts then we can merge.

mayooot:
> @mayooot if you can resolve conflicts then we can merge.

OK, Please review the last commit.

lizardruss (approved):

FabianKramm (approved):

PR #2813, opened 2025-05-22
delete external ETCD after migration

hidalgopl:
* allow for external etcd statefulset & services deletion after migration to embedded
* pass workload client to embedded etcd function

/kind bugfix

part of ENG-5634

add RBAC needed for external etcd deletion after migration to embedded

this is needed for deleting external etcd statefulset & services after successful migration to embedded etcd (implemented in the pro)


FabianKramm (changes requested):
See https://linear.app/loft/issue/ENG-5634/bug-enabling-migratefromdeployedetcd-still-has-the-external-etcd-pod#comment-9557dfe4, this is not a good idea.

PR #2834, opened 2025-05-26
fix: service cidr detection

FabianKramm:
/kind bugfix

resolves ENG-6989

Fixed an issue where vcluster wouldn't detect the service cidr correctly in v1.33 Kubernetes clusters

### DIFF
--- a/pkg/util/servicecidr/servicecidr.go
+++ b/pkg/util/servicecidr/servicecidr.go
@@ -1,133 +1,245 @@
 package servicecidr
 
 import (
+	"bytes"
 	"context"
-	"errors"
+	"encoding/binary"
 	"fmt"
+	"math/big"
 	"net"
 	"strings"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/loft-sh/vcluster/config"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/klog/v2"
 )
 
 const (
-	ErrorMessageFind = "The range of valid IPs is "
-	FallbackCIDR     = "10.96.0.0/12"
+	ServiceCIDRAnnotation = "vcluster.loft.sh/service-cidr"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/util/servicecidr/servicecidr.go
```diff
@@ -2,132 +2,244 @@ package servicecidr
...
-		}()
-
-		// check if this is dual stack, and which family is default
-		if len(testService.Spec.IPFamilies) > 0 {
-			if testService.Spec.IPFamilies[0] == corev1.IPv4Protocol {
-				// IPv4 is the default
-				return fmt.Sprintf("%s,%s", ipv4CIDR, ipv6CIDR), ""
+		}, metav1.CreateOptions{})
+		if err == nil {
+			err := client.CoreV1().Services(vClusterNamespace).Delete(ctx, testService.GetName(), metav1.DeleteOptions{})
```
kale-amruta, line 74: Do we want to `defer` the deletion of services? We wouldnt want any any dangling test services

pkg/util/servicecidr/servicecidr.go
```diff
@@ -2,132 +2,244 @@ package servicecidr
...
+	// check if we need to use the new service cidr detection
+	vClusterService, err := client.CoreV1().Services(vClusterNamespace).Get(ctx, vClusterServiceName, metav1.GetOptions{})
+	if err != nil {
+		return "", fmt.Errorf("failed to get vCluster service %s: %v", vClusterServiceName, err)
 	}
 
-	if client == nil {
-		panic("client is nil")
+	// check if we already found the service cidr
+	if vClusterService.Annotations[ServiceCIDRAnnotation] != "" {
```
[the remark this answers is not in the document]
FabianKramm: Thats actually not needed, golang maps can do a lookup even on nil maps: https://go.dev/play/p/P3e6a2hymIB

pkg/util/servicecidr/servicecidr.go
FabianKramm: Whats the benefit of a defer here instead of deleting it directly? In either case the service gets deleted

kale-amruta: deferring will make sure , even if there are panics, the test service still will get cleaned up.

kale-amruta (approved):
lgtm

PR #2844, opened 2025-05-28
ENG-6979 | Fixes for namespace sync validation

janekbaraniewski:
/kind enhancement

### DIFF
--- a/pkg/config/validation.go
+++ b/pkg/config/validation.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/loft-sh/vcluster/config"
 	"github.com/loft-sh/vcluster/pkg/constants"
-	"github.com/loft-sh/vcluster/pkg/util/clienthelper"
 	"github.com/loft-sh/vcluster/pkg/util/namespaces"
 	"github.com/loft-sh/vcluster/pkg/util/toleration"
 )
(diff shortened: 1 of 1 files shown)


pkg/util/namespaces/util.go
```diff
@@ -0,0 +1,47 @@
...
+
+// ProcessNamespaceName returns namespace name after applying all pre-processing to it
+func ProcessNamespaceName(namespaceName string, vclusterName string) string {
+	return replaceNamePlaceholder(namespaceName, vclusterName)
+}
+
+// replaceNamePlaceholder replaces ${name} placeholder in given templateString with vclusterName
+func replaceNamePlaceholder(templateString string, vclusterName string) string {
+	return strings.ReplaceAll(templateString, NamePlaceholder, vclusterName)
+}
```
zerbitx, line 47:
```suggestion
// ProcessNamespaceName returns namespace name after applying all pre-processing to it
func ProcessNamespaceName(namespaceName string, vclusterName string) string {
	return strings.ReplaceAll(namespaceName, NamePlaceholder, vclusterName)
}
```

janekbaraniewski: yeah I guess you're right, no point in additional abstraction if that's the only processing step 👍

pkg/util/namespaces/validation.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	configPathIdentifier := "config.sync.toHost.namespaces.mappings.byName"
+
+	if len(c.Sync.ToHost.Namespaces.Mappings.ByName) == 0 {
+		return fmt.Errorf("%s are empty", configPathIdentifier)
+	}
+
+	virtualNamespaceNames := make([]string, 0, len(c.Sync.ToHost.Namespaces.Mappings.ByName))
+	hostNamespaceNames := make([]string, 0, len(c.Sync.ToHost.Namespaces.Mappings.ByName))
```
zerbitx, line 23: There's [set implementation](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set) in apimachinery I just recently learned about from another PR, that you could use to check for duplicates in a single pass.

janekbaraniewski: that's great! thx for this, didn't know about it! Updated to use sets 🫡

pkg/util/namespaces/util.go
janekbaraniewski: updated 👍

PR #2846, opened 2025-05-28
requeue endpoint reconcile when the service is created after the endpoint

### LINKED ISSUE #2719: Headless Service Endpoint and EndpointSlice not syncing from virtual to host
### What happened?

I'm attempting route a Service to an external IP address so I created the Service headless with `ClusterIP: None` and also created an Endpoint resource for the external IPv4 address and port (matching the port in Service).  The names of Service and Endpoint match exactly.

When deployed to the vcluster the EndpointSlice is created for the Endpoint, as expected.  The problem is both the Endpoint and EndpointSlice are not synced to the host cluster, unlike other Service resources using selectors for pods which are synced.  The headless Service is synced though.

I expected the Endpoint and EndpointSlice resources to sync to the host cluster so the corresponding service's ingress will route though to the external IP.

1. Create a vCluster using v0.24.0 helm chart, and connect to this new cluster.
2. Deploy a Service resource to the vcluster, specify `clusterIP: None` and don't provide a selector.
3. Deploy an Endpoint resource to the vcluster, name it the same as the Service, and give an arbitrary IP address but use the same port as the targetPort in the Service.
4. Use kubectl to confirm an EndpointSlice was created automagically.
5. Connect to the host cluster and confirm the Service is synced but the Endpoint and EndpointSlice are not.

_No response_

<details>

```console
Client Version: v1.29.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.0
```

</details>

### vcluster version

<details>

```console
vcluster version 0.24.1
```

</details>

### VCluster Config

<details>

```
USER-SUPPLIED VALUES:
controlPlane:
  proxy:
    extraSANs:
    - tb-idp-dmp1.dmp1.svc
exportKubeConfig:
  secret:
    name: tb-idp-dmp1-kubeconfig-secret
    namespace: dmp1
  server: https://tb-idp-dmp1.dmp1.svc
integrations:
  metricsServer:
    enabled: true
    nodes: true
    pods: true
networking:
  replicateServices:
    fromHost:
    - from: observability/opentelemetry-collector
      to: observability/opentelemetry-collector
    - from: redis/redis
      to: redis/redis
plugin:
  cert-manager-plugin:
    image: ghcr.io/loft-sh/vcluster-plugins/cert-manager-plugin:0.3.0
    imagePullPolicy: IfNotPresent
    rbac:
      clusterRole:
        extraRules:
        - apiGroups:
          - apiextensions.k8s.io
          resources:
          - customresourcedefinitions
          verbs:
          - get
          - list
          - watch
      role:
        extraRules:
        - apiGroups:
          - cert-manager.io
          resources:
          - issuers
          - certificates
          verbs:
          - create
          - delete
          - patch
          - update
          - get
          - list
          - watch
sync:
  toHost:
    endpoints:
      enabled: true
    ingresses:
      enabled: true
    persistentVolumes:
      enabled: true
    serviceAccounts:
      enabled: true
    services:
      enabled: true
    storageClasses:
      enabled: true
```

</details>

Prakhar824:
/kind bugfix

resolves #2719 
fixes ENG-6661

Fixed an issue where vcluster does not sync the endpoint to host cluster if the endpoint is created before the service.

When an endpoint is created before its corresponding service, it does not sync to the host cluster because reconciliation is skipped due to the absence of the service in the vcluster. This PR adds the code to trigger endpoint reconciliation after the service is created, ensuring the associated endpoint is also synced to the host cluster.

### DIFF
--- a/pkg/controllers/resources/endpoints/syncer.go
+++ b/pkg/controllers/resources/endpoints/syncer.go
@@ -1,9 +1,16 @@
 package endpoints
 
 import (
+	"context"
 	"errors"
 	"fmt"
-	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 
 	"github.com/loft-sh/vcluster/pkg/mappings"
 	"github.com/loft-sh/vcluster/pkg/patcher"
@@ -14,13 +21,8 @@ import (
 	"github.com/loft-sh/vcluster/pkg/syncer/translator"
 	syncertypes "github.com/loft-sh/vcluster/pkg/syncer/types"
 	"github.com/loft-sh/vcluster/pkg/util/translate"
-	corev1 "k8s.io/api/core/v1"
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/klog/v2"
-	ctrl "sigs.k8s.io/controller-runtime"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/resources/endpoints/syncer.go
```diff
@@ -56,6 +59,48 @@ func (s *endpointsSyncer) Syncer() syncertypes.Sync[client.Object] {
...
+            },
+        }
+    }))
+
+    return bld, nil
+}
+
 func (s *endpointsSyncer) SyncToHost(ctx *synccontext.SyncContext, event *synccontext.SyncToHostEvent[*corev1.Endpoints]) (ctrl.Result, error) {
 	if event.HostOld != nil {
 		return patcher.DeleteVirtualObject(ctx, event.Virtual, event.HostOld, "host object was deleted")
```
cbron: Did you ever find out why the endpoint itself doesn't sync here ? Thinking about this further, you should be able to sync the endpoint even if a service doesn't exist. The docs say this, but not the other way around: 
> Service resources are also enabled by default and require Endpoint resources.

pkg/controllers/resources/endpoints/syncer.go
```diff
@@ -56,6 +59,48 @@ func (s *endpointsSyncer) Syncer() syncertypes.Sync[client.Object] {
...
+			Name:      service.Name,
+		}, endpoint)
+		if err != nil {
+			if kerrors.IsNotFound(err) {
+				klog.Infof("Endpoints for Service %s/%s not found, skipping", service.Namespace, service.Name)
+				return []ctrl.Request{}
+			}
+			klog.Errorf("Error retrieving Endpoints for Service %s/%s: %v", service.Namespace, service.Name, err)
+			return []ctrl.Request{}
+		}
```
lizardruss: Is loading the endpoint needed? IIRC, the syncer should do this before determining whether to call `Sync`, `SyncToHost` and `SyncToVirtual`. If this is required, it may mean that the logic in these methods may need updating.

Prakhar824: I had a look at it again and found out that loading the endpoints here is not necessary, syncer will detect automatically if the endpoint resource exists. Based on this I have removed this piece of code and manually tested the scenarios and it worked as expected.

pkg/controllers/resources/endpoints/syncer.go
Prakhar824: The reason for endpoint not syncing if the service does not exist is due to [this line of code](https://github.com/loft-sh/vcluster/blob/07381bf36faf905c171b1213fd59b92c6451d726/pkg/controllers/resources/endpoints/syncer.go#L123). Endpoint in themselves don't have any functionality without the corresponding service, service is considered as the parent object of endpoint and is responsible to manage it's lifecycle.

pkg/controllers/resources/endpoints/syncer.go
```diff
@@ -57,6 +62,31 @@ func (s *endpointsSyncer) Syncer() syncertypes.Sync[client.Object] {
...
+
+	// Watch for changes to Services and reconcile Endpoints
+	return bld.Watches(&corev1.Service{}, handler.EnqueueRequestsFromMapFunc(func(_ context.Context, obj client.Object) []ctrl.Request {
+		service, ok := obj.(*corev1.Service)
+		if !ok || service == nil {
+			klog.Info("Received an object that is not a Service or is nil, skipping")
+			return []ctrl.Request{}
+		}
+
+		klog.Infof("Enqueuing reconciliation request of Endpoint for Service: %s/%s", service.Namespace, service.Name)
```
FabianKramm: This will pollute the logs, please remove this

FabianKramm (changes requested):
Looks good, but please fix the above

pkg/controllers/resources/endpoints/syncer.go
Prakhar824: I have fixed the same. Could you please re-review?

PR #2855, opened 2025-05-30
Add fields in values.yaml for probes to be overriden

kale-amruta:
/kind enhancement

resolves ENG-6645

Adds fields in values.yaml to override the default values for `liveness,readiness and startupProbes`. Users will now be able to create vcluster with specifying custom values for probes in values.yaml
```
controlPlane:
  statefulSet:
    probes:
         # LivenessProbe specifies if the liveness probe for the container should be enabled
         livenessProbe:
           # Enabled defines if this option should be enabled.
           enabled: true
           failureThreshold: 80
           initialDelaySeconds: 80
           timeoutSeconds: 8
           periodSeconds: 9
         # ReadinessProbe specifies if the readiness probe for the container should be enabled
         readinessProbe:
           # Enabled defines if this option should be enabled.
           enabled: true
           failureThreshold: 80
           timeoutSeconds: 9
           periodSeconds: 7
         # StartupProbe specifies if the startup probe for the container should be enabled
         startupProbe:
           # Enabled defines if this option should be enabled.
           enabled: true
           failureThreshold: 80
           timeoutSeconds: 8
           periodSeconds: 10
``` 
       

neogopher (dismissed):

kale-amruta:
> Why make this so complex in the helm chart? We set defaults via values.yaml should be there only.

I thought, we should always set defaults, even when someone removes these fields from values.yaml, hence I had added the defaults in helpers. But if we dont care about anyone removing them then its fine. I have removed that part.

config/config.go
```diff
@@ -1977,13 +1977,21 @@ type ControlPlaneWorkloadServiceAccount struct {
...
 	// ReadinessProbe specifies if the readiness probe for the container should be enabled
-	ReadinessProbe EnableSwitch `json:"readinessProbe,omitempty"`
+	ReadinessProbe Probe `json:"readinessProbe,omitempty"`
 
 	// StartupProbe specifies if the startup probe for the container should be enabled
-	StartupProbe EnableSwitch `json:"startupProbe,omitempty"`
+	StartupProbe Probe `json:"startupProbe,omitempty"`
+}
+
+type Probe struct {
```
FabianKramm: Please add a comment to each of these fields

chart/templates/statefulset.yaml
```diff
@@ -146,36 +146,42 @@ spec:
           args:
 {{ toYaml .Values.controlPlane.statefulSet.args | indent 12 }}
           {{- end }}
-          {{- if .Values.controlPlane.statefulSet.probes.livenessProbe.enabled }}
+         {{- with .Values.controlPlane.statefulSet.probes.livenessProbe }}
```
FabianKramm: Instead of doing with here, just use the full path, makes it a lot easier, also do this for all paths below

config/config.go
```diff
@@ -1977,13 +1977,21 @@ type ControlPlaneWorkloadServiceAccount struct {
 
 type ControlPlaneProbes struct {
 	// LivenessProbe specifies if the liveness probe for the container should be enabled
-	LivenessProbe EnableSwitch `json:"livenessProbe,omitempty"`
+	LivenessProbe Probe `json:"livenessProbe,omitempty"`
```
FabianKramm: This is all the same type, but we don't actually allow all fields for liveness, readiness and startupProbe in the helm chart. Either you allow them in the helm chart or make different types for each probe.

chart/tests/statefulset_test.yaml
```diff
@@ -805,3 +805,56 @@ tests:
...
+          value: 8
+      - equal:
+          path: spec.template.spec.containers[0].startupProbe.failureThreshold
+          value: 350
+      - equal:
+          path: spec.template.spec.containers[0].startupProbe.timeoutSeconds
+          value: 7
+      - equal:
+          path: spec.template.spec.containers[0].startupProbe.periodSeconds
+          value: 8
```
FabianKramm, line 859: Can you add a test to check the defaults when not specifying anything?

FabianKramm (approved):

PR #2856, opened 2025-05-30
Add imported marked annotation to prevent deletion of imported namespaces

janekbaraniewski:
/kind enhancement

resolves ENG-6976

Adds annotation to mark namespaces that were imported from host. This way we can prevent deletion of such namespaces on host in case they get removed from vCluster or if the vCluster itself gets deleted.

### DIFF
--- a/pkg/controllers/resources/namespaces/syncer.go
+++ b/pkg/controllers/resources/namespaces/syncer.go
@@ -10,6 +10,7 @@ import (
 	"github.com/loft-sh/vcluster/pkg/syncer/synccontext"
 	"github.com/loft-sh/vcluster/pkg/syncer/translator"
 	syncertypes "github.com/loft-sh/vcluster/pkg/syncer/types"
+	"github.com/loft-sh/vcluster/pkg/util/translate"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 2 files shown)


pkg/controllers/resources/namespaces/cleanup.go
```diff
@@ -0,0 +1,155 @@
...
+
+			if currentRuleMatches {
+				logger.Infof("Attempting to delete virtual cluster namespace %s.", hostNs.Name)
+				err := k8sClient.CoreV1().Namespaces().Delete(ctx, hostNs.Name, metav1.DeleteOptions{})
+				if err != nil {
+					return fmt.Errorf("delete virtual cluster namespace %s: %w", hostNs.Name, err)
+				}
+				logger.Donef("Successfully deleted virtual cluster namespace %s.", hostNs.Name)
+				// This namespace has been handled. Skip other mappings and move to the next one.
+				goto nextHostNs
```
hidalgopl: If I understand correctly, this can be replaced with `break`, right? Because you want to break from the inner loop (that loops over `mappingsConfig.ByName`) without breaking from the outer loop (that loops over `hostNamespaces.Items`)

pkg/controllers/resources/namespaces/cleanup.go
```diff
@@ -0,0 +1,155 @@
...
+			var currentRuleMatches bool
+			if namespaces.IsPattern(processedHostTargetPattern) {
+				_, currentRuleMatches = namespaces.MatchAndExtractWildcard(hostNs.Name, processedHostTargetPattern)
+			} else {
+				currentRuleMatches = (hostNs.Name == processedHostTargetPattern)
+			}
+
+			if currentRuleMatches {
+				logger.Infof("Attempting to delete virtual cluster namespace %s.", hostNs.Name)
+				err := k8sClient.CoreV1().Namespaces().Delete(ctx, hostNs.Name, metav1.DeleteOptions{})
```
hidalgopl: I'd collect the errors from `Delete()` in the `var errs []error` and then check if there are any errors in the `errs` when we're done with the loop. This way you'd cleanup most namespaces instead of erroring when one of them cannot be deleted.

pkg/controllers/resources/namespaces/cleanup.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/loft-sh/log"
+	"github.com/loft-sh/vcluster/config"
+	"github.com/loft-sh/vcluster/pkg/util/namespaces"
+	"github.com/loft-sh/vcluster/pkg/util/translate"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// CleanupHandler defines the function signature for namespace cleanup operations.
+type CleanupHandler func(
```
hidalgopl: It seems to be used only in the CLI, right? I'd move this file to `pkg/cli` then

pkg/helm/helm.go
```diff
@@ -39,6 +39,13 @@ type UpgradeOptions struct {
 	Debug    bool
 }
 
+// GetValuesOptions holds the options for retrieving values for a release.
+type GetValuesOptions struct {
+	// All indicates whether to retrieve all values, including those set by the chart by default.
+	// Corresponds to the `--all` flag in Helm.
+	All bool
```
hidalgopl: nit: if the only option is bool, it may be as well the function argument in `GetValues` :)

pkg/controllers/resources/namespaces/cleanup.go
janekbaraniewski: yeah can be true 👍

pkg/helm/helm.go
janekbaraniewski: yup yup, cleaned up 👍

pkg/controllers/resources/namespaces/cleanup.go
janekbaraniewski: moved to `pkg/cli/cleanup_namespaces.go` 👍

pkg/controllers/resources/namespaces/cleanup.go
janekbaraniewski: yeah good point 👍  updated error handling, thx!

hidalgopl (approved):

hidalgopl (approved):

hidalgopl (approved):

PR #2857, opened 2025-05-30
fix: vcluster upgrade of k3s distro

jjaferson:
/kind bugfix

resolves ENG-6422

Fixed an issue when upgrading vcluster with different distro than the default one

### DIFF
--- a/pkg/cli/create_helm.go
+++ b/pkg/cli/create_helm.go
@@ -239,6 +239,12 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 			if err := currentVClusterConfig.UnmarshalYAMLStrict([]byte(migratedValues)); err != nil {
 				return err
 			}
+		} else {
+			// When a vCluster is not legacy, there should be a config secret and we will fetch the values from the secret
+			currentVClusterConfig, err = getConfigfileFromSecret(ctx, vClusterName, cmd.Namespace)
+			if err != nil {
+				return err
+			}
 		}
 
 		if len(cmd.Values) == 0 {
@@ -342,11 +348,6 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 	verb := "created"
 	if isVClusterDeployed(release) {
 		verb = "upgraded"
-		currentVClusterConfig, err = getConfigfileFromSecret(ctx, vClusterName, cmd.Namespace, cmd.log)
-		if err != nil {
-			return err
-		}
-
 		// While certain backing store changes are allowed we prohibit changes to another distro.
 		if err := config.ValidateChanges(currentVClusterConfig, vClusterConfig); err != nil {
 			return err
@@ -952,7 +953,7 @@ func (cmd *createHelm) getVClusterConfigFromSnapshot(ctx context.Context) (strin
 	return "", nil
 }
 
-func getConfigfileFromSecret(ctx context.Context, name, namespace string, log log.Logger) (*config.Config, error) {
+func getConfigfileFromSecret(ctx context.Context, name, namespace string) (*config.Config, error) {
 	secretName := "vc-config-" + name
 
 	kConf := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(clientcmd.NewDefaultClientConfigLoadingRules(), &clientcmd.ConfigOverrides{})
@@ -968,10 +969,6 @@ func getConfigfileFromSecret(ctx context.Context, name, namespace string, log lo
 
 	secret, err := clientset.CoreV1().Secrets(namespace).Get(ctx, secretName, metav1.GetOptions{})
 	if err != nil {
-		if kerrors.IsNotFound(err) {
-			log.Warnf("Secret %s not found, returning empty vCluster config", secretName)
-			return &config.Config{}, nil
-		}
 		return nil, err
 	}
 


jjaferson:
I've tested the upgrade from 0.19.x -> 0.25.x

```
./vclusterv0.19.9 create  test-19-25 -n test-19-25
 warn There is a newer version of vcluster: v0.25.0. Run `vcluster upgrade` to upgrade to the newest version.

info Creating namespace test-19-25
done Switched active kube context to vcluster_test-19-25_test-19-25_kind-test-vcluster
- Use `vcluster disconnect` to return to your previous kube context
- Use `kubectl get namespaces` to access the vcluster


./vclusterv0.19.9 ls

       NAME    |      CLUSTER       | NAMESPACE  | STATUS  | VERSION | CONNECTED |            CREATED            |  AGE  | DISTRO
  -------------+--------------------+------------+---------+---------+-----------+-------------------------------+-------+---------
    test-19-25 | kind-test-vcluster | test-19-25 | Running | 0.19.9  |           | 2025-05-30 19:25:17 -0300 -03 | 2m26s | OSS


./vcluster-main convert config --distro k3s -f vclusterv19.config.yaml > vcluster25.config.yaml

cat vcluster25.config.yaml                                                                                                                                                
"
controlPlane:
  distro:
    k3s:
      enabled: true
      image:
        tag: v1.29.0-k3s1
  service:
    spec:
      type: NodePort
  statefulSet:
    scheduling:
      podManagementPolicy: OrderedReady
sync:
  fromHost:
    nodes:
      enabled: true
"

./vcluster-main create --upgrade test-19-25 -f vcluster25.config.yaml -n test-19-25
info Upgrade vcluster test-19-25
...
done Successfully upgraded virtual cluster test-19-25 in namespace test-19-25

./vcluster-main ls

       NAME    | NAMESPACE  | STATUS  | VERSION | CONNECTED |  AGE
  -------------+------------+---------+---------+-----------+---------
    test-19-25 | test-19-25 | Running | 0.25.0  | True      | 10m39s
    
 and 0.24.0 -> 0.25.0

./vclusterv0.24.0 create test-24-25 -n test-24-25 -f vcluster25.config.yaml      
warn There is a newer version of vcluster: v0.25.0. Run `vcluster upgrade` to upgrade to the newest version.
info Create vcluster test-24-25...

./vcluster-main ls

       NAME    | NAMESPACE  | STATUS  | VERSION | CONNECTED |  AGE
  -------------+------------+---------+---------+-----------+---------
    test-19-25 | test-19-25 | Running | 0.25.0  |           | 16m59s
    test-24-25 | test-24-25 | Running | 0.24.0  | True      | 2m20s
    
./vcluster-main create --upgrade test-24-25 -f vcluster25.config.yaml -n test-24-25
info Upgrade vcluster test-24-25...
done Successfully upgraded virtual cluster test-24-25 in namespace test-24-25

./vcluster-main ls 

       NAME    | NAMESPACE  | STATUS  | VERSION | CONNECTED |  AGE
  -------------+------------+---------+---------+-----------+---------
    test-19-25 | test-19-25 | Running | 0.25.0  |           | 18m16s
    test-24-25 | test-24-25 | Running | 0.25.0  | True      | 3m37s

File: pkg/cli/create_helm.go [hunk 1]
```diff
@@ -239,6 +239,13 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 			if err := currentVClusterConfig.UnmarshalYAMLStrict([]byte(migratedValues)); err != nil {
 				return err
 			}
+		} else if !currentVClusterConfig.ControlPlane.Distro.K3S.Enabled && !currentVClusterConfig.ControlPlane.Distro.K8S.Enabled {
```
rmweir:
```suggestion
		} else {
```
This has the same effect as using else. If the block above wasn't entered, the config will be empty and have these default values.

@cbron the block above this has logic to infer distro and store and then put it in to a config. However, the old position of `getConfigfileFromSecret` would overwrite these values. Now, we only get the config in the case that we didn't already detect it was legacy and needed the inferred values.

pkg/cli/create_helm.go
```diff
@@ -239,6 +239,13 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
...
+			// When the both supported distro (k3s & k8s) are disabled, it means that the current vcluster config is empty
+			// And for none legacy vCluster version we need to fetch the vcluster config values from a secret
```
rmweir: This comment has some typos but given what I mentioned above, I don't think it would make sense anyway. It is the other way around, the vcluster config is empty so k3s/k8s is going to appear to be disabled.

More accurately: the vCluster is not legacy. Therefore, it should have a config secret. We will attempt to fetch the secret.

rmweir:
Also, I do not think we should ever return an empty config from `getConfigfileFromSecret`. It makes assumption that can be wrong and consequently creates side effects. We simply cannot validate if there is no config secret. We can return an error for now since the known cases are now covered.

pkg/cli/create_helm.go
cbron: Thanks for explanation. 

> This has the same effect as using else. If the block above wasn't entered, the config will be empty and have these default values.


~Wouldn't removing the conditionals actually catch all other deployed clusters ? The point of conditional would be to get deployed clusters that _don't_ have information stored about them.~
Nvm, I misunderstood, all clusters _would_ end up in this block.

pkg/cli/create_helm.go
```diff
@@ -342,11 +349,6 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
...
 	if isVClusterDeployed(release) {
 		verb = "upgraded"
-		currentVClusterConfig, err = getConfigfileFromSecret(ctx, vClusterName, cmd.Namespace, cmd.log)
-		if err != nil {
-			return err
-		}
-
 		// While certain backing store changes are allowed we prohibit changes to another distro.
 		if err := config.ValidateChanges(currentVClusterConfig, vClusterConfig); err != nil {
 			return err
```
cbron, line 353: Re-posting ricardo's comment: 

> Also, I do not think we should ever return an empty config from getConfigfileFromSecret. It makes assumption that can be wrong and consequently creates side effects. We simply cannot validate if there is no config secret. We can return an error for now since the known cases are now covered.

This I agree with, we need to adjust that part. Though I don't want to block users who are in that spot. Ideally we would error out in the function, then log an error in the command, but let it go through if possible.

cbron: Something like this could work: https://github.com/loft-sh/vcluster/pull/2854/files#diff-7cdd5ad4fe3fde3dc8096e61b051bbddae8a1354fb23417e1ba8aaf057f698f2R355

Or we just hard error and fail the command

pkg/cli/create_helm.go
jjaferson: Yes, I just added these two checks to enforce that the config is empty and we are not overwritten any value related to the distro struct

pkg/cli/create_helm.go
jjaferson: I also agree with @rmweir on that, we shouldn't really return an empty config but I don't think we should skip validation as it would also end up leading to some side effects as the one we are facing at the moment.

pkg/cli/create_helm.go
jjaferson: That we know by reading the code, I just wanted to add that comment there to explain the reason for having the check there

jjaferson: I rephrased the sentence: When a vCluster is not legacy, there should be a config secret. To ensure the old vCluster config is empty, we check that both distros (k3s and k8s) are disabled.

pkg/cli/create_helm.go
jjaferson: Removed the distro check to avoid confusion

pkg/cli/create_helm.go
jjaferson: We decided to return the error and fail the command if secret is missing as it shouldn't be a valid case

cbron (approved):

rmweir (approved):
LGTM

PR #2860, opened 2025-06-01
[ENG-7068] fix(classes): Add validation to *Class selectors

ulyssessouza:
/kind bugfix

resolves #ENG-7068

Add validation to *Class selectors

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -792,13 +792,17 @@ type SyncFromHost struct {
 
 type StandardLabelSelector v1.LabelSelector
 
-func (s StandardLabelSelector) Matches(obj client.Object) bool {
-	ls := v1.LabelSelector(s)
-	selector, err := v1.LabelSelectorAsSelector(&ls)
+func (s StandardLabelSelector) Matches(obj client.Object) (bool, error) {
+	selector, err := s.ToSelector()
 	if err != nil {
-		return false
+		return false, fmt.Errorf("failed to convert label selector: %w", err)
 	}
-	return selector.Matches(labels.Set(obj.GetLabels()))
+	return selector.Matches(labels.Set(obj.GetLabels())), nil
+}
+
+func (s StandardLabelSelector) ToSelector() (labels.Selector, error) {
+	ls := v1.LabelSelector(s)
+	return v1.LabelSelectorAsSelector(&ls)
 }
 
 type EnableSwitchWithPatchesAndSelector struct {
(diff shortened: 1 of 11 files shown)


pkg/controllers/resources/ingresses/syncer.go
```diff
@@ -137,7 +137,12 @@ func (s *ingressSyncer) applyLimitByClass(ctx *synccontext.SyncContext, virtual
 			s.EventRecorder().Eventf(virtual, "Warning", "SyncWarning", "did not sync ingress %q to host because the ingress class %q couldn't be reached in the host: %s", virtual.GetName(), *virtual.Spec.IngressClassName, err)
 			return true
 		}
-		if !ctx.Config.Sync.FromHost.IngressClasses.Selector.Matches(pIngressClass) {
+		matches, err := ctx.Config.Sync.FromHost.IngressClasses.Selector.Matches(pIngressClass)
+		if err != nil {
+			s.EventRecorder().Eventf(virtual, "Warning", "SyncWarning", "did not sync ingress %q to host because the ingress class %q in the host could not be checked against the selector under 'sync.fromHost.ingressClasses.selector': %s", virtual.GetName(), pIngressClass.GetName(), err)
```
ulyssessouza, line 142: I was almost making a function to make this less repetitive.
Please tell me if I should make a function or a "method" to `syncertypes.GenericTranslator` (a bit agressive).

config/config.go
```diff
@@ -792,13 +792,17 @@ type SyncFromHost struct {
...
 type StandardLabelSelector v1.LabelSelector
 
-func (s StandardLabelSelector) Matches(obj client.Object) bool {
-	ls := v1.LabelSelector(s)
-	selector, err := v1.LabelSelectorAsSelector(&ls)
+func (s StandardLabelSelector) Matches(obj client.Object) (bool, error) {
+	selector, err := s.ToSelector()
 	if err != nil {
-		return false
+		return false, fmt.Errorf("failed to convert label selector: %w", err)
```
FabianKramm, line 798: When can this fail? What configuration would be needed in vcluster.yaml to make this fail?

pkg/controllers/resources/ingresses/syncer.go
FabianKramm: I think thats okay, we don't need to centralize that as its mostly just error messages and not real functionality.

config/config.go
ulyssessouza: This is for the case were the user uses an unknown string in the operator part of the `sync.fromHost.storageClasses.selector.matchExpressions`. 
Valid operators are:
- `In`
- `NotIn`
- `Exists`
- `DoesNotExist`

So if the user uses anything else that should cause this error. Ex:
```yaml
sync:
  fromHost:
    storageClasses:
      enabled: true
      selector:
        matchExpressions:
          - key: "myLabel"
            operator: DifferentFrom
            values: ["myValue"]
```

FabianKramm (approved):

PR #2874, opened 2025-06-04
feat: add standalone

FabianKramm:
Changes:
* Refactor constants to allow changing the values and move all filepath references to 
* Since standalone mode has no host cluster, skip certain parts of the code where host client is required
* Change the way how we currently execute api-server, controller-manager etc to allow restarting the vCluster process without left-over processes


lizardruss (approved):
lgtm

PR #2879, opened 2025-06-04
Block unwanted virtual scheduling

nprokopic:
/kind bugfix

resolves ENG-6916
resolves ENG-6921

Fixed a hybrid scheduling issue, caused by incorrectly configured hybrid scheduling, where a virtual pod is undesirably scheduled by a scheduler in the virtual cluster, while it should be scheduled by a scheduler in the host cluster.


lizardruss (approved):
lgtm

PR #2882, opened 2025-06-05
Cleanup metadata of imported namespces when removing vCluster

janekbaraniewski:
/kind bugfix

resolves ENG-6976

Fixed an issue where vcluster was not importing a namespace that was previously imported by a different, already uninstalled vcluster.

### DIFF
--- a/pkg/cli/cleanup_namespaces.go
+++ b/pkg/cli/cleanup_namespaces.go
@@ -4,13 +4,19 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/loft-sh/log"
 	"github.com/loft-sh/vcluster/config"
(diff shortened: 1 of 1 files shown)


hidalgopl (commented):
How will it work if I have multiple vClusters in the same host cluster? Won't it delete all synced namespaces in the host (not only from virtual cluster A, but also from virtual cluster B)?

janekbaraniewski:
@hidalgopl no, since we only iterate over namespaces that are matching `"vcluster.loft.sh/managed-by" = translate.SafeConcatName(mainPhysicalNamespace, "x", vClusterName)` label selector - so we only cleanup namespaces imported by our vCluster

hidalgopl (approved):

PR #2886, opened 2025-06-06
[ENG-7165, ENG-7207] fix(fromHost): Mark and only sync resources that are created by vCluster

ulyssessouza:
The intent of this PR is to solve the problem of vCluster trying to sync/manage resources that are not created by itself.

/kind bugfix

resolves #ENG-7165

Fixed an issue where vcluster tries to delete/update resources that are not created by vCluster. This PR avoids it by marking the resources created by us and checking for it before the sync

### DIFF
--- a/pkg/controllers/resources/ingressclasses/syncer.go
+++ b/pkg/controllers/resources/ingressclasses/syncer.go
@@ -58,7 +58,8 @@ func (i *ingressClassSyncer) SyncToVirtual(ctx *synccontext.SyncContext, event *
 		return ctrl.Result{}, nil
 	}
 
-	vObj := translate.CopyObjectWithName(event.Host, types.NamespacedName{Name: event.Host.Name, Namespace: event.Host.Namespace}, false)
+	vObj := translate.VirtualMetadata(event.Host, types.NamespacedName{Name: event.Host.Name, Namespace: event.Host.Namespace})
+	vObj.SetAnnotations(translate.HostAnnotations(vObj, event.Host))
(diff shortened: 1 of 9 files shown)


pkg/controllers/resources/ingressclasses/syncer.go
```diff
@@ -71,6 +72,11 @@ func (i *ingressClassSyncer) SyncToVirtual(ctx *synccontext.SyncContext, event *
 }
 
 func (i *ingressClassSyncer) Sync(ctx *synccontext.SyncContext, event *synccontext.SyncEvent[*networkingv1.IngressClass]) (_ ctrl.Result, retErr error) {
+	// If the virtual object has the name annotation, it means it was created by vCluster and we can safely manage it.
+	if _, ok := event.Virtual.GetAnnotations()[translate.HostNameAnnotation]; !ok {
```
nprokopic, line 76: 1-2 months ago we decided not to use annotations and labels in this way, where annotations and labels impact the behavior (I made a similar change, maybe even smaller), IIRC because users can make changes to those and it's not secure.

cc @FabianKramm can you take a look here?

FabianKramm: Yes that is the wrong approach here, we cannot use annotations on every virtual object just to mark it.

ulyssessouza:
Closing the PR, since the behavior it prevents seems to be expected.
A follow-up PR on docs will explain the behavior.

PR #2893, opened 2025-06-09
fix(issues/2889): Close os.Stdout when using `vcluster connect <vcluster-id> --print`

### LINKED ISSUE #2889: vcluster --print with port forwarding creates invalid kubeconfig
### What happened?

kubeconfig is invalid because it has trailing `Forwarding from ...` lines

eg:
```
$ vcluster connect vctest --print | tee ~/.kube/vcluster.yaml
...
VktLS0tLQo=
Forwarding from 127.0.0.1:12933 -> 8443
Forwarding from [::1]:12933 -> 8443
```

I don't have docker running. 

A valid kubeconfig.

See above.

_No response_

<details>

```console
$ kubectl version
# paste output here
```

</details>

### vcluster version

<details>

```console
$ vcluster --version
vcluster version 0.25.0
```

</details>

### VCluster Config

<details>

```
```

</details>

mayooot:

/kind bugfix

resolve #2889 

Fixed an issue where running `vcluster connect <vcluster-id> --print | tee ./vcluster.yaml` appended unwanted port fowarding output to the generated file, such as: 
~~~
Forwarding from 127.0.0.1:10641 -> 8443
Forwarding from [::1]:10641 -> 8443
~~~

I found the unwanted output coming from [portforward.go](https://github.com/loft-sh/vcluster/blob/74e339e4dce0e3ff0befaae05be10f3fdffb66bb/pkg/util/portforward/portforward.go#L339-L341), which is part of kubernetes source code. Changing the source code directly would be a pain to maintain, so I tackled the issue in the `--print` logic of the `connect` command instead.
The code is as follows:

~~~go
	if options.Print {
		_, err = os.Stdout.Write(out)
		if err != nil {
			return err
		}
		_ = os.Stdout.Close() // Add it 
	}
~~~

I'm not sure what the unintended error of `os.Stdout.Close()` would be.

### DIFF
--- a/pkg/cli/connect_helm.go
+++ b/pkg/cli/connect_helm.go
@@ -167,6 +167,7 @@ func writeKubeConfig(kubeConfig *clientcmdapi.Config, vClusterName string, optio
 		if err != nil {
 			return err
 		}
+		_ = os.Stdout.Close()
 	} else if options.UpdateCurrent {
 		var clusterConfig *clientcmdapi.Cluster
 		for _, c := range kubeConfig.Clusters {


lizardruss:
Hello! Thanks for the PR! Unfortunately closing `os.Stdout` will have other effects. I created [another PR](https://github.com/loft-sh/vcluster/pull/2896) that will hopefully solve the issue for you.

Closing this PR in favor of that one.

PR #2898, opened 2025-06-10
Add an option to set the vCluster logging format to json

ApsTomar:
/kind enhancement
/kind feature

resolves ENG-7039

Add an option to set the vCluster logging format to json

**Test Artifacts:**

Here's how the json logs would look like:
Screenshot from 2025-06-10 18-15-13

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -1154,7 +1154,7 @@ telemetry:
   # Enabled specifies that the telemetry for the vCluster control plane should be enabled.
   enabled: true
 
-# Configures structured logging within vCluster
+# Logging provides structured logging options
 logging:
   # Encoding specifies the format of vCluster logs, it can either be json or console.
   encoding: console
--- a/chart/tests/statefulset_test.yaml
+++ b/chart/tests/statefulset_test.yaml
@@ -929,3 +929,19 @@ tests:
       - equal:
           path: spec.template.spec.containers[0].startupProbe.failureThreshold
           value: 300
+
+  - it: should set logging format to json
+    set:
+      logging:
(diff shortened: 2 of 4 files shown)


hidalgopl (approved):

chart/templates/statefulset.yaml
```diff
@@ -181,6 +181,10 @@ spec:
           env:
             - name: VCLUSTER_NAME
               value: "{{ .Release.Name }}"
+            {{- if .Values.logging }}
+            - name: LOFT_LOG_ENCODING
+              value: {{ default "console" .Values.logging.encoding }}
```
cbron, line 186: Was curious how this was actually working, looks like its part of logging framework: https://github.com/loft-sh/log/blob/c2f046135b12ac51d9446dc329fa2067853e7f93/logr/logr.go#L183

chart/values.yaml
```diff
@@ -1153,3 +1153,8 @@ experimental:
 telemetry:
   # Enabled specifies that the telemetry for the vCluster control plane should be enabled.
   enabled: true
+
+# Logging provides structured logging options
+logging:
+  # Encoding specifies the format of vCluster logs, it can either be json or console.
+  encoding: console
```
cbron, line 1160: My main thought here was this could go into somwhere not top-level, like into the statefulset config. But platform uses this location, so we can leave here to make it consistent.

cbron (approved):

lucasfcnunes:
My logs are in json and I haven't set the `logging.encoding` to json (so it's in the default settings i.e. console). I'm on v0.26.0.

image

image

PR #2903, opened 2025-06-12
[ENG-7207] fix: Adds an allowlist to avoid to evaluate native priorityClasses

ulyssessouza:
/kind bugfix

Resolves ENG-7207


zerbitx (approved):
lgtm

PR #2910, opened 2025-06-13
Fix re-importing objects from imported namespaces, remove cleanup strategies when deleting vcluster

janekbaraniewski:
/kind enhancement

resolves ENG-7264

### DIFF
--- a/pkg/cli/cleanup_namespaces.go
+++ b/pkg/cli/cleanup_namespaces.go
@@ -2,29 +2,39 @@ package cli
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/loft-sh/log"
-	"github.com/loft-sh/vcluster/config"
 	"github.com/loft-sh/vcluster/pkg/util/translate"
 	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 )
 
 const (
 	vClusterMetadataPrefix = "vcluster.loft.sh/"
 )
 
-// CleanupSyncedNamespaces handles deletion of namespaces for the 'synced' policy.
-// It deletes namespaces from the host cluster that were created as a result of syncing process from vCluster,
+// CleanupSyncedNamespaces identifies all physical namespaces that were created or managed by vCluster.
+//  1. Namespaces created by the vCluster: These are deleted directly.
+//  2. Namespaces "imported" into the vCluster: For these, we remove all vCluster-related
+//     metadata from the namespace itself and all resources within it.
 func CleanupSyncedNamespaces(
 	ctx context.Context,
 	mainPhysicalNamespace string,
 	vClusterName string,
-	_ config.SyncToHostNamespaces,
+	restConfig *rest.Config,
 	k8sClient *kubernetes.Clientset,
 	logger log.Logger,
 ) error {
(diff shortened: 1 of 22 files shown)


pkg/cli/cleanup_namespaces.go
```diff
@@ -19,61 +18,9 @@ const (
...
-
-	if len(errs) > 0 {
-		return fmt.Errorf("metadata cleanup for vCluster '%s' namespaces finished with errors: %w", vClusterName, errors.Join(errs...))
-	}
-
-	logger.Infof("Metadata cleanup for vCluster '%s' namespaces finished.", vClusterName)
-	return nil
-}
-
 // cleanupSyncedNamespaces handles deletion of namespaces for the 'synced' policy.
```
jjaferson:
```suggestion
// CleanupSyncedNamespaces handles deletion of namespaces for the 'synced' policy.
```

pkg/syncer/syncer.go
```diff
@@ -199,7 +199,11 @@ func (r *SyncController) Reconcile(ctx context.Context, vReq reconcile.Request)
 	if vObj != nil && pObj != nil {
 		// make sure the object uid matches
 		pAnnotations := pObj.GetAnnotations()
-		if !r.options.DisableUIDDeletion && pAnnotations[translate.UIDAnnotation] != "" && pAnnotations[translate.UIDAnnotation] != string(vObj.GetUID()) {
+
+		if !r.config.Sync.ToHost.Namespaces.Enabled && // if we're running with namespace sync enabled, we don't want to remove any host object based on UID because they might have been previously imported by other vclusters.
```
FabianKramm: What if a pod gets super quickly created and then deleted again? Wouldn't this cause issues when namespace sync is enabled as the pod would get outdated on the host? Basically this:
1. Create pod in vCluster
2. Sync pod to Host Cluster
3. Super quickly delete & create pod in vCluster
4. Now host cluster has a pod with that name and vCluster, but the spec could be different.
How do we solve this?

pkg/cli/cleanup_namespaces.go
janekbaraniewski: oh yeah, nice catch! thx!

pkg/util/translate/translate.go
```diff
@@ -149,13 +150,18 @@ func addHostAnnotations(retMap map[string]string, vObj, pObj client.Object) {
...
+func ShouldDeleteHostObject(ctx *synccontext.SyncContext, pObj client.Object) bool {
 	// if host object is deleting we should delete it
 	if pObj.GetDeletionTimestamp() != nil {
 		return true
 	}
 
-	// if host object was synced before we should delete it as well
+	// if we're running with namespace sync enabled, we don't want to delete host synced objects
+	if ctx.Config.Config.Sync.ToHost.Namespaces.Enabled {
+		return false
```
FabianKramm: Wait so that means we never delete any objects on the host when using namespace sync? This seems not correct to me

FabianKramm (changes requested):

pkg/util/translate/translate.go
janekbaraniewski: no no, we'd only skip checks for annotations matching/not matching our vcluster instance name. reason being that if we re-deploy vcluster with the same name and re-import objects, we'll here mark them for deletion.  

Deletion of objects still removes them from host:

```
» k run nginx-from-vcluster -n default --image nginx 
pod/nginx-from-vcluster created
------------------------------------------------------------------------------------------------------------------------------------------
» k get po --context kind-vcluster-pro -n vcluster-my-vc-2 
NAME                                             READY   STATUS    RESTARTS   AGE
coredns-94f599b5-msccz-x-kube-system-x-my-vc-2   0/1     Running   0          61s
my-vc-2-0                                        1/1     Running   0          104s
nginx-from-vcluster-x-default-x-my-vc-2          1/1     Running   0          9s
------------------------------------------------------------------------------------------------------------------------------------------
» k delete po nginx-from-vcluster     (vcluster_my-vc-2_vcluster-my-vc-2_kind-vcluster-pro/default)
pod "nginx-from-vcluster" deleted
------------------------------------------------------------------------------------------------------------------------------------------
» k get po --context kind-vcluster-pro -n vcluster-my-vc-2
NAME                                             READY   STATUS    RESTARTS   AGE
coredns-94f599b5-msccz-x-kube-system-x-my-vc-2   0/1     Running   0          75s
my-vc-2-0                                        1/1     Running   0          118s
```

Same for objects in synced namespaces:

```
» k create ns sync-from-vcluster      (vcluster_my-vc-2_vcluster-my-vc-2_kind-vcluster-pro/default)
namespace/sync-from-vcluster created
------------------------------------------------------------------------------------------------------------------------------------------
» k -n sync-from-vcluster run nginx-from-vcluster --image nginx 
pod/nginx-from-vcluster created
------------------------------------------------------------------------------------------------------------------------------------------
» k --context kind-vcluster-pro -n host-from-vcluster get po                                  
NAME                  READY   STATUS    RESTARTS   AGE
nginx-from-vcluster   1/1     Running   0          11s
------------------------------------------------------------------------------------------------------------------------------------------
» k -n sync-from-vcluster delete po nginx-from-vcluster 
pod "nginx-from-vcluster" deleted
------------------------------------------------------------------------------------------------------------------------------------------
» k --context kind-vcluster-pro -n host-from-vcluster get po 
No resources found in host-from-vcluster namespace.
------------------------------------------------------------------------------------------------------------------------------------------
```

pkg/syncer/syncer.go
janekbaraniewski: then we probably should just explicitly cleanup objects when removing vCluster instead of overwriting. this way we don't need to change UID checks logic.

pkg/controllers/resources/register.go
```diff
@@ -44,8 +44,8 @@ func getSyncers(ctx *synccontext.RegisterContext) []BuildController {
 	return append([]BuildController{
 		isEnabled(ctx.Config.Sync.ToHost.Services.Enabled, services.New),
 		isEnabled(ctx.Config.Sync.ToHost.ConfigMaps.Enabled, configmaps.New),
-		isEnabled(ctx.Config.Sync.FromHost.ConfigMaps.Enabled, configmaps.NewFromHost),
-		isEnabled(ctx.Config.Sync.FromHost.Secrets.Enabled, secrets.NewFromHost),
```
FabianKramm: That seems odd to me, why is that necessary?

FabianKramm (approved):

PR #2917, opened 2025-06-16
[ENG-7336, ENG-7342] fix: Ignore the *Class limiting when no selector is specified

ulyssessouza:
This avoids applying exclusion in the following cases:
- When selector is not specified
- On Pods when priority class is a native one

/kind bugfix

Resolves ENG-7336
Resolves ENG-7342

Then no `selector` is specified no filtering is applied.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -982,6 +982,11 @@ type SyncFromHost struct {
 
 type StandardLabelSelector v1.LabelSelector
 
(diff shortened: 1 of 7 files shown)


johannesfrey (approved):
lgtm. But would be nice if one of the originally requested reviewers would also take a look.

PR #2919, opened 2025-06-16
fix: add control plane pods to certs

FabianKramm:
Fixes an issue where node ips were not correctly added to the server certificate when using a NodePort service with vCluster


lizardruss (approved):
lgtm

PR #2956, opened 2025-06-25
Delete coreDNS components during vcluster delete op

ApsTomar:
/kind bugfix

resolves ENG-7242

Fixed an issue where coreDNS components don't get deleted during vcluster delete operation

### DIFF
--- a/pkg/cli/delete_helm.go
+++ b/pkg/cli/delete_helm.go
@@ -189,7 +189,7 @@ func DeleteHelm(ctx context.Context, platformClient platform.Client, options *De
 	err = coredns.DeleteCoreDNSComponents(ctx, cmd.kubeClient, cmd.Namespace)
 	cmd.log.Info("Deleting CoreDNS components...")
 	if err != nil {
-		return fmt.Errorf("delete coreDNS components: %w", err)
+		cmd.log.Warnf("delete coreDNS components: %v", err)
 	}
 
 	// check if there are any other vclusters in the namespace you are deleting vcluster in.
--- a/pkg/cli/delete_helm.go
+++ b/pkg/cli/delete_helm.go
@@ -14,6 +14,7 @@ import (
 	"github.com/loft-sh/vcluster/pkg/cli/find"
 	"github.com/loft-sh/vcluster/pkg/cli/flags"
 	"github.com/loft-sh/vcluster/pkg/cli/localkubernetes"
+	"github.com/loft-sh/vcluster/pkg/coredns"
 	"github.com/loft-sh/vcluster/pkg/helm"
 	"github.com/loft-sh/vcluster/pkg/platform"
 	"github.com/loft-sh/vcluster/pkg/util/clihelper"
@@ -182,6 +183,13 @@ func DeleteHelm(ctx context.Context, platformClient platform.Client, options *De
 		} else {
 			cmd.log.Donef("Successfully deleted ConfigMap %s in namespace %s", configMapName, cmd.Namespace)
 		}
+	}
+
+	// delete coreDNS components since they're separately deployed and not with the vCluster helm chart
+	err = coredns.DeleteCoreDNSComponents(ctx, cmd.kubeClient, cmd.Namespace)
+	cmd.log.Info("Deleting CoreDNS components...")
+	if err != nil {
+		return fmt.Errorf("delete coreDNS components: %w", err)
 	}
 
 	// check if there are any other vclusters in the namespace you are deleting vcluster in.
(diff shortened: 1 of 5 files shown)


pkg/coredns/coredns.go
```diff
@@ -100,3 +103,37 @@ func GetUserID() int64 {
 
 	return int64(uid)
 }
+
+func DeleteCoreDNSComponents(ctx context.Context, client *kubernetes.Clientset, namespace string) error {
+	labelSelector := labels.SelectorFromSet(labels.Set{"k8s-app": "vcluster-kube-dns"}).String()
```
hidalgopl: minor / nit
```go
labelSelector := labels.FormatLabels(map[string]string{"k8s-app": "vcluster-kube-dns"})
```

hidalgopl: We're using `"k8s-app": "vcluster-kube-dns"` in multiple places and this may become a problem when we want to change it. Could you make it a constant and store in the `pkg/constants/coredns.go` and replace all occurrences of it in the code? Thanks

pkg/coredns/coredns.go
```diff
@@ -100,3 +103,37 @@ func GetUserID() int64 {
...
+					err = client.CoreV1().Services(namespace).Delete(ctx, svc.Name, metav1.DeleteOptions{})
+					if err != nil {
+						return err
+					}
+				}
+			}
+			return nil
+		},
+	}
+
```
hidalgopl, line 114: I'd rather do the best-effort cleanup by always running all delete commands and returning an error at the end:
```go
var errs []error
for _, deleteFn := range deleteFuncs {
		if err := deleteFn(); err != nil {
			errs = append(errs, err)
		}
	}
return errors.Join(errs...)
```

pkg/cli/delete_helm.go
```diff
@@ -184,6 +185,13 @@ func DeleteHelm(ctx context.Context, platformClient platform.Client, options *De
 		}
 	}
 
+	// delete coreDNS components since they're separately deployed and not with the vCluster helm chart
+	err = coredns.DeleteCoreDNSComponents(ctx, cmd.kubeClient, cmd.Namespace)
+	cmd.log.Info("Deleting CoreDNS components...")
+	if err != nil {
+		return fmt.Errorf("delete coreDNS components: %w", err)
```
hidalgopl: If the cleanup fails on any step later in the process (e.g. `find.ListVClusters`) then you cannot re-run it to clean remaining resources, because it will fail here (because coredns was deleted in the previous run. I think we can consider just warning here instead of erroring out

pkg/coredns/coredns.go
ApsTomar: Sure, I've created a const. Though most of the occurrences are in yaml files or templates where we can't replace them with consts, rest everywhere I've replaced.
Since, this label is specifically concerned with coreDNS, hence I've prefixed the label key and value with `coreDNSLabel`, let me know if you have a different opinion.

pkg/coredns/coredns.go
ApsTomar: Agreed, good point.

hidalgopl (dismissed):

pkg/coredns/coredns.go
```diff
@@ -100,3 +104,38 @@ func GetUserID() int64 {
...
+	var errs []error
+	for _, deleteFn := range deleteFuncs {
```
FabianKramm: I think we can simplify this, we can just do
```
 err := client.AppsV1().Deployments(namespace).DeleteCollection(ctx, metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: labelSelector})
if err != nil {
  errs = append(errs, err)
}

 err = client.CoreV1().Pods(namespace).DeleteCollection(ctx, metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: labelSelector})
if err != nil {
  errs = append(errs, err)
}
			services, err := client.CoreV1().Services(namespace).List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
			if err != nil {
				return err
			}
			if len(services.Items) != 0 {
				for _, svc := range services.Items {
					err = client.CoreV1().Services(namespace).Delete(ctx, svc.Name, metav1.DeleteOptions{})
					if err != nil {
  errs = append(errs, err)
					}
				}
```
I don't see the reason why we need an array of functions here to iterate over, just makes the code harder to understand.

FabianKramm (changes requested):

pkg/coredns/coredns.go
ApsTomar: @FabianKramm Made the suggested changes, please take a look.

FabianKramm (approved):

PR #2963, opened 2025-06-30
enhancement: Cleans crdName from version on Helm RBAC

ulyssessouza:
/kind enhancement

Part of ENG-6273

This is part of another issue implemented in `vcluster-pro`


chart/templates/_rbac.tpl
```diff
@@ -140,8 +140,9 @@
 {{- if .Values.sync.toHost.customResources }}
 {{- range $crdName, $rule := .Values.sync.toHost.customResources }}
 {{- if $rule.enabled }}
-- resources: [ "{{ (splitn "." 2 $crdName)._0 }}" ]
-  apiGroups: [ "{{ (splitn "." 2 $crdName)._1 }}" ]
+{{- $crdNameWithoutVersion := (split "/" $crdName)._0 -}}  # Takes part before "/"
+- resources: [ "{{ (splitn "." 2 $crdNameWithoutVersion)._0 }}" ]
+  apiGroups: [ "{{ (splitn "." 2 $crdNameWithoutVersion)._1 }}" ]
```
lizardruss, line 145: Could we add a test case in [role_test.yaml](https://github.com/loft-sh/vcluster/blob/main/chart/tests/role_test.yaml) to verify these cases?

chart/templates/_rbac.tpl
```diff
@@ -155,8 +156,9 @@
 {{- if .Values.sync.fromHost.customResources }}
 {{- range $crdName, $rule := .Values.sync.fromHost.customResources }}
 {{- if $rule.enabled }}
-- resources: [ "{{ (splitn "." 2 $crdName)._0 }}" ]
-  apiGroups: [ "{{ (splitn "." 2 $crdName)._1 }}" ]
+{{- $crdNameWithoutVersion := (split "/" $crdName)._0 -}}  # Takes part before "/"
+- resources: [ "{{ (splitn "." 2 $crdNameWithoutVersion)._0 }}" ]
+  apiGroups: [ "{{ (splitn "." 2 $crdNameWithoutVersion)._1 }}" ]
```
lizardruss, line 161: Could we add a test case in [clusterrole_test.yaml](https://github.com/loft-sh/vcluster/blob/main/chart/tests/clusterrole_test.yaml) to verify these cases?

chart/templates/_rbac.tpl
ulyssessouza: I added version to already existing tests

chart/templates/_rbac.tpl
ulyssessouza: I added version to already existing tests

chart/tests/clusterrole_test.yaml
```diff
@@ -459,7 +459,7 @@ tests:
       sync:
         fromHost:
           customResources:
-            test.test-group:
+            test.test-group/v1alpha1:
```
hidalgopl: is it backward compatible? I mean, can you still add the custom resource without the version? If yes, please add a test cases here & in the `role_test.yaml` for it

ulyssessouza: Yes. I'll remove the version on the test just above. We don't need in all of them

ulyssessouza: Done. For the `role_test.yaml` there is already a case without the version. `test.my-group`

johannesfrey (approved):

ulyssessouza:
@FabianKramm Not needed. It just strips the version if present.
I'll add a new entry under the `customResources` for the versioned ones

PR #2975, opened 2025-07-08
fix: Clean up Released PVs in virtual cluster

### LINKED ISSUE #2939: After deleting the PVC, the corresponding PV was not deleted
### What happened?

After deleting a PVC in the virtual cluster (vcluster), I observed that the PVC on the virtual cluster side was deleted, but the PV was not. Meanwhile, on the host cluster side, the corresponding PVC and PV could be deleted normally.

All PVCs and PVs should be purged from both the virtual cluster and the host cluster.

```
#Create a Pod that declares a PVC using the YAML file below.

# pod-with-pvc.yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: app-data-pvc
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
  storageClassName: nfs-csi
---
apiVersion: v1
kind: Pod
metadata:
  name: web-app-pod
  labels:
    app: web-server
spec:
  containers:
  - name: nginx-container
    image: nginx:latest
    ports:
    - containerPort: 80
    volumeMounts:
    - name: app-storage
      mountPath: "/usr/share/nginx/html"  
  volumes:
  - name: app-storage
    persistentVolumeClaim:
      claimName: app-data-pvc

# then delete this resource
kubectl delete -f pod-with-pvc.yaml
```

_No response_

<details>

```console
$ kubectl version
Client Version: v1.30.6
```

</details>

### vcluster version

<details>

```console
$ vcluster --version
vcluster version 0.23.0
```

</details>

### VCluster Config

<details>

```

controlPlane:
  advanced:
    globalMetadata: {}
    headlessService: {}
    virtualScheduler: {}
    workloadServiceAccount: {}
  backingStore:
    database:
      embedded: {}
      external: {}
    etcd:
      deploy:
        enabled: true
        headlessService: {}
        service: {}
        statefulSet:
          highAvailability:
            replicas: 3
          image: {}
          persistence:
            volumeClaim:
              enabled: true
              size: 50Gi
          resources:
            limits:
              cpu: "2"
              memory: 6Gi
            requests:
              cpu: 200m
              memory: 1Gi
          security: {}
      embedded: {}
  coredns:
    deployment:
      pods: {}
      replicas: 2
      resources: {}
    enabled: true
    service: {}
  distro:
    k0s:
      image: {}
    k3s:
      image: {}
    k8s:
      apiServer:
        image:
          tag: v1.30.6
      controllerManager:
        image:
          tag: v1.30.6
      enabled: true
      scheduler:
        image:
          tag: v1.30.6
      version: v1.30.6
  hostPathMapper: {}
  ingress: {}
  proxy:
    extraSANs:
    - 10.237.234.50
  service: {}
  serviceMonitor: {}
  statefulSet:
    highAvailability:
      replicas: 3
    persistence:
      volumeClaim:
        enabled: true
    probes:
      livenessProbe: {}
      readinessProbe: {}
      startupProbe: {}
    resources:
      limits:
        cpu: "4"
        memory: 6Gi
      requests:
        cpu: 300m
        memory: 1Gi
    security: {}
experimental:
  deploy:
    host: {}
    vcluster: {}
  genericSync:
    clusterRole: {}
    role: {}
  isolatedControlPlane: {}
  multiNamespaceMode: {}
  syncSettings: {}
  virtualClusterKubeConfig: {}
integrations:
  certManager:
    sync:
      fromHost:
        clusterIssuers:
          selector: {}
      toHost:
        certificates: {}
        issuers: {}
  externalSecrets:
    sync:
      clusterStores:
        selector: {}
      externalSecrets: {}
      stores: {}
    webhook: {}
  kubeVirt:
    apiService:
      service: {}
    sync:
      dataVolumes: {}
      virtualMachineClones: {}
      virtualMachineInstanceMigrations: {}
      virtualMachineInstances: {}
      virtualMachinePools: {}
      virtualMachines: {}
    webhook: {}
  metricsServer:
    apiService:
      service: {}
networking:
  advanced:
    proxyKubelets: {}
  replicateServices: {}
policies:
  centralAdmission: {}
  limitRange:
    default:
      cpu: 100m
      ephemeral-storage: 10Gi
      memory: 200Mi
    defaultRequest:
      cpu: "0"
      ephemeral-storage: 0Gi
      memory: "0"
    enabled: true
  networkPolicy:
    enabled: true
    outgoingConnections:
      ipBlock:
        cidr: 0.0.0.0/0
        except:
        - 172.26.0.0/16
  resourceQuota:
    enabled: true
    quota:
      count/configmaps: 1000
      count/endpoints: 1000
      count/persistentvolumeclaims: 1000
      count/pods: 1000
      count/secrets: 1000
      count/services: 1000
      limits.cpu: 1000
      limits.ephemeral-storage: 10000Gi
      limits.memory: 2000Gi
      requests.cpu: 500
      requests.ephemeral-storage: 2000Gi
      requests.memory: 4000Gi
      requests.storage: 10000Gi
      services.loadbalancers: 100
      services.nodeports: 100
rbac:
  clusterRole: {}
  role: {}
sync:
  fromHost:
    configMaps:
      mappings: {}
    csiDrivers: {}
    csiNodes: {}
    csiStorageCapacities: {}
    events: {}
    ingressClasses: {}
    nodes:
      clearImageStatus: true
    priorityClasses: {}
    runtimeClasses: {}
    secrets:
      mappings: {}
    storageClasses:
      enabled: true
    volumeSnapshotClasses: {}
  toHost:
    configMaps: {}
    endpoints: {}
    ingresses: {}
    networkPolicies:
      enabled: true
    persistentVolumeClaims:
      enabled: true
    persistentVolumes:
      enabled: true
    podDisruptionBudgets: {}
    pods:
      rewriteHosts:
        initContainer:
          resources: {}
    priorityClasses:
      enabled: true
    secrets: {}
    serviceAccounts: {}
    services: {}
    storageClasses: {}
    volumeSnapshotContents: {}
    volumeSnapshots: {}
telemetry: {}
```

</details>

After I added a label to the PV using the command below, the PV was triggered for deletion.
`kubectl label pv pvc-a5f95675-b9a9-4183-860e-0a347a53f55e leon=test`

squ94wk:
A Released PV would not trigger a sync on the Delete event from the host.

/kind bugfix

resolves #2939 

Fixed an issue where the deletion of a Released PV in the host cluster would not be propagate to the virtual cluster until vcluster was restarted or the object modified again.

This change does not change anything in the behavior of the syncer itself.
It changes what events are enqueued for reconciliation through the `shouldSync` and therefore `IsManaged` function of the PV syncer.
Therefore, its unit tests are not the right place for a regression test.

### DIFF
--- a/pkg/controllers/resources/persistentvolumes/syncer.go
+++ b/pkg/controllers/resources/persistentvolumes/syncer.go
@@ -309,6 +309,9 @@ func (s *persistentVolumeSyncer) shouldSync(ctx *synccontext.SyncContext, pObj *
 		} else if translate.Default.IsManaged(ctx, pObj) {
 			return true, nil, nil
 		}
+		if translate.Default.IsTargetedNamespace(ctx, pObj.Spec.ClaimRef.Namespace) && pObj.Status.Phase == corev1.VolumeReleased {
+			return true, nil, nil
+		}
 
 		return translate.Default.IsTargetedNamespace(ctx, pObj.Spec.ClaimRef.Namespace) && pObj.Spec.PersistentVolumeReclaimPolicy == corev1.PersistentVolumeReclaimRetain, nil, nil
 	}
--- a/test/e2e/syncer/pvc/pvc.go
+++ b/test/e2e/syncer/pvc/pvc.go
@@ -44,7 +44,7 @@ var _ = ginkgo.Describe("Persistent volume synced from host cluster", func() {
 		framework.ExpectNoError(err)
 	})
 
-	ginkgo.It("Test pvc provisioned successfully and is synced back to vcluster", func() {
+	ginkgo.It("PVC binds to dynamically provisioned PV, is synced back to vcluster and cleaned up properly", func() {
 		pvcName := "test"
 		podName := "nginx-test"
 
@@ -114,5 +114,17 @@ var _ = ginkgo.Describe("Persistent volume synced from host cluster", func() {
 		framework.ExpectNoError(err)
 
 		framework.ExpectEqual(vpvc.Status, pvc.Status)
+
+		err = f.VClusterClient.CoreV1().PersistentVolumeClaims(ns).Delete(f.Context, pvcName, metav1.DeleteOptions{})
+		framework.ExpectNoError(err)
+
+		err = f.VClusterClient.CoreV1().Pods(ns).Delete(f.Context, podName, metav1.DeleteOptions{})
+		framework.ExpectNoError(err)
+
+		err = f.WaitForPVCDeletion(ns, pvcName)
+		framework.ExpectNoError(err)
+
+		err = f.WaitForPVDeletion(vpvc.Spec.VolumeName)
+		framework.ExpectNoError(err)
 	})
 })
--- a/test/e2e/values.yaml
+++ b/test/e2e/values.yaml
@@ -69,3 +69,10 @@ sync:
           "from-host-sync-test-2/dummy": "barfoo2/dummy"
           "default/my-secret": "barfoo2/secret-my"
           "/my-secret-in-default-ns": "barfoo2/secret-from-default-ns"
+    storageClasses:
+      enabled: true
+  toHost:
+    persistentVolumeClaims:
+      enabled: true
+    persistentVolumes:
+      enabled: true
--- a/test/framework/util.go
+++ b/test/framework/util.go
@@ -180,6 +180,32 @@ func (f *Framework) WaitForServiceToUpdate(client *kubernetes.Clientset, service
 	})
 }
 
+func (f *Framework) WaitForPVCDeletion(namespace, name string) error {
+	return wait.PollUntilContextTimeout(f.Context, time.Second*5, PollTimeout, true, func(ctx context.Context) (bool, error) {
+		_, err := f.VClusterClient.CoreV1().PersistentVolumeClaims(namespace).Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			if kerrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, err
+		}
+		return false, nil
+	})
+}
+
+func (f *Framework) WaitForPVDeletion(name string) error {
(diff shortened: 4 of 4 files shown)


pkg/controllers/resources/persistentvolumes/syncer.go
```diff
@@ -309,6 +309,9 @@ func (s *persistentVolumeSyncer) shouldSync(ctx *synccontext.SyncContext, pObj *
 		} else if translate.Default.IsManaged(ctx, pObj) {
 			return true, nil, nil
 		}
+		if translate.Default.IsTargetedNamespace(ctx, pObj.Spec.ClaimRef.Namespace) && pObj.Status.Phase == corev1.VolumeReleased {
```
hidalgopl, line 312: would you mind adding a test case to the `pkg/controllers/resources/persistentvolumes/syncer_test.go`?
basically add a test case here: https://github.com/loft-sh/vcluster/blob/main/pkg/controllers/resources/persistentvolumes/syncer_test.go#L317-L333
It seems that you will need to add PV to InitialPhysicalState there and somehow force virtual client returning an error from `syncer..virtualClient.Get()`

Did you manage to test this manually / locally?

hidalgopl: If it's easy to reproduce, maybe it would make sense to have a new e2e tests for it in `test/e2e/syncer/pv`

squ94wk: The other tests call Sync directly.
The problem is that before Sync (or in this case relevant is SyncToHost), the reconciler checks if the resource is managed, which in the case of PVs specifically involves this `shouldSync` function.
Before this change, shouldSync would return `false`, since the previously bound PVC no longer exists.
The Delete event would not be enqueued and SyncToHost would not run, which now it does.

But if I call SyncToHost directly in my test, then it's not testing anything new.

It's easy to test it manually. But the bug only happens if the Delete event is registered. The object is deleted nonetheless when vcluster restarts or it's modified in the virtual cluster as this triggers a virtual event, which does not involve the shouldSync function.

We could write a new test for IsManaged() of the PV syncer, but then we should probably refactor that logic, too.
It's very convoluted.

PR #1561, opened 2024-02-26
now makes sure that the secret references localhost instead of old service name

facchettos:
**What issue type does this pull request address?** (keep at least one, remove the others) 
/kind bugfix

**What does this pull request do? Which issues does it resolve?** (use `resolves #<issue_number>` if possible) 
resolves eng-2901 

**Please provide a short message that should be published in the vcluster release notes**
Fixed an issue where vcluster would not be able to go from 0.18 to 0.19

**What else do we need to know?**

PR #2976, opened 2025-07-08
Cert rotation

johannesfrey:
/kind feature

resolves ENG-7787
resolves ENG-7788
resolves ENG-7789

Add vcluster cert rotation commands

This PR introduces following new commands:
- `vcluster certs rotate`: Rotate client/server leaf certificates using the current CA.
- `vcluster certs rorate-ca`: Rotate all certificates including the CA.
- `vcluster certs check`: Display basic infos about the current certificates.

Most of the actual rotation functionality is under `cmd/vcluster/cmd/certs` resp. `/pkg/certs/rotate.go` in order to be able to rotate on the file system level, which makes it compatible to "standalone", which has no notion of a host cluster. Nevertheless the cert secret is still updated in case we are non-standalone because otherwise virtual clusters without persistence would not see the new certificates after the restart.
The CLI counterparts under `cmd/vclusterctl/cmd/certs` are merely responsible to exec the commands above either by pausing and resuming the virtual cluster and starting a standalone pod or directly executing inside the virtual cluster pod (depending if there's persistence or not).

This PR already provides some rudimentary e2e tests. More will be added to this [branch](https://github.com/loft-sh/vcluster/pull/2985) in parallel in order to "unblock" reviews of the main PR here..

### DIFF
--- a/cmd/vcluster/cmd/certs/check.go
+++ b/cmd/vcluster/cmd/certs/check.go
@@ -80,7 +80,7 @@ func (cmd *checkCmd) Run() error {
 	})
 
 	if err != nil {
-		return fmt.Errorf("walking PKI dir: %w", err)
+		return fmt.Errorf("finding certificate information: %w", err)
 	}
 
 	if err := json.NewEncoder(os.Stdout).Encode(certificateInfos); err != nil {
--- a/cmd/vcluster/cmd/certs/check.go
+++ b/cmd/vcluster/cmd/certs/check.go
@@ -1,0 +1,102 @@
+package certs
+
+import (
+	"crypto/x509"
+	"encoding/json"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/loft-sh/log"
+	"github.com/loft-sh/vcluster/pkg/certs"
+	"github.com/loft-sh/vcluster/pkg/constants"
+	"github.com/spf13/cobra"
+	certutil "k8s.io/client-go/util/cert"
+)
+
+type checkCmd struct {
+	Path string
+	log  log.Logger
+}
+
+func check() *cobra.Command {
+	cmd := &checkCmd{
+		log: log.GetInstance(),
+	}
+
+	checkCmd := &cobra.Command{
+		Use:   "check",
+		Short: "Checks the current certificates",
+		Args:  cobra.NoArgs,
+		RunE: func(_ *cobra.Command, _ []string) error {
+			return cmd.Run()
+		}}
+
+	checkCmd.Flags().StringVar(&cmd.Path, "path", constants.PKIDir, "Destination path to the PKI directory")
+
+	return checkCmd
+}
+
+// Run checks the current certificates in the PKI directory and returns base information about those.
+func (cmd *checkCmd) Run() error {
+	now := time.Now()
+	var certificateInfos []certs.Info
+	err := filepath.WalkDir(cmd.Path, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if d.IsDir() {
+			return nil
+		}
+
+		if filepath.Ext(path) != ".crt" {
+			return nil
+		}
+
+		c, err := os.ReadFile(path)
+		if err != nil {
+			return fmt.Errorf("reading file %s", path)
+		}
+
+		crts, err := certutil.ParseCertsPEM(c)
+		if err != nil {
+			return err
+		}
+
+		for _, crt := range crts {
+			certificateInfos = append(certificateInfos, certs.Info{
+				Filename:   d.Name(),
+				Subject:    crt.Subject.CommonName,
+				Issuer:     crt.Issuer.CommonName,
+				ExpiryTime: crt.NotAfter,
+				Status:     certStatus(crt, now),
+			})
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		return fmt.Errorf("walking PKI dir: %w", err)
+	}
+
+	if err := json.NewEncoder(os.Stdout).Encode(certificateInfos); err != nil {
+		return fmt.Errorf("encoding JSON: %w", err)
+	}
+
+	return nil
+}
+
+func certStatus(cert *x509.Certificate, now time.Time) string {
+	if now.Before(cert.NotBefore) {
+		return "NOT YET VALID"
+	}
+	if now.After(cert.NotAfter) {
+		return "EXPIRED"
+	}
+
+	return "OK"
+}
--- a/cmd/vcluster/cmd/certs/rotate.go
+++ b/cmd/vcluster/cmd/certs/rotate.go
@@ -1,0 +1,69 @@
+package certs
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/loft-sh/log"
+	"github.com/loft-sh/vcluster/pkg/certs"
+	"github.com/loft-sh/vcluster/pkg/config"
+	"github.com/loft-sh/vcluster/pkg/constants"
+	"github.com/loft-sh/vcluster/pkg/pro"
+	"github.com/loft-sh/vcluster/pkg/setup"
+	"github.com/spf13/cobra"
+)
+
+type rotateCmd struct {
+	log log.Logger
+}
+
+func rotate() *cobra.Command {
+	cmd := &rotateCmd{
+		log: log.GetInstance(),
+	}
+
+	rotateCmd := &cobra.Command{
+		Use:   "rotate",
+		Short: "Rotates control-plane client and server certs",
+		Args:  cobra.NoArgs,
+		RunE: func(cobraCmd *cobra.Command, _ []string) error {
+			return cmd.Run(cobraCmd.Context(), false)
+		}}
+
+	return rotateCmd
+}
+
+func rotateCA() *cobra.Command {
+	cmd := &rotateCmd{
+		log: log.GetInstance(),
+	}
+
+	rotateCACmd := &cobra.Command{
+		Use:   "rotate-ca",
+		Short: "Rotates the CA certificate",
+		Args:  cobra.NoArgs,
+		RunE: func(cobraCmd *cobra.Command, _ []string) error {
+			return cmd.Run(cobraCmd.Context(), true)
+		}}
+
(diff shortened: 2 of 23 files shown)


pkg/certs/constants.go
```diff
@@ -140,6 +140,10 @@ const (
 	DefaultAPIServerBindAddress = "0.0.0.0"
 )
 
+func CertMap() map[string]string {
```
zerbitx: Any caller of this will be able to modify the `certMap`.  Maybe clone the map if you don't want that, or just export the map directly?

johannesfrey: Good point. I'll restructure so that most of the logic moves to `pkg/certs` so there's no need anymore to export this.

cmd/vcluster/cmd/certs/check.go
```diff
@@ -0,0 +1,102 @@
...
+				ExpiryTime: crt.NotAfter,
+				Status:     certStatus(crt, now),
+			})
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		return fmt.Errorf("walking PKI dir: %w", err)
```
zerbitx: For cli output "walking PKI dir" could be more user friendly.  Maybe something more about finding or determining certificate information.

cmd/vcluster/cmd/certs/rotate.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func rotate() *cobra.Command {
+	cmd := &rotateCmd{
+		log: log.GetInstance(),
+	}
+
+	rotateCmd := &cobra.Command{
+		Use:   "rotate",
+		Short: "Rotates control-plane client and server certs",
```
zerbitx, line 28: Do you think it would be worthwhile to add a "Long" description to these as well, like [this one](https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/platform/start.go#L43-L57)

zerbitx: Ah, just scrolled down and saw that you have the "long" on a few others.

cmd/vclusterctl/cmd/certs/rotate.go
```diff
@@ -0,0 +1,83 @@
...
+		Short: "Rotates the CA certificate",
+		Long: `##############################################################
+################## vcluster certs rotate-ca ##################
+##############################################################
+Rotates the CA certificates of the given virtual cluster using
+the current CA certificates.
+The CA files (ca.{crt,key}) can be placed in the PKI directory
+(either /data/pki or /var/lib/vcluster/pki) to issue new leaf
+certificates to be signed by that CA.
+If the ca.crt file is a bundle containing multiple certificates
```
zerbitx, line 69: Great bit of information!

pkg/certs/rotate.go
```diff
@@ -0,0 +1,219 @@
...
+	Issuer     string    `json:"issuer,omitempty"`
+	ExpiryTime time.Time `json:"expiryTime"`
+	Status     string    `json:"status,omitempty"` // "OK", "EXPIRED"
+}
+
+// Rotate rotates the certificates in the PKI directory.
+// If running non-standalone it also updates the cert secret to contain the newly created certificates.
+// Depending on the withCA argument this either means rotation the leaf certificates (withCA=false)
+// or the whole PKI infra (withCA=true). In both cases the current SA pub and private key are untouched.
+// IMPORTANT NOTE: The function expects the clients in vConfig to be populated.
```
zerbitx: Rather than just noting this in a comment, maybe check for the fields and do the [setup](https://github.com/loft-sh/vcluster/pull/2976/files?w=1#diff-9fee33c88e62e8cce171f5f22d07b3d49f773c07df181c833ab7727285225b86R54-R67) here if they're missing, or just always do the setup here instead of passing in the `vConfig`?

pkg/certs/rotate.go
```diff
@@ -0,0 +1,219 @@
...
+		d, err := os.ReadFile(filepath.Join(pkiPath, k))
+		if err != nil {
+			return fmt.Errorf("reading file %s: %w", k, err)
+		}
+
+		data[v] = d
+	}
+	secret.Data = data
+
+	_, err = client.CoreV1().Secrets(secretNamespace).Update(ctx, secret, metav1.UpdateOptions{})
```
zerbitx: nit: prefer a [patch](https://github.com/loft-sh/vcluster/blob/main/pkg/k8s/migrate.go#L203-L208) to an update.

pkg/util/podhelper/pod.go
```diff
@@ -0,0 +1,237 @@
...
+		})
+	}
+
+	// build the pod spec
+	newPod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: fmt.Sprintf("vcluster-%s-", containerName),
+			Namespace:    vCluster.Namespace,
+			Labels: map[string]string{
+				"app": fmt.Sprintf("vcluster-%ss", containerName),
```
zerbitx: Is there supposed to be a trailing "s" here?

pkg/util/podhelper/wait.go
```diff
@@ -0,0 +1,121 @@
...
+					reason = containerStatus.State.Terminated.Reason
+					message = containerStatus.State.Terminated.Message
+				} else if containerStatus.State.Waiting != nil {
+					reason = containerStatus.State.Waiting.Reason
+					message = containerStatus.State.Waiting.Message
+				}
+
+				out, err := kubeClient.CoreV1().Pods(namespace).GetLogs(pod.Name, &corev1.PodLogOptions{
+					Container: container,
+				}).Do(context.Background()).Raw()
```
zerbitx: Should this be `ctx` instead of `context.Background()`?  Not sure if the polling method cancels that context at an inopportune time messing up the log output.

pkg/util/podhelper/wait.go
```diff
@@ -0,0 +1,121 @@
...
+			// this is a fatal
+			return false, fmt.Errorf("error trying to retrieve pod %s/%s: %w", namespace, name, err)
+		}
+
+		for _, containerStatus := range pod.Status.ContainerStatuses {
+			if containerStatus.Name != container {
+				continue
+			}
+
+			if containerStatus.State.Running != nil {
```
zerbitx, line 95: nit: since each of these conditionals return, you flatten the `else if`s

cmd/vclusterctl/cmd/certs/rotate.go
johannesfrey: Currently up for [discussion](https://www.notion.so/vCluster-Certificate-Rotation-21510940806980f19ad9fba9fda9439b?d=22c10940806980ce8df5001c096d7e56&source=copy_link#21510940806980108d14c72a8c6ecc4f) if we want to support resp. document so prominently.

pkg/certs/rotate.go
johannesfrey: Makes sense. Though, this requires a bit more restructuring because this would currently introduce an import cycle.  I would need `setup` here while `setup` already imports `certs`.

johannesfrey: Went ahead and did the restructure.

pkg/cli/certs/rotate_helm.go
```diff
@@ -0,0 +1,198 @@
...
+	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		return fmt.Errorf("creating kubernetes client: %w", err)
+	}
+
+	var restartETCD bool
+	sts, err := kubeClient.AppsV1().StatefulSets(vCluster.Namespace).List(ctx, metav1.ListOptions{LabelSelector: "app=vcluster-etcd,release=" + vCluster.Name})
+	if err != nil && !kerrors.IsNotFound(err) {
+		return fmt.Errorf("listing potential etcd statefulsets: %w", err)
+	}
```
zerbitx: If I'm understanding, we only care about the replica count if they're trying to use the `RotationCmdCACerts` command.  If that's the case, you could move this lookup into the `RotationCmdCACerts` case of the switch statement and avoid looking up the sts for any other command. So the switch would look more like

```go
switch rotationCmd {
case RotationCmdCerts:
	cmd = fmt.Sprintf("%s/vcluster certs %s", cmd, RotationCmdCerts)
case RotationCmdCACerts:
	sts, err := kubeClient.AppsV1().StatefulSets(vCluster.Namespace).List(ctx, metav1.ListOptions{LabelSelector: "app=vcluster-etcd,release=" + vCluster.Name})
	if err != nil && !kerrors.IsNotFound(err) {
		return fmt.Errorf("listing potential etcd statefulsets: %w", err)
	}
	if len(sts.Items) > 0 {
                 // 👇 only has to check replicas since we're in the CA case
		if sts.Items[0].Status.AvailableReplicas > 1 {
			return fmt.Errorf("cert rotation with CA is currently not supported for deployed etcd in HA mode")
		}
		restartETCD = true
	}
	cmd = fmt.Sprintf("%s/vcluster certs %s", cmd, RotationCmdCACerts)
default:
	return fmt.Errorf("unknown rotation command: %s", rotationCmd)
}
```

zerbitx (dismissed):
Left a suggestion that shouldn't be a blocker, so approving.  Happy to take another look if you want to make the change.

pkg/cli/certs/rotate_helm.go
johannesfrey: Yes. That make sense 👌

zerbitx (approved):
lgtm

pkg/certs/ensure.go
```diff
@@ -75,63 +136,12 @@ func EnsureCerts(
 	secretName := CertSecretName(options.Name)
 	secret, err := currentNamespaceClient.CoreV1().Secrets(currentNamespace).Get(ctx, secretName, metav1.GetOptions{})
 	if err == nil {
-		// download certs from secret
```
FabianKramm, line 78: Thats all not needed anymore because of k0s not being there anymore correct?

johannesfrey: It looks like this has been introduced by this PR here: https://github.com/loft-sh/vcluster/pull/1561
It had something to do with upgrading from 0.18 to 0.19. Which also shouldn't be relevant anymore. But it could also be something related to k0s, but I'm not sure about that.

PR #2983, opened 2025-07-09
enhancement: Retrieves an already existing CRD with another version and adds the new version into it.

ulyssessouza:
This sets the newly added version as storage version

/kind enhancement

resolves #ENG-7766

### DIFF
--- a/pkg/util/translate/translate.go
+++ b/pkg/util/translate/translate.go
@@ -332,6 +332,10 @@ func applyMaps(fromMap, toMap map[string]string, opts ApplyMapsOptions) (map[str
 	return retMap, managedKeysStr
 }
 
+func hasStatus(version apiextensionsv1.CustomResourceDefinitionVersion) bool {
+	return version.Subresources != nil && version.Subresources.Status != nil
+}
+
 func getCrdVersionByName(crdVersions []apiextensionsv1.CustomResourceDefinitionVersion, versionName string) *apiextensionsv1.CustomResourceDefinitionVersion {
 	for _, version := range crdVersions {
 		if version.Name == versionName {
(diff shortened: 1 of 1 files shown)


pkg/util/translate/translate.go
```diff
@@ -369,42 +400,60 @@ func EnsureCRDFromPhysicalCluster(ctx context.Context, pConfig *rest.Config, vCo
...
+			newVersions = append(newVersions, *newVersion)
+			vCrdDefinition.Spec.Versions = newVersions
+			// Update the CRD in the virtual cluster
+			klog.FromContext(ctx).Info("Updating CRD in virtual cluster with new version", "crd", vCrdDefinition.Name, "version", groupVersionKind.Version)
+			_, err = vClient.ApiextensionsV1().CustomResourceDefinitions().Update(ctx, vCrdDefinition, metav1.UpdateOptions{})
+			if err != nil {
+				return isClusterScoped, hasStatusSubresource, fmt.Errorf("update crd in virtual cluster: %w", err)
+			}
+			// Check if the status subresource is set
+			hasStatusSubresource = newVersion.Subresources != nil && newVersion.Subresources.Status != nil
```
hidalgopl: this is a check duplicated many times in this file. Let's make a helper function `func hasStatus(version) bool` and use it here and other places in this file.

pkg/util/translate/translate.go
```diff
@@ -332,15 +332,46 @@ func applyMaps(fromMap, toMap map[string]string, opts ApplyMapsOptions) (map[str
...
+	pCrdDefinition, err := pClient.ApiextensionsV1().CustomResourceDefinitions().Get(ctx, groupVersionResource.GroupResource().String(), metav1.GetOptions{})
+	if err != nil {
+		return isClusterScoped, hasStatusSubresource, errors.Wrap(err, "retrieve crd in host cluster")
+	}
+
+	apiResource, err := KindExists(vConfig, groupVersionKind)
+	if err != nil && !kerrors.IsNotFound(err) { // If the kind does not exist, we will create it in the virtual cluster
+		return isClusterScoped, hasStatusSubresource, fmt.Errorf("check virtual cluster kind: %w", err)
+	}
+	if err == nil { // Exact GVK match
```
hidalgopl: can we improve readability of this?

it seems there are 3 cases here:
1. err is nil
2. err is not nil, and it isn't "is not found"
3. err is "is not found".

we can have one if / else if / else block or a switch statement here.
also, I think you don't need to check if error is nil here:
```go
if err != nil && !kerrors.IsNotFound(err)
```
and just do `!kerrors.IsNotFound(err)`. If nil is  passed to this function, it will return false.

hidalgopl: basically, it seems there are 4 cases:
```go
	switch {
	case err != nil && !kerrors.IsNotFound(err):
		// cannot find CRD in virtual, return error
	case err == nil:
		// kind exists, just get existing version of the virtual CRD
	case kerrors.IsNotFound(err) && "crd found in virtual":
		// exact version does not exist in virtual,
		// update virtual CRD with new version
	case kerrors.IsNotFound(err) && "crd not found in virtual"	
	}
```
for the blocks that end with return, could you move it to the separate function and call it there? It's quite hard to read and comprehend currently

hidalgopl (changes requested):

pkg/util/translate/translate.go
```diff
@@ -369,42 +400,60 @@ func EnsureCRDFromPhysicalCluster(ctx context.Context, pConfig *rest.Config, vCo
...
 	}
-
-	// get resource from kind name in physical cluster
-	groupVersionResource, err := ConvertKindToResource(pConfig, groupVersionKind)
-	if err != nil {
-		if kerrors.IsNotFound(err) {
-			return isClusterScoped, hasStatusSubresource, fmt.Errorf("seems like resource %s is not available in the physical cluster or vcluster has no access to it", groupVersionKind.String())
+	if kerrors.IsNotFound(err) { // Try to find an already exiting CRD in the virtual cluster that has other version. Since exact GVK match failed, we will try to find a CRD with the same group and kind, but different version.
+		vCrdDefinition, err := vClient.ApiextensionsV1().CustomResourceDefinitions().Get(ctx, groupVersionResource.GroupResource().String(), metav1.GetOptions{})
+		if err != nil && !kerrors.IsNotFound(err) {
```
hidalgopl: it seems that the case when `kerrors.IsNotFound(err)` is implicitly not handled at all

ulyssessouza: This is actually the case where the CRD doesn't exist.
So this is the nominal case where the CRD doesn't exist and will be created down in the file.

pkg/util/translate/translate.go
ulyssessouza: Actually there are 3 cases. The change I did is to:

```golang
switch {
case exactMatchInVCluster: // CRD exists in the physical cluster and in the virtual cluster with the same GVK
	return checkSubresourceStatus(ctx, vClient, apiResource, groupVersionKind)
case vCrdExists: // CRD exists in the virtual cluster but needs an update to add the new version
	return crdUpdateWithNewVersion(ctx, vClient, pCrdDefinition, vCrdDefinition, groupVersionKind)
default: // CRD does not exist in the virtual cluster, need to create it
	return createCrdFromPhysicalCluster(ctx, vClient, pCrdDefinition, groupVersionResource, groupVersionKind)
}
```

pkg/util/translate/translate.go
```diff
@@ -332,97 +332,123 @@ func applyMaps(fromMap, toMap map[string]string, opts ApplyMapsOptions) (map[str
...
+	return isClusterScoped, hasStatusSubresource, err
+}
 
-	// get resource from kind name in physical cluster
-	groupVersionResource, err := ConvertKindToResource(pConfig, groupVersionKind)
-	if err != nil {
-		if kerrors.IsNotFound(err) {
-			return isClusterScoped, hasStatusSubresource, fmt.Errorf("seems like resource %s is not available in the physical cluster or vcluster has no access to it", groupVersionKind.String())
-		}
+func crdUpdateWithNewVersion(ctx context.Context, vClient *apiextensionsv1clientset.Clientset, pCrdDefinition, vCrdDefinition *apiextensionsv1.CustomResourceDefinition, groupVersionKind schema.GroupVersionKind) (bool, bool, error) {
```
zerbitx, line 379: (nits non-blocking): Recommend a shorter name like `addCRDVersion` and passing in the version by name you get from `getCrdVersionByName` on [398](https://github.com/loft-sh/vcluster/pull/2983/files?w=1#diff-554810cc82c273edb027c1636c5110a42b7de94605d3fe86ebb22212fd12cdb2R398) instead of `pCrdDefinition` just to use it for that call.

zerbitx (approved):
lgtm

hidalgopl (approved):

PR #2987, opened 2025-07-11
DRAFT: vCluster autoscaling

FabianKramm:


config/config.go
```diff
@@ -109,6 +109,163 @@ type PrivateNodes struct {
...
+	//
+	// The feature can also be used to allow maximum time limits for long-running jobs which can delay node termination with preStop hooks.
+	// Defaults to 30s. Set to Never to wait indefinitely for pods to be drained.
+	TerminationGracePeriod string `json:"terminationGracePeriod,omitempty"`
+
+	// The amount of time a Node can live on the cluster before being removed
+	ExpireAfter string `json:"expireAfter,omitempty"`
+
+	// Weight is the weight of this node pool.
+	Weight int `json:"weight,omitempty"`
```
pascalbreuninger, line 184: Is this equivalent to `cost` or different?

FabianKramm: No this is the nodePool weight which is a karpenter concept described here: https://karpenter.sh/docs/concepts/nodepools/#specweight

config/config.go
```diff
@@ -183,10 +340,22 @@ type Standalone struct {
...
 	Bundle string `json:"bundle,omitempty"`
 
+	// Nodes is a list of nodes to deploy for standalone mode.
+	Nodes StandaloneNodes `json:"nodes,omitempty"`
+
 	// JoinNode holds configuration for the standalone control plane node.
 	JoinNode StandaloneJoinNode `json:"joinNode,omitempty"`
 }
 
+type StandaloneNodes struct {
```
pascalbreuninger, line 350: Don't we need `terminationGracePeriod`, `NodeLabels` and `Taints` for this as well?

FabianKramm: This can be defined via `controlPlane.standalone.joinNode.taints` etc.

FabianKramm: Termination grace period we don't need as this will not be managed by karpenter

config/config.go
pascalbreuninger: Is this correct then:
- `cost` ranks NodeTypes against one another when deciding between two equivalent (in terms of resources and requirements) NodeTypes 
- `weight` ranks NodePools inside of the virtual cluster when deciding which NodePool to use for new nodes

config/config.go
pascalbreuninger: And `NodeLabels`?

FabianKramm: Since this will be not done by karpenter node labels can be added via kubelet args:
```
controlPlane:
  standalone:
    joinNode:
      nodeRegistration:
        kubeletExtraArgs:
        - name: node-labels
           value: my-label-1=my-value-1,my-label-2=my-value-2
```

pascalbreuninger (approved):

PR #2988, opened 2025-07-11
Clarifies how default k8s version is set

cbron:
Follow up to 
* ENG-6743
* https://github.com/loft-sh/vcluster/pull/2762


zerbitx (approved):
lgtm

matskiv (approved):

PR #2992, opened 2025-07-15
feat: RewriteHosts initContainer image may use override image

squ94wk:
BREAKING CHANGE: Change config.sync.toHost.pods.rewriteHosts.initContainer.image from string to struct.

/kind feature

This changes the config for the "rewriteHosts" feature to be in line with other places.
The init container will now also use the `config.controlPlane.advanced.defaultRegistry`.

Allow to use a custom registry for the rewriteHosts init container.
BREAKING CHANGE: replaced the config string `sync.toHost.pods.rewriteHosts.initContainer.image` with object that takes `registry`, `repository`, `tag`.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1315,7 +1315,7 @@ type SyncRewriteHosts struct {
 
 type SyncRewriteHostsInitContainer struct {
 	// Image is the image virtual cluster should use to rewrite this FQDN.
-	Image string `json:"image,omitempty"`
+	Image Image `json:"image,omitempty"`
 
 	// Resources are the resources that should be assigned to the init container for each stateful set init container.
 	Resources Resources `json:"resources,omitempty"`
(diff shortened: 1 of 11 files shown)


zerbitx (approved):
lgtm

config/config.go
```diff
@@ -1722,17 +1722,20 @@ type StatefulSetImage struct {
...
+	// Repository is the repository of the container image reference, e.g. my-repo/my-image.
 	Repository string `json:"repository,omitempty"`
 
-	// Tag is the tag of the container image, and is the default version.
+	// Tag is the tag of the container image reference, and is the default version.
 	Tag string `json:"tag,omitempty"`
 }
 
+func (i Image) String() string {
+	return fmt.Sprintf("%s/%s:%s", strings.TrimSuffix(i.Registry, "/"), i.Repository, i.Tag)
```
FabianKramm: What happens if registry or tag is empty? Do we not want to allow this? E.g. "alpine" is a valid image reference. Since this should be a generic function, we should add this.

config/config.go
```diff
@@ -1722,17 +1722,20 @@ type StatefulSetImage struct {
 }
 
 type Image struct {
-	// Registry is the registry of the container image, e.g. my-registry.com or ghcr.io. This setting can be globally
-	// overridden via the controlPlane.advanced.defaultImageRegistry option. Empty means docker hub.
+	// Registry is the registry of the container image reference, e.g. ghcr.io.
```
FabianKramm: Whats the reasoning behind changing the message here? I think we consciously added the part about setting this globally

config/values.yaml
```diff
@@ -28,7 +28,10 @@ sync:
       rewriteHosts:
         enabled: true
         initContainer:
-          image: "library/alpine:3.20"
+          image:
+            registry: ""
+            repository: library/alpine
+            tag: "3.20"
```
FabianKramm, line 34: Do you see a way to reuse the vCluster image here? Then we wouldn't even need that option and a separate image anymore?

FabianKramm (changes requested):
I think the PR looks good, but shouldn't we consider reusing the vCluster image as the init container as that would simplify the setup in air-gapped?

config/values.yaml
squ94wk: I can imagine users don't want a >300M image where <10M suffice.
If we mirrored alpine or busybox, we could replace the docker hub reference.

config/config.go
squ94wk: I wasn't aware at the time that this translates directly to the docs.
Since this was a generic type, I thought it should be generic.
But it makes sense for the docs.
I'll revert this.

FabianKramm (approved):

PR #2999, opened 2025-07-17
chore(ci): use pull_request_target for fork E2E testing

sydorovdmytro:
Switch from pull_request to pull_request_target to enable E2E testing on PRs from forks while maintaining security through repository-level approval controls.

/kind enhancement

related to OPS-237

With this change, we can later use DockerHub credentials to avoid rate limiting on pulls.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -3,7 +3,7 @@ name: E2E CI
 on:
   release:
     types: [created]
-  pull_request:
+  pull_request_target:
     branches:
       - main
       - v*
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yaml
```diff
@@ -3,7 +3,7 @@ name: E2E CI
 on:
   release:
     types: [created]
-  pull_request:
+  pull_request_target:
```
cbron, line 6: > Switch from pull_request to pull_request_target to enable E2E testing on PRs from forks while maintaining security through repository-level approval controls.

What are the security controls here ? `pull_request_target` is notorious for allowing untrusted users on OSS repos access to secrets. More info [here](https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git) and [here](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target)

sydorovdmytro: Good catch, thanks. Let me have another look into it.

Piotr1215: Agree with @cbron `pull_request_target` is a no-go for public repositories.

PR #3003, opened 2025-07-18
fix: also restart etcd when doing leaf cert rotation

johannesfrey:
/kind bugfix

resolves #

Furthermore, this simplifies the code by simply exiting early when the replica count is higher than 1.


zerbitx (dismissed):
lgtm

PR #3007, opened 2025-07-22
Fix: config change breaks upgrade to 0.27.0

squ94wk:
/kind bugfix

resolves #ENG-8029

Fixed upgrade of vclusters created with <0.27.0.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1730,14 +1730,13 @@ func (i *Image) UnmarshalJSON(data []byte) error {
 		return nil
 	}
 
-	type Alias Image // Prevent infinite recursion
-	aux := &struct {
-		*Alias
-	}{
-		Alias: (*Alias)(i),
+	type Alias Image
+	var aux Alias
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
 	}
-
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -1721,7 +1722,46 @@ type Image struct {
...
+	}
+
+	type Alias Image // Prevent infinite recursion
+	aux := &struct {
+		*Alias
+	}{
+		Alias: (*Alias)(i),
+	}
+
+	return json.Unmarshal(data, &aux)
```
lizardruss: Is there an advantage to the embedded struct pointer vs this?
```golang
type Alias Image
var aux Alias
err := json.Unmarshal(data, &aux)
if err != nil {
    return err
}
*i = Image(aux)
return nil
```

lizardruss (approved):
Added a nit / question comment, but should not hold up approval.

johannesfrey (approved):

config/config.go
squ94wk: Don't see really a meaningful difference.
It's maybe a bit more explicit.

johannesfrey (approved):

PR #3013, opened 2025-07-23
fix: also set loftHost if host and no-tunnel flags are set at the sam…

johannesfrey:
…e time

/kind bugfix

resolves ENG-7988

Fixed an issue where vcluster would not set loftHost if platform start is called with no-tunnel. 


zerbitx (approved):
lgtm

lizardruss (approved):

PR #3014, opened 2025-07-23
Fix: advanced.defaultRegistry takes precedence over image.registry field

squ94wk:
/kind bugfix

resolves ENG-8053

### DIFF
--- a/pkg/controllers/resources/pods/translate/translator_test.go
+++ b/pkg/controllers/resources/pods/translate/translator_test.go
@@ -336,7 +336,7 @@ func TestRewriteHostsTranslation(t *testing.T) {
 			},
 		},
 		{
-			name: "Use default registry if specified",
+			name: "Default registry has precedence over specified registry",
 			pod: corev1.Pod{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "pod-name",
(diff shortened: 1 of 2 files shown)


sydorovdmytro (approved):

pkg/controllers/resources/pods/translate/translator_test.go
```diff
@@ -348,6 +348,7 @@ func TestRewriteHostsTranslation(t *testing.T) {
 			},
 			configMod: func(c *config.VirtualClusterConfig) {
 				c.ControlPlane.Advanced.DefaultImageRegistry = "my-registry"
+				c.Sync.ToHost.Pods.RewriteHosts.InitContainer.Image.Registry = "mirror.gcr.io"
```
Piotr1215, line 351: @squ94wk I wonder if this has any implications for the default installation logic, should we mention this in the docs?

squ94wk: This is a test file and we corrected the expectation in the test and the code.

squ94wk: We haven't made any changes yet to default config.
Only that it's possible now to specify the image in its parts.
That is already documented in ENG-7954.

cbron (approved):
lgtm, this was net-new code for 0.27 anyway

pkg/controllers/resources/pods/translate/translator_test.go
cbron: That one is ongoing here: https://github.com/loft-sh/vcluster/pull/3002

PR #3018, opened 2025-07-23
Snapshot List and Delete

lizardruss:
/kind feature

resolves ENG-7418
resolves ENG-7434

Add commands to list and delete vcluster snapshots

~~Currently delete does not work for ghcr.io registries, since they do not support the OCI spec for deleting tags. PR will remain in draft until a workaround can be implemented.~~

### DIFF
--- a/cmd/vcluster/cmd/snapshot/list.go
+++ b/cmd/vcluster/cmd/snapshot/list.go
@@ -15,7 +15,7 @@ func NewListCmd() *cobra.Command {
 		Short: "list vCluster snapshots",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			options := &SnapshotOptions{}
+			options := &Options{}
 			envOptions, err := parseOptionsFromEnv()
 			if err != nil {
 				klog.Warningf("Error parsing environment variables: %v", err)
--- a/cmd/vcluster/cmd/snapshot/list.go
+++ b/cmd/vcluster/cmd/snapshot/list.go
@@ -1,0 +1,44 @@
+package snapshot
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	"k8s.io/klog/v2"
+)
+
+func NewListCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "list vCluster snapshots",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			options := &SnapshotOptions{}
+			envOptions, err := parseOptionsFromEnv()
+			if err != nil {
+				klog.Warningf("Error parsing environment variables: %v", err)
+			} else {
+				options.Snapshot = *envOptions
+			}
+
+			snapshots, err := options.List(cmd.Context())
+			if err != nil {
+				return fmt.Errorf("list snapshots: %w", err)
+			}
+
+			encodedBytes, err := json.Marshal(snapshots)
+			if err != nil {
+				return fmt.Errorf("json marshal: %w", err)
+			}
+
+			if _, err := os.Stdout.Write(encodedBytes); err != nil {
+				return err
+			}
+			return nil
+		},
+	}
+
+	return cmd
+}
--- a/pkg/snapshot/oci/ghcr/ghcr.go
+++ b/pkg/snapshot/oci/ghcr/ghcr.go
@@ -1,0 +1,144 @@
+package ghcr
+
+import (
+	"context"
+	"fmt"
+	"net/http"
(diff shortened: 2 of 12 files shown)


lizardruss:
Moving back to draft, pending product decision regarding OCI registry delete support.

pkg/snapshot/oci/ghcr/ghcr.go
```diff
@@ -0,0 +1,144 @@
...
+		_, err = pkgs.PackageDeleteVersion(ctx, orgOrUser, containerPackageType, pkg, version.GetID())
+		return err
+	}
+	return fmt.Errorf("image not found: %s", ref.String())
+}
+
+func (i *imageDeleter) getPackageVersions(ctx context.Context, orgOrUser, pkg string) (packageClient, []*github.PackageVersion, error) {
+	var all []*github.PackageVersion
+
+	options := &github.PackageListOptions{
```
zerbitx, line 67: Thoughts on making a "type" of the `PackageGetAllVersions` funcs and looping to reduce repetions?  Something like

```go
	type getter func(ctx context.Context, org, packageType, packageName string, opts *github.PackageListOptions) ([]*github.PackageVersion, *github.Response, error)
	
	for which, pGetter := range map[string]getter{"org": i.client.Organizations.PackageGetAllVersions, "user": i.client.Users.PackageGetAllVersions} {
                //                        👇
		versions, resp, err := pGetter(ctx, orgOrUser, containerPackageType, pkg, options)
		if resp != nil && resp.StatusCode == http.StatusNotFound {
			break
		} else if err != nil {
			return nil, nil, err
		} else if resp == nil {
			return nil, nil, fmt.Errorf("no response listing package %s versions", pkg)
		}
		
		all = append(all, versions...)
		if resp.NextPage == 0 {
			break
		}
		
		options.Page = resp.NextPage
		if len(all) > 0 {
			switch which {
			case "org":
				return i.client.Organizations, all, nil
			case "user":
				return i.client.Users, all, nil
			}
		}
	}
```

cmd/vcluster/cmd/snapshot/list.go
```diff
@@ -0,0 +1,44 @@
...
+func NewListCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "list vCluster snapshots",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			options := &Options{}
+			envOptions, err := parseOptionsFromEnv()
+			if err != nil {
+				klog.Warningf("Error parsing environment variables: %v", err)
```
jjaferson, line 21: nit and question: should we use low case for all logging messages? 

Also do we just log the error when parsing env variables or should we return the error?

pkg/snapshot/s3/store.go
```diff
@@ -299,6 +299,45 @@ func (o *ObjectStore) GetObject(ctx context.Context) (io.ReadCloser, error) {
...
+				continue
+			}
+
+			if !strings.HasSuffix(*obj.Key, "tar.gz") {
+				continue
+			}
+
+			snapshots = append(snapshots, types.Snapshot{
+				ID:        *obj.Key,
+				URL:       "s3://" + o.bucket,
```
jjaferson, line 325: Should we add the key to the URL as well? I think vcluster validates that when parsing the URL https://github.com/loft-sh/vcluster/blob/main/pkg/snapshot/snapshot.go#L133

lizardruss:
Closing this PR in favor of https://github.com/loft-sh/vcluster/pull/3021 (no OCI delete support)

cmd/vcluster/cmd/snapshot/list.go
lizardruss: Mostly taken from the [existing snapshot command](https://github.com/loft-sh/vcluster/blob/main/cmd/vcluster/cmd/snapshot.go#L39). I think we only do lowercase for error messages, and not log messages.

pkg/snapshot/oci/ghcr/ghcr.go
lizardruss: This part is getting removed because supporting OCI delete has complications (behavior is dependent on each registry provider), so removing it for now until it can be discussed.

pkg/snapshot/s3/store.go
lizardruss: It's somewhat unclear what's expected via the UI, since the Figma just says "S3 Bucket". My thought was that it would be easier to group them without parsing the bucket URL.

For `container://` and `oci://` implementations, the list config contains a directory / repository without a tag, and the `URL` in those matches that. I'll need to update the validation so it doesn't fail in those cases.

Either way I would not expect list output to be passed back through to be validated. It should be easy enough to construct the full URL for delete by joining the URL + ID.

PR #3020, opened 2025-07-24
fix: abort cert commands in case the virtual cluster is non-running

johannesfrey:
/kind enhancement

resolves #

Fixed an issue where vcluster cert rotation was performed even though the virtual cluster was non-running.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -115,7 +115,7 @@ sync:
       enabled: false
       # MappingsOnly defines if creation of namespaces not matched by mappings should be allowed.
       mappingsOnly: false
-
+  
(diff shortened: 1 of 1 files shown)


FabianKramm:
Just FYI the Lint issue is because line endings seem to have changed with this PR: https://github.com/loft-sh/vcluster/pull/3002 . We somehow need to revert that.

johannesfrey:
> Just FYI the Lint issue is because line endings seem to have changed with this PR: #3002 . We somehow need to revert that.

Ah. Thx for the info. Would a `go run ./hack/schema/main.go` fix this?

PR #3021, opened 2025-07-24
Snapshot List and Delete

lizardruss:
/kind feature

resolves ENG-7418
resolves ENG-7434

Add commands to list and delete vcluster snapshots

Deleting OCI snapshots will not be supported with this PR, but continuing without to unblock work on the platform side. Typically deleting an image tag does not clean up the associated layers, so further discussion will be needed to determine how to proceed.


pkg/snapshot/s3/store.go
```diff
@@ -299,7 +296,69 @@ func (o *ObjectStore) GetObject(ctx context.Context) (io.ReadCloser, error) {
...
+	return err
+}
+
 // this is required because os pipes cause trouble with aws uploader
 type wrapper struct {
 	io.Reader
 }
+
+func toS3URL(bucket, key, region string) string {
+	url := "s3://" + bucket + "/" + key
```
jjaferson, line 364: We need to decide how we will be handling the `key` in the case a user wants to set a path in the key we will be including the `/` in the name

jjaferson (dismissed):

pkg/snapshot/s3/store.go
lizardruss: I may not follow what you mean, but in my testing I found that the API did not like keys with a leading `/`. In the case of `vcluster snapshot list`, an empty string key (`""`) is valid, and means the root of the bucket. We could accept it as `"/"` and parse it out, but felt that might be unexpected. In other cases, listing a nested path worked as long as it did not start with a `/`. For example, `snapshots/vcluster-1` works fine, but `/snapshots/vcluster-1` "works" but has no results.

zerbitx (commented):
lgtm

FabianKramm (approved):

PR #3023, opened 2025-07-25
Added few e2e tests for vCluster cert rotation

akhilaoommen:
/kind test

resolves #
ENG-7793

Added few e2e tests for some P0 use-cases in vCluster Cert Rotate.


test/e2e_certs/certs/rotate.go
```diff
@@ -326,6 +327,134 @@ var _ = ginkgo.Describe("vCluster cert rotation kube config tests", ginkgo.Order
...
+						"container %d in pod %s should be running", i, pod.Name)
+					gomega.Expect(container.Ready).To(gomega.BeTrue(),
+						"container %d in pod %s should be ready", i, pod.Name)
+				}
+			}
+
+			return nil
+		}).WithPolling(time.Second).
+			WithTimeout(framework.PollTimeoutLong).
+			Should(gomega.Succeed())
```
johannesfrey: You have to pass in the `g gomega.Gomega` object here as described [here](https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode). Otherwise, it won't intercept any failures and the test will just fail directly (also for all occurrences here as well).

```suggestion
		gomega.Eventually(func(g gomega.Gomega) error {
			pods, err := f.HostClient.CoreV1().Pods(f.VClusterNamespace).List(f.Context, metav1.ListOptions{
				LabelSelector: "app=vcluster,release=" + f.VClusterName,
			})
			g.Expect(err).NotTo(gomega.HaveOccurred())
			g.Expect(pods.Items).NotTo(gomega.BeEmpty())

			for _, pod := range pods.Items {
				g.Expect(pod.Status.ContainerStatuses).NotTo(gomega.BeEmpty(),
					"pod %s should have container statuses", pod.Name)

				for i, container := range pod.Status.ContainerStatuses {
					g.Expect(container.State.Running).NotTo(gomega.BeNil(),
						"container %d in pod %s should be running", i, pod.Name)
					g.Expect(container.Ready).To(gomega.BeTrue(),
						"container %d in pod %s should be ready", i, pod.Name)
				}
			}

			return nil
		}).WithPolling(time.Second).
			WithTimeout(framework.PollTimeoutLong).
			Should(gomega.Succeed())
```

test/e2e_certs/certs/rotate.go
```diff
@@ -326,6 +327,134 @@ var _ = ginkgo.Describe("vCluster cert rotation kube config tests", ginkgo.Order
...
+
+			if cert.NotAfter.Before(time.Now()) {
+				return nil
+			}
+			return fmt.Errorf("CA cert not expired yet (expires at %s)", cert.NotAfter)
+		}).
+			WithPolling(time.Second).
+			WithTimeout(framework.PollTimeoutLong).
+			Should(gomega.Succeed())
+	})
```
johannesfrey: Same as my comment below regarding the need to pass in `g gomega.Gomega`. By that you can also do assertions vs returning nil or err:

```suggestion
		gomega.Eventually(func(g gomega.Gomega) error {
			secret, err := f.HostClient.CoreV1().Secrets(f.VClusterNamespace).Get(
				f.Context, certs.CertSecretName(f.VClusterName), metav1.GetOptions{})
			g.Expect(err).NotTo(gomega.HaveOccurred())

			certPEM := secret.Data["ca.crt"]
			block, _ := pem.Decode(certPEM)
			g.Expect(block).NotTo(gomega.BeNil())
			// ...
		}).
			WithPolling(time.Second).
			WithTimeout(framework.PollTimeoutLong).
			Should(gomega.Succeed())
	})
```

test/e2e_certs/certs/rotate.go
```diff
@@ -326,6 +327,134 @@ var _ = ginkgo.Describe("vCluster cert rotation kube config tests", ginkgo.Order
...
+
+	ginkgo.It("should obtain the current cert secret of vCluster", func() {
+		_, err := f.HostClient.CoreV1().Secrets(f.VClusterNamespace).Get(f.Context, certs.CertSecretName(f.VClusterName), metav1.GetOptions{})
+		framework.ExpectNoError(err)
+	})
+
+	ginkgo.It("setting validity of ca cert of vCluster to 30 seconds", func() {
+		os.Setenv("DEVELOPMENT", "true")
+		os.Setenv("VCLUSTER_CERTS_VALIDITYPERIOD", "30s")
```
johannesfrey: Can we lower this value? If I understand correctly the test below would take 30 seconds to complete?

johannesfrey (changes requested):

test/e2e_certs/certs/rotate.go
akhilaoommen: Understood, I have changed the code as you suggested.

test/e2e_certs/certs/rotate.go
akhilaoommen: I will try with 10 secs.

akhilaoommen: Well, with 10 sec is too less as I am checking the ready state of vCluster after cert rotate, hence set expiration time as 15secs. Now it takes 10 secs to check the expired status. Results [here](https://github.com/loft-sh/vcluster/actions/runs/16624369046/job/47036996564?pr=3023#step:12:326).

johannesfrey: Does you answer still apply. I'm somehow not able to see where the ready check is after the rotation.

akhilaoommen: Sorry I removed that, this comment is not valid, anyway, I set it as 15 secs, otherwise tests were flaky.

johannesfrey: Ah, I see. I think the test might be flaky when setting the expiry time really low because the expiry check operates on the secret which will directly be expired even before the virtual cluster is running again after the rotation. And the `vcluster certs check` command requires the virtual cluster to be in Running state (and it might still be in an `Init` state of the init containers).
So the test might be more robust to check the virtual cluster Pod status to be `Running` (no container checks necessary, as those will not be ready because of the expired certs) before doing the `vcluster certs check` command. With this change I think we could even drop the validity period to like 1 second. WDYT?

test/e2e_certs/certs/rotate.go
```diff
@@ -198,6 +199,160 @@ var _ = ginkgo.Describe("vCluster cert rotation tests", ginkgo.Ordered, func() {
...
+
+		certsCmd := certscmd.NewCertsCmd(&flags.GlobalFlags{Namespace: f.VClusterNamespace})
+		certsCmd.SetArgs([]string{"rotate-ca", f.VClusterName})
+
+		err := certsCmd.Execute()
+		framework.ExpectNoError(err)
+	})
+
+	ginkgo.It("should wait until the vCluster is ready again", func() {
+		framework.ExpectNoError(f.WaitForVClusterReady())
```
johannesfrey: I think you have to remove this check here because `f.WaitForVClusterReady()` takes the StatefulSet into account and waits until all replicas are available, which won't be the case because the syncer container status won't be running.

akhilaoommen: Understood, I also removed printing expired status using `certs check` command. Now tests are running without any error.

test/e2e_certs/certs/rotate.go
```diff
@@ -198,6 +199,152 @@ var _ = ginkgo.Describe("vCluster cert rotation tests", ginkgo.Ordered, func() {
...
+			pods, err := f.HostClient.CoreV1().Pods(f.VClusterNamespace).List(f.Context, metav1.ListOptions{
+				LabelSelector: "app=vcluster,release=" + f.VClusterName,
+			})
+			g.Expect(err).NotTo(gomega.HaveOccurred())
+			g.Expect(pods.Items).NotTo(gomega.BeEmpty())
+
+			for _, pod := range pods.Items {
+				if pod.Status.Phase != corev1.PodRunning {
+					return fmt.Errorf("pod %s is not in Running phase, current phase: %s", pod.Name, pod.Status.Phase)
+				}
```
johannesfrey: Small last nit because of consistency, consider as non-blocking. That would not work?!

```suggestion
				g.Expect(pod.Status.Phase).To(gomega.Equal(corev1.PodRunning))
```

test/e2e_certs/certs/rotate.go
```diff
@@ -198,6 +199,152 @@ var _ = ginkgo.Describe("vCluster cert rotation tests", ginkgo.Ordered, func() {
...
+	var (
+		f *framework.Framework
+	)
+
+	ginkgo.JustBeforeEach(func() {
+		f = framework.DefaultFramework
+	})
+
+	ginkgo.It("should obtain the current cert secret of vCluster", func() {
+		_, err := f.HostClient.CoreV1().Secrets(f.VClusterNamespace).Get(f.Context, certs.CertSecretName(f.VClusterName), metav1.GetOptions{})
```
adriankabala: nit: Line 212 and 217 is basically the same

akhilaoommen: I removed secrets check and now test checks the current validity, thank you.

johannesfrey (approved):

PR #3024, opened 2025-07-25
fix: update pod hostIP syncing to better handle missing node service

lizardruss:
/kind bugfix

resolves ENG-8084

Fixed an issue where vcluster creates an invalid patch when syncing pod hostIPs

This is a timing issue between the pod & node service creation. The node / node service is created lazily when the first pod on that node is created, but the pod syncer did not handle the node service not found error, created an invalid patch, and the error resulted in an event getting recorded.

### DIFF
--- a/pkg/controllers/resources/pods/syncer.go
+++ b/pkg/controllers/resources/pods/syncer.go
@@ -254,13 +254,16 @@ func (s *podSyncer) SyncToHost(ctx *synccontext.SyncContext, event *synccontext.
 		}
 
 		if s.fakeKubeletIPs {
-			nodeIP, err := s.getNodeIP(ctx, pPod.Spec.NodeName)
+			nodeService, err := s.ensureNodeService(ctx, pPod)
 			if err != nil {
+				if kerrors.IsNotFound(err) {
+					return ctrl.Result{RequeueAfter: time.Second}, nil
+				}
 				return ctrl.Result{}, err
 			}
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/resources/pods/syncer.go
```diff
@@ -536,30 +547,15 @@ func (s *podSyncer) assignNodeToPod(ctx *synccontext.SyncContext, pObj *corev1.P
...
-func (s *podSyncer) getNodeIP(ctx *synccontext.SyncContext, name string) (string, error) {
-	serviceName := translate.SafeConcatName(translate.VClusterName, "node", strings.ReplaceAll(name, ".", "-"))
+func (s *podSyncer) ensureNodeService(ctx *synccontext.SyncContext, pPod *corev1.Pod) (*corev1.Service, error) {
+	serviceName := translate.SafeConcatName(translate.VClusterName, "node", strings.ReplaceAll(pPod.Spec.NodeName, ".", "-"))
 
 	nodeService := &corev1.Service{}
 	err := ctx.CurrentNamespaceClient.Get(ctx.Context, types.NamespacedName{Name: serviceName, Namespace: ctx.CurrentNamespace}, nodeService)
-	if err != nil && !kerrors.IsNotFound(err) {
-		return "", fmt.Errorf("list services: %w", err)
+	if err != nil {
```
FabianKramm, line 555: Does this work when `networking.advanced.proxyKubelets.byIP: false`?

FabianKramm (approved):

PR #3028, opened 2025-07-28
chore(ci): decouple helm release action

sydorovdmytro:
Use a standalone workflow to publish the Helm chart. This makes it possible to be triggered by another workflow or from external repositories.

/kind enhancement

related to OPS-206

### DIFF
--- a/.github/workflows/publish-chart.yaml
+++ b/.github/workflows/publish-chart.yaml
@@ -1,0 +1,44 @@
+name: Publish Chart
+
+on:
+  workflow_call:
+    inputs:
+      release_version:
+        description: "The release version to publish"
+        required: true
+        type: string
+      ref:
+        description: "The git ref to checkout"
+        required: true
+        type: string
+
+jobs:
+  publish-chart:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Validate semantic version
+        id: semver
+        uses: loft-sh/github-actions/.github/actions/semver-validation@semver-validation/v1
+        with:
+          version: "${{ inputs.release_version }}"
+      - name: Check validation result
+        run: '[[ "${{ steps.semver.outputs.is_valid }}" == "true" ]] || (echo "Invalid version: ${{ inputs.release_version }}" && exit 1)'
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref }}
+      - uses: azure/setup-helm@v4
+        with:
+          version: "v3.0.2"
+      - name: Publish Helm chart
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-chart.yaml
```diff
@@ -0,0 +1,34 @@
...
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref }}
+      - uses: azure/setup-helm@v4
+        with:
+          version: "v3.0.2"
+      - run: |
+          helm plugin install https://github.com/chartmuseum/helm-push.git
```
Piotr1215: This will fail if runner already has it. It's not very likely that they will add it to the image, but it's a quick check.

```bash
➜ helm plugin install https://github.com/chartmuseum/helm-push.git
Error: plugin already exists
loft-enterprise [ main][$]
✗ echo $?
1
```

.github/workflows/publish-chart.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref }}
+      - uses: azure/setup-helm@v4
+        with:
+          version: "v3.0.2"
+      - run: |
```
Piotr1215: It's always a good idea to add some bash settings in the script block
eg `set -euo pipefail`

.github/workflows/publish-chart.yaml
```diff
@@ -0,0 +1,34 @@
+name: Publish Chart
+
+on:
+  workflow_call:
+    inputs:
+      release_version:
```
Piotr1215, line 6: Does the version have to follow semver? If so adding basic regex validation would be good.

.github/workflows/publish-chart.yaml
Piotr1215, line 1: Add some logging statements with variables for better troubleshooting.

.github/workflows/publish-chart.yaml
sydorovdmytro: added via `loft-sh/github-actions/.github/actions/semver-validation@semver-validation/v1`

.github/workflows/publish-chart.yaml
sydorovdmytro: Done

.github/workflows/publish-chart.yaml
sydorovdmytro: Fixed

.github/workflows/publish-chart.yaml
sydorovdmytro: Done

cbron (approved):

PR #3035, opened 2025-07-28
[v0.24] Snapshot List and Delete (#3021)

lizardruss:
Backport from `main` to `v0.24`

Original PR Nr.: #3021

### Backported Commits:
- 1e45f69f Snapshot List and Delete (#3021)

* feat: implement vcluster snapshot list command

* feat: implement vcluster snapshot delete command

* fix: update snapshot validation for list command

* fix: modify commands to error if the configuration cannot be parsed from the environment

* fix: use full URL in list results

* fix: only list objects in the &quot;current directory&quot; of the bucket

The default s3 behavior is to list everything under a prefix. If a bucket is used for multiple purposes, this could include non-snapshot resources.

* fix: handle list input with file key at the bucket root

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -225,7 +225,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        distribution: ["k3s", "k8s", "k0s"]
+        # Skip k0s upgrades since we dropped support in later CLI versions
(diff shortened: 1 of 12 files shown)


pascalbreuninger:
We can cancel this one as 0.24 is at the end of it's support lifecycle @lizardruss

PR #3039, opened 2025-07-29
chore(ci): decouple notify release action

sydorovdmytro:
Use a standalone workflow to send notifications about the new release. This makes it possible to be triggered by another workflow or from external repositories.

/kind enhancement

part of OPS-206


.github/workflows/notify-release.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+jobs:
+  notify_release:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ inputs.ref }}
+      - id: get_base_branch
```
Piotr1215, line 27: This should probably be moved to reusable actions.

Piotr1215 (approved):
Approving with a non-blocking comment.

.github/workflows/notify-release.yaml
sydorovdmytro: Will do as a future improvement

PR #3040, opened 2025-07-29
Fail early with advisory message if node deletion will fail due to th…

zerbitx:
…e node being in an unknown status

/kind enhancement

resolves #ENG-8061

### DIFF
--- a/cmd/vclusterctl/cmd/node/delete.go
+++ b/cmd/vclusterctl/cmd/node/delete.go
@@ -9,8 +9,10 @@ import (
 	"github.com/loft-sh/log"
 	"github.com/loft-sh/vcluster/pkg/cli/flags"
 	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


cmd/vclusterctl/cmd/node/delete.go
```diff
@@ -108,3 +119,20 @@ func (o *DeleteOptions) Run(ctx context.Context, args []string) error {
 	o.Log.Infof("Successfully deleted node %s", nodeName)
 	return nil
 }
+
+func (o *DeleteOptions) pollNodeStatus(ctx context.Context, kubeClient *kubernetes.Clientset, nodeName string, errc chan error) {
+	for range time.Tick(time.Second * 5) {
+		node, err := kubeClient.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+		if err != nil {
```
johannesfrey, line 128: Just for my understanding. Are we not returning on error here because this would lead to aborting the drain even before `RunNodeDrain` does have a chance to run?

zerbitx: Correct.  I'll add a comment to explain that.

johannesfrey (approved):
lgtm

PR #3047, opened 2025-07-30
fix(ci): better change detection for E2E

sydorovdmytro:
Fixes an issue when PR status checks did not report the status. This is a common case when changed files are not in the list and workflow didn't trigger.

/kind enhancement

related-to OPS-97


Piotr1215 (commented):
Looks good

.github/workflows/e2e.yaml
```diff
@@ -32,9 +21,24 @@ env:
...
+        - "test/**" # include test files in e2e again
+        - "!**.md"
+        - "Dockerfile.release"
+        - ".github/workflows/e2e.yaml"
+        - "chart/**"
+        - "manifests/**"
   build:
     runs-on: ubuntu-latest
-    if: github.repository_owner == 'loft-sh' # do not run on forks
+    if: github.repository_owner == 'loft-sh' && needs.detect_changes.outputs.has_changed == 'true'
```
Piotr1215, line 40: This change will cause all jobs that depend on 'build' to be skipped when no changes are detected. Jobs like vcluster-install-delete, test-vcluster-k3s, and test-vcluster need to also check detect_changes and have appropriate conditions, otherwise they'll be skipped defeating the purpose of always reporting status.

sydorovdmytro: Done

Piotr1215 (approved):

PR #3048, opened 2025-07-30
Fix platform login UX

ApsTomar:
/kind enhancement

resolves ENG-4877

Fixed an issue where vcluster platform login opened a browser every time to login. Now, if access-key is stored in platform config, it will use that to login. Only if the access-key is not provided in the command line args and also not stored in the platform config, then it will open the browser to login.

### DIFF
--- a/cmd/vclusterctl/cmd/platform/login.go
+++ b/cmd/vclusterctl/cmd/platform/login.go
@@ -124,6 +124,9 @@ func (cmd *LoginCmd) Run(ctx context.Context, args []string) error {
 	var err error
 	if cmd.AccessKey != "" {
 		err = loginClient.LoginWithAccessKey(url, cmd.AccessKey, cmd.Insecure)
+	} else if cfg.Platform.AccessKey != "" {
+		cmd.AccessKey = cfg.Platform.AccessKey
+		err = loginClient.LoginWithAccessKey(url, cmd.AccessKey, cmd.Insecure)
 	} else {
 		err = loginClient.Login(url, cmd.Insecure, cmd.Log)
 	}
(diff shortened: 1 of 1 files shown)


cmd/vclusterctl/cmd/platform/login.go
```diff
@@ -124,6 +124,9 @@ func (cmd *LoginCmd) Run(ctx context.Context, args []string) error {
 	var err error
 	if cmd.AccessKey != "" {
 		err = loginClient.LoginWithAccessKey(url, cmd.AccessKey, cmd.Insecure)
+	} else if cfg.Platform.AccessKey != "" {
+		cmd.AccessKey = cfg.Platform.AccessKey
```
lizardruss: Is this required, or could you pass `cmd.Platform.AccessKey` instead?

Mostly wondering because there's already [this note](https://github.com/loft-sh/vcluster/pull/3048/files#diff-71ff6622ab3d1a1610edf7d2289b50195624ee90bc948853fc8f33a7880efb40R161-R164), where we'd like to avoid this type value copying, since it can make it harder to debug.

lizardruss (commented):
Added a question, but otherwise looks like an easy fix.

cmd/vclusterctl/cmd/platform/login.go
ApsTomar: yes, I earlier thought that cmd.AccessKey would also be getting used later in the loginWithAccessKey method but it was not used in that method. So, I've removed the assignment and directly used `cfg.Platform.AccessKey`

lizardruss (approved):

PR #3056, opened 2025-08-01
fix: Differenciate between new segments and map keys

ulyssessouza:
This preserves the notation of a map after the parsing to be able to distinguish when accessing.

Ex: `mymap["mykey"]` parses to:
Before: `["mymap", "mykey"]`
After:  `["mymap", "[mykey]"]`

The goal is to make `Translate` be able to add a new value for a key that doesn't exist.

/kind bugfix

resolves #ENG-8116


pkg/util/patch/patch.go
```diff
@@ -56,20 +56,20 @@ func (p Patch) Clear() {
 	}
 }
 
-func (p Patch) MustTranslate(path string, translate func(path string, val interface{}) (interface{}, error)) {
+func (p Patch) MustTranslate(path string, translate func(path string, val interface{}, exists bool) (interface{}, error)) {
```
lizardruss: nit: could we define a type for the `translate` func to avoid copying the signature in multiple places?

lizardruss (approved):
one suggestion, but optional. Otherwise looks good!

pkg/util/patch/patch.go
ulyssessouza: Sure! I will do that!

lizardruss (approved):

PR #3065, opened 2025-08-07
feat(helm chart): update livenessProbe and startProbe for etcd Statef…

mayooot:
/kind feature

Update startupProbe to aviod extra checks and make pod enter running state faster.

Update livenessProbe to improve accuracy and ensure service availability when appending pod IP to Services.Endpoints.

References:
- https://github.com/kubernetes/kubeadm/issues/2567
- https://github.com/kubernetes/kubernetes/pull/110744

I have already tested it in my environment.

### DIFF
--- a/chart/templates/etcd-statefulset.yaml
+++ b/chart/templates/etcd-statefulset.yaml
@@ -187,7 +187,7 @@ spec:
 {{ toYaml $externalEtcd.resources | indent 10 }}
         livenessProbe:
           httpGet:
-            path: /health
+            path: /health?exclude=NOSPACE&serializable=true
             port: 2381
             scheme: HTTP
           initialDelaySeconds: 10
@@ -197,7 +197,7 @@ spec:
           failureThreshold: 8
         startupProbe:
           httpGet:
-            path: /health
+            path: /health?serializable=false
             port: 2381
             scheme: HTTP
           initialDelaySeconds: 10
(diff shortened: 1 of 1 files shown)


zerbitx:
Thanks for this submission and the reference links.  After checking those out I found this later issue: https://github.com/kubernetes/kubeadm/issues/3039, where it seems we can accomplish the same thing with `/livez` and `/readyz` instead.  Could you try that out? It looks like [here](https://github.com/kubernetes/kubernetes/commit/eeac2dda7091c1f1ea100e7637291bba2345bb0b) is where they made that change, if that's of use.

mayooot:
@zerbitx Yep, that sounds great. I’ll work on this soon.
BTW, do you think we should provide a function that can dynamically check the etcd version and update the chart template for the etcd section? It sounds like it might require more work.

zerbitx:
> @zerbitx Yep, that sounds great. I’ll work on this soon. BTW, do you think we should provide a function that can dynamically check the etcd version and update the chart template for the etcd section? It sounds like it might require more work.

If you mean to check the version to ensure it has those new endpoints, I think this should be good as is.  Those were added back in [2023](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.5.md#v3511-2023-12-07), and future versions of vCluster can look more to the present and future.

mayooot:
@zerbitx Obviously, the etcd version is >= `v3.5.11` in the latest vcluster chart.
https://github.com/loft-sh/vcluster/blob/9dba31d6f5328c571cd29a6682934f60a8c0b46b/chart/values.yaml#L380

So, I just update the probe endpoints, which I believe makes it easier to review and merge.

PR #3074, opened 2025-08-11
Add CSI volume snapshots support

nprokopic:
/kind feature

resolves #

Add support for creating CSI volume snapshots and restoring volumes from them.


nprokopic:
The PR size here can be reduced by upgrading `github.com/kubernetes-csi/external-snapshotter/client/v4 v4.2.0` to `github.com/kubernetes-csi/external-snapshotter/client/v8 v8.2.0` in main 🤔

PR #3075, opened 2025-08-11
fix: use a recursive copy instead of archive

zerbitx:
/kind bugfix

resolves #ENG-8214

I looked into making this conditional, like only changing the flag if they have a security context set since it might mess with permissions, but even without it, it would fail silently for the IronBank case, while `-r` worked consistently for me.

### DIFF
--- a/chart/templates/_init-containers.tpl
+++ b/chart/templates/_init-containers.tpl
@@ -34,7 +34,7 @@
   command:
     - cp
   args:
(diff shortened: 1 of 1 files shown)


jjaferson (approved):
Looks good

lizardruss (approved):

johannesfrey (approved):

PR #3100, opened 2025-08-18
feat: add cloud controller manager config & remove standalone syncConfig, bundle & bundleRepository

FabianKramm:


pkg/k8s/k8s.go
```diff
@@ -156,6 +156,9 @@ func StartK8S(ctx context.Context, serviceCIDR string, vConfig *config.VirtualCl
 					args = append(args, "--controllers=*,bootstrapsigner,tokencleaner")
 					args = append(args, "--allocate-node-cidrs=true")
 					args = append(args, "--cluster-cidr="+vConfig.Networking.PodCIDR)
+					// we set cloud provider to external as we either want to use an external cloud controller manager
+					// such as AWS or GCP or we fallback to our in-built cloud controller manager.
+					args = append(args, "--cloud-provider=external")
```
pascalbreuninger, line 161: Where do we configure the built in cloud controller manager based on the vcluster.yaml config?

FabianKramm: Will be in vCluster pro

PR #3108, opened 2025-08-20
Update README.md

pescerosso:
Right now the Getting Started guide shows brew install ... and then immediately vcluster create .... For someone brand new, they might assume vcluster is self-contained and will “just run” without realizing it requires a Kubernetes cluster underneath. Adding a prerequisite to prevent confusion. Otherwise, first-time users who run the command on a laptop without a cluster will just get connection errors from the CLI.

/kind bugfix
/kind enhancement
/kind feature
/kind documentation
/kind test

resolves #

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,19 @@
 
 Deploy your first virtual cluster with minimal effort:
 
+#### Step 1: Install vCluster CLI
 ```bash
 brew install loft-sh/tap/vcluster
+```
+#### Prerequisite: Set Up a Kubernetes Cluster
+Before creating a virtual cluster, ensure you have access to a running Kubernetes cluster. Examples of lightweight clusters you can use:
+- [kind](https://kind.sigs.k8s.io/)
+- [minikube](https://minikube.sigs.k8s.io/)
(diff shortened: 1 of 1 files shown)


saiyam1814:
@pescerosso if this is a prerequisite step then can we have it before step 1?

pescerosso:
@saiyam1814 I was thinking about that too, but technically this cmd "brew install loft-sh/tap/vcluster" doesn't require a K8s cluster running. Only the cmd does.

README.md
```diff
@@ -28,10 +28,26 @@
 
 Deploy your first virtual cluster with minimal effort:
 
+#### Step 1: Install vCluster CLI
 ```bash
 brew install loft-sh/tap/vcluster
+```
+#### Prerequisite: Set Up a Kubernetes Cluster
```
cbron: Goal of this top section is to be as minimal as possible. We should remove all these external links as they are distracting IMO, and people are generally familiar with k8s solutions. I would add this line at beginning of step 2: 
> Before creating a virtual cluster, ensure you have access to a running Kubernetes cluster.

pescerosso: Ok, removed and rephrased it to make it clear and concise.

PR #3112, opened 2025-08-20
refactor: remove isolated control plane feature

FabianKramm:
Resolves ENG-8195


FabianKramm:
@lizardruss yeah I'll remove those as well then

PR #3130, opened 2025-08-23
fix: Check for platform project existence on CLI

ulyssessouza:
/kind bugfix

This is a follow up PR for https://github.com/loft-sh/loft-enterprise/pull/4654 which is the platform checking of the same logic as this one

resolves #ENG-7334

### DIFF
--- a/pkg/config/validation.go
+++ b/pkg/config/validation.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"context"
 	"crypto/x509"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


pkg/config/validation.go
```diff
@@ -268,6 +273,26 @@ func validatePatches(patchesValidation ...patchesValidation) error {
 	return nil
 }
 
+func ValidatePlatformProject(ctx context.Context, config *config.Config, loadedConfig *cliconfig.CLI) error {
+	platformConfig, err := config.GetPlatformConfig()
+	if err != nil {
+		return fmt.Errorf("get platform config: %w", err)
```
FabianKramm, line 279: This will render vCluster CLI OSS not usable without a platform anymore

ulyssessouza: I'm not sure if I understood it because I tested locally without a running platform.

Errors returned from `config.GetPlatformConfig()` are only errors related to yaml parsing.
https://github.com/loft-sh/vcluster/blob/main/config/config.go#L849-L857

FabianKramm: Ah yeah true, my bad

PR #3147, opened 2025-08-27
fix: patch virtual instead of physical and always add host ip annotations to physical

johannesfrey:
/kind bugfix

resolves ENG-8629

When `proxyKubelets.byIP` is set to true (which is the default), attempting to patch the status of the physical pod there is an edge case that when running kubelet in version < 1.28 with an apiserver >= 1.28 that it only patches `hostIP` and not `hostIPs[0]`, which will then fail at the validation with following error:
```
is invalid: status.hostIPs[0].ip: Invalid value: \"1.2.3.4\": must be equal to `hostIP`
```

As the hostIPs are only needed in the virtual context, those have been moved there. Furthermore, they physical pod will have the hostIP in its annotations regardless of virtual scheduling being enabled or not.

### DIFF
--- a/pkg/controllers/resources/pods/syncer.go
+++ b/pkg/controllers/resources/pods/syncer.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"reflect"
 	"slices"
-	"strings"
 	"time"
 
 	nodev1 "k8s.io/api/node/v1"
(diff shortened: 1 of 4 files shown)


johannesfrey:
~~Should `enableScheduler` also be removed from here: https://github.com/loft-sh/vcluster/blob/main/pkg/controllers/resources/pods/translate/translator.go#L610?!~~

FabianKramm (approved):

FabianKramm (approved):

PR #3156, opened 2025-08-29
fix: increase ephemeral-storage requests

lizardruss:
/kind bugfix

resolves ENG-8713

Fixed an issue where vcluster could be evicted due to not enough ephemeral-storage requests.

Especially important on GKE Autopilot clusters where limits are set equal to requests


FabianKramm:
@lizardruss to prevent this from happening like every other month, can we make this 1Gi?

FabianKramm (approved):

PR #3164, opened 2025-09-02
Add `vcluster snapshot create` command

nprokopic:
/kind enhancement

resolves ENG-8530 and ENG-8531

Add `vcluster snapshot create` command that creates snapshots asynchronously.

`vcluster snapshot create` creates a snapshot request, which is then reconciled by a vCluster snapshot controller that creates snapshots in the background.

- Snapshot request consists of two resources:
  - ConfigMap: vCluster snapshot controller reconciles the ConfigMaps with a special `vcluster.loft.sh/snapshot-request` label (ATM the ConfigMap is used mostly 
  - Secret: Used to temporarily save snapshot options (including credentials), so that vCluster snapshot controller can access the snapshot location. The snapshot request Secret is deleted as soon as the snapshot request has been processed (either completed successfully or failed)
- `vcluster snapshot create` command sends a POST request to the vCluster `/vcluster/snapshots` route
- `/vcluster/snapshots` route handler creates snapshot request resources
- for vCluster with shared nodes, snapshot request resources are created in the host cluster
- for vCluster with private nodes, snapshot request resources are created in the virtual cluster

Old `vcluster snapshot` command has been deprecated and it continues to work like before.

e2e snapshot test have been updated to test both the old CLI-based `vcluster snapshot` command and the new controller-based `vcluster snapshot create` command.
- The latest platform release uses the old `vcluster snapshot` command, so it's important to keep running the tests for the old command, until the platform is updated to use the new `vcluster snapshot create` command.
- Tests specs are identical for both CLI-based and controller-based snapshot command.
- The test for `vcluster snapshot create` creates the snapshot in the container on the path `container:///snapshot-data/snapshot.tar.gz`. The snapshot container path is from a PVC (created in the CI) that is mounted to vcluster container, so that restore pod can use it when restoring. The `/data/...` container path could not have been be used, because some e2e tests (e.g. HA setup) use `emptyDir` for `/data`, which is not accessible by a separate restore pod.


pkg/util/clihelper/virtual.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/loft-sh/vcluster/pkg/util/portforward"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func GetVClusterKubeConfig(ctx context.Context, kubeConfig *rest.Config, kubeClient *kubernetes.Clientset, vCluster *find.VCluster, log log.Logger) (*rest.Config, error) {
```
nprokopic: Moved here from `cmd/vclusterctl/cmd/debug/collect.go` so it can be reused in multiple places.

pkg/cli/snapshot_helm.go
```diff
@@ -44,8 +50,62 @@ func Snapshot(ctx context.Context, args []string, globalFlags *flags.GlobalFlags
...
+	}
+	snapshotRequestDataJSON, err := json.Marshal(snapshotRequestData)
+	if err != nil {
+		return fmt.Errorf("failed to marshal snapshot request data: %w", err)
+	}
+	httpRequest, err := http.NewRequest(http.MethodPost, requestURL, bytes.NewReader(snapshotRequestDataJSON))
+	if err != nil {
+		return fmt.Errorf("failed to create http request: %w", err)
+	}
+	response, err := virtualHTTPClient.Do(httpRequest)
```
hidalgopl: This will fail for any virtual cluster below v0.29, correct?
Just wondering if there's a way to provide a better UX here (so either checking a virtual cluster version, and for `<v0.29` avoid all these operations and just output "async snapshots not supported for this vCluster version, please upgrade" or if we don't have a way to know vCluster version here, have a special handling for an error returned by pre-v0.29 (I guess the `/vcluster/snapshots` will return 404 in this case, right?)

pkg/snapshot/controller.go
```diff
@@ -0,0 +1,473 @@
...
+		if objLabels == nil {
+			return false
+		}
+		_, ok := objLabels[RequestLabel]
+		return ok
+	})
+
+	return ctrl.NewControllerManagedBy(c.manager).
+		WithOptions(controller.Options{
+			CacheSyncTimeout: constants.DefaultCacheSyncTimeout,
```
hidalgopl: Shall we set `MaxConcurrentReconciles: 1` here?

pkg/server/routes/snapshots.go
```diff
@@ -0,0 +1,113 @@
...
+		if vConfig.Config.PrivateNodes.Enabled {
+			kubeClient = uncachedVirtualClient
+			snapshotRequestNamespace = "kube-system"
+		} else {
+			kubeClient = uncachedLocalClient
+			snapshotRequestNamespace = vConfig.HostNamespace
+		}
+
+		// create snapshot request Secret
+		secret, err := snapshot.CreateSnapshotOptionsSecret(snapshotRequestNamespace, &snapshotOptions)
```
hidalgopl: Not sure if it's out of scope, but I think it would be good to have some kind of rate-limiting mechanism, either here or in the CLI. Basically to reject a request for creating a new snapshot if the last one is not done yet. WDYT?

pkg/cli/snapshot_helm.go
```diff
@@ -44,8 +50,62 @@ func Snapshot(ctx context.Context, args []string, globalFlags *flags.GlobalFlags
...
+		return fmt.Errorf("failed to read response body: %w", err)
+	}
+	snapshotRequestResultJSON := responseBody.Bytes()
+	var snapshotRequestResult snapshot.Request
+	err = json.Unmarshal(snapshotRequestResultJSON, &snapshotRequestResult)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal snapshot request result: %w", err)
+	}
+
+	fmt.Printf("Created snapshot request %s\n", snapshotRequestResult.Name)
```
hidalgopl: how do I know where to get it from, once the snapshot is done?

FabianKramm: We shouldn't use fmt.Printf in general and instead the cli logger

pkg/server/routes/snapshots.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+const (
+	PostSnapshotsURL = "/vcluster/snapshots"
+)
+
+func WithSnapshotsCreate(ctx context.Context, uncachedLocalClient, uncachedVirtualClient client.Client, vConfig *config.VirtualClusterConfig) http.Handler {
+	logger := log.GetInstance()
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		if req.Method != http.MethodPost {
```
FabianKramm: How is this request currently authenticated and authorized?

pkg/snapshot/controller.go
```diff
@@ -0,0 +1,473 @@
...
+	return &Reconciler{
+		vConfig:       registerContext.Config,
+		manager:       manager,
+		logger:        logger,
+		eventRecorder: manager.GetEventRecorderFor(controllerName),
+	}, nil
+}
+
+func (c *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, retErr error) {
+	c.logger.Infof("reconciling snapshot request ConfigMap %s", req.NamespacedName)
```
FabianKramm: Please be consistent in the log messages, above you use capital messages, so this should be "Reconciling..."

pkg/snapshot/controller.go
```diff
@@ -0,0 +1,473 @@
...
+
+	var configMap corev1.ConfigMap
+	err := c.client().Get(ctx, req.NamespacedName, &configMap)
+	if kerrors.IsNotFound(err) {
+		c.logger.Infof("snapshot request ConfigMap %s not found", req.NamespacedName)
+		return ctrl.Result{}, nil
+	} else if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get snapshot request ConfigMap %s/%s: %w", req.Namespace, req.Name, err)
+	}
+	c.logger.Infof("found ConfigMap %s/%s with vcluster snapshot request", configMap.Namespace, configMap.Name)
```
FabianKramm: Same here

FabianKramm (changes requested):

pkg/cli/snapshot_helm.go
nprokopic: Thanks, I missed to update this, will fix.

pkg/server/routes/snapshots.go
nprokopic: Route handler is added here to the existing `http.ServeMux` / `http.Handler` instance https://github.com/loft-sh/vcluster/blob/877ea2fe32b164d11bfd483d41aaea10c89008f9/pkg/server/server.go#L179

And IIUC authorization for that same HTTP handler is already configured here https://github.com/loft-sh/vcluster/blob/b3ace5e7819d8968fa4e09a1c441a40ef8c5d10a/pkg/server/server.go#L317.

More details here https://github.com/loft-sh/vcluster/blob/b3ace5e7819d8968fa4e09a1c441a40ef8c5d10a/pkg/server/server.go#L255-L290

Then on the client side, the HTTP client is created from the vCluster rest config here https://github.com/loft-sh/vcluster/blob/877ea2fe32b164d11bfd483d41aaea10c89008f9/pkg/cli/snapshot_helm.go#L58-L70 so the HTTP client gets automatically configured with required auth config

If I just use an unauthenticated HTTP client (without creating it from vCluster rest config that has required client auth config), e.g. default HTTP client, server rejects the request and returns an error like this:
```
16:31:35 fatal failed to post snapshot request: Post "https://localhost:12018/vcluster/snapshots": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

pkg/cli/snapshot_helm.go
nprokopic: > This will fail for any virtual cluster below v0.29, correct?

Yes, thanks for the reminder, forgot to add version check here!

> I guess the /vcluster/snapshots will return 404 in this case, right?

Yeah, I think it would return 404. IIRC it would route the request to k8s API server, which returns an error saying that such resource type does not exist.

pkg/cli/snapshot_helm.go
nprokopic: Same like for the existing command.

New command example would be:
```
vcluster snapshot create my-vcluster "s3://my-snapshots-bucket/my-vcluster"
```

And you restore the backup like before with:
```
vcluster restore my-vcluster "s3://my-snapshots-bucket/my-vcluster"
```

nprokopic: I will most probably add a new command to check the progress of the in-progress snapshot.

Currently you could check the in-progress snapshot by checking the snapshot request ConfigMap:
```
kubectl get configmaps -n $VCLUSTER_NAME snapshot-request-clqkf -o yaml
apiVersion: v1
data:
  snapshotRequest: '{"name":"snapshot-request-clqkf","spec":{},"status":{"phase":"Completed"}}'
kind: ConfigMap
metadata:
  creationTimestamp: "2025-09-10T14:35:39Z"
  labels:
    vcluster.loft.sh/snapshot-request: ""
  name: snapshot-request-clqkf
  namespace: nik215
  resourceVersion: "1297"
  uid: a85e2ffe-fa8d-4782-8ad8-d984ff6938e8
```

Wiring the above into something like `vcluster snapshot list` and/or `vcluster snapshot get` would be nicer UX definitely.

pkg/server/routes/snapshots.go
nprokopic: It's definitely in-scope for the project, just out-of-scope for this PR.

I would like to get the controller-based snapshot creation to work, so I can add back (on top of this) the PV snapshots implementation, and then polish it all together, making improvements in incremental steps, rather than implementing lots of details and optimizations right away here (even this PR got way bigger than I wanted 🙈).

pkg/snapshot/controller.go
nprokopic: Done ✅

pkg/cli/snapshot_helm.go
nprokopic: Updated ✅

pkg/snapshot/controller.go
nprokopic: Fixed capitalization ✅

chart/templates/role.yaml
```diff
@@ -44,8 +44,11 @@ rules:
     resources: ["statefulsets", "replicasets", "deployments"]
     verbs: ["get", "list", "watch"]
   - apiGroups: [""]
-    resources: ["endpoints", "events", "pods/log"]
+    resources: ["endpoints", "pods/log"]
     verbs: ["get", "list", "watch"]
+  - apiGroups: [ "" ]
+    resources: [ "events", ]
```
zerbitx:
```suggestion
    resources: ["events"]
```

pkg/cli/snapshot_helm.go
```diff
@@ -44,8 +52,66 @@ func Snapshot(ctx context.Context, args []string, globalFlags *flags.GlobalFlags
...
+	_, err = responseBody.ReadFrom(response.Body)
+	if err != nil {
+		return fmt.Errorf("failed to read response body: %w", err)
+	}
+	snapshotRequestResultJSON := responseBody.Bytes()
+	var snapshotRequestResult snapshot.Request
+	err = json.Unmarshal(snapshotRequestResultJSON, &snapshotRequestResult)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal snapshot request result: %w", err)
+	}
```
zerbitx: nit: You can accomplish the body reading and unmarshalling with fewer intermediate steps.  Something like 

```go
var snapshotRequestResult snapshot.Request
if err = json.NewDecoder(response.Body).Decode(&snapshotRequestResult); err != nil {
  return fmt.Errorf("failed to unmarshal snapshot request result: %w", err)
}
```

pkg/cli/snapshot_helm.go
nprokopic: Added (a lightweight) version check, which fails only if it determines that the version is old (<0.29). In case it cannot determine the version (not set for some reason), or it sees v0.0.1, which is hard-coded and may appear in local builds, it just prints a warning.

pkg/cli/snapshot_helm.go
nprokopic: Thanks, will check this out!

pkg/cli/snapshot_helm.go
```diff
@@ -44,8 +55,86 @@ func Snapshot(ctx context.Context, args []string, globalFlags *flags.GlobalFlags
 		}
 	}
 
-	// run snapshot pod
-	return pod.RunSnapshotPod(ctx, restConfig, kubeClient, []string{"/vcluster", "snapshot"}, vCluster, podOptions, snapshotOpts, log)
+	if !async {
+		// run snapshot pod
+		return pod.RunSnapshotPod(ctx, restConfig, kubeClient, []string{"/vcluster", "snapshot"}, vCluster, podOptions, snapshotOpts, log)
```
jjaferson, line 56: This will be deprecated but are we going to still support it for while?

pkg/server/routes/snapshots.go
jjaferson: yeah, this is something that will be important for scheduled snapshot too

pkg/server/routes/snapshots.go
```diff
@@ -0,0 +1,113 @@
...
+		var snapshotOptions snapshot.Options
+		err = json.Unmarshal(snapshotRequestData[snapshot.OptionsKey], &snapshotOptions)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		// create the snapshot request Secret and ConfigMap
+		// - for shared nodes, create resources in the host cluster in the vCluster namespace
+		// - for private nodes, create resources in the virtual cluster in the kube-system namespace
```
jjaferson: For private nodes can we not use the vCluster namespace?

pkg/server/routes/snapshots.go
```diff
@@ -0,0 +1,113 @@
...
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		if snapshotRequest.Name != "" {
+			// snapshot request name is specified
+			secret.Name = snapshotRequest.Name
+		} else {
+			// snapshot request name is auto-generated
+			secret.GenerateName = "snapshot-request-"
```
jjaferson: wouldn't it be a good idea to generate the snapshot request with the vcluster name too? It will be easier to identify to which vcluster the CM belongs to

pkg/snapshot/request.go
```diff
@@ -0,0 +1,142 @@
...
+	snapshotRequestJSON, err := json.Marshal(snapshotRequest)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal snapshot request: %w", err)
+	}
+
+	configMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: vClusterNamespace,
+			Labels: map[string]string{
+				RequestLabel: "",
```
jjaferson: Wouldn't it be good to have the vcluster name in the label values? in case we want to filter down snapshots request from an specific vCluster?

pkg/util/clihelper/virtual.go
```diff
@@ -0,0 +1,105 @@
...
+	log.Infof("Start port-forwarding to virtual cluster")
+	vKubeConfig, err := GetKubeConfig(ctx, kubeClient, vCluster.Name, vCluster.Namespace, log)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse kube config: %w", err)
+	}
+
+	localPort := RandomPort()
+	errorChan := make(chan error)
+	go func() {
+		errorChan <- portforward.StartPortForwardingWithRestart(ctx, kubeConfig, "127.0.0.1", podName, vCluster.Namespace, strconv.Itoa(localPort), "8443", make(chan struct{}), portForwardingOptions.StdOut, portForwardingOptions.StdErr, log)
```
jjaferson, line 70: What would happen with the port forwarding if the command is interrupted? Would this process get cancelled?

pkg/cli/snapshot_helm.go
nprokopic: Yes, this PR deprecates the `vcluster snapshot` command. We have to support it for a while, especially because the platform still uses it, and in general IMHO no rush to remove it.

Command description is also updated:
```
vcluster --help | grep snapshot
  restore     Restores a virtual cluster from snapshot
  snapshot    Snapshot a virtual cluster (deprecated, use 'vcluster snapshot create' instead)
```

I have updated the snapshot e2e tests here to run them for both old and new command.

pkg/server/routes/snapshots.go
nprokopic: For private nodes the snapshot request resources are created inside of the **virtual** cluster, so theoretically we can create them in any namespace. I would save them in some namespace like `vcluster-system`, but AFAIK we don't have that, so I went with `kube-system`.

For shared nodes OTOH, the snapshot request resources are created inside of the **host** cluster, so vCluster namespace makes the most sense.

P.S. Theoretically we could also use vCluster namespace on the host for private nodes as well, but during the design phase (multiple meets) we have decided to use the virtual cluster for private nodes, and the host cluster for shared nodes.

pkg/server/routes/snapshots.go
nprokopic: Good idea to add the vcluster name, I'll add it as a prefix.

These snapshot request resources will be always created in the vCluster namespace, so it should be easy to identify them.

P.S. The controller reconciles only the ConfigMaps from the vCluster namespace that also have `vcluster.loft.sh/snapshot-request` label, see [here](https://github.com/loft-sh/vcluster/blob/e4a96bc59111403836881b311e99834f097001ef/pkg/snapshot/controller.go#L147).

pkg/snapshot/request.go
nprokopic: You can filter them by vcluster namespace, e.g. 

```
kubectl get configmaps -n $VCLUSTER_NAME -l vcluster.loft.sh/snapshot-request
NAME                     DATA   AGE
snapshot-request-2jvb9   1      15h
snapshot-request-6cxg8   1      15s
snapshot-request-bv7pm   1      14h
snapshot-request-c8dbt   1      18s
snapshot-request-clqkf   1      19h
snapshot-request-ftnhd   1      18h
snapshot-request-n6rc9   1      15h
snapshot-request-vm5dg   1      14h
```

Also we should also add a new command for listing the snapshots (and in-progress snapshot requests), e.g. `vcluster snapshot list`, so checking the snapshots should be even easier.

pkg/util/clihelper/virtual.go
nprokopic: Not sure TBH, I have just used the existing function (moved it here so it can be reused) for getting the virtual cluster config, so I didn't dig into it to see how it works in detail.

pkg/snapshot/request.go
nprokopic: we can add the vcluster label in any case

jjaferson: > Also we should also add a new command for listing the snapshots

That is the reason I was asking but I guess many snapshots can be happening in parallel and even to see the history of snapshots that have been taken

pkg/util/clihelper/virtual.go
jjaferson: I see, I was wondering because we have an issue where the vcluster was being paused and not resumed due to the restore command being interrupted https://linear.app/loft/issue/ENG-8545/bug-vcluster-restore-leaves-pod-scaled-down-if-interrupted-preventing

pkg/server/routes/snapshots.go
FabianKramm: The error you posted just shows that the certificate is self-signed, has nothing to do with auth. Only certain resources in the chain will be authorized and authenticated before sending upstream to the virtual kubernetes server. If you are not authenticated vCluster will treat the request as anonymous and since you don't have a specific authorization check everyone will be able to do snapshots which is hugely probelmatic

pkg/snapshot/controller.go
```diff
@@ -0,0 +1,474 @@
...
+		// something went wrong, recorde error and update snapshot request phase to Failed
+		c.eventRecorder.Eventf(&configMap, corev1.EventTypeWarning, "SnapshotRequestFailed", "Snapshot request %s/%s has failed with error: %v", configMap.Namespace, configMap.Name, retErr)
+		_, err = c.updateRequestPhase(ctx, &configMap, *snapshotRequest, RequestPhaseFailed)
+		if err != nil {
+			retErr = fmt.Errorf("failed to update snapshot request phase to %s: %w", RequestPhaseFailed, err)
+		}
+	}()
+
+	switch snapshotRequest.Status.Phase {
+	case "":
```
FabianKramm, line 124: Okay maybe I overlooked this, but we immediately delete the configMap when its failed or completed correct?

nprokopic: The Secret (where the snapshot options are, including potentially saved credentials) is always deleted immediately after the request is processed (whether it completed or failed).

Initially I was also deleting the ConfigMap immediately after processing the request (Completed or Failed), but that made it harder to develop, especially in case of errors, so I changed it to preserve the ConfigMap when the snapshot creation fails, so you can more easily check what was the error.

Then I also changed it to keep the ConfigMap for the completed requests as well, as it was just easier to develop and see if everything worked correctly (should have reverted this).

I would leave at least failed request ConfigMaps, so it's easier to see what failed, and then maybe auto-delete them after some TTL like 1 hour (similarly to Events). For completed request ConfigMaps, I can change it to delete it immediately, or also after some TTL.

pkg/util/clihelper/virtual.go
nprokopic: Here specifically, if you interrupt the CLI command while it's trying to get the kubeconfig for accessing the virtual cluster, then the snapshot request shouldn't be even sent, because it first has to connect in order to get the access to the virtual server.

test/e2e/snapshot/snapshot.go
```diff
@@ -20,576 +20,661 @@ import (
...
-			WithTimeout(framework.PollTimeout).
-			Should(gomega.Equal(configMapToRestore.Data))
 
-		gomega.Eventually(func() map[string][]byte {
-			secrets, err := f.VClusterClient.CoreV1().Secrets(defaultNamespace).List(f.Context, metav1.ListOptions{
-				LabelSelector: "snapshot=restore",
-			})
+			// check deployment is deleted
+			Eventually(func() error {
+				_, err := f.VClusterClient.CoreV1().Secrets(testNamespaceName).List(f.Context, metav1.ListOptions{
```
jjaferson, line 493: I guess we should check if the deployment was deleted here

test/e2e/snapshot/snapshot.go
```diff
@@ -20,576 +20,661 @@ import (
...
+			Eventually(func() error {
+				_, err := f.VClusterClient.CoreV1().ConfigMaps(testNamespaceName).List(f.Context, metav1.ListOptions{
+					LabelSelector: "snapshot=restore",
+				})
 
-				for _, container := range pod.Status.ContainerStatuses {
-					if container.State.Running == nil || !container.Ready {
-						return fmt.Errorf("pod %s container %s is not running", pod.Name, container.Name)
-					}
+				if err != nil {
```
jjaferson, line 463: Should we check for not found?

test/e2e/snapshot/snapshot.go
```diff
@@ -20,576 +20,661 @@ import (
...
+					ObjectMeta: metav1.ObjectMeta{
+						Labels: map[string]string{
+							"snapshot": "restore",
+						},
+					},
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{
+							{
+								Name:  "example-container",
+								Image: "nginx:1.25.0",
```
jjaferson, line 121: Nit: move this into a constant

test/e2e/snapshot/snapshot.go
```diff
@@ -20,576 +20,661 @@ import (
...
+					Namespace: vClusterDefaultNamespace,
 				},
-			},
-		},
-	}
+				Spec: corev1.PersistentVolumeClaimSpec{
+					AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
+					Resources: corev1.VolumeResourceRequirements{
+						Requests: corev1.ResourceList{
+							corev1.ResourceStorage: resource.MustParse("5Gi"),
```
jjaferson, line 163: Nit: move this to a constant

test/e2e/snapshot/snapshot.go
```diff
@@ -20,576 +20,661 @@ import (
...
+			By("Delete resources that going to be restored")
+			err := f.VClusterClient.CoreV1().ConfigMaps(testNamespaceName).Delete(f.Context, configMapToRestore.Name, metav1.DeleteOptions{})
 			framework.ExpectNoError(err)
 
-			for _, pod := range pods.Items {
-				if len(pod.Status.ContainerStatuses) == 0 {
-					return fmt.Errorf("pod %s has no container status", pod.Name)
-				}
+			// check configmap is deleted
+			Eventually(func() error {
```
jjaferson, line 458: maybe we can move this function to a generic one `verifyResourceDeleted(resourceType) error `

pkg/server/routes/snapshots.go
nprokopic: As we talked - I have changed the implementation to create the snapshot request resources in the host cluster. Server route has been removed now.

pkg/server/routes/snapshots.go
nprokopic: Added vcluster name as a prefix for the generated name:

```
kubectl get cm -A -l vcluster.loft.sh/snapshot-request
NAMESPACE   NAME                            DATA   AGE
nik217      nik217-snapshot-request-cpncc   1      8m56s
```

FabianKramm (approved):

PR #3173, opened 2025-09-05
vcluster-config tag has to be in format vcluster-vMAJOR.MINOR

hidalgopl:
/kind bugfix
resolves #

### DIFF
--- a/.github/workflows/sync-config-schema.yaml
+++ b/.github/workflows/sync-config-schema.yaml
@@ -105,7 +105,7 @@ jobs:
           if [[ "${RELEASE_TAG}" == *"-rc"* ]]; then
             TARGET_VERSION="${MINOR_VERSION_WITH_PATCH}"
             echo "RC release detected, using versioned folder: ${TARGET_VERSION}"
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-config-schema.yaml
```diff
@@ -105,7 +105,7 @@ jobs:
           if [[ "${RELEASE_TAG}" == *"-rc"* ]]; then
             TARGET_VERSION="${MINOR_VERSION_WITH_PATCH}"
             echo "RC release detected, using versioned folder: ${TARGET_VERSION}"
-            VCLUSTER_CONFIG_BRANCH="vcluster-v${TARGET_VERSION}"
+            VCLUSTER_CONFIG_BRANCH="vcluster-v${MAJOR_VERSION}.${MINOR_VERSION}"
```
cbron, line 108: So this just removes the patch number

cbron (approved):

.github/workflows/sync-config-schema.yaml
hidalgopl: yes

Issue #3017, opened 2025-07-23
Ingressclass Missed on Ingress Template


valkiriaaquatica:
### Is your feature request related to a problem?

Not at all, but it would be fine if the in the template it's added an ingressClassName to place the ingress we want

### Which solution do you suggest?

Add ingressClassName: .{{{ - Values.controlPlane.ingress.spec.ingressClassName }}  in the ingress template https://github.com/loft-sh/vcluster/blob/main/chart/templates/ingress.yaml   and then ingressClassName: in the values.yaml  https://github.com/loft-sh/vcluster/blob/main/chart/values.yaml#L535 .

### Which alternative solutions exist?

Manually edit the ingress and add it if we want to specify it in case we have multiple ingressclass in the cluster.

### Additional context

Not too much, just wondering if this blank was know thing/purpose. If not, I can just PR with the change mentioned!

Thanks in advanced!

cbron:
Hi @valkiriaaquatica I think that makes sense, but we should add it to the config here, and have it be optional: https://www.vcluster.com/docs/vcluster/next/configure/vcluster-yaml/control-plane/deployment/ingress
If you want to give it a shot you can, otherwise I will add it to our enhancement backlog.

valkiriaaquatica:
Hi @cbron  just opened both PR in this repo and on vcluster-docs repo

cbron:
Closing this as solution was given on PR, feel free to re-open if you have questions.

PR #3174, opened 2025-09-06
feat(values): add ingressclassname option in ingress

valkiriaaquatica:
/kind enhancement

resolves # https://github.com/loft-sh/vcluster/issues/3017#issuecomment-3260089993

Enchance the ingress configuration to add an option option to add ingressclassname so user can choose the ingress controller that will handle the ingrerss.

There is also this PR opened on https://github.com/loft-sh/vcluster-docs  to enchance the docs https://github.com/loft-sh/vcluster-docs/pull/1127

### DIFF
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -18,6 +18,9 @@ metadata:
 {{ toYaml .Values.controlPlane.ingress.labels | indent 4 }}
   {{- end }}
 spec:
+  {{- if .Values.controlPlane.ingress.ingressClassName }}
+  ingressClassName: {{ .Values.controlPlane.ingress.ingressClassName | quote }}
+  {{- end }}
   {{- with .Values.controlPlane.ingress.spec }}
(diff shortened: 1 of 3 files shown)


chart/values.yaml
```diff
@@ -547,6 +547,9 @@ controlPlane:
     # Spec allows you to configure extra ingress options.
     spec:
       tls: []
+      # Optional set the ingressClassName if you want to target a specific ingress controller of the cluster
+      # Example: niginx, cloudflare, traefik, ...
+      ingressClassName: ""
```
cbron: This will need to be added in config.go: https://github.com/loft-sh/vcluster/blob/main/config/config.go#L2274 , and then run `just generate-config-schema` to add to both places. Also you'd need to add to actual ingress file.

valkiriaaquatica: Thanks for the information! I have updated the PR

cbron:
Going to close this, as Fabians solution should be correct pattern to follow.

PR #3188, opened 2025-09-10
fix: preventing vcluster from being recreated when pod is down

jjaferson:
/kind bugfix

resolves ENG-8545
fixes ENG-8545

Fixed preventing vcluster from being recreated from a snapshot when vcluster pod is scale down

### DIFF
--- a/pkg/cli/create_helm.go
+++ b/pkg/cli/create_helm.go
@@ -43,6 +43,7 @@ import (
 	"github.com/loft-sh/vcluster/pkg/constants"
 	"github.com/loft-sh/vcluster/pkg/embed"
 	"github.com/loft-sh/vcluster/pkg/helm"
+	"github.com/loft-sh/vcluster/pkg/lifecycle"
 	"github.com/loft-sh/vcluster/pkg/platform"
 	platformclihelper "github.com/loft-sh/vcluster/pkg/platform/clihelper"
 	"github.com/loft-sh/vcluster/pkg/platform/sleepmode"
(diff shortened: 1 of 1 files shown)


pkg/cli/create_helm.go
```diff
@@ -180,6 +181,12 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 	if !cmd.Upgrade {
 		if isVClusterDeployed(release) {
 			if cmd.Restore != "" {
+				log.Infof("Resuming vCluster %s after it was paused", vClusterName)
+				err = lifecycle.ResumeVCluster(ctx, cmd.kubeClient, vClusterName, cmd.Namespace, true, log)
+				if err != nil {
+					log.Infof("Skipped resuming vCluster %s", vClusterName)
```
hidalgopl, line 187: do we want to continue on error here? If so, let's at least log original error, because otherwise there is no way to tell why vCluster wasn't resumed

jjaferson: Yeah, it's fine only saying that resuming vCluster was skipped in this phase, the [error message there](https://github.com/loft-sh/vcluster/blob/main/pkg/lifecycle/lifecycle.go#L251) will only say that the vCluster was not paused and if it wasn't paused we want to skip, if the restore command is not interrupt vCluster instance will be resumed. 

We are adding this step there to recover from a bad state where the snapshot restore command was interrupt by a terminal exit signal and did not resume the vCluster, so when the command is retried we can resume the vCluster instance before the [initSnapshotCommand](https://github.com/loft-sh/vcluster/blob/main/pkg/cli/restore_helm.go#L26C43-L26C62) fails due to the vCluster instance not being running

hidalgopl (approved):

PR #3192, opened 2025-09-11
Volume snapshots config and upstream controller and CRDs

nprokopic:
/kind feature

towards ENG-7763 and ENG-5938

This is a smaller PR that just adds new config `deploy.volumeSnapshotController.enabled` and deploys upstream [volume snapshot controller](https://github.com/kubernetes-csi/external-snapshotter) when `deploy.volumeSnapshotController.enabled` is set to `true` (only with private nodes).

### DIFF
--- a/pkg/cli/create_helm.go
+++ b/pkg/cli/create_helm.go
@@ -329,6 +329,11 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 		return err
 	}
 
+	err = pkgconfig.ValidateVolumeSnapshotController(vClusterConfig.Deploy.VolumeSnapshotController, vClusterConfig.PrivateNodes)
+	if err != nil {
+		return err
+	}
+
 	warnings := pkgconfig.Lint(*vClusterConfig)
 	for _, warning := range warnings {
 		cmd.log.Warnf(warning)
--- a/pkg/config/validation.go
+++ b/pkg/config/validation.go
@@ -200,7 +200,7 @@ func ValidateConfigAndSetDefaults(vConfig *VirtualClusterConfig) error {
 	}
 
 	// validate deploy.volumeSnapshotController
-	err = validateVolumeSnapshotController(vConfig)
+	err = ValidateVolumeSnapshotController(vConfig.Config.Deploy.VolumeSnapshotController, vConfig.PrivateNodes)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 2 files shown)


pkg/snapshot/controller.go
```diff
@@ -0,0 +1,480 @@
...
+		return ctrl.Result{}, fmt.Errorf("failed to unmarshal vcluster snapshot request from ConfigMap %s/%s: %w", configMap.Namespace, configMap.Name, err)
+	}
+
+	// Not done? Add finalizer! 🔒
+	if !snapshotRequest.Done() {
+		updated, err := c.addFinalizer(ctx, &configMap)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to add vCluster snapshot controller finalizer to the snapshot request ConfigMap %s/%s: %w", configMap.Namespace, configMap.Name, err)
+		}
+		if updated {
```
jjaferson: Will there ever be a case where it won't be updated and no error is returned? If so then the reconciler will continue the reconciliation?

pkg/snapshot/controller.go
```diff
@@ -0,0 +1,480 @@
...
+	// Find snapshot request secret, it contains snapshot options (with the storage credentials) 🪪
+	var secret corev1.Secret
+	secretObjectKey := client.ObjectKey{
+		Namespace: configMap.Namespace,
+		Name:      configMap.Name,
+	}
+	err := c.client().Get(ctx, secretObjectKey, &secret)
+	if kerrors.IsNotFound(err) {
+		// Too soon? Requeue if this is a recently created snapshot request.
+		if time.Since(configMap.CreationTimestamp.Time) < 10*time.Second {
```
jjaferson: nit: What if it takes longer than 10 second to create the secret or something went wrong on the creation of the secret? I think controllers should be idempotent and return the same result for a state regardless of the time it takes. 

In this case I'd log that the secret wasn't found but return true as the reconciliation still in progress

jjaferson (commented):
Code looks

pkg/snapshot/controller.go
nprokopic: Secret is always created **before** the ConfigMap, so it should already exist when the ConfigMap is reconciled. Forgot now why I added this initially, IIRC it was because cached clients needs a second to see the Secret, will check.

> In this case I'd log that the secret wasn't found but return true as the reconciliation still in progress

I was letting it to requeue always when the Secret is not found, but then changed it while debugging the reconciliation issue (failing fast was easier to work with), and it stayed then like that 🙈 

I'll make a change here, need to add more observability, so when this happens you can see more easily that it's missing the secret. I would also add a separate snapshot request phase, e.g. `IncompleteRequest`, or something like that.

pkg/snapshot/controller.go
nprokopic: > Will there ever be a case where it won't be updated and no error is returned?

Yes. It sets the finalizer only on the first reconciliation, and when it sets the finalizer it immediately returns (because the ConfigMap got updated, so new reconciliation is immediately scheduled).

>  If so then the reconciler will continue the reconciliation?
Every next reconciliation is no-op here, because the finalizer has been already set, so it just proceeds with the rest of the reconciliation.

nprokopic:
Updated config, now there is only `deploy.volumeSnapshotController.enabled`

FabianKramm (approved):

PR #3200, opened 2025-09-12
Upgrade upstream volume snapshots client module to v8

nprokopic:
/kind enhancement

towards ENG-7763 and ENG-5938

### DIFF
--- a/pkg/snapshot/volumes/csi/snapshotter.go
+++ b/pkg/snapshot/volumes/csi/snapshotter.go
@@ -1,6 +1,0 @@
-package csi
-
-import (
-	_ "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1"
-	_ "github.com/kubernetes-csi/external-snapshotter/client/v8/clientset/versioned"
-)


jjaferson (approved):

pkg/snapshot/volumes/csi/snapshotter.go
```diff
@@ -0,0 +1,6 @@
+package csi
+
+import (
+	_ "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1"
+	_ "github.com/kubernetes-csi/external-snapshotter/client/v8/clientset/versioned"
+)
```
FabianKramm: Why is that needed?

nprokopic: Not needed in this PR, it's used in the next stacked PR here https://github.com/loft-sh/vcluster/pull/3206.

VolumeSnapshots are not part of the core API, so in order to work with them we need this module.

I added a separate PR for this (like for manifests in previous PR) to keep the PRs with actual feature code smaller (without vendor dir, big YAMLs, etc).

nprokopic:
> @nprokopic we already have v4 as a dependency and I don't like that now we have two different versions as a dependency, either use v4 or upgrade the existing stuff to use v8 as well

Upgraded everything to use just v8, and completely removed v4.

PR #3202, opened 2025-09-12
Fix sync-config-schema.yaml

hidalgopl:
/kind bugfix

resolves #

Fixed an issue where vcluster were pushing only tag to the vcluster-config without the commit and it is causing issue


.github/workflows/sync-config-schema.yaml
```diff
@@ -219,6 +219,8 @@ jobs:
 
           # Update the existing tag to point to the new commit
           git tag -f "${TARGET_TAG}"
+          # push the commit
+          git push origin
```
Piotr1215: This will fail in CI, because the current branch has no upstream branch.

```suggestion
          git push origin HEAD:main
```

PR #3205, opened 2025-09-14
Make sure chart stores chart version in labels safely

### LINKED ISSUE #3204: Chart incorrectly assumes that .Chart.Version is always usable in .metadata.labels.
### What happened?

```
 StatefulSet.apps "vcluster" is invalid: metadata.labels: Invalid value: "vcluster-0.28.0+570022cfeb6c": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```
We tried using Flux's helm-controller to deploy vcluster but we got the error above.

[Here](https://github.com/loft-sh/vcluster/blob/main/chart/templates/service.yaml#L12) you can see 
```
  chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
```
This is a problem and it is in several manifests.

It is a problem because .Chart.Version is a semantic version and not all semantic version values can be used in .metadata.labels.    e.g.   A semantic version may contain the '+' character but .metadata.labels values cannot.

Something like the following should be used to sanitize the version if it is going to be used in label.   
```
{{- define "mychart.version.label" -}}
{{- $rawVersion := .Chart.Version -}}
{{- $sanitized := replace "+" "_" $rawVersion | replace "@" "_" -}}
{{- if gt (len $sanitized) 63 -}}
{{- $sanitized | trunc 63 -}}
{{- else -}}
{{- $sanitized -}}
{{- end -}}
{{- end -}}
```

I expected my helm release to be healthy 

Try using a chart version that is a valid semantic version and that contains a '+' character.

_No response_

Client Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.3", GitCommit:"9e644106593f3f4aa98f8a84b23db5fa378900bd", GitTreeState:"clean", BuildDate:"2023-03-15T13:40:17Z", GoVersion:"go1.19.7", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"31", GitVersion:"v1.31.11", GitCommit:"7e94e1ce5c71407a782db8ef701e290bb0246da0", GitTreeState:"clean", BuildDate:"2025-07-15T18:05:33Z", GoVersion:"go1.23.10", Compiler:"gc", Platform:"linux/amd64"}
WARNING: version difference between client (1.26) and server (1.31) exceeds the supported minor version skew of +/-1

### vcluster version

0.28.0

### VCluster Config

0.28.0

dee0sap:
/kind bugfix

resolves #
Resolves #3204 

Fix chart deploy fail when chart version does meet k8s label value restrictions.  e.g. It contains '+' char


dee0sap:
Hey @FabianKramm 

Thanks for looking at this.

If you are installing helm charts with Flux then when Flux is creating making the release it makes the chart version 
```
[version from chart.yaml]+[revision from source repo]
```

In my organization we are using Flux for our gitops based deployments consequently vcluster is undeployable as is. ( See error I mentioned in the ticket )

chart/Chart.yaml
```diff
@@ -23,4 +23,4 @@ sources:
   - https://github.com/loft-sh/vcluster
 type: application
 
-version: 0.0.1 # version is auto-generated by release pipeline
+version: 0.28.0 # version is auto-generated by release pipeline
```
FabianKramm: Please change this back, this will be automatically filled in the pipeline

FabianKramm: This also makes the unit test fail

dee0sap: Thanks @FabianKramm 

Reverted.

FabianKramm (approved):

chart/Chart.yaml
FabianKramm: Great thanks as soon as the tests pass I will merge this

PR #3206, opened 2025-09-14
Add volume snapshots support to `vcluster snapshot create` command

nprokopic:
/kind feature

closes ENG-7763, ENG-5938, ENG-9121, ENG-8836

Add volume snapshots support to `vcluster snapshot create` and `vcluster restore` command.

This PR adds support for creating volume snapshots and restoring volumes for vCluster with shared and private nodes. Support for standalone will be added in one of the subsequent PRs.

Remaining TODOs:
- [x] Clean up snapshot/restore request ConfigMap 24h after the snapshot request reconciliation has been finished (Completed / Failed / PartiallyFailed)
- [x] Add new RBAC config for volume snapshots with `auto` value (enabled with shared nodes, disabled with private nodes)
- [x] ~Clean up VolumeSnapshot and VolumeSnapshotContent resources that vCluster creates when creating snapshots and restoring volumes~ Created issue ENG-9271 for this.
- [ ] Add e2e tests (in progress)


pkg/snapshot/controller.go
```diff
@@ -97,37 +142,87 @@ func (c *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
+		// reconcile volume snapshots
+		volumeSnapshotsRequest := &snapshotRequest.Spec.VolumeSnapshots
+		previousVolumeSnapshotsRequestPhase := volumeSnapshotsRequest.Status.Phase
+		err = c.volumeSnapshotter.Reconcile(ctx, snapshotRequest.Name, volumeSnapshotsRequest)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to reconcile volume snapshots: %w", err)
+		}
+
+		// check volume snapshots' status
+		switch volumeSnapshotsRequest.Status.Phase {
```
jjaferson: nit: to make this a bit more readable I would move this to a func 
```
processVolumeSnapshotStatus(&volumeSnapshotsRequest.Status.Phase, previousVolumeSnapshotsRequestPhase) (ctrl.Result, error)
```

nprokopic: I will tidy up these long funcs. More than few changes were happening lately, so I wanted to wait for that to stabilize a bit, before moving around code.

pkg/cli/snapshot_helm.go
```diff
@@ -57,13 +61,81 @@ func Snapshot(ctx context.Context, args []string, globalFlags *flags.GlobalFlags
...
+	restClient, err := vCluster.ClientFactory.ClientConfig()
+	if err != nil {
+		return fmt.Errorf("failed to get rest client for vcluster %s: %w", vClusterName, err)
+	}
+	kubeClient, err := kubernetes.NewForConfig(restClient)
+	if err != nil {
+		return fmt.Errorf("failed to create kubernetes client for vcluster %s: %w", vClusterName, err)
+	}
+	// get snapshot request ConfigMap
+	snapshotRequestName := args[1]
```
jjaferson: Should we validate if the snapshot request name has been passed?

pkg/snapshot/restorecontroller.go
```diff
@@ -0,0 +1,267 @@
...
+func NewRestoreController(registerContext *synccontext.RegisterContext) (*RestoreReconciler, error) {
+	logger := loghelper.New(restoreControllerName)
+
+	if registerContext == nil {
+		return nil, errors.New("register context is nil")
+	}
+	if registerContext.Config == nil {
+		return nil, errors.New("virtual cluster config is nil")
+	}
+	isHostMode, err := IsSnapshotRequestCreatedInHostCluster(registerContext.Config)
```
jjaferson, line 43: Will snapshot PV work for standalone clusters?

pkg/snapshot/volumes/csi/snapshotinprogress.go
```diff
@@ -0,0 +1,260 @@
...
+		Name:      volumeSnapshotRequest.PersistentVolumeClaim.Name,
+	}
+	if volumeSnapshotStatus.Phase != volumes.RequestPhaseInProgress {
+		return updatedStatus(fmt.Errorf("invalid volume snapshot request phase %s for PVC %s, expected %s, got %s", volumeSnapshotStatus.Phase, pvcName.String(), volumes.RequestPhaseInProgress, volumeSnapshotStatus.Phase))
+	}
+
+	// Check if VolumeSnapshot has been created
+	volumeSnapshotName := fmt.Sprintf("%s-%s", volumeSnapshotRequest.PersistentVolumeClaim.Name, requestName)
+	volumeSnapshot, err := s.snapshotsClient.SnapshotV1().VolumeSnapshots(pvcName.Namespace).Get(ctx, volumeSnapshotName, metav1.GetOptions{})
+	if kerrors.IsNotFound(err) {
```
jjaferson: it's down to preference but I find it better to return the error as early as possible 

```
if err != nil {

 if !kerrors.IsNotFound(err) {
 return updatedStatus(fmt.Errorf("failed to get VolumeSnapshot %s/%s: %w", volumeSnapshot.Namespace, volumeSnapshot.Name, err)) 
 }
 
 ...
  
 createVolumeSnapshotResource() ... 
}
```

cmd/vclusterctl/cmd/snapshot/root.go
```diff
@@ -58,6 +58,8 @@ vcluster snapshot my-vcluster container:///data/my-local-snapshot.tar.gz
 
 	// add subcommands
 	cobraCmd.AddCommand(NewCreateCmd(globalFlags))
+	cobraCmd.AddCommand(NewGetRequestCmd(globalFlags))
+	cobraCmd.AddCommand(NewListRequestsCmd(globalFlags))
```
jjaferson: We need to expose these commands to the vCluster binary as well https://github.com/loft-sh/vcluster/blob/main/cmd/vcluster/cmd/snapshot/snapshot.go#L27 

Not sure how we want to handle it as we have vcluster snapshot list and vcluster snapshot delete exposed via the vCluster binary

chart/templates/role.yaml
```diff
@@ -74,10 +74,13 @@ rules:
...
     verbs: ["create", "delete", "patch", "update", "get", "list", "watch"]
   {{- end }}
+  - apiGroups: [ "snapshot.storage.k8s.io" ]
+    resources: [ "volumesnapshots" ]
+    verbs: ["create", "delete", "patch", "update", "get", "list"]
   {{- if .Values.sync.toHost.volumeSnapshots.enabled }}
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshots"]
-    verbs: ["create", "delete", "patch", "update", "get", "list", "watch"]
+    verbs: ["watch"]
```
FabianKramm, line 90: Are you sure you don't need the watch here?

chart/templates/clusterrole.yaml
```diff
@@ -63,10 +63,13 @@ rules:
     resources: ["csistoragecapacities"]
```
FabianKramm, line 63: Since the cluster role is not always available, snapshot sometimes doesn't work like you expect I guess, have you thought about this?

cmd/vclusterctl/cmd/snapshot/getrequest.go
```diff
@@ -0,0 +1,44 @@
...
+	Log log.Logger
+}
+
+func NewGetRequestCmd(globalFlags *flags.GlobalFlags) *cobra.Command {
+	getRequestCmd := &GetRequestCmd{
+		GlobalFlags: globalFlags,
+		Log:         log.GetInstance(),
+	}
+
+	_, nameValidator := util.NamedPositionalArgsValidator(true, false, "VCLUSTER_NAME")
```
FabianKramm: This hould be called "get" instead of "get-request"

cmd/vclusterctl/cmd/snapshot/listrequests.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func NewListRequestsCmd(globalFlags *flags.GlobalFlags) *cobra.Command {
+	listRequestsCmd := &ListRequestsCmd{
+		GlobalFlags: globalFlags,
+		Log:         log.GetInstance(),
+	}
+
+	_, nameValidator := util.NamedPositionalArgsValidator(true, false, "VCLUSTER_NAME")
+	cobraCmd := &cobra.Command{
```
FabianKramm: Should be "list" instead of "list-request", if this is just about currently in progress snapshots, then we shouldn't have this command

pkg/controllers/resources/persistentvolumeclaims/syncer.go
```diff
@@ -235,6 +260,35 @@ func (s *persistentVolumeClaimSyncer) ensurePersistentVolume(ctx *synccontext.Sy
 	return false, nil
 }
 
+func (s *persistentVolumeClaimSyncer) isHostVolumeRestoreInProgress(ctx *synccontext.SyncContext, pObj *corev1.PersistentVolumeClaim) (bool, error) {
+	configMaps := &corev1.ConfigMapList{}
+	err := ctx.HostClient.List(ctx.Context, configMaps, client.InNamespace(ctx.Config.HostNamespace), client.MatchingLabels{
+		snapshotMeta.RestoreRequestLabel: "",
```
FabianKramm: This looks weird, why is the label empty?

pkg/snapshot/controllerbase.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	c.logger.Infof(
+		"Added vCluster %s controller finalizer %s to the %s request ConfigMap %s/%s",
+		c.kind,
+		c.finalizer,
+		c.kind,
+		configMap.Namespace,
+		configMap.Name)
+	return true, nil
```
FabianKramm, line 86: Please don't do so verbose logging here, rather make this a V(1)

pkg/snapshot/controllerbase.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	c.logger.Infof(
+		"Removed vCluster %s controller finalizer %s from the %s request ConfigMap %s/%s",
+		c.kind,
+		c.finalizer,
+		c.kind,
+		configMap.Namespace,
+		configMap.Name)
+	return nil
```
FabianKramm, line 121: Less verbose logging here, don't think adding/reoving a finalizer needs two log messages

pkg/snapshot/meta/labels.go
```diff
@@ -0,0 +1,6 @@
+package meta
+
+const (
+	RequestLabel        = "vcluster.loft.sh/snapshot-request"
+	RestoreRequestLabel = "vcluster.loft.sh/restore-request"
+)
```
FabianKramm: Move this to constants package

FabianKramm (changes requested):

pkg/cli/snapshot_helm.go
```diff
@@ -57,13 +61,81 @@ func Snapshot(ctx context.Context, args []string, globalFlags *flags.GlobalFlags
 	}
 
 	// creating snapshot request with 'vcluster snapshot create' command
-	err = createSnapshotRequest(ctx, vCluster, kubeClient, snapshotOpts, log)
+	err = createSnapshotRequest(ctx, vCluster, kubeClient, snapshotOpts, log, includeVolumes)
```
jjaferson, line 59: We will need this function to be invoked by the vcluster binary too, shouldn't we move it to the snapshot client [here](https://github.com/loft-sh/vcluster/blob/main/pkg/snapshot/client.go)

pkg/cli/snapshot_helm.go
```diff
@@ -57,13 +61,81 @@ func Snapshot(ctx context.Context, args []string, globalFlags *flags.GlobalFlags
...
 	if err != nil {
 		return err
 	}
 	return nil
 }
 
+func GetRequest(ctx context.Context, args []string, globalFlags *flags.GlobalFlags, log log.Logger) error {
```
jjaferson: Same, shouldn't it be part of the snapshot client?

pkg/cli/snapshot_helm.go
nprokopic: Yes!

pkg/snapshot/restorecontroller.go
nprokopic: Not in this PR, there is an explicit check for that which fails.

cmd/vclusterctl/cmd/snapshot/root.go
nprokopic: Will look into this, how to expose these, but wouldn't do that in this PR.

nprokopic: For all the additions needed for the integration with auto snapshots, I would rather make those in separate PRs, so the scope does not explode here.

chart/templates/role.yaml
nprokopic: Yup. vCluster is not watching VolumeSnapshots, it's the upstream snapshot-controller that is watching them.

We could implement watching VolumeSnapshots, as an optimization, so that vCluster reacts directly to VolumeSnapshot changes, as opposed to periodically checking them. If we want to do this, I would rather make that change in a separate PR, as this is already big enough and we want QA to start testing as soon as possible.

chart/templates/clusterrole.yaml
nprokopic: Good reminder, thanks!

Yes, I was looking into this, that's remaining to be done, forgot to add it to the TODO list in the PR description.

The original idea was to use the config for enabling volume snapshots to check if the role has to be deployed, but we don't have such config anymore for volume snapshots.

So now I think that we have to always deploy the ClusterRole (also Role), because otherwise volume snapshots will not work. These roles would then always have minimal permissions with just VolumeSnapshot, VolumeSnapshotContent and VolumeSnapshotClass resources being specified.

nprokopic: We can also add back some explicit config for this, so users have to enable it explicitly (e.g. like deploying upstream snapshot-controller with private nodes).

cmd/vclusterctl/cmd/snapshot/listrequests.go
nprokopic: > if this is just about currently in progress snapshots, then we shouldn't have this command

Yes, this was the idea. Mostly added so it's easier to check the status while the snapshot is being created (especially for QA).

In this case I would just remove these commands here and add `vcluster snapshot get` and `vcluster snapshot list` is another PR, to keep the scope limited in this PR.

pkg/controllers/resources/persistentvolumeclaims/syncer.go
nprokopic: There is no label value, it's just to label the ConfigMap as a snapshot request. We can put some dummy value, but technically it's not required.

pkg/cli/snapshot_helm.go
nprokopic: Another PR changes this https://github.com/loft-sh/vcluster/pull/3234

pkg/cli/snapshot_helm.go
nprokopic: I will probably move all this from this PR, and then add `vcluster snapshot get` and `vcluster snapshot list` commands in a new PR. See this thread https://github.com/loft-sh/vcluster/pull/3206#discussion_r2381776956.

cmd/vclusterctl/cmd/snapshot/listrequests.go
jjaferson: How would the list and get work? We already have a [list](https://github.com/loft-sh/vcluster/blob/main/pkg/snapshot/client.go#L60) snapshots command that connects to the backing store to check whether the snapshot exists or not AFAIK the `GetRequest` and `ListRequest` commands just checks the CM with the snapshot request metadata

nprokopic: Get and list commands would have to show both uploaded snapshots and ongoing snapshot requests, preferably in some unified way, and show the status of all of them, so users can see clearly if some snapshot is uploaded, still in progress, or failed.

jjaferson: I agree but I guess it will be out of scope for this release, right? if so should we create an issue to get it implemented?

pkg/snapshot/controllerbase.go
nprokopic: Changed it to `Debugf` (this logger does not have V(1) func).

pkg/snapshot/controllerbase.go
nprokopic: Yeah, these are leftovers from initial debugging, forgot to remove some of these logs. Removed one now.

pkg/snapshot/meta/labels.go
nprokopic: Moved it ✅

cmd/vclusterctl/cmd/snapshot/listrequests.go
nprokopic: > but I guess it will be out of scope for this release, right? 

Not sure about that. I think we can have it also for this release (v0.30), but let's discuss this next week.

> if so should we create an issue to get it implemented?

These commands are mentioned in some of the existing issues, but I created a separate issue now (ENG-9242), to track it more easily.

nprokopic: I have removed get-request and list-requests snapshot commands.

pkg/cli/snapshot_helm.go
nprokopic: I have removed get-request and list-requests snapshot commands.

cmd/vclusterctl/cmd/snapshot/getrequest.go
nprokopic: I have removed get-request and list-requests snapshot commands.

pkg/cli/snapshot_helm.go
```diff
@@ -129,7 +129,7 @@ func initSnapshotCommand(
 	return vCluster, kubeClient, restClient, nil
 }
 
-func createSnapshotRequest(ctx context.Context, vCluster *find.VCluster, kubeClient *kubernetes.Clientset, snapshotOpts *snapshot.Options, log log.Logger) error {
+func createSnapshotRequest(ctx context.Context, vCluster *find.VCluster, kubeClient *kubernetes.Clientset, snapshotOpts *snapshot.Options, log log.Logger, includeVolumes bool) error {
```
jjaferson, line 131: Sorry I approved the other PR https://github.com/loft-sh/vcluster/pull/3234 with the change to expose `createSnapshotRequest` but now that I look at it again, should not this whole function be exposed with the validations? Otherwise whoever uses that function will need to duplicate the validations

jjaferson: Also should the [CreateSnapshotRequestResources](https://github.com/loft-sh/vcluster/pull/3234/files#diff-dcbd5f8718a4eb4303410c42a5e3c138d46250ac84f3035fa0525d16f5497f07R47) be part of the [snapshot client](https://github.com/loft-sh/vcluster/blob/main/pkg/snapshot/client.go)?

chart/templates/clusterrole.yaml
nprokopic: I have added new `rbac.enableVolumeSnapshotRules.enabled` config, which is set to `auto`. Auto behavior is:
- `true` for shared nodes (vCluster is working with volume snapshot resources in the host cluster)
- `false` for private nodes (vCluster is working with volume snapshot resources in the virtual cluster)

pkg/cli/snapshot_helm.go
nprokopic: There are 2 pieces of code that stayed in `createSnapshotRequest` and are not moved to `CreateSnapshotRequestResources`:
- getting vCluster config, because it's done differently based on who is the caller (so the `CreateSnapshotRequestResources` caller should pass the config to the func)
  - CLI here fetches the vCluster config from ConfigMap or Secret
  - vcluster binary reads it from a file
- validation
  - CLI needs to do the validation, because the latest CLI can be used with older vcluster
  - new vcluster binary (where we are adding call to `CreateSnapshotRequestResources`) does not need to validate vCluster version because new vCluster (that will call `CreateSnapshotRequestResources`) supports snapshots

nprokopic: We can move the standalone check into the func though (will move that). But this check should be removed anyway, as standalone support should be added in one of the next PRs.

jjaferson: Yup, any feature validation for creating snapshot request should be handled in the function scope

pkg/snapshot/volumes/csi/snapshotinprogress.go
nprokopic: I also prefer that when it's simpler error handling.

Changed it here as suggested. Now IMO it looks a bit verbose, but it's fine 👍

pkg/cli/snapshot_helm.go
nprokopic: Moved the standalone check into the CreateSnapshotRequestResources func.

pkg/snapshot/controller.go
nprokopic: moved whole block from `case RequestPhaseCreatingVolumeSnapshots` into a separate func

pkg/cli/snapshot_helm.go
nprokopic: This command is now removed.

chart/templates/clusterrole.yaml
FabianKramm: Great that works

FabianKramm (approved):

PR #3207, opened 2025-09-14
Add support for restoring volumes from snapshots

nprokopic:
/kind feature

resolves ENG-7763

Add support for restoring volumes from snapshots.

This PR adds support for restoring volumes from snapshots only for private nodes. Support for shared nodes will be added in one of the subsequent PRs.


pkg/controllers/register.go
```diff
@@ -276,3 +280,20 @@ func registerSnapshotController(registerContext *synccontext.RegisterContext) er
 
 	return nil
 }
+
+func registerRestoreController(registerContext *synccontext.RegisterContext) error {
+	if !registerContext.Config.VolumeSnapshots.Enabled {
```
jjaferson: What happens if there were ongoing snapshot being taken and vcluster gets redeployed with volume snapshots disabled?

pkg/snapshot/controllerbase.go
```diff
@@ -0,0 +1,245 @@
...
+		c.kind,
+		c.finalizer,
+		c.kind,
+		configMap.Namespace,
+		configMap.Name)
+	return nil
+}
+
+// reconcileCompletedRequest cleans up the completed snapshot/restore request resources.
+func (c *reconcilerBase) reconcileCompletedRequest(ctx context.Context, configMap *corev1.ConfigMap) error {
```
jjaferson, line 132: What is the difference between `reconcileCompletedRequest` and `reconcileFailedRequest` func? They both invoke  `reconcileDoneRequest` and just outputs different logs, could it not be handled [here](https://github.com/loft-sh/vcluster/blob/main/pkg/snapshot/controller.go#L139)

pkg/snapshot/controllerbase.go
```diff
@@ -0,0 +1,245 @@
...
+	name := configMap.Name
+	c.logger.Debugf("Deleting %s request Secret %s/%s", c.kind, namespace, name)
+
+	// find snapshot/restore request secret
+	var secret corev1.Secret
+	secretObjectKey := client.ObjectKey{
+		Namespace: namespace,
+		Name:      name,
+	}
+	err := c.client().Get(ctx, secretObjectKey, &secret)
```
jjaferson, line 207: Do we need to check if the secret exists before before deleting it? I guess the `Delete()` func will return an `not found` error if it doesn't exists

pkg/controllers/register.go
nprokopic: This config got removed now, so volume snapshots are enabled via CLI flag (therefore cannot be disabled for a in-progress snapshot/restore request).

pkg/snapshot/controllerbase.go
nprokopic: Here there is no difference, but will add few things:
- event recording
- snapshot request ConfigMap deletion (currently snapshot request resources are kept, as it's easier to debug issues during the development)
  - The idea is to delete the snapshot request ConfigMap after the snapshot request has been completed or failed
  - Completed requests will be maybe deleted immediately (not sure yet)
  - Failed snapshot request will be deleted after some TTL, like events (e.g. by default 1h or maybe 1 day). So users can more easily check issues in case of failures. We could maybe leave the ConfigMap in case of failures, and let users delete it when needed.

I want to have Completed and Failed explicitly handled in code for now, since we may add other differences there, like retries for Failed requests.

pkg/snapshot/controllerbase.go
nprokopic: > I guess the Delete() func will return an not found error if it doesn't exists

right, will check that

PR #3211, opened 2025-09-17
Added platform login info to describe the persisted login credentials location

cbalan:
/kind enhancement

resolves #ENG-4853

Let the user know that the login credentials are stored in the local vcluster config. Mainly relevant for new users who might attempt to reuse an instance(e.g. delete the cluster, deploy another). The Platform login details of persists.

### DIFF
--- a/cmd/vclusterctl/cmd/platform/login.go
+++ b/cmd/vclusterctl/cmd/platform/login.go
@@ -133,6 +133,7 @@ func (cmd *LoginCmd) Run(ctx context.Context, args []string) error {
 	if err != nil {
 		return err
 	}
+	cmd.Log.Infof("Login credentials are stored in %q", cmd.Config)
 	cmd.Log.Donef(product.Replace("Successfully logged into Loft instance %s"), ansi.Color(url, "white+b"))
 
 	// skip log into docker registries?


cmd/vclusterctl/cmd/platform/login.go
```diff
@@ -133,6 +133,7 @@ func (cmd *LoginCmd) Run(ctx context.Context, args []string) error {
 	if err != nil {
 		return err
 	}
+	cmd.Log.Infof("Login credentials are stored in %q", cmd.Config)
```
lizardruss: what do you think about combining these messages? Just thinking that it might get missed. Example:
```suggestion
	cmd.Log.Donef(product.Replace("Successfully logged into Loft instance %s, credentials are stored in %q"), ansi.Color(url, "white+b"), cmd.Config)
```

lizardruss (commented):
Left a question, but should get opinions from Denise / Sowmya

cmd/vclusterctl/cmd/platform/login.go
cbalan: Thank you. I was thinking about combining these, but I was afraid of introducing unexpected side effects. 
1. The comma after the url might be confused as part of the url. 
2. Potential automation might be broken, as those might be looking for the exact format with the URL at the end.
3. The log message gets quite long. The information related to the credentials location might not be visible. For instance, the following has 156 characters. 
23:15:48 done Successfully logged into vCluster Platform instance https://loft.rocks, credentials are stored in "/Users/catalin.balan/.vcluster/config.json"

There is a good chance that these are non-problems. I'm happy to change these. :)

lizardruss: Fair points. I don't have a strong opinion. Could you comment on the linear ticket with the new output and ask for feedback?

FabianKramm (changes requested):
I don't think we should do this, why is this relevant? When I do docker login it also doesn't tell me my credentials are stored at ~/.docker/config.json

PR #3234, opened 2025-09-26
Move snapshot request resources creation into snapshot pkg

nprokopic:
/kind enhancement

towards ENG-9108

Move code that creates snapshot request resources into snapshot pkg so it can be reused for the vcluster binary command.


jjaferson (approved):
LGTM

PR #3235, opened 2025-09-28
feat: Add support for vcluster describe user supplied vcluster.yaml

cbalan:
/kind feature

resolves #ENG-9175

User supplied vcluster.yaml can be retrieved via vcluster describe NAME -o yaml --all=false


FabianKramm (changes requested):
This is unfortunately wrong, please take a look at how `helm get values` works as this issue is about doing that. Right now we take the values from config secret which we shouldn't and instead take it from the helm secret, which is also much easier to do.

cbalan:
Closing due to requirements change. A new PR will be summited with reduced scope.

PR #3237, opened 2025-09-29
feat(snapshot): expose snapshot create command to vcluster binary

jjaferson:
kind enhancement

resolves ENG-9273

Exposes `snapshot create` command to vcluster binary

### DIFF
--- a/cmd/vcluster/cmd/snapshot/create.go
+++ b/cmd/vcluster/cmd/snapshot/create.go
@@ -1,0 +1,62 @@
+package snapshot
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/loft-sh/vcluster/pkg/config"
+	"github.com/loft-sh/vcluster/pkg/constants"
+	setupconfig "github.com/loft-sh/vcluster/pkg/setup/config"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/loft-sh/vcluster/pkg/snapshot"
+	"github.com/spf13/cobra"
+)
+
+func NewCreateCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "create vCluster snapshots",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			// parse vCluster config
+			vConfig, err := config.ParseConfig(constants.DefaultVClusterConfigLocation, os.Getenv("VCLUSTER_NAME"), nil)
+			if err != nil {
+				return err
+			}
+
+			envOptions, err := snapshot.ParseOptionsFromEnv()
+			if err != nil {
+				return fmt.Errorf("failed to parse options from environment: %w", err)
+			}
+
+			restClient, vClusterNamespace, err := setupconfig.InitClientConfig()
+			if err != nil {
+				return fmt.Errorf("failed to init client config: %w", err)
+			}
+
+			kubeClient, err := kubernetes.NewForConfig(restClient)
+			if err != nil {
+				return fmt.Errorf("failed to create kube client: %w", err)
+			}
+
+			request, err := snapshot.CreateSnapshotRequestResources(cmd.Context(), vClusterNamespace, vConfig.Name, vConfig, envOptions, kubeClient)
+			if err != nil {
+				return fmt.Errorf("failed to create snapshot request resources: %w", err)
+			}
+
+			encodedBytes, err := json.Marshal(request)
+			if err != nil {
+				return fmt.Errorf("failed to marshal json: %w", err)
+			}
+
+			if _, err := os.Stdout.Write(encodedBytes); err != nil {
+				return err
+			}
+			return nil
+		},
+	}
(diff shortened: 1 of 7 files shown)


cmd/vcluster/cmd/snapshot/create.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+
+			envOptions, err := snapshot.ParseOptionsFromEnv()
+			if err != nil {
+				return fmt.Errorf("failed to parse options from environment: %w", err)
+			}
+
+			restClient, vClusterNamespace, err := setupconfig.InitClientConfig()
+			if err != nil {
+				return err
```
nprokopic: Error wrapping is inconsistent in the file.

Any reason not to wrap error here with `fmt.Errorf(...)`? (same for other places with just `return err`)

cmd/vcluster/cmd/snapshot/create.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+
+			kubeClient, err := kubernetes.NewForConfig(restClient)
+			if err != nil {
+				return err
+			}
+
+			requet, err := snapshot.CreateSnapshotRequestResources(cmd.Context(), vConfig.Name, vClusterNamespace, envOptions, kubeClient)
```
nprokopic: nit: typo in var name requet -> request 😬

nprokopic: parameters order is different, vCluster namespace goes before name

```suggestion
			requet, err := snapshot.CreateSnapshotRequestResources(cmd.Context(), vClusterNamespace, vConfig.Name, envOptions, kubeClient)
```

nprokopic: Did you maybe check if `vCluster.Name` is set here to a non-empty value? IIRC I tried using the name (or namespace) from the config (in the CLI code though), but it was empty.

jjaferson: Good catch, thanks! 

When I tested it vCluster name and namespace had the same so I didn't see it happening :+1:

jjaferson:
> Why can we not just reuse the CLI one 1:1? We should also document in the command why we need it on the server side

We discussed about using the CLI when we were designing auto snapshot, but as the CLI used to create a pod to run the snapshot we decided to not go on that path also we would need to add the vCluster CLI as dependency in the platform so to avoid these things we decided to use the vcluster binary instead.

For now we are just integrating PVs to the platform but when we start working on Auto snapshot GA we can go back and discuss it again

cmd/vcluster/cmd/snapshot/create.go
jjaferson: Yes, it gets vCluster name from the `VCLUSTER_NAME` env var

```
Name:myvcluster2 HostNamespace: Path:/var/lib/vcluster/config.yaml
```

cmd/vcluster/cmd/snapshot/create.go
jjaferson: Not really but I'll wrap the error to follow the same standard

nprokopic:
> > Why can we not just reuse the CLI one 1:1? We should also document in the command why we need it on the server side
> 
> We discussed about using the CLI when we were designing auto snapshot, but as the CLI used to create a pod to run the snapshot we decided to not go on that path also we would need to add the vCluster CLI as dependency in the platform so to avoid these things we decided to use the vcluster binary instead.
> 
> For now we are just integrating PVs to the platform but when we start working on Auto snapshot GA we can go back and discuss it again

Maybe we can just bundle the CLI binary into the vCluster container (when building vCluster), so it's just there and can be used. More or less, merge parts of [this](https://github.com/loft-sh/vcluster/blob/main/Dockerfile.cli) into [this](https://github.com/loft-sh/vcluster/blob/main/Dockerfile).

jjaferson:
> > > Why can we not just reuse the CLI one 1:1? We should also document in the command why we need it on the server side
> > 
> > 
> > We discussed about using the CLI when we were designing auto snapshot, but as the CLI used to create a pod to run the snapshot we decided to not go on that path also we would need to add the vCluster CLI as dependency in the platform so to avoid these things we decided to use the vcluster binary instead.
> > For now we are just integrating PVs to the platform but when we start working on Auto snapshot GA we can go back and discuss it again
> 
> Maybe we can just bundle the CLI binary into the vCluster container (when building vCluster), so it's just there and can be used. More or less, merge parts of [this](https://github.com/loft-sh/vcluster/blob/main/Dockerfile.cli) into [this](https://github.com/loft-sh/vcluster/blob/main/Dockerfile).

There are also things that vCluster binary does that the CLI doesn't like `list` and `delete` snapshots from the backing store as you can see [here](https://github.com/loft-sh/vcluster/blob/main/cmd/vcluster/cmd/snapshot/snapshot.go#L27)

jjaferson:
@nprokopic I've also added the `--include-volume` flag as part of the `Option` struct so we can leverage the flag when exposing `VCLUSTER_STORAGE_OPTIONS` env var

PR #3243, opened 2025-10-01
Fix endpointSlice deletion issue

ApsTomar:
/kind bugfix

resolves #

Fixed an issue where endpointSlices deleted from virtual cluster does not sync to host i.e. host eps still persisted. This PR resolves that.

### DIFF
--- a/pkg/controllers/resources/endpointslices/syncer.go
+++ b/pkg/controllers/resources/endpointslices/syncer.go
@@ -127,26 +127,24 @@ func (s *endpointSliceSyncer) ReconcileStart(ctx *synccontext.SyncContext, req c
 			err = ctx.HostClient.Get(ctx, s.VirtualToHost(ctx, req.NamespacedName, nil), hostEps)
 			if err != nil {
 				if !kerrors.IsNotFound(err) {
-					klog.Infof("Error retrieving host endpointSlice: %v", err)
+					return true, fmt.Errorf("error retrieving host endpointSlice: %w", err)
 				}
 				return true, nil
 			}
 			err = ctx.HostClient.Delete(ctx, hostEps)
 			if err != nil {
-				klog.Infof("Error deleting endpointSlice from host %s: %v", eps.Name, err)
+				return true, fmt.Errorf("error deleting endpointSlice from host %s: %w", eps.Name, err)
 			}
 			return true, nil
 		}
 
-		klog.Infof("Error retrieving endpointslice: %v", err)
-		return true, nil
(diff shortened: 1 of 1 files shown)


pkg/controllers/resources/endpointslices/syncer.go
```diff
@@ -121,6 +121,23 @@ func (s *endpointSliceSyncer) ReconcileStart(ctx *synccontext.SyncContext, req c
 		Name:      req.Name,
 	}, eps)
 	if err != nil {
+		// if endpointSlice is not found on virtual cluster, then remove it from host as well
+		if kerrors.IsNotFound(err) {
+			hostEps := &discoveryv1.EndpointSlice{}
+			err = ctx.HostClient.Get(ctx, s.VirtualToHost(ctx, req.NamespacedName, nil), hostEps)
+			if err != nil {
+				if !kerrors.IsNotFound(err) {
```
nprokopic, line 129: This hides the error from the `ReconcileStart` caller, as it returns `true, nil` even when there is an error (different than NotFound).


Why not wrap and return the error with something like

```golang
return true, fmt.Errorf("Error retrieving host endpointSlice: %w", err)
```

And let the caller deal with error handling?

nprokopic: In few other places in the file error (different than NotFound) is returned.

pkg/controllers/resources/endpointslices/syncer.go
```diff
@@ -121,6 +121,23 @@ func (s *endpointSliceSyncer) ReconcileStart(ctx *synccontext.SyncContext, req c
...
+					klog.Infof("Error retrieving host endpointSlice: %v", err)
+				}
+				return true, nil
+			}
+			err = ctx.HostClient.Delete(ctx, hostEps)
+			if err != nil {
+				klog.Infof("Error deleting endpointSlice from host %s: %v", eps.Name, err)
```
nprokopic: Same as above, why not wrap and return error (when different than NotFound)?

```golang
return true, fmt.Errorf("Error deleting endpointSlice from host %s: %w", eps.Name, err)
```

pkg/controllers/resources/endpointslices/syncer.go
ApsTomar: makes sense, made the suggested changes.

pkg/controllers/resources/endpointslices/syncer.go
ApsTomar: made the suggested changes

nprokopic (approved):

PR #3245, opened 2025-10-02
fix getting namespace for cert rotation in standalone

hidalgopl:
/kind bugfix

resolves #

fix getting namespace for cert rotation in standalone

### DIFF
--- a/pkg/setup/config/config.go
+++ b/pkg/setup/config/config.go
@@ -28,7 +28,7 @@ const (
 	AnnotationStore  = "vcluster.loft.sh/store"
 )
 
-func InitClientConfig() (*rest.Config, string, error) {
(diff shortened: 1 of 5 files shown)


pkg/setup/config/config.go
```diff
@@ -58,9 +58,14 @@ func InitClientConfig() (*rest.Config, string, error) {
 	}
 	inClusterConfig.Timeout = time.Duration(timeout) * time.Second
 	// get current namespace
-	currentNamespace, err := clienthelper.CurrentNamespace()
-	if err != nil {
-		return nil, "", err
+	var currentNamespace string
```
FabianKramm, line 61: Why can we not do this via the env variable NAMESPACE?

FabianKramm (changes requested):

PR #3254, opened 2025-10-07
Add volume snapshots e2e tests

nprokopic:
/kind test

resolves ENG-9291

Besides adding basic e2e tests, there are few other minor changes here:
- PVC syncer: check if there is an ongoing restore earlier in PVC syncing
- vcluster restore: when restoring a PVC resource, and when that PVC will be restored from the snapshot in the host cluster, delete PVC's volume name, because the new volume has to be created

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -318,6 +318,28 @@ jobs:
           echo "kubectl config current-context: $(kubectl config current-context)"
           echo "KUBECONFIG env var: ${KUBECONFIG}"
 
+      - name: Setup snapshot-controller and CSI driver
+        run: |
+          set -x
+
+          echo "Deploy upstream CSI volume snapshot CRDs and snapshot-controller"
+          kubectl kustomize https://github.com/kubernetes-csi/external-snapshotter/client/config/crd | kubectl create -f -
+          kubectl kustomize https://github.com/kubernetes-csi/external-snapshotter/deploy/kubernetes/snapshot-controller | kubectl create -f -
+
+          echo "Deploy CSI driver"
+          temp_git_dir=$(mktemp -d)
+          git clone https://github.com/kubernetes-csi/csi-driver-host-path.git $temp_git_dir
+          $temp_git_dir/deploy/kubernetes-latest/deploy.sh
+
+          echo "Deploy StorageClass and VolumeSnapshotClass"
+          kubectl apply -f $temp_git_dir/examples/csi-storageclass.yaml
+          kubectl apply -f $temp_git_dir/examples/csi-volumesnapshotclass.yaml
+          kubectl annotate volumesnapshotclass csi-hostpath-snapclass snapshot.storage.kubernetes.io/is-default-class="true"
+          rm -rf $temp_git_dir
+
+          echo "Wait for snapshot-controller to be ready"
+          kubectl wait --for=condition=Available -n kube-system deploy/snapshot-controller --timeout=60s
+
       - name: Download vcluster cli
         uses: actions/download-artifact@v5
         with:
--- a/Justfile
+++ b/Justfile
@@ -86,8 +86,26 @@ embed-chart version="0.0.0":
 test-chart:
   helm unittest chart
 
+setup-csi-volume-snapshots:
+  # Deploy upstream CSI volume snapshot CRDs and snapshot-controller
+  kubectl kustomize https://github.com/kubernetes-csi/external-snapshotter/client/config/crd | kubectl create -f -
+  kubectl kustomize https://github.com/kubernetes-csi/external-snapshotter/deploy/kubernetes/snapshot-controller | kubectl create -f -
+
+  # Deploy CSI driver, StorageClass and VolumeSnapshotClass
+  temp_git_dir=$(mktemp -d) && \
+    git clone https://github.com/kubernetes-csi/csi-driver-host-path.git $temp_git_dir && \
+    $temp_git_dir/deploy/kubernetes-latest/deploy.sh && \
+    kubectl apply -f $temp_git_dir/examples/csi-storageclass.yaml && \
+    kubectl apply -f $temp_git_dir/examples/csi-volumesnapshotclass.yaml && \
+    kubectl annotate volumesnapshotclass csi-hostpath-snapclass \
+      snapshot.storage.kubernetes.io/is-default-class="true" && \
+    rm -rf $temp_git_dir
+
+  # wait for snapshot-controller to be ready
+  kubectl wait --for=condition=Available -n kube-system deploy/snapshot-controller --timeout=60s
+
 # Run e2e tests
-e2e distribution="k3s" path="./test/e2e" multinamespace="false": create-kind && delete-kind
+e2e distribution="k3s" path="./test/e2e" multinamespace="false": create-kind setup-csi-volume-snapshots && delete-kind
   echo "Execute test suites ({{ distribution }}, {{ path }}, {{ multinamespace }})"
 
   TELEMETRY_PRIVATE_KEY="" goreleaser build --snapshot --clean
--- a/test/e2e/snapshot/common.go
+++ b/test/e2e/snapshot/common.go
@@ -17,6 +17,42 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
+func createSnapshot(f *framework.Framework, useNewCommand bool, snapshotPath string, includeVolumes bool) {
+	By("Snapshot vcluster")
+	var cmd *exec.Cmd
+	if useNewCommand {
+		// snapshots created asynchronously by the controller
+		args := []string{
+			"snapshot",
+			"create",
+			f.VClusterName,
+			snapshotPath,
+			"-n", f.VClusterNamespace,
+		}
+		if includeVolumes {
+			args = append(args, "--include-volumes")
+		}
+		cmd = exec.Command(
+			"vcluster",
+			args...,
+		)
+	} else {
+		// snapshots created synchronously by the CLI
+		cmd = exec.Command(
+			"vcluster",
+			"snapshot",
+			f.VClusterName,
+			snapshotPath,
+			"-n", f.VClusterNamespace,
+			"--pod-mount", "pvc:snapshot-pvc:/snapshot-pvc",
+		)
+	}
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err := cmd.Run()
+	framework.ExpectNoError(err)
+}
+
 func restoreVCluster(f *framework.Framework, snapshotPath string, controllerBasedSnapshot, restoreVolumes bool) {
 	By("Restore vCluster")
 	restoreArgs := []string{
--- a/test/e2e/snapshot/common.go
+++ b/test/e2e/snapshot/common.go
@@ -1,0 +1,140 @@
+package snapshot
+
+import (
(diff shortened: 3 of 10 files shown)


.github/workflows/e2e.yaml
```diff
@@ -318,6 +318,28 @@ jobs:
...
+          set -x
+
+          echo "Deploy upstream CSI volume snapshot CRDs and snapshot-controller"
+          kubectl kustomize https://github.com/kubernetes-csi/external-snapshotter/client/config/crd | kubectl create -f -
+          kubectl kustomize https://github.com/kubernetes-csi/external-snapshotter/deploy/kubernetes/snapshot-controller | kubectl create -f -
+
+          echo "Deploy CSI driver"
+          temp_git_dir=$(mktemp -d)
+          git clone https://github.com/kubernetes-csi/csi-driver-host-path.git $temp_git_dir
+          $temp_git_dir/deploy/kubernetes-latest/deploy.sh
```
lizardruss: is the clone necessary, or could we use the raw url in the commands below?

example:
```shell
kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/csi-driver-host-path/refs/heads/master/examples/csi-storageclass.yaml
kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/csi-driver-host-path/refs/heads/master/examples/csi-volumesnapshotclass.yaml
```

Justfile
```diff
@@ -86,8 +86,26 @@ embed-chart version="0.0.0":
...
   helm unittest chart
 
+setup-csi-volume-snapshots:
+  # Deploy upstream CSI volume snapshot CRDs and snapshot-controller
+  kubectl kustomize https://github.com/kubernetes-csi/external-snapshotter/client/config/crd | kubectl create -f -
+  kubectl kustomize https://github.com/kubernetes-csi/external-snapshotter/deploy/kubernetes/snapshot-controller | kubectl create -f -
+
+  # Deploy CSI driver, StorageClass and VolumeSnapshotClass
+  temp_git_dir=$(mktemp -d) && \
+    git clone https://github.com/kubernetes-csi/csi-driver-host-path.git $temp_git_dir && \
```
lizardruss, line 96: similar question as the workflow file

test/e2e/snapshot/common.go
```diff
@@ -0,0 +1,176 @@
...
+					rp = rp + 1
+				}
+			}
+			if rp == p {
+				return true, nil
+			}
+		}
+		return false, nil
+	})
+	framework.ExpectNoError(err)
```
lizardruss: why use `wait.PollUntilContextTimeout` instead of `Eventually`?

test/e2e/snapshot/common.go
```diff
@@ -0,0 +1,176 @@
...
+			for _, container := range pod.Status.ContainerStatuses {
+				if container.State.Running == nil || !container.Ready {
+					return fmt.Errorf("pod %s container %s is not running", pod.Name, container.Name)
+				}
+			}
+		}
+		return nil
+	}).WithPolling(time.Second).
+		WithTimeout(framework.PollTimeout).
+		Should(Succeed())
```
lizardruss: using `framework.ExpectNoError(err)` inside `Eventually` will exit immediately instead of retrying. Instead, I'd suggest using a locally scoped `Gomega` to do assertions. In most cases this also helps with readability. Here's an example:
```golang
	Eventually(func(g Gomega) {
		pods, err := f.HostClient.CoreV1().Pods(f.VClusterNamespace).List(f.Context, metav1.ListOptions{
			LabelSelector: "app=vcluster,release=" + f.VClusterName,
		})
		g.Expect(err).NotTo(HaveOccurred())

		for _, pod := range pods.Items {
		    g.Expect(pod.Status.ContainerStatuses).NotTo(BeEmpty(), "pod %s has no container status", pod.Name)

			for _, container := range pod.Status.ContainerStatuses {
			    g.Expect(container.State.Running).NotTo(BeNil(), "pod %s container %s is not running", pod.Name, container.Name)
			    g.Expect(container.Ready).To(BeTrue(), "pod %s container %s is not ready", pod.Name, container.Name)
			}
		}
	}).WithPolling(time.Second).
		WithTimeout(framework.PollTimeout).
		Should(Succeed())
```

test/e2e/snapshot/common.go
```diff
@@ -0,0 +1,176 @@
...
+type unmarshalRequestFunc[T snapshot.LongRunningRequest] func(request *corev1.ConfigMap) (T, error)
+
+func waitForRequestToFinish[T snapshot.LongRunningRequest](f *framework.Framework, requestLabel string, unmarshal unmarshalRequestFunc[T], timeout time.Duration) {
+	Eventually(func() error {
+		listOptions := metav1.ListOptions{
+			LabelSelector: requestLabel,
+		}
+		requestConfigMaps, err := f.HostClient.CoreV1().ConfigMaps(f.VClusterNamespace).List(f.Context, listOptions)
+		framework.ExpectNoError(err)
+		Expect(requestConfigMaps.Items).To(HaveLen(1))
```
lizardruss: `Expect` here has a similar issue as `framework.ExpectNoError(err)`, since it's essentially a global panic, so similar suggestion would be to use `Eventually(func (g Gomega) {})` instead.

.github/workflows/e2e.yaml
```diff
@@ -318,6 +318,28 @@
...
+        run: |
+          set -x
+
+          echo "Deploy upstream CSI volume snapshot CRDs and snapshot-controller"
+          kubectl kustomize https://github.com/kubernetes-csi/external-snapshotter/client/config/crd | kubectl create -f -
+          kubectl kustomize https://github.com/kubernetes-csi/external-snapshotter/deploy/kubernetes/snapshot-controller | kubectl create -f -
+
+          echo "Deploy CSI driver"
+          temp_git_dir=$(mktemp -d)
+          git clone https://github.com/kubernetes-csi/csi-driver-host-path.git $temp_git_dir
```
Piotr1215: Let's make actionlint happy and quote the variable
```suggestion
git clone https://github.com/kubernetes-csi/csi-driver-host-path.git "$temp_git_dir"
```

.github/workflows/e2e.yaml
```diff
@@ -318,6 +318,28 @@
...
+          $temp_git_dir/deploy/kubernetes-latest/deploy.sh
```
Piotr1215: Let's make actionlint happy and quote the variable
```suggestion
"$temp_git_dir/deploy/kubernetes-latest/deploy.sh"
```

.github/workflows/e2e.yaml
```diff
@@ -318,6 +318,28 @@
...
+
+          echo "Deploy StorageClass and VolumeSnapshotClass"
+          kubectl apply -f $temp_git_dir/examples/csi-storageclass.yaml
```
Piotr1215: Let's make actionlint happy and quote the variable
```suggestion
kubectl apply -f "$temp_git_dir/examples/csi-storageclass.yaml"
```

.github/workflows/e2e.yaml
```diff
@@ -318,6 +318,28 @@
...
+          kubectl apply -f $temp_git_dir/examples/csi-volumesnapshotclass.yaml
```
Piotr1215: Let's make actionlint happy and quote the variable
```suggestion
kubectl apply -f "$temp_git_dir/examples/csi-volumesnapshotclass.yaml"
```

.github/workflows/e2e.yaml
```diff
@@ -318,6 +318,28 @@
...
+          kubectl annotate volumesnapshotclass csi-hostpath-snapclass snapshot.storage.kubernetes.io/is-default-class="true"
+          rm -rf $temp_git_dir
```
Piotr1215: Let's make actionlint happy and quote the variable
```suggestion
rm -rf "$temp_git_dir"
```

test/e2e/snapshot/common.go
lizardruss: One more optional suggestion: If you pass a `ctx` to `Eventually`, or use `WithContext(ctx)`, you'll be able to see which assertions are failing as it's retrying. It should help with debugging things, but I think the `ctx` needs to be the one passed by Ginkgo and not `f.Context`, so to use it here would be a larger change.

.github/workflows/e2e.yaml
nprokopic: Ah, forgot to fix this here (fixed it before in vcluster-pro PR). I'll change it here (looks like above code suggestion is removing white space in the beginning of the line).

.github/workflows/e2e.yaml
nprokopic: I don't think clone can be avoided, as in the `deploy/kubernetes-latest/deploy.sh` path `kubernetes-latest` and `deploy.sh` path parts are actually symlinks, and ATM the actual script path is `deploy/util/deploy-hostpath.sh`.

I could use the above resolved script path, but my concern with that is that it might break easily (e.g. with new k8s version), since the deploy docs are using the symlink path.

What I can do here is add `--depth 1` in got clone, so it doesn't unnecessarily pull the entire revision history.

test/e2e/snapshot/common.go
nprokopic: Good question, not sure, this was used in the test before, I just moved this code into a func that you see here. I used `Eventually(...)` in most of (all?) the cases where I was adding _wait for X_ code.

I'll try `Eventually(...)` here as well, it should work.

test/e2e/snapshot/common.go
nprokopic: Thanks for the reminder to use Gomega matchers, will change to use them! `framework.ExpectNoError(err)` is still present quite a lot in the tests, and IMO tests would be cleaner with Gomega's `Expect()...` matchers.

Will look into the context usage as well.

Justfile
nprokopic: same answer :) https://github.com/loft-sh/vcluster/pull/3254/files#r2417807491

.github/workflows/e2e.yaml
lizardruss: bah, skimmed over the script. my bad.

.github/workflows/e2e.yaml
nprokopic: Fixed.

.github/workflows/e2e.yaml
nprokopic: Fixed.

.github/workflows/e2e.yaml
nprokopic: Fixed.

.github/workflows/e2e.yaml
nprokopic: Fixed.

.github/workflows/e2e.yaml
nprokopic: Fixed.

test/e2e/snapshot/common.go
nprokopic: Changed to Eventually

test/e2e/snapshot/common.go
nprokopic: Updated to use locally scoped Gomega and context passed by Gomega

test/e2e/snapshot/common.go
nprokopic: Updated to use locally scoped Gomega

lizardruss (approved):
lgtm

PR #3262, opened 2025-10-10
feat: Add support for ingress network policies at vcluster creation time

cbalan:
/kind feature

resolves #ENG-9473

Ingress network policies rules can be set at vcluster creation time, in addition to egress network policies.

---

> [!NOTE]
> Adds ingress-capable NetworkPolicies, restructures config into controlPlane/workload (with publicEgress), updates templates, schema, config, migration, and tests.
> 
> - **Helm templates (`chart/templates/networkpolicy.yaml`)**:
>   - Add `policyTypes` including `Ingress` and expand rules for workloads and control plane (DNS, control plane ports, platform access, etc.).
>   - Generate unified metadata via merged `labels`/`annotations` and apply to all policies.
>   - Introduce separate policies for `vc-kube-dns-*` (when CoreDNS not embedded) and `vc-snapshot-*`.
> - **Values & Schema (`chart/values.yaml`, `chart/values.schema.json`)**:
>   - Replace `outgoingConnections`/`extra*Rules` with structured `networkPolicy.controlPlane.{ingress,egress}` and `workload.{ingress,egress,publicEgress{enabled,cidr,except}}`.
>   - Add `networkPolicy.{labels,annotations}`; remove legacy `IPBlock`/`OutgoingConnections` defs.
> - **Go config (`config/config.go`)**:
>   - Mirror new NetworkPolicy model (controlPlane/workload/publicEgress, labels/annotations); deprecate legacy `OutgoingConnections`/`IPBlock` types.
>   - Add migration mapping old `Isolation.NetworkPolicy.OutgoingConnections` to `workload.publicEgress`.
> - **Tests (`chart/tests/networkpolicy_test.yaml`)**:
>   - Update expectations to 4 documents and new ingress/egress counts.
>   - Assert metadata propagation and support for extra ingress/egress rules.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5304fc4b84585ab57419695e186b2bd49cec05da. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/chart/tests/networkpolicy_test.yaml
+++ b/chart/tests/networkpolicy_test.yaml
@@ -75,7 +75,16 @@ tests:
     release:
       name: my-release
       namespace: my-namespace
+    chart:
+      version: 0.0.1
     set:
+      controlPlane:
+        advanced:
+          globalMetadata:
+            annotations:
+              global-annotation-key-1: global-annotation-value-1
+              global-annotation-key-2: global-annotation-value-2
+              annotation-key-to-be-overridden: annotation-key-overridden-from-global
       policies:
         networkPolicy:
           enabled: true
@@ -83,6 +92,7 @@ tests:
             label-key-1: label-value-1
             label-key-2: label-value-2
           annotations:
+            annotation-key-to-be-overridden: value-overridden-from-global
             annotation-key-1: annotation-value-1
             annotation-key-2: annotation-value-2
     asserts:
@@ -93,9 +103,16 @@ tests:
           content:
             label-key-1: label-value-1
             label-key-2: label-value-2
+            app: vcluster
+            chart: vcluster-0.0.1
+            release: my-release
+            heritage: Helm
       - isSubset:
           path: metadata.annotations
           content:
+            global-annotation-key-1: global-annotation-value-1
+            global-annotation-key-2: global-annotation-value-2
+            annotation-key-to-be-overridden: annotation-key-overridden-from-global
             annotation-key-1: annotation-value-1
             annotation-key-2: annotation-value-2
 


config/config.go
```diff
@@ -2694,28 +2694,85 @@ type NetworkPolicy struct {
...
+	// Workload network policy rules
+	Workload NetworkPolicyWorkload `json:"workload,omitempty"`
+
 	// OutgoingConnections are the outgoing connections options for the vCluster workloads.
+	// Deprecated: use policy.networkPolicy.workload.publicEgress instead.
 	OutgoingConnections OutgoingConnections `json:"outgoingConnections,omitempty"`
 
 	// ExtraControlPlaneRules are extra allowed rules for the vCluster control plane.
+	// Deprecated: use policy.networkPolicy.controlPlane.egress instead.
 	ExtraControlPlaneRules []map[string]interface{} `json:"extraControlPlaneRules,omitempty"`
```
FabianKramm: Please remove this we are fine with a breaking change

config/config.go
```diff
@@ -2694,28 +2694,85 @@ type NetworkPolicy struct {
...
 
 	// ExtraWorkloadRules are extra allowed rules for the vCluster workloads.
+	// Deprecated: use policy.networkPolicy.workload.egress instead.
 	ExtraWorkloadRules []map[string]interface{} `json:"extraWorkloadRules,omitempty"`
```
FabianKramm: Please remove this we are fine with a breaking change

config/config.go
```diff
@@ -2694,28 +2694,85 @@ type NetworkPolicy struct {
...
 
+	// ControlPlane network policy rules
+	ControlPlane NetworkPolicyControlPlane `json:"controlPlane,omitempty"`
+
+	// Workload network policy rules
+	Workload NetworkPolicyWorkload `json:"workload,omitempty"`
+
 	// OutgoingConnections are the outgoing connections options for the vCluster workloads.
+	// Deprecated: use policy.networkPolicy.workload.publicEgress instead.
 	OutgoingConnections OutgoingConnections `json:"outgoingConnections,omitempty"`
```
FabianKramm: Please remove this we are fine with a breaking change

chart/tests/networkpolicy_test.yaml
```diff
@@ -18,106 +18,246 @@ tests:
...
       policies:
         networkPolicy:
           enabled: true
-          extraControlPlaneRules:
-            - ports:
-              - port: 443
-              - port: 8443
-              - port: 6443
+          controlPlane:
+            denyAllByDefault: false
```
FabianKramm: Whats the use case for this one?

chart/tests/networkpolicy_test.yaml
```diff
@@ -18,106 +18,246 @@ tests:
...
+            allowPlatform: false
+            allowCommonHostApiServer: false
```
FabianKramm: This breaks the vCluster whats the use case here?

chart/tests/networkpolicy_test.yaml
```diff
@@ -18,106 +18,246 @@ tests:
...
+          workload:
+            allowVirtualCluster: false
```
FabianKramm: This breaks the vCluster whats the use case here?

cbalan: The assumption was that `workload.allowVirtualCluster: false` would be used in combination with `sync.toHost.networkPolicies.enabled: true`, to allow workload network policy validation. 

As far as I understand, synced to host Virtual cluster network policies would be ignored if pod-to-pod communication is always allowed.

chart/tests/networkpolicy_test.yaml
cbalan: `denyAllByDefault:true` denies traffic for all pods in the current namespace, if no other allow policy was matched. This is mainly useful for virtual cluster e2e tests/development, to validate that all pods are covered by the current policies.

This flag is disabled by default.

Advanced users might use this flag to enforce network isolation, if other pods are created next to the virtual cluster control plane(ex. via experimental.deploy.host.manifests or platform host namespace resources).

chart/tests/networkpolicy_test.yaml
cbalan: The original motivation for the `allow*` flags was to enable virtual clusters advanced users to fine tune the default policies, by disabling some of the convenience policies and define custom rules via the `ingress/egress` sections. 

For this iteration, in order to reduce the number of moving parts, we can remove most of these flags. 
Based on user feedback, we may reconsider adding these back on need basis.

chart/tests/networkpolicy_test.yaml
cbalan: The allowVirtualCluster flag was removed.

chart/tests/networkpolicy_test.yaml
cbalan: The allowCommonHostApiServer flag was removed.

chart/tests/networkpolicy_test.yaml
cbalan: The denyAllByDefault flag was removed. For the dev/e2e use cases, the default deny all policy can be added as part of the setup stage.

config/config.go
cbalan: OutgoingConnections field was removed from the policies.networkPolicy object. 

The config.OutgoingConnections type was still kept, as it is used by the legacyconfig package. To be removed via  future clean up work.

chart/templates/networkpolicy.yaml
```diff
@@ -1,108 +1,251 @@
 {{- if .Values.policies.networkPolicy.enabled }}
+{{- $labels := merge dict (dict
+      "app" "vcluster"
+      "chart" (include "vcluster.version.label" $)
+      "release" .Release.Name
+      "heritage" .Release.Service) .Values.policies.networkPolicy.labels -}}
+{{- $annotations := merge dict .Values.controlPlane.advanced.globalMetadata.annotations .Values.policies.networkPolicy.annotations -}}
+{{- $labelsAndAnnotations := dict "labels" $labels "annotations" $annotations -}}
+---
```
FabianKramm, line 9: Can we add a test for this?

chart/templates/networkpolicy.yaml
```diff
@@ -1,108 +1,251 @@
...
 spec:
   podSelector:
     matchLabels:
       vcluster.loft.sh/managed-by: {{ .Release.Name }}
+  policyTypes:
+    - Egress
+    - Ingress
   egress:
-    # Allows outgoing connections to the vcluster control plane
+    # Allow egress to vcluster DNS and control plane.
```
FabianKramm, line 24: So as far as I understood we do the following:

Egress:
* Allow dns / control plane
* Allow workloads
* Allow public egress

Ingress:
* Allow control plane
* Allow workloads

Why do we not need dns and public for ingress?

chart/templates/networkpolicy.yaml
```diff
@@ -1,108 +1,251 @@
...
     matchLabels:
       release: {{ .Release.Name }}
+  policyTypes:
+    - Egress
+    - Ingress
   egress:
-    # Allows outgoing connections to all pods with
-    # port 443, 8443 or 6443. This is needed for host Kubernetes
-    # access
+    # Allow egress to host kube-dns.
```
FabianKramm, line 87: Same question here:

Egress:
* Kube-dns
* Kube-api
* vCluster peers
* vCluster Platform
* vCluster workloads

Ingress:
* vCluster peers
* vCluster workloads
* vCluster snapshot
* vCluster platform

Why is it different between egress and ingress?

chart/templates/networkpolicy.yaml
```diff
@@ -1,108 +1,251 @@
...
+
+{{- if and .Values.controlPlane.coredns.enabled (not .Values.controlPlane.coredns.embedded) }}
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: vc-kube-dns-{{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
+{{ toYaml $labelsAndAnnotations | indent 2 }}
+spec:
```
FabianKramm, line 173: Why does this need to be separate?

chart/templates/networkpolicy.yaml
```diff
@@ -1,108 +1,251 @@
...
+{{- end }}
+
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: vc-snapshot-{{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
+{{ toYaml $labelsAndAnnotations | indent 2 }}
+spec:
```
FabianKramm, line 220: Why does this need to be separate?

chart/templates/networkpolicy.yaml
cbalan: dns and public egress rules are not required on the ingress side as reply traffic is implicitly allowed for the already covered egress traffic.

chart/templates/networkpolicy.yaml
cbalan: In addition to the rules defined by the vc-work policy, which would match due to the managed-by label, vcluster-kube-dns requires access to the host kube-dns service. This is the reason why this rule is separated. 

Given its importance, the critical access rules are also defined, just in case the vc-work rules are updated. ie. dropping workload to workload rule.

chart/templates/networkpolicy.yaml
cbalan: vc-snapshot-* pods have only the `app` label, which is not enough to allow the vc-cp rule to match it. Maybe we follow up on a different thread on the best way to label control plane related pods. 

Today, it is a little awkward to set the release label on the vc-snapshot-* pods.

cbalan:
bugbot run

PR #3264, opened 2025-10-10
reverts: set patches

zerbitx:
/kind bugfix

Fixed an issue where vcluster defining both an `expression` and `reverseExpression` on a sync patch would cause an infinite back and forth or errors.  ⚠️ Also reverts the functionality of creating non-existent keys through patch definitions.

### DIFF
--- a/pkg/syncer/from_host_syncer.go
+++ b/pkg/syncer/from_host_syncer.go
@@ -117,7 +117,7 @@ func (s *genericFromHostSyncer) SyncToVirtual(ctx *synccontext.SyncContext, even
 		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 	}
 
-	err = pro.ApplyPatchesVirtualObject(ctx, vObj, event.Host, s.GetProPatches(ctx.Config.Config), false)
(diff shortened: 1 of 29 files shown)


pkg/syncer/from_host_syncer.go
```diff
@@ -117,7 +117,7 @@ func (s *genericFromHostSyncer) SyncToVirtual(ctx *synccontext.SyncContext, even
 		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 	}
 
-	err = pro.ApplyPatchesVirtualObject(ctx, vObj, event.Host, s.GetProPatches(ctx.Config.Config), false)
+	err = pro.ApplyPatchesVirtualObject(ctx, nil, vObj, event.Host, s.GetProPatches(ctx.Config.Config), false)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
```
cbron, line 123: What about everything from util package ? https://github.com/loft-sh/vcluster/pull/3186/files#diff-5b45287ef380139c50dc991df682e835f5bad4c3c5ea91a32838803fbfc51737R137

PR #3266, opened 2025-10-10
fix: Restore passing the "beforeObj" to revert set patches

zerbitx:
/kind bugfix

### DIFF
--- a/pkg/syncer/from_host_syncer.go
+++ b/pkg/syncer/from_host_syncer.go
@@ -117,7 +117,7 @@ func (s *genericFromHostSyncer) SyncToVirtual(ctx *synccontext.SyncContext, even
 		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 	}
 
-	err = pro.ApplyPatchesVirtualObject(ctx, vObj, event.Host, s.GetProPatches(ctx.Config.Config), false)
+	err = pro.ApplyPatchesVirtualObject(ctx, nil, vObj, event.Host, s.GetProPatches(ctx.Config.Config), false)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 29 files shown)


pkg/syncer/from_host_syncer.go
```diff
@@ -117,7 +117,7 @@ func (s *genericFromHostSyncer) SyncToVirtual(ctx *synccontext.SyncContext, even
 		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 	}
 
-	err = pro.ApplyPatchesVirtualObject(ctx, vObj, event.Host, s.GetProPatches(ctx.Config.Config), false)
+	err = pro.ApplyPatchesVirtualObject(ctx, nil, vObj, event.Host, s.GetProPatches(ctx.Config.Config), false)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
```
cbron, line 123: There are some other changes from [original PR](https://github.com/loft-sh/vcluster/pull/3186/files) not included here. Specifically patch_test.go around `"new-key\` and then functions in patch.go. Are those unrelated enough to leave in ? The test one appears to be setting a new key.

zerbitx: Those weren't part of the issue with `reverseExpressions`.  Leaving in the [isBracketEnclosed](https://github.com/loft-sh/vcluster/pull/3186/files?w=1#diff-5b45287ef380139c50dc991df682e835f5bad4c3c5ea91a32838803fbfc51737R137) parts allows it to continue to parse paths like `metadata.annotations["test"]` instead of `metadata.annotations.test`.  The bug was in how the state was compared and the patches were applied on the pro side in `translate.go` and restoring that comparison depends on re-adding the beforeObj parameter to the `ApplyPatchesHostObject` and `ApplyPatchesVirtualObject` methods.

PR #3272, opened 2025-10-16
feat: Add --external for platform add

cbalan:
/kind feature

resolves #ENG-8877

Setting the --external flag to false allows the platform to take over the vcluster instance management.


FabianKramm (changes requested):
I don't think this will work, have you tested this? If you set this as a non-external you will need to provider cluster name, namespace and virtual cluster name

cbalan:
I tested that the flag was reaching the api. Indeed I have not looked pass that. Thank you. I'll follow up on this.

cbalan:
In my test, I validated that this change would be equivalent with the steps described on https://www.vcluster.com/docs/platform/use-platform/virtual-clusters/add-virtual-clusters#post-deployment-platform-registration

Which is lacking context regarding the actual vcluster instance host cluster. Just by following these steps, a new vcluster will be created using the default platform host cluster. 

I'll add the additional flags. and follow up on the docs as well.

PR #3275, opened 2025-10-19
Clean up volume snapshot resources after taking snapshot

nprokopic:
/kind enhancement

resolves ENG-9271

Delete VolumeSnapshot and VolumeSnapshotContent resources after snapshot and restore.

VolumeSnapshotContent is deleted with `spec.deletionPolicy=Retain`, so the volume snapshot is retained in the storage backend.

### DIFF
--- a/pkg/snapshot/volumes/csi/restoreinprogress.go
+++ b/pkg/snapshot/volumes/csi/restoreinprogress.go
@@ -72,6 +72,23 @@ func (r *Restorer) reconcileInProgress(ctx context.Context, requestObj runtime.O
 		case volumes.RequestPhaseCompletedCleaningUp:
 			fallthrough
 		case volumes.RequestPhaseFailedCleaningUp:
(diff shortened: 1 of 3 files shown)


pkg/snapshot/volumes/csi/restoreinprogress.go
```diff
@@ -59,15 +60,54 @@ func (r *Restorer) reconcileInProgress(ctx context.Context, requestObj runtime.O
...
 				hasInProgressRestores = true
 				continue
-			case volumes.RequestPhaseFailed:
-				failedRestoresCount++
-			case volumes.RequestPhaseCompleted:
-				hasCompletedRestores = true
+			case volumes.RequestPhaseCompletedCleaningUp:
+				fallthrough
+			case volumes.RequestPhaseFailedCleaningUp:
+				cleaningUpSnapshots = true
```
jjaferson, line 66: Why do we set `cleaningUpSnapshots` to true when it fails? And what is the overall status when the deletion of volume fails

PR #3279, opened 2025-10-20
skip restoring pvc(s) with failed snapshot

neogopher:
/kind bugfix

resolves ENG-9341
ENG-9645

### DIFF
--- a/pkg/snapshot/restoreclient.go
+++ b/pkg/snapshot/restoreclient.go
@@ -24,6 +24,7 @@ import (
 	"github.com/loft-sh/vcluster/pkg/scheme"
 	setupconfig "github.com/loft-sh/vcluster/pkg/setup/config"
 	"github.com/loft-sh/vcluster/pkg/snapshot/volumes"
+	"github.com/loft-sh/vcluster/pkg/util/translate"
 	"go.etcd.io/etcd/server/v3/storage/backend"
 	"go.etcd.io/etcd/server/v3/storage/mvcc"
 	"go.etcd.io/etcd/server/v3/storage/schema"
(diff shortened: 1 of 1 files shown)


pkg/snapshot/restoreclient.go
```diff
@@ -314,28 +315,42 @@ func (o *RestoreClient) createRestoreRequest(ctx context.Context, vConfig *confi
...
+			volumeName := strings.TrimPrefix(key, pvPrefix)
+			for _, snapshotSpec := range o.snapshotRequest.Spec.VolumeSnapshots.Requests {
+				if snapshotSpec.PersistentVolumeClaim.Spec.VolumeName == volumeName {
+					return true
+				}
+			}
+		}
+	} else {
 		// Restore skips restoring PVs and PVCs only in private mode.
 		// In the shared mode, PVs and PVCs are restored in virtual cluster, and volumes are restored in the host
```
nprokopic: This comment should now be updated (part of it should also be moved to the if block for private nodes), because it changed for shared nodes, as now with shared nodes it also skips PVCs for which the snapshots have failed.

pkg/snapshot/restoreclient.go
```diff
@@ -354,7 +369,11 @@ func (o *RestoreClient) isPVCThatShouldBeRestoredInHost(key string) bool {
 	}
 
 	pvcName := strings.TrimPrefix(key, pvcPrefix)
-	status, ok := o.snapshotRequest.Status.VolumeSnapshots.Snapshots[pvcName]
+	translatedPVCName := getTranslatedPVCName(pvcName)
```
nprokopic: Currently `isPVCThatShouldBeRestoredInHost` func is called in the wrong place/moment, i.e. even when volume restore is not enabled, as `o.RestoreVolumes` check is missing here, so I'm afraid that fixing this might have even worse unintended side effected 🙈

Did you notice some bug so this fix was needed?

Could you add also `o.RestoreVolumes` check here in the beginning? Or maybe not change this if not really required for fixing ENG-9341 (skipping PVCs for failed snapshots).

pkg/snapshot/restoreclient.go
```diff
@@ -314,28 +315,42 @@ func (o *RestoreClient) createRestoreRequest(ctx context.Context, vConfig *confi
 }
 
 func (o *RestoreClient) skipKey(key string) bool {
-	if !o.vConfig.PrivateNodes.Enabled {
+	if o.vConfig.PrivateNodes.Enabled {
```
nprokopic, line 322: Since we're skipping PVCs/PVs only when restoring volumes, can you please add the missing `o.RestoreVolumes` check here in the beginning of func?

Something like:
```
if !o.RestoreVolumes {
    return false
}
```

pkg/snapshot/restoreclient.go
neogopher: > Currently isPVCThatShouldBeRestoredInHost func is called in the wrong place/moment, i.e. even when volume restore is not enabled, as o.RestoreVolumes check is missing here, so I'm afraid that fixing this might have even worse unintended side effected

Oh ok. 

> Did you notice some bug so this fix was needed?

Yes, with shared nodes, pvcName variable contains the name of the pvc as it appears inside the virtual cluster. However, the keys under `o.snapshotRequest.Status.VolumeSnapshots.Snapshots` are in the translated format as on the host cluster. Hence, with shared nodes, it never matches and `ok` is always false, thus returning with false on line 350. i.e. it never skips pvc restore no matter the status.

> Could you add also o.RestoreVolumes check here in the beginning? Or maybe not change this if not really required for fixing ENG-9341 (skipping PVCs for failed snapshots).

Sure, I will retest now if removing all my recent changes from only the `isPVCThatShouldBeRestoredInHost` func still works to skip the failed PVCs. I think it should, but I will confirm.

pkg/snapshot/restoreclient.go
neogopher: sure. Will make these changes.

pkg/snapshot/restoreclient.go
neogopher: 👍

pkg/snapshot/restoreclient.go
neogopher: Just confirmed by removing changes under `isPVCThatShouldBeRestoredInHost` function and it was successful.

PR #3286, opened 2025-10-21
ENG-9633 Add controlPlane.advanced.kubeVip

squ94wk:
/kind feature

resolves #ENG-9633

Add config for kube-vip layer 2 endpoint IP announcement.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1799,10 +1799,6 @@ type KubeVip struct {
 	Gateway string `json:"gateway,omitempty"`
 }
 
-func (k KubeVip) JSONSchemaExtend(base *jsonschema.Schema) {
-	addProToJSONSchema(base, reflect.TypeOf(k))
-}
-
 type ControlPlaneStatefulSet struct {
 	// HighAvailability holds options related to high availability.
 	HighAvailability ControlPlaneHighAvailability `json:"highAvailability,omitempty"`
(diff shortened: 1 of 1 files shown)


chart/values.yaml
```diff
@@ -793,6 +793,10 @@ controlPlane:
       anonymousPull: true
       # Config is the regular docker registry config. See https://distribution.github.io/distribution/about/configuration/ for more details.
       config: {}
+    # KubeVip holds configuration for embedded kube-vip that announces the virtual cluster endpoint IP on layer 2.
+    kubeVip:
+      # Enabled defines if embedded kube-vip should be enabled. This requires a pro license.
```
cbron: Won't require pro license

cmd/vcluster/cmd/start.go
```diff
@@ -177,6 +177,13 @@ func StartInCluster(ctx context.Context, options *StartOptions) error {
 		}
 	}
 
+	// should start embedded kube-vip?
+	if vConfig.ControlPlane.Advanced.KubeVip.Enabled {
```
FabianKramm, line 181: Only start this if privateNodes is true as well

FabianKramm (changes requested):

chart/templates/statefulset.yaml
```diff
@@ -171,9 +171,20 @@ spec:
...
+            {{- $kubeVipSecContext = dict "capabilities" (dict "add" (list "NET_ADMIN" "NET_RAW")) }}
+            {{- end }}
+            {{- with merge (dict) $kubeVipSecContext .Values.controlPlane.statefulSet.security.containerSecurityContext }}
+              {{- if .capabilities }}
+                {{- if .capabilities.add }}
+                {{- $_ := set .capabilities "add" (uniq .capabilities.add) }}
+                {{- end }}
+              {{- end }}
+{{ toYaml . | indent 12 }}
+            {{- end }}
```
FabianKramm, line 183: Would be great if we add a simple helm unit test for this just to make sure this is set when kubeVip is enabled and not set when its disabled

PR #3293, opened 2025-10-22
Add snapshot Deleting phase

nprokopic:
/kind enhancement

resolves ENG-9335, ENG-9602

Add snapshot Deleting phase which is used to delete volume snapshots.

vcluster (binary) snapshot delete command recreates a snapshot request in Deleting phase (@jjaferson is working on that part here #3290), and then the vCluster snapshot controller reconciles the snapshot deletion.

TODO:
- [x] Delete volume snapshots
- [x] Delete snapshot tarball
- [x] Add e2e tests

### DIFF
--- a/pkg/snapshot/volumes/csi/snapshotdeleting.go
+++ b/pkg/snapshot/volumes/csi/snapshotdeleting.go
@@ -46,64 +46,51 @@ func (s *VolumeSnapshotter) reconcileDeleting(ctx context.Context, requestObj ru
 		}
 
 		volumeSnapshotName := fmt.Sprintf("%s-%s", volumeSnapshotRequest.PersistentVolumeClaim.Name, requestName)
-		if volumeSnapshotStatus.IsVolumeSnapshotMaybeCreated() {
-			// The volume snapshot could have been created, the deletion has not been started, so
-			// trigger deletion here.
-			deleted, err := s.deleteVolumeSnapshot(
+		if volumeSnapshotStatus.IsVolumeSnapshotMaybeCreated() || volumeSnapshotStatus.IsDeletingVolumeSnapshot() {
+			// Re-create VolumeSnapshot and VolumeSnapshotContent resources if the following conditions are met:
+			// - this is the Deletion request
+			// - volume snapshot is already being deleted (because then, if VolumeSnapshot and VolumeSnapshotContent
+			//   resources are not found, it means that they have been already deleted)
+			recreateResourceIfNotFound :=
+				status.RecreateVolumeSnapshotsWhenDeleting() &&
+					!volumeSnapshotStatus.IsDeletingVolumeSnapshot()
+			deletedResources, err := s.deleteVolumeSnapshot(
 				ctx,
 				constants.SnapshotRequestLabel,
 				requestName,
 				volumeSnapshotRequest,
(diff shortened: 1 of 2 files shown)


pkg/snapshot/volumes/request.go
```diff
@@ -142,3 +156,22 @@ func (s SnapshotStatus) Done() bool {
...
+
+func (s SnapshotStatus) IsDeletingVolumeSnapshot() bool {
+	return s.Phase == RequestPhaseDeleting || s.Phase == RequestPhaseCanceling
+}
+
+func (s SnapshotStatus) RecreateVolumeSnapshotWhenDeleting() bool {
+	return s.Phase == RequestPhaseDeleting
+}
+
+func (s SnapshotStatus) IsVolumeSnapshotMaybeCreated() bool {
```
jjaferson, line 168: nit: I would just call it `IsVolumeSnapshotCreated` specially because it returns a boolean so it's either VolumeSnapshot was created or not there no "maybe"

nprokopic: Actually there is a "maybe". The above is important for canceling volume snapshots, where the snapshot request can switch from InProgress to Canceling, so if the phase is InProgress for some volume, it depends at which point in time of InProgress phase you do the check.

For example:
- volume snapshot moves to InProgress
- controller creates VolumeSnapshot resource
- controller returns requeues in 30 seconds, because VolumeSnapshot resource has just been created, so it's being reconciled by the upstream snapshot-controller and CSI driver
- the above finishes after 10 seconds, so volume snapshots have been created
- snapshot request is still in InProgress phase, because the next reconciliation has not yet happened
- snapshot request gets Canceled

In this example you want to delete the volume snapshots from the storage backend.

OTOH the snapshot request can get canceled when the volume snapshot has not yet been created, so Canceling will just delete the VolumeSnapshot resource in that case, or maybe not even that, but in any case we try.

nprokopic: when `IsVolumeSnapshotMaybeCreated()` returns false, it means that the volume snapshot has definitely not been created

nprokopic:
> LGTM
> 
> I'll approve once the tarball deletion is added

Added it, tested locally, looks good ✅

nprokopic:
Snapshot canceling test is failing. After looking into it, it looks like it fails because Cancleing is still in progress. So it either needs a longer Eventually timeout, or VolumeSnapshot resource is stuck in deletion for some reason.

Added detailed test logging, so it's clear what is the issue.

nprokopic:
> Snapshot canceling test is failing. After looking into it, it looks like it fails because Cancleing is still in progress. So it either needs a longer Eventually timeout, or VolumeSnapshot resource is stuck in deletion for some reason.
> 
> Added detailed test logging, so it's clear what is the issue.

There was a bug in the snapshot canceling e2e test.

jjaferson:
@nprokopic do you wanna add my [PR](https://github.com/loft-sh/vcluster/pull/3290/files) as part of your change and we can merge it all together in one PR?

nprokopic:
- Fixed canceling bug
- Fix a bug in the restore test (test bug, not a restore bug)

Now the test uncovered a bug in snapshot deletion. Testing fix locally.

cbron (approved):

PR #3302, opened 2025-10-23
Fix race condition in restore volumes e2e test

nprokopic:
/kind test

There is a race condition bug that sometimes happens, which makes the test to fail. The restore happens too quickly after the PVC has been deleted, so the PVC syncer probably doesn't have to fully handle the PVC deletion.


cbron (approved):
Should reconsider this long term

nprokopic:
> Should reconsider this long term

Yeah, definitely. Long term we should try to improve how the syncer works. And if we cannot improve the syncer, we should try to understand why the syncer triggers the additional deletion, and then adjust the test here according to that (so it doesn't wait fix time, but it actually waits for \*some condition\* to be met)

PR #3306, opened 2025-10-23
fix: snapshot delete request to deleting phase

jjaferson:
/kind bugfix
/kind enhancement
/kind feature
/kind documentation
/kind test

resolves #

### DIFF
--- a/pkg/snapshot/request.go
+++ b/pkg/snapshot/request.go
@@ -373,6 +373,7 @@ func DeleteSnapshotRequestResources(ctx context.Context, vClusterNamespace, vClu
 	}
 
 	// update the snapshot request status to indicate that the snapshot request is in the cleaning up phase
+	savedSnapshotRequest.CreationTimestamp = metav1.Now()
 	savedSnapshotRequest.Status.Phase = RequestPhase(volumes.RequestPhaseDeleting)
 
 	// first create the snapshot options Secret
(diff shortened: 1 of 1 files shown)


pkg/snapshot/request.go
nprokopic, line 391: Before this line you should also update `savedSnapshotRequest.Name` to `secret.Name`

nprokopic: also reset the creation timestamp `savedSnapshotRequest.CreationTimestamp = metav1.Now()`

nprokopic: so the snapshot request that gets saved into the ConfigMap has correct name and timestamp

jjaferson: > Before this line you should also update savedSnapshotRequest.Name to secret.Name

Should we rename the snapshot request? The secret and ConfigMap are already renamed but I thought the original snapshot request name should remain the same?

nprokopic: The name of the serialized snapshot request should match the ConfigMap and Secret, as they are all originally created with the same name. Not sure what would be the consequence if the serialized snapshot request (in ConfigMap.data) has a different name than the ConfigMap itself, but something might break.

nprokopic: Forgot now why the name was required in the snapshot request struct, will see if we can remove it entirely and just use ConfigMap name (might be a bigger code change, so I wouldn't do that these days 🙈 ).

jjaferson: Sure, I'll push the changes then

nprokopic (approved):

nprokopic:
The failing test should be fixed in another PR

cbron:
Hard merging as failure is around test that _should_ be fixed already, and this PR just wasn't rebased.

PR #3326, opened 2025-10-29
chore: extend the test framework to manage host kubeconfig file

roehrijn:
/kind test

in order to create a Helm client to deploy something to the host cluster within a test for example, it is useful to have access to the host kubeconfig file through framework.

### DIFF
--- a/test/framework/framework.go
+++ b/test/framework/framework.go
@@ -2,6 +2,7 @@ package framework
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


test/framework/framework.go
```diff
@@ -65,6 +66,11 @@ type Framework struct {
 	// host kubernetes cluster were we are testing in
 	HostCRClient client.Client
 
+	// HostKubeConfigFile is a file containing kube config
+	// of the current host kubernetes cluster were we are testing in.
+	// This file shall be deleted in the end of the test suite execution.
+	HostKubeConfigFile *os.File
```
zerbitx, line 72: nit: This could just be a string of the path since the file is closed right after opening it, before adding it to the struct and only the `.Name()` is used in the cleanup.

roehrijn: agree. But just did it like that because we already have https://github.com/loft-sh/vcluster/blob/2addef3573850d5caff5e55a5ac90832004406c6/test/framework/framework.go#L86-L89

PR #3373, opened 2025-11-26
add a fallback command for checking helm binary version (v4.x)

neogopher:
/kind bugfix

resolves ENG-10292

Fixed an issue where vcluster CLI would not work with Helm v4

---

> [!NOTE]
> Add fallback Helm version check without --client to ensure compatibility with Helm v4.
> 
> - **CLI**:
>   - Update Helm version detection in `pkg/cli/create_helm.go` to retry without `--client` when the initial `helm version --client` call fails, adding debug logging for the fallback (supports Helm v4).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6f32f065a88f68a91a38d4fbf2498675612a5711. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/cli/create_helm.go
+++ b/pkg/cli/create_helm.go
@@ -139,6 +139,8 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 
 	output, err := exec.Command(helmBinaryPath, "version", "--client", "--template", "{{.Version}}").Output()
 	if err != nil {
+		log.Debugf("error getting helm version: '%v'. Attempting again with commandline flags compatible with Helm v4.x..", err)
+
 		// Helm v4.x does not support the --client flag
 		output, err = exec.Command(helmBinaryPath, "version", "--template", "{{.Version}}").Output()
 		if err != nil {


pkg/cli/create_helm.go
```diff
@@ -139,7 +139,11 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 
 	output, err := exec.Command(helmBinaryPath, "version", "--client", "--template", "{{.Version}}").Output()
 	if err != nil {
```
hidalgopl, line 141: is there a way to check for this particular error here (helm v4.x not supporting client flag) and fallback only on it?

neogopher: When run with helmv4, the attempt on line 140 fails with an "exit status 1" err.
I can match against that string, but again it won't be very specific even in that case.

hidalgopl: I see, then maybe let's add `log.Debug` with `err` from the command with `--client` and message saying that we try helm v4 version of the command. Otherwise it can be hard to debug in case if it fails for any other reason than being v4

neogopher: Yes, that's a really good idea. Done!!

hidalgopl (approved):

PR #3386, opened 2025-12-04
fix: returning an error when vcluster config schema changed

jjaferson:
/kind bugfix

resolves ENG-9408

Fixed an issue where vcluster was not being deleted due to the schema difference between vcluster v0.29 and v0.30

---

> [!NOTE]
> Switch deletion to parse Helm values dynamically and check `sync.toHost.namespaces.enabled` via `unstructured` to conditionally run namespace cleanup without relying on old config schema.
> 
> - **CLI (vcluster deletion)**:
>   - Parse Helm values into a generic `map[string]any` and use `unstructured.NestedBool` to read `sync.toHost.namespaces.enabled`.
>   - Gate `CleanupSyncedNamespaces(...)` on the detected flag, defaulting to disabled on parse errors.
>   - Remove dependency on `config.Config` schema and import; add `unstructured` import and debug logs for parse/lookup errors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit db2244a04eade97ac74742259435cd80949798dc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/cli/delete_helm.go
+++ b/pkg/cli/delete_helm.go
@@ -10,7 +10,6 @@ import (
 	"github.com/ghodss/yaml"
 	managementv1 "github.com/loft-sh/api/v4/pkg/apis/management/v1"
 	"github.com/loft-sh/log"
-	"github.com/loft-sh/vcluster/config"
 	"github.com/loft-sh/vcluster/pkg/cli/find"
 	"github.com/loft-sh/vcluster/pkg/cli/flags"
 	"github.com/loft-sh/vcluster/pkg/cli/localkubernetes"
@@ -23,6 +22,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -127,10 +127,17 @@ func DeleteHelm(ctx context.Context, platformClient platform.Client, options *De
 		return err
 	}
 
-	vclusterConfig := &config.Config{}
-	err = yaml.Unmarshal(values, vclusterConfig)
-	if err != nil {
-		return err
+	var configValues map[string]any
+	namespacesSyncEnabled := false
+	if err := yaml.Unmarshal(values, &configValues); err != nil {
+		cmd.log.Debugf("Error parsing vcluster config from Helm values: %v. Namespace sync will be disabled", err)
+	} else {
+		enabled, isFound, err := unstructured.NestedBool(configValues, "sync", "toHost", "namespaces", "enabled")
+		if err != nil {
+			cmd.log.Debugf("Error getting namespace sync enabled flag: %v. Namespace sync will be disabled", err)
+		} else if isFound {
+			namespacesSyncEnabled = enabled
+		}
 	}
 
 	// we have to delete the chart
@@ -203,7 +210,7 @@ func DeleteHelm(ctx context.Context, platformClient platform.Client, options *De
 	}
 
 	// if namespace sync is enabled, use cleanup handlers to handle namespace cleanup
-	if vclusterConfig.Sync.ToHost.Namespaces.Enabled {
+	if namespacesSyncEnabled {
 		if err := CleanupSyncedNamespaces(ctx, cmd.Namespace, vClusterName, cmd.restConfig, cmd.kubeClient, cmd.log); err != nil {
 			return fmt.Errorf("run namespace cleanup: %w", err)
 		}


pkg/cli/delete_helm.go
```diff
@@ -130,7 +130,9 @@ func DeleteHelm(ctx context.Context, platformClient platform.Client, options *De
 	vclusterConfig := &config.Config{}
 	err = yaml.Unmarshal(values, vclusterConfig)
 	if err != nil {
-		return err
+		cmd.log.Warnf("Failed to parse vcluster config from Helm values: %v. ", err)
+		cmd.log.Infof("Continuing with deletion...")
+		vclusterConfig = nil
```
mfranczy: This approach is error prone, the code shouldn’t continue like this, especially since we already have cleanup logic implemented for synced namespaces. Moreover, other potential configuration issues (e.g., invalid `sync.toHost` fields) would go undetected.

I recommend implementing a custom UnmarshalJSON method for [PrivateNodesAutoNodes](https://github.com/loft-sh/vcluster/blob/main/config/config.go#L142) that attempts to unmarshal the data and, if unsuccessful, returns a dedicated error (e.g., `ErrInvalidAutoNodesConfig`). Later, check  whether the error from yaml.Unmarshal indicates an invalid auto nodes configuration if so, continue as normal, otherwise return the error.

Additionally, please add a comment explaining that we tolerate an invalid auto nodes configuration in the delete action because no cleanup or other follow-up action is required.

mfranczy (changes requested):

pkg/cli/delete_helm.go
jjaferson: I agree, I was trying to just fix the issue by warning the user that the vcluster config could not be parsed but the vcluster chart would be deleted.

NVM it won't work when we backport to v0.29 version 
~~If we want we could make `PrivateNodes` backward compatible with older versions by creating a custom `UnmarshalJSON` that transform `PrivateNodes.AutoNodes` from an object to an array, instead of returning an dedicated error, what do you think?~~

mfranczy: Yes, the error is enough.

config/config.go
```diff
@@ -118,6 +119,18 @@ type PrivateNodes struct {
 	VPN PrivateNodesVPN `json:"vpn,omitempty"`
 }
 
+// UnmarshalJSON makes the schema change return a custom error for invalid private nodes config
+func (p *PrivateNodes) UnmarshalJSON(data []byte) error {
+	type PrivateNodesAlias PrivateNodes
+	var alias PrivateNodesAlias
+	if err := json.Unmarshal(data, &alias); err != nil {
```
flomedja: (question) Do we want just to validate the marshalling or a stricter control on the input keys (by using `DisallowUnknownFields` with a json decoder) ?

jjaferson: There was a change in the vcluster config schema from v0.29 to v0.30 and when the config yaml is being unmarshalled via the CLI it stops the execution and the vcluster helm chart is not deleted so the goal here is to identify that there were an error while unmarshalling vcluster config due to the schema change and warn the user instead of interrupt the execution.

pkg/cli/delete_helm.go
```diff
@@ -130,7 +131,15 @@ func DeleteHelm(ctx context.Context, platformClient platform.Client, options *De
 	vclusterConfig := &config.Config{}
 	err = yaml.Unmarshal(values, vclusterConfig)
 	if err != nil {
-		return err
+		privateNodesError := errors.Is(err, config.ErrInvalidPrivateNodesConfig) ||
+			strings.Contains(err.Error(), config.ErrInvalidPrivateNodesConfig.Error())
+		if privateNodesError {
+			cmd.log.Warnf("Failed to parse vcluster config from Helm values: %v. ", err)
+			cmd.log.Infof("Continuing with deletion...")
```
mfranczy: I'd suggest keeping only the first warning here. The second log doesn’t add much context and might confuse readers. It sounds informational but doesn’t clearly indicate that the previous step failed. A single warning is enough to convey the issue.

jjaferson: Why would it confuse users?  IMO it helps to indicate that the vcluster config parse failed but the vcluster deletion continued

config/config.go
flomedja: Thanks for the detail :)

pkg/cli/delete_helm.go
mfranczy: Because it creates unnecessary noise and breaks the one log per event principle, which is a good practice. Most core K8s components stick to this rule, in case of errors, you're informed about the next step in the same log, e.g. kubelet:

> Error updating node status, will retry: ...

or controller-runtime:

> unable to create kubeconfig file, continuing on without it

etc.

Another reason why it’s a good practice: when debugging a cluster with high verbosity, you may want to filter out only warnings and/or errors. In a similar case like this one, you would lose information about whether the deletion stopped or continued.

Is there any specific reason why you don't want to just log:
```
cmd.log.Warnf("Failed to parse vcluster config from Helm values: %v, continuing with deletion", err)
```
?

mfranczy:
@FabianKramm 

> The solution is to fix that we error out when deleting a vCluster, not sure why parsing the vCluster config there is needed anyways

The config parsing in the cli was introduced because of the [CleanUpNamespace](https://github.com/loft-sh/vcluster/blob/main/pkg/cli/cleanup_namespaces.go#L38), which depends on the `vclusterConfig.Sync.ToHost.Namespaces.Enabled` setting in the config. The config is parsed only to determine whether namespace cleanup should be executed.

If we remove the config parsing, the cleanup function would need to run unconditionally which should be fine, since the function already checks whether managed namespaces exist.

>  There is a bigger fundamental problem which is that we try to parse the config when deleting a vCluster again

I understand that you're essentially saying we should not rely on the vCluster config when managing resources from cli, but rather on the resource specs themselves, which were generated based on that config is that correct?

FabianKramm:
@mfranczy yes then lets do it like this, don't parse the vCluster config anymore or just parse the `sync.toHost.namespaces.enabled` unstrictly in a generic struct

Yes the CLI needs to be vCluster version independent and that also means independent of the vCluster config

FabianKramm (approved):

PR #3395, opened 2025-12-10
Deprecate ingress nginx

rmweir:
/kind bugfix
/kind enhancement
/kind feature
/kind documentation
/kind test

resolves #ENG10314

---

> [!NOTE]
> Marks ingress-nginx as deprecated across config/schema and switches the CLI install prompt to a deprecated message with a No default.
> 
> - **Config/Schema**:
>   - Mark `deploy.ingressNginx` as deprecated in `chart/values.schema.json` and `config/config.go`.
>   - Add deprecation comments around `deploy.ingressNginx` and ingress annotations in `chart/values.yaml` and `config/values.yaml`.
> - **CLI**:
>   - Update ingress controller install question to `[DEPRECATED]` and change default from `Yes` to `No` in `pkg/platform/clihelper/clihelper.go`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5597a691e574fffe3bc66bf41414122063e7997f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -548,9 +548,9 @@ controlPlane:
     pathType: ImplementationSpecific
     labels: {}
     annotations:
-      nginx.ingress.kubernetes.io/backend-protocol: HTTPS
-      nginx.ingress.kubernetes.io/ssl-passthrough: "true"
-      nginx.ingress.kubernetes.io/ssl-redirect: "true"
+      nginx.ingress.kubernetes.io/backend-protocol: HTTPS # ingress-nginx has been deprecated
+      nginx.ingress.kubernetes.io/ssl-passthrough: "true" # ingress-nginx has been deprecated
+      nginx.ingress.kubernetes.io/ssl-redirect: "true" # ingress-nginx has been deprecated
     # Spec allows you to configure extra ingress options.
     spec:
       tls: []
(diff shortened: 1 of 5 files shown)


chart/values.yaml
```diff
@@ -890,6 +890,7 @@ deploy:
       l2Advertisement: true
```
cbron, line 890: Need to deprecate annotations on line 522 above. 
https://github.com/loft-sh/vcluster/blob/0518cb55658e1ff8f3acffdef70af442646d0883/chart/values.yaml#L551-L553

rmweir: @cbron I think we have a process for deprecating the default values of a field. I don't think we should deprecate the field `ingress.annotations` since it applies to any ingress and could be generally useful.

cbron: Not the field, the literal `nginx.ingress` annotations: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#backend-protocol

rmweir: Yes but how do we deprecate specific annotations? I don't think we are ready to remove them.

cbron (approved):

PR #3405, opened 2025-12-15
bugfix(cli): use the helm version fallback command in the delete subcommand as well

neogopher:
/kind bugfix

resolves ENG-10647

Fixed an issue where vCluster CLI would fail with a fatal error when deleting a virtual cluster

---

> [!NOTE]
> Use `helm version --template "{{.Version}}"` (no `--client`) for version checks in both create and delete commands, removing fallback logic.
> 
> - **CLI**
>   - **Helm version detection**:
>     - Use `helm version --template "{{.Version}}"` in `pkg/cli/create_helm.go` and `pkg/cli/delete_helm.go`.
>     - Remove deprecated `--client` flag usage and associated fallback/retry logic.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5db334e21c7ddc120693e5720a6d6b254b1472d4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/cli/create_helm.go
+++ b/pkg/cli/create_helm.go
@@ -137,7 +137,7 @@ func CreateHelm(ctx context.Context, options *CreateOptions, globalFlags *flags.
 		return err
 	}
 
-	output, err := exec.Command(helmBinaryPath, "version", "--client", "--template", "{{.Version}}").Output()
+	output, err := exec.Command(helmBinaryPath, "version", "--template", "{{.Version}}").Output()
 	if err != nil {
 		return err
 	}
--- a/pkg/cli/delete_helm.go
+++ b/pkg/cli/delete_helm.go
@@ -100,7 +100,7 @@ func DeleteHelm(ctx context.Context, platformClient platform.Client, options *De
 		return err
 	}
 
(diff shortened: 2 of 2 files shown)


pkg/helm/util.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"fmt"
+	"os/exec"
+
+	"github.com/loft-sh/log"
+)
+
+func GetBinaryVersion(helmBinaryPath string, log log.Logger) ([]byte, error) {
+	output, err := exec.Command(helmBinaryPath, "version", "--client", "--template", "{{.Version}}").Output()
```
lizardruss: Looks like the `--client` has been [ignored / hidden](https://github.com/helm/helm/blob/release-3.0/cmd/helm/version.go#L64-L65) since Helm 3.0, so we could likely drop it entirely.

lizardruss (changes requested):
I think this could be simplified since the `--client` flag was probably kept for users upgrading from helm v2.

pkg/helm/util.go
neogopher: Thanks @lizardruss. Yes, then it makes sense to remove it. 
Done 👍

PR #3436, opened 2025-12-29
ENG-10923 | Add config validation for experimental custom resource proxy

janekbaraniewski:
Fixes ENG-10923

Example:

```bash
➜ vc create -n test test-config -f client-vcluster.yaml
13:39:36 warn There is a newer version of vcluster: v0.30.4. Run `vcluster upgrade` to upgrade to the newest version.

13:39:36 info Using vCluster driver 'helm' to create your virtual clusters, which means the vCluster CLI is running helm commands directly
13:39:36 info If you prefer to use the vCluster platform API instead, use the flag '--driver platform' or run 'vcluster use driver platform' to change the default
13:39:37 fatal experimental.proxy.customResources['workflows.alpha.sh/v1'].targetVirtualCluster is required
```

### DIFF
--- a/pkg/config/validation.go
+++ b/pkg/config/validation.go
@@ -893,23 +893,22 @@ func validateAdvancedControlPlaneConfig(controlPlaneAdvanced config.ControlPlane
 }
 
 func ValidateExperimentalProxyCustomResourcesConfig(cfg map[string]config.CustomResourceProxy) error {
-	if len(cfg) > 0 {
-		for resourcePath, resourceConfig := range cfg {
-			basePath := fmt.Sprintf("experimental.proxy.customResources['%s']", resourcePath)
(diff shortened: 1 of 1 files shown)


pkg/config/validation.go
```diff
@@ -891,6 +892,27 @@ func validateAdvancedControlPlaneConfig(controlPlaneAdvanced config.ControlPlane
 	return nil
 }
 
+func ValidateExperimentalProxyCustomResourcesConfig(cfg map[string]config.CustomResourceProxy) error {
+	if len(cfg) > 0 {
```
zerbitx: (nit non-blocking): The range will skip the loop anyway if the len is 0.

janekbaraniewski: haha yeah i think this was looking a bit different at some point 😄  thx, cleaned up!

PR #3452, opened 2026-01-08
Allow control plane ingress from vcluster rancher operator agent install pods

cbalan:
/kind bugfix

related #ENG-10941
related #ENG-10779

Allow vcluster rancher operator agent install job pods to connect to the virtual cluster control plane when network policies are enabled

---

> [!NOTE]
> Enables rancher operator agent install pods to reach the virtual cluster control plane when NetworkPolicies are enabled.
> 
> - Adds a new `from` rule in `vc-cp-<release>` NetworkPolicy matching `app: rancher-agent-install` in `networkpolicy.yaml`
> - Updates tests in `networkpolicy_test.yaml`: increases expected control plane ingress rule count and shifts indices for extra ingress assertions
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 36cf255a99d9c1fba5fc35c270659283f2602c21. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/chart/templates/networkpolicy.yaml
+++ b/chart/templates/networkpolicy.yaml
@@ -151,6 +151,12 @@ spec:
             matchLabels:
               app: vcluster-snapshot
 
+    # Allow ingress from vcluster rancher operator agent install pods.
+    - from:
+        - podSelector:
+            matchLabels:
+              app: rancher-agent-install
+
     # Allow ingress from vcluster platform.
     - from:
         - podSelector:
--- a/chart/tests/networkpolicy_test.yaml
+++ b/chart/tests/networkpolicy_test.yaml
@@ -57,7 +57,7 @@ tests:
           value: vc-cp-my-release
         lengthEqual:
           path: spec.ingress
-          count: 4
+          count: 5
       - documentSelector:
           path: metadata.name
           value: vc-kube-dns-my-release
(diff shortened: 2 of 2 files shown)


chart/templates/networkpolicy.yaml
```diff
@@ -151,6 +151,12 @@ spec:
             matchLabels:
               app: vcluster-snapshot
 
+    # Allow ingress from vcluster rancher operator agent install pods.
+    - from:
+        - podSelector:
+            matchLabels:
+              app: rancher-agent-install
```
FabianKramm, line 158: This feels weird, is our pattern to add every application that is out there like this? I feel like there needs to be a better / more generic solution

FabianKramm (changes requested):

chart/templates/networkpolicy.yaml
cbalan: Indeed, at this stage, we do not have a generic way to identify various vcluster apps. 
A better approach could be to introduce a tier label that could be used here. label that could be used by all vcluster apps/components. ex. tier: control-plane/workload

What do you think about getting this one in, just so we can simplify the vcluster rancher operator for 0.31. I can fill in a follow up issue for a generic tier based solution that would need to be applied for all vcluster apps.

PR #3454, opened 2026-01-08
fix(assets scripts): support multiarch images

sydorovdmytro:
Fixes an issue when a single architecture image was pushed to the private registry

/kind bugfix

resolves ENG-10853

---

> [!NOTE]
> Adds multi-arch handling to image download/push scripts with per-arch tagging and manifest publishing.
> 
> - `download-images.sh`: replaces single `--platform` with `--platforms` (comma-separated), pulls each arch, tags as `image-linux-<arch>`, aggregates and validates before `docker save`, with improved errors
> - `push-images.sh`: supports `--platforms` and `--insecure`, loads archive, pushes per-arch tags, then creates/pushes a multi-arch manifest list; trims `docker.io` as well; adds robust error handling
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6e3d817977e7541c1f9e5c56bbca65a7675dcaf3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


assets/download-images.sh
```diff
@@ -1,17 +1,16 @@
 #!/bin/bash
 list="images.txt"
 images="vcluster-images.tar.gz"
-platform="linux/amd64"
+platforms="linux/amd64,linux/arm64"
```
FabianKramm: I'm against this, this makes the bundle a lot bigger, why would we want to do this? In most cases you only need linux/amd64 and not arm64. And even if you would need it you could just adjust this parameter

sydorovdmytro: It can be a problem for clusters that have both node types: amd64 and arm64. We publish a multi-arch image originally, but then push a single arch image to the private registry.

Another option could be keeping a single item in the list `linux/amd64`, and document the option of adding `linux/arm64` if needed.

sydorovdmytro: @FabianKramm it's changed to use a single architecture by default, but with an option to extend the list.

PR #3494, opened 2026-01-16
feat: sync containers resources inplace resize on host cluster

flomedja:
/kind enhancement

resolves ENG-11208

Sync the virtual cluster resource resizing in place to the host cluster for kubernetes version `1.35.0` and greater.

### DIFF
--- a/chart/templates/role.yaml
+++ b/chart/templates/role.yaml
@@ -40,6 +40,10 @@ rules:
     resources: ["pods/status", "pods/ephemeralcontainers"]
     verbs: ["patch", "update"]
   {{- end }}
+  {{- if ge (.Capabilities.KubeVersion.Minor|int) 35 }}
+  - apiGroups: [""]
+    resources: ["pods/resize"]
+  {{- end }}
   - apiGroups: ["apps"]
     resources: ["statefulsets", "replicasets", "deployments"]
     verbs: ["get", "list", "watch"]
--- a/pkg/controllers/resources/pods/syncer.go
+++ b/pkg/controllers/resources/pods/syncer.go
@@ -2,6 +2,7 @@ package pods
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"slices"
@@ -9,8 +10,8 @@ import (
 
 	nodev1 "k8s.io/api/node/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
-
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	utilversion "k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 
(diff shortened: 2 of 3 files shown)


devspace.yaml
```diff
@@ -32,6 +32,8 @@ deployments:
           distro:
             k8s:
               enabled: true
+              image:
+                tag: v1.35.0
```
FabianKramm: I think we should remove that here

chart/templates/role.yaml
```diff
@@ -40,6 +40,9 @@ rules:
     resources: ["pods/status", "pods/ephemeralcontainers"]
     verbs: ["patch", "update"]
   {{- end }}
+  - apiGroups: [""]
```
FabianKramm, line 44: Can you use .Capabilities.KubeVersion.Minor here?

pkg/controllers/resources/pods/syncer.go
```diff
@@ -405,6 +412,36 @@ func (s *podSyncer) Sync(ctx *synccontext.SyncContext, event *synccontext.SyncEv
 	return ctrl.Result{}, nil
 }
 
+func (s *podSyncer) resizeHostPodContainerResourcesInPlace(ctx *synccontext.SyncContext, event *synccontext.SyncEvent[*corev1.Pod]) error {
+	hostClusterVersion, err := ctx.Config.HostClient.Discovery().ServerVersion()
```
FabianKramm: This is very inefficient doing that for every reconcile

flomedja: Fixed

chart/templates/role.yaml
flomedja: @FabianKramm  Do you mean ?

```yaml
  {{- if ge (.Capabilities.KubeVersion.Minor|int) 35 }}
  - apiGroups: [""]
    resources: ["pods/resize"]
  {{- end }}
  ```

File: devspace.yaml (hunk 1 above)
Comment (flomedja) [reply devspace.yaml]: fixed

File: chart/templates/role.yaml (hunk 2 above)
Comment (FabianKramm) [reply chart/templates/role.yaml:44]: Yep

File: pkg/controllers/resources/pods/syncer.go [hunk 4]
```diff
@@ -119,6 +132,9 @@ func New(ctx *synccontext.RegisterContext) (syncertypes.Object, error) {
 		nodeSelector:          nodeSelector,
 		tolerations:           tolerations,
 
+		hostClusterVersionInfo: hostClusterVersionInfo,
```
FabianKramm: Why does this need to be in the podSyncer struct, doesn't look like you are using that anywhere

pkg/controllers/resources/pods/syncer.go
```diff
@@ -445,6 +487,59 @@ func setSATokenSecretAsOwner(ctx *synccontext.SyncContext, pClient client.Client
...
+		p, ok := hostContainers[v.Name]
+		if !ok {
+			continue
+		}
+		if equality.Semantic.DeepEqual(p.Resources, v.Resources) {
+			continue
+		}
+
+		var resources corev1.ResourceRequirements
+		v.Resources.DeepCopyInto(&resources)
```
FabianKramm: Why do deepcopy here, why can't we use v.Resources directly?

chart/templates/role.yaml
flomedja: Fixed

pkg/controllers/resources/pods/syncer.go
flomedja: It is used only to compute the `hostClusterVersion`. I remove it from the struct.

pkg/controllers/resources/pods/syncer.go
flomedja: We don't need it because we return a json.Marshall instead of a reference to the object. Removed

FabianKramm (approved):

PR #3527, opened 2026-01-26
fix: skip error event on resourceversion conflict in pod syncer

ymichaelson:
/kind bugfix

resolves #<issue_number> 

Fixed an issue where vcluster pod syncer would record unnecessary SyncError events for transient resource version conflicts during patch operations, which could mislead users into thinking the operation failed when it would actually succeed on retry.

## Summary

This PR fixes the handling of conflict errors in the pod syncer's patch operation. When a resource version conflict occurs during patching (e.g., due to concurrent updates from webhooks or other controllers), controller-runtime will automatically retry the reconcile operation and eventually maintain data consistency. However, the previous implementation would record a SyncError event for these transient conflicts, which could mislead users into thinking the operation had failed when it would actually succeed on retry.

## Problem

When a resource version conflict causes a patch operation to fail:
- Controller-runtime automatically retries the reconcile operation
- The operation eventually succeeds and data consistency is maintained
- However, recording a SyncError event for these transient conflicts creates a false impression of failure to users

## Changes

- Added conflict error detection in `pkg/controllers/resources/pods/syncer.go`
- When a conflict error is detected during patch operation, the error is aggregated but not recorded as a SyncError event
- This prevents misleading error events while still allowing controller-runtime to handle retries automatically

## Technical Details

The fix adds a check for `kerrors.IsConflict(err)` before recording the error. When a resource version conflict is detected:
1. The error is aggregated with any existing errors
2. The function returns early without recording a SyncError event
3. Controller-runtime's retry mechanism will automatically retry the reconcile operation
4. The operation will eventually succeed and maintain data consistency

This prevents unnecessary and misleading error events from being recorded in the virtual cluster for transient conflicts that are expected to be resolved on retry.

## Related Issues

image

### DIFF
--- a/pkg/controllers/resources/pods/syncer.go
+++ b/pkg/controllers/resources/pods/syncer.go
@@ -384,6 +384,10 @@ func (s *podSyncer) Sync(ctx *synccontext.SyncContext, event *synccontext.SyncEv
 
 	defer func() {
 		if err := patch.Patch(ctx, event.Host, event.Virtual); err != nil {
+			if kerrors.IsConflict(err) {
+				retErr = utilerrors.NewAggregate([]error{retErr, err})
+				return
+			}
 			retErr = utilerrors.NewAggregate([]error{retErr, err})
 		}
 


FabianKramm:
@ymichaelson thanks for the PR! Its rather weird this happens on a patch request, but I guess it doesn't hurt to not print the error here.

ymichaelson:
Thanks for the quick review and merge @FabianKramm !  
Appreciate the feedback — glad this small but annoying user-facing noise is fixed now.

Looking forward to it being part of the next release 🚀

PR #3564, opened 2026-02-08
fix(docker): remove use of mount namespace

mhrabovcin:
/kind bugfix

resolves #
Resolving path for containerd or docker sockes rely on having `netstat` and `awk` installed on the host system, which may not be the case. This PR removes use of the host mount namespace and it will use mentioned binaries from the alpine container that runs the resolution command.

Fixed an issue where vcluster create with docker provider runs on a machine that doesn't have `netstat` installed.


FabianKramm:
@mhrabovcin thanks for this PR! I think we settled with an alternative when netstat and awk are not there which should always work as well in here https://github.com/loft-sh/vcluster/pull/3567. Your approach seems more sophisticated but I'm unsure if not entering the mount namespace of the host will basically lead to issues with the paths from netstat

mhrabovcin:
Thanks @FabianKramm , I've tested it on my system and the fix worked.

PR #3567, opened 2026-02-09
fix: missing netstat

FabianKramm:
/kind bugfix

### DIFF
--- a/pkg/cli/create_docker.go
+++ b/pkg/cli/create_docker.go
@@ -1134,7 +1134,7 @@ func getDockerSocketPath(ctx context.Context) (string, error) {
         if command -v netstat >/dev/null 2>&1; then
             netstat -xlp | awk '$NF ~ /\/docker\.sock(\.real)?$/ {print $NF}'
         else
-            for p in /var/run/docker.sock /var/run/docker.sock.real; do
+            for p in /var/run/docker.sock.real /var/run/docker.sock; do
                 if [ -S "$p" ]; then
                     echo "$p"
                 fi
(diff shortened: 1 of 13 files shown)


pkg/cli/create_docker.go
```diff
@@ -1130,27 +1130,40 @@ func findTailIPs(ctx context.Context, networkName string, tailSize int) ([]strin
 func getDockerSocketPath(ctx context.Context) (string, error) {
```
roehrijn, line 1130: `getContainerdSocketPath()` has a very similar issue. Would you mind fixing that as well? 
I'm currently working on an Ubuntu 25.10 and `netstat` (aka `net-tools` package) is not installed by default. I experimented a bit with other options and one thing which worked for me is: 

    docker run -q --rm --privileged --pid=host alpine nsenter -t 1 -m -p -u -i -n sh -c "lsof -U -F n | sed -n 's/^n\(\/.*\/containerd\.sock\) .*/\1/p'"

I'm not 100% sure, but this could also be more distribution/platform independent.

pkg/cli/create_docker.go
```diff
@@ -1130,27 +1130,40 @@ func findTailIPs(ctx context.Context, networkName string, tailSize int) ([]strin
...
+            for p in /var/run/docker.sock /var/run/docker.sock.real; do
+                if [ -S "$p" ]; then
+                    echo "$p"
+                fi
+            done
+        fi
+    `
 
-	out, err := exec.CommandContext(ctx, "docker", "run", "-q", "--rm", "--privileged", "--pid=host", "alpine", "nsenter", "-t", "1", "-m", "-p", "-u", "-i", "-n", "sh", "-c", cmdStr).CombinedOutput()
+	out, err := exec.CommandContext(ctx, "docker", "run", "-q", "--rm", "--privileged", "--pid=host", "alpine", "nsenter", "-t", "1", "-m", "-p", "-u", "-i", "-n", "sh", "-c", cmdStr).Output()
```
roehrijn, line 1145: not sure if it makes sense to handle `stderr` and `stdout` with two independent buffers. WDYT?

FabianKramm: So golang wraps the stderr into a error which is then used below, so we are not losing that information when an error occurs

pkg/cli/create_docker.go
FabianKramm: Yes true, I will change that

PR #3572, opened 2026-02-10
feature(debug-shell): introduce a debug shell command

mfranczy:
/kind feature

resolves ENGPROV-122


pkg/util/debugshell/embedded_etcd.go
```diff
@@ -0,0 +1,72 @@
...
+func BuildEtcdEndpoints(vClusterName string, podCount int) string {
+	debugEndpoints := "https://localhost:2379"
+	if podCount <= 1 {
+		return debugEndpoints
+	}
+
+	remoteEndpoints := make([]string, 0, podCount)
+	for i := 0; i < podCount; i++ {
+		remoteEndpoints = append(remoteEndpoints, fmt.Sprintf("https://%s-%d.%s-headless:2379", vClusterName, i, vClusterName))
+	}
```
zerbitx, line 70:
```suggestion
	remoteEndpoints := make([]string, 0, podCount)
	for i := 0; i < podCount; i++ {
		remoteEndpoints[i] = fmt.Sprintf("https://%s-%d.%s-headless:2379", vClusterName, i, vClusterName)
	}
```

mfranczy: This will give me index out of range error :)
Should be: `remoteEndpoints := make([]string, podCount)`
Although, I would not change this, performance wise there is no visible difference, and using append is idiomatic.

PR #3607, opened 2026-02-26
Implement DeleteAllOf in blockingcacheclient and pluginhookclient

behrooz:
## What does this PR do?

Implements `DeleteAllOf` explicitly on the wrapper clients so the previously unimplemented method is no longer a TODO.

## Changes

- **pkg/util/blockingcacheclient/client.go**: Add `DeleteAllOf` that delegates to the embedded `client.Client`. Removes the `// TODO: implement DeleteAllOf` comment.
- **pkg/util/pluginhookclient/client.go**: Add `DeleteAllOf` that delegates to the embedded `client.Client`. Removes the `// TODO: implement DeleteAllOf` comment.

## Behavior

No behavior change. Both wrappers already exposed the embedded client’s `DeleteAllOf` via struct embedding. This makes that behavior explicit and resolves the TODOs.

### DIFF
--- a/pkg/util/blockingcacheclient/client.go
+++ b/pkg/util/blockingcacheclient/client.go
@@ -263,7 +263,9 @@ func (c *CacheClient) blockApply(ctx context.Context, obj runtime.ApplyConfigura
 	})
 }
 
-// TODO: implement DeleteAllOf
+func (c *CacheClient) DeleteAllOf(ctx context.Context, obj client.Object, opts ...client.DeleteAllOfOption) error {
+	return c.Client.DeleteAllOf(ctx, obj, opts...)
+}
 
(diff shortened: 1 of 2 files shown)


flomedja:
As you said both structs already get `DeleteAllOf` promoted from their embedded `client.Client` field. If we want to implement `DeleteAllOf` to be compliant with the `Delete` implementation that will require cache-blocking and check for pluginHooks on a group of objects present in the cache. 
IMHO just calling the wrapped function have no real value however I agree that we could make the TODO comment more explicit for a future implementation.

WDYT @FabianKramm

FabianKramm:
Totally agree here @flomedja, this PR doesn't really improve the situation.

PR #3610, opened 2026-02-27
fix(docker): wait for systemd readiness before installing vcluster

Piotr1215:
## Summary

Fixes intermittent "Failed to connect to bus: No such file or directory" errors when creating vCluster clusters with the Docker driver, especially on CI runners (GitHub Actions).

## Problem

When `vcluster create` runs with the Docker driver, the sequence is:
1. `docker run` starts a container with systemd as PID 1
2. Immediately, `docker exec` runs `install-standalone.sh` inside the container
3. The install script calls `systemctl daemon-reload` and `systemctl enable --now vcluster.service`
4. If systemd hasn't finished starting dbus inside the container, `systemctl` fails with "Failed to connect to bus: No such file or directory"

This is a race condition between container startup and the install script. It's particularly common on GitHub Actions runners due to:
- Nested containerization overhead (Docker inside the runner VM)
- cgroup v2 with systemd cgroup driver
- Resource contention from parallel workflows

For reference, KinD avoids this by waiting for systemd readiness before running kubeadm inside its containers.

## Fix

Add `waitForSystemd()` between `runControlPlaneContainer()` and `installVClusterStandalone()`. It polls `systemctl is-system-running` every 500ms for up to 60s, exiting as soon as systemd reports "running" or "degraded". On a healthy system this adds <2s; it only becomes meaningful when the race would otherwise cause a failure.

## Testing

- Verified compilation: `go build ./pkg/cli/...`
- The `loft-sh/setup-vind` GitHub Action (DEVOPS-591) has been hitting this race consistently in CI — this fix addresses the root cause
- Test matrix at [vClusterLabs-Experiments/setup-vind-test](https://github.com/vClusterLabs-Experiments/setup-vind-test) exercises single and multi-cluster creation on GitHub runners

References DEVOPS-591

### DIFF
--- a/pkg/cli/create_docker.go
+++ b/pkg/cli/create_docker.go
@@ -213,16 +213,6 @@ func CreateDocker(ctx context.Context, options *CreateOptions, globalFlags *flag
 		return err
 	}
 
-	// wait for systemd to be ready inside the container before running install
-	// script, which calls systemctl. Without this, the install can race with
-	// dbus startup and fail with "Failed to connect to bus".
-	if !exists {
-		err = waitForSystemd(ctx, vClusterName, log)
-		if err != nil {
-			return err
-		}
-	}
-
 	// install vCluster standalone
 	if !exists {
 		err = installVClusterStandalone(ctx, vClusterName, vClusterVersion, extraVClusterArgs, log)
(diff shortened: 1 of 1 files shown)


FabianKramm (changes requested):
Can we move that check into the script that tries to install the control plane / join the node?

Piotr1215:
Moved the systemd readiness check into the `installVClusterStandalone()` bash command as requested. The wait loop now runs inline before the `curl | sh`, in a single `docker exec`:

```bash
set -e -o pipefail; for i in $(seq 1 120); do state=$(systemctl is-system-running 2>/dev/null || true); [ "$state" = "running" ] || [ "$state" = "degraded" ] && break; sleep 0.5; done; curl -sfLk "..." | sh -s -- ...
```

Removed the standalone `waitForSystemd()` Go function and its separate call site (-41 lines).

**Local testing:**
- `go build ./cmd/vclusterctl` — compiles clean
- `go test ./pkg/cli/...` — all tests pass
- Verified the bash wait loop against the actual `ghcr.io/loft-sh/vm-container` image:
  ```
  $ docker run -d --name test --privileged --tmpfs /run --tmpfs /tmp ghcr.io/loft-sh/vm-container
  $ docker exec test bash -c 'for i in $(seq 1 120); do state=$(systemctl is-system-running 2>/dev/null || true); [ "$state" = "running" ] || [ "$state" = "degraded" ] && echo "ready: $state" && break; sleep 0.5; done'
  ready: running
  ```
- Verified the combined wait+command flow completes without errors

PR #3613, opened 2026-03-02
fix(ci): skip homebrew upload for non-latest stable releases

Piotr1215:
## Summary

Prevents older maintenance releases (e.g. v0.31.1) from overwriting newer stable versions (v0.32.0) in the `loft-sh/homebrew-tap` formulas.

## What changed

Added a step to the release workflow that fetches the current version from `loft-sh/homebrew-tap/Formula/vcluster.rb` before running GoReleaser. If the release would downgrade the tap, `--skip=homebrew` is passed to GoReleaser. All other release assets (Docker images, binaries, chart) are still published normally.

Fallback: if the tap can't be reached, the upload proceeds (safe default). Pre-releases are unaffected — goreleaser's existing `skip_upload: auto` handles them.

## Test plan

- [ ] Release a maintenance patch while tap has a newer version — homebrew upload skipped
- [ ] Release the latest stable version — homebrew upload proceeds
- [ ] Release a pre-release — handled by goreleaser's `skip_upload: auto`
- [ ] Network failure fetching tap — falls back to allowing upload

Closes DEVOPS-630

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -87,22 +87,32 @@ jobs:
           # Only check stable releases. Pre-releases are handled by
           # goreleaser's skip_upload: auto in .goreleaser.yaml.
           if [[ "$CURRENT_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-            # Fetch the current version from the homebrew tap formula.
-            # Only skip if this release would downgrade what's in the tap.
-            TAP_VERSION=$(curl -sfL --max-time 10 \
-              "https://raw.githubusercontent.com/loft-sh/homebrew-tap/main/Formula/vcluster.rb" \
-              | grep -oP 'version "\K[^"]+' || echo "")
-            if [ -n "$TAP_VERSION" ]; then
-              CURRENT_VERSION="${CURRENT_TAG#v}"
-              if [ "$(printf '%s\n' "$CURRENT_VERSION" "$TAP_VERSION" | sort -V | tail -1)" != "$CURRENT_VERSION" ]; then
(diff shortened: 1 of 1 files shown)


Piotr1215:
## Semver logic clarification

This PR does **not** skip all maintenance/patch releases from Homebrew. It only skips releases where the tag is not the highest stable semver in the repository.

### How the check works

The new `check_latest_stable` step compares the current release tag against all stable tags (matching `vX.Y.Z` — no pre-release suffix) using `sort -V`:

```bash
LATEST_STABLE=$(git tag -l 'v*' | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -1)
```

### Examples

| Release sequence | Current tag | Latest stable tag | Homebrew upload? |
|---|---|---|---|
| v0.32.0 released first, then v0.31.1 | `v0.31.1` | `v0.32.0` | **Skipped** — v0.31.1 < v0.32.0 |
| v0.32.0 is the newest stable | `v0.32.0` | `v0.32.0` | **Uploaded** — matches latest |
| v0.32.1 released after v0.32.0 | `v0.32.1` | `v0.32.1` | **Uploaded** — it IS the new latest |
| v0.33.0-rc.1 pre-release | `v0.33.0-rc.1` | `v0.32.0` | Deferred to goreleaser's `skip_upload: auto` (skips pre-releases) |

The key scenario this fixes: v0.32.0 was released Feb 18 and correctly updated the tap. v0.31.1 was released Feb 19 and **overwrote** the tap back to 0.31.1 because goreleaser's `skip_upload: auto` only guards against pre-releases, not older stable versions.

PR #3614, opened 2026-03-02
ci(e2e): migrate from kind to vind for e2e test clusters

Piotr1215:
## Summary

Replace `loft-sh/setup-kind` with `loft-sh/setup-vind@v1` across all three e2e workflow jobs (install-delete, upgrade-test, e2e-tests).

The vind Docker driver provides native Docker image access, eliminating the need for explicit `kind load image-archive` steps. Node hostname detection is done dynamically since vind uses the cluster name as hostname (not the KinD `<name>-control-plane` pattern).

## Changes

- `setup-kind@master` → `setup-vind@v1` with `kubernetes-version` input (replaces `version` + `image`)
- Remove `kind load image-archive` commands (vind shares host Docker daemon images natively)
- Detect host node name dynamically via `kubectl get nodes` for the node sync selector in `commonValues.yaml`
- Add `HOST_NODE_NAME` env var to e2e test runner with backwards-compatible fallback to `KIND_NAME` for local KinD setups
- Minimal Go test change in `test/e2e/node/node.go` to support the new env var
- Skip vind cluster deletion in post-job cleanup (avoids driver mismatch with helm)

## Blocker

⚠️ **Draft**: Blocked on vcluster v0.32.1 release containing the systemd/dbus readiness fix ([`0707af6`](https://github.com/loft-sh/vcluster/commit/0707af67903caa17db619f8215f8ce825a08c7bd) — `fix(docker): wait for systemd readiness before installing vcluster`).

Without this fix, `vcluster create` with the Docker driver intermittently fails with `Failed to connect to bus: No such file or directory` when systemd hasn't started dbus yet inside the container. The `setup-vind` action retries 3 times but some CI jobs still fail all attempts.

**Passing jobs** (10/15): install-delete, upgrade, e2e_ha, e2e_certs, e2e_pause_resume, e2e_metrics_proxy, e2e_plugin, e2e_limit_classes, lint, build
**Failing jobs** (5/15): e2e_node, e2e_cli, e2e_rootless, e2e_isolation_mode, e2e_scheduler — all due to dbus race or downstream timing effects

## Context

Part of DEVOPS-579 (KinD → vind migration). This is the CI/e2e workflow PR. Justfile and docs changes follow in separate PRs.

Depends on [`loft-sh/setup-vind@v1`](https://github.com/loft-sh/setup-vind) (DEVOPS-591).

Closes DEVOPS-579

### DIFF
--- a/test/e2e/node/node.go
+++ b/test/e2e/node/node.go
@@ -19,7 +19,7 @@ var _ = ginkgo.Describe("Node sync", func() {
 
 		hostname := "kind-control-plane"
 
-		if nodeName, ok := os.LookupEnv("HOST_NODE_NAME"); ok {
+		if nodeName, ok := os.LookupEnv("HOST_NODE_NAME"); ok && nodeName != "" {
 			hostname = nodeName
 		} else if kindName, ok := os.LookupEnv("KIND_NAME"); ok {
 			hostname = kindName + "-control-plane"


Piotr1215:
Marked as draft — blocked on the systemd/dbus readiness fix landing in a release ([`0707af6`](https://github.com/loft-sh/vcluster/commit/0707af67903caa17db619f8215f8ce825a08c7bd)). Without it, `vcluster create` with the Docker driver intermittently fails with `Failed to connect to bus: No such file or directory` during CI. 10/15 e2e jobs pass but the remaining 5 hit this race condition. Will mark ready once v0.32.1 ships with the fix.

.github/workflows/e2e.yaml
```diff
@@ -116,14 +116,18 @@ jobs:
...
           version: "v3.11.0"
 
-      - name: Set up kind k8s cluster
-        uses: loft-sh/setup-kind@master
+      - name: Set up vind cluster
+        uses: loft-sh/setup-vind@v1
         with:
-          version: "v0.30.0"
-          image: kindest/node:v1.35.0@sha256:4613778f3cfcd10e615029370f5786704559103cf27bef934597ba562b269661
+          version: "v0.32.2-rc.1"
```
sydorovdmytro: why do we use `rc` here?

.github/workflows/e2e.yaml
```diff
@@ -143,7 +147,8 @@ jobs:
 
       - name: Setup environment
         run: |
-          kind load image-archive vcluster_syncer
+          docker load -i vcluster_syncer
```
sydorovdmytro, line 150: Do we still need this step for vind? I think it can use a local docker cache now

sydorovdmytro (changes requested):
Generally I'm not sure about using `rc` version for tests and we should test if the docker image import is still needed, as one of our marketing features for vind is a local docker cache.

One more common patter for all jobs here: we setup Vind with the external action, then reset the driver back to Helm and see the cluster info. Shall we just merge these steps into `loft-sh/setup-vind`?

.github/workflows/e2e.yaml
Piotr1215: It is required because the fix that we need is not yet released

.github/workflows/e2e.yaml
Piotr1215: Yes we do because in order to use the image cache with vind on linux we need to reconfigure docker deamon.

Piotr1215:
https://github.com/loft-sh/vcluster/pull/3642 Fix from this pr needs to land first.

Piotr1215:
Fix is merged, this needs to be re-run when the fix is first released

Piotr1215:
Addressed review feedback:

1. **RC version replaced** — `v0.32.2-rc.2` → `v0.33.0` (latest stable, includes the kine socket restore fix from #3642).

2. **Docker image import removed** — dropped the `docker save | docker exec ctr` workaround in all 3 jobs. vind natively mirrors the host Docker image cache, so `docker load` is sufficient. This matches the pattern already working in `e2e-ginkgo.yaml` (the `run-ginkgo-e2e` action only calls `docker load`). The setup-vind README migration table also confirms: `kind load docker-image` → `*(not needed)*`.

3. **Merge setup-vind + driver-reset** — valid suggestion, but out of scope for this PR. The `vcluster use driver helm` step after setup-vind could be an input like `driver: helm` on the action. Will track as a follow-up in `loft-sh/setup-vind`.

/retest

PR #3617, opened 2026-03-02
feat: Ability to set custom labels in kubernetes resources deployed by vcluster helm chart

scheibinger:
/kind enhancement

resolves #

Adds a way to set custom labels for kubernetes resources deployed by vcluster helm chart.

Problem: In my kubernetes environment, I need to set custom labels which are enforced by the policies in the cluster. 

Ability to set custom labels in kubernetes resources deployed by vcluster helm chart


chart/templates/clusterrole.yaml
```diff
@@ -8,6 +8,10 @@ metadata:
     chart: "{{ include "vcluster.version.label"  $ }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
+  {{- $labels := merge dict .Values.controlPlane.statefulSet.labels }}
+  {{- if $labels }}
+{{ toYaml $labels | indent 4 }}
+  {{- end }}
```
FabianKramm, line 14: I would rather want to do ` .Values.controlPlane.advanced.globalMetadata.labels` here than reuse the statefulSet labels on a ClusterRole

chart/templates/clusterrolebinding.yaml
```diff
@@ -8,6 +8,10 @@ metadata:
     chart: "{{ include "vcluster.version.label"  $ }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
+  {{- $labels := merge dict .Values.controlPlane.statefulSet.labels }}
```
FabianKramm, line 11: Same as above

chart/templates/config-secret.yaml
```diff
@@ -8,6 +8,10 @@ metadata:
     chart: "{{ include "vcluster.version.label"  $ }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
+  {{- $labels := merge dict .Values.controlPlane.statefulSet.labels }}
```
FabianKramm, line 11: Same as above

FabianKramm (changes requested):
I think what you want is a global labels similar to `.Values.controlPlane.advanced.globalMetadata.annotations` which fits better than reusing the statefulset labels

PR #3627, opened 2026-03-03
fix: ensure that toleration update are propagated to host cluster

flomedja:

/kind enhancement

resolves ENGTEN-122


flomedja:
@codex review

flomedja:
@codex review

flomedja:
@codex review

flomedja:
@codex review

flomedja:
@codex review

pkg/controllers/resources/pods/translate/translator.go
```diff
@@ -417,6 +424,13 @@ func (t *translator) Translate(ctx *synccontext.SyncContext, vPod *corev1.Pod, s
 		}
 	}
 
+	// apply enforced tolerations from vcluster config, skipping any already present
+	for _, toleration := range t.enforcedTolerations {
+		if !hasTolerationForSameTaint(pPod.Spec.Tolerations, toleration) {
```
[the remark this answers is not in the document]
flomedja: @FabianKramm I would like your input on this. Should we use the enforceTolerations or overwrite it with a similar toleration with a different `tolerationSeconds` define in the virtual pod ?

pkg/controllers/resources/pods/translate/diff.go
```diff
@@ -50,6 +51,49 @@ func (t *translator) Diff(ctx *synccontext.SyncContext, event *synccontext.SyncE
...
+			// by a virtual or enforced toleration.
+			// Uses taint-target equality (ignoring TolerationSeconds) so that a stale host
+			// variant of a toleration whose TolerationSeconds was just updated in the virtual
+			// pod does not accumulate alongside the updated one.
+			if !hasTolerationForSameTaint(newTolerations, hostTol) {
+				newTolerations = append(newTolerations, hostTol)
+			}
+		}
+		pPod.Spec.Tolerations = newTolerations
+	}
```
FabianKramm: This is too complex just do:
```
if KubernetesSupportsIt() {
  pPod.Spec.Tolerations = vPod.Spec.Tolerations

  // add the ctx.Config.Sync.ToHost.Pods.EnforceTolerations
}
```

And move this into t.calcSpecDiff

flomedja: I am fine for removing the part to determine if we need to sync tolerations. It's an optimization that is not critical. However, It is important to check if there is tolerations on the Host cluster (added by some admission controller or other mechanisms on the host cluster) that should persisted. 
Tolerations can't be removed but just added. And tolerationsSeconds on a toleration with `NoExcecute` taint can be increase.

so WDYT about :

```go
	newTolerations := append([]corev1.Toleration{}, vPod.Spec.Tolerations...)
	for _, toleration := range t.enforcedTolerations {
		// Use taint-target equality so that a virtual pod carrying the enforced taint
		// with a different TolerationSeconds does not produce a duplicate entry.
		if !hasTolerationForSameTaint(newTolerations, toleration) {
			newTolerations = append(newTolerations, toleration)
		}
	}
	for _, hostTol := range event.Host.Spec.Tolerations {
		// Carry forward host-only tolerations that are not already covered
		// by a virtual or enforced toleration.
		// Uses taint-target equality (ignoring TolerationSeconds) so that a stale host
		// variant of a toleration whose TolerationSeconds was just updated in the virtual
		// pod does not accumulate alongside the updated one.
		if !hasTolerationForSameTaint(newTolerations, hostTol) {
			newTolerations = append(newTolerations, hostTol)
		}
	}
```

FabianKramm: Make sure you preserve the order as much as possible other than that thats okay I guess

flomedja: My question [here](https://github.com/loft-sh/vcluster/pull/3627/changes/BASE..085a7eb4ede7b96d28ac7e746a530b9942ffedb4#r2887488252) is still relevant.

When a vPop define the same tolerations as a enforceTolerations with a different `tolerationsSeconds` do we want to take the tolerations in the `vPop` or the one defined in the `enforcedTolerations`. 

PS: we don't [parse](https://github.com/loft-sh/vcluster/blob/18243ba6be71871dfc297850dc231bf4cb4bd0f6/pkg/util/toleration/toleration.go#L11) the `tolerationsSeconds` from the enforcedTolerations

flomedja: @FabianKramm What do you mean by preserving the order ?

flomedja: I avoid duplicate toleration in my implementation, Do you mean that we should allow duplicate tolerations but ensure that they are always added in the same order ?

PS: Kubernetes don't avoid duplicate tolerations even if they can lead to weird behaviors in some case

FabianKramm: With order I mean first add host tolerations then enforced tolerations currently its the other way around

FabianKramm: yes just allow duplicate tolerations if kubernetes is not failing I don'T see a reason why we would not

FabianKramm: And get rid of the test suite just for that and make some unit tests instead or make it part of the already existing test suites

pkg/controllers/resources/pods/translate/translator.go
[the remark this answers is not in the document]
flomedja: Discussed. We want to allow duplicate.

pkg/controllers/resources/pods/translate/diff.go
flomedja: Fixed

PR #3424, opened 2025-12-18
(feat): add snapshot support for vCluster Standalone

neogopher:
**What issue type does this pull request address?** (keep at least one, remove the others) 
/kind feature

**What does this pull request do? Which issues does it resolve?** (use `resolves #<issue_number>` if possible) 
resolves ENG-9254

**Please provide a short message that should be published in the vcluster release notes**
Fixed an issue where vcluster ...

**What else do we need to know?**

PR #3658, opened 2026-03-09
Add snapshots support for standalone

nprokopic:
/kind enhancement

resolves ENGOX-155

Add snapshot and restore support for vCluster Standalone.

Based on Sachin's PR https://github.com/loft-sh/vcluster/pull/3424.

### DIFF
--- a/cmd/vcluster/cmd/snapshot/restore.go
+++ b/cmd/vcluster/cmd/snapshot/restore.go
@@ -2,7 +2,9 @@ package snapshot
 
 import (
 	"fmt"
+	"os"
 
+	"github.com/loft-sh/vcluster/pkg/config"
 	"github.com/loft-sh/vcluster/pkg/snapshot"
 	"github.com/spf13/cobra"
 )
@@ -18,12 +20,17 @@ func NewRestoreCommand() *cobra.Command {
 		Short: "Restore a vCluster",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, _ []string) error {
+			vConfig, err := config.LoadAutoDetectedRuntimeConfig(os.Getenv("VCLUSTER_NAME"))
+			if err != nil {
+				return err
+			}
+
 			envOptions, err := snapshot.ParseOptionsFromEnv()
 			if err != nil {
 				return fmt.Errorf("failed to parse options from environment: %w", err)
 			}
 			restoreClient := snapshot.NewRestoreClient(*envOptions, restoreVolumes, newVCluster)
-			return restoreClient.Run(cmd.Context())
+			return restoreClient.Run(cmd.Context(), vConfig)
 		},
 	}
 
--- a/pkg/cli/find/find.go
+++ b/pkg/cli/find/find.go
@@ -3,6 +3,8 @@ package find
 import (
 	"context"
 	"fmt"
+	"os"
+	"os/exec"
 	"regexp"
 	"strings"
 	"time"
@@ -11,10 +13,12 @@ import (
 	"github.com/loft-sh/log"
 	"github.com/loft-sh/log/survey"
 	"github.com/loft-sh/log/terminal"
+	vclusterconfig "github.com/loft-sh/vcluster/pkg/config"
 	"github.com/loft-sh/vcluster/pkg/constants"
 	"github.com/loft-sh/vcluster/pkg/platform"
 	"github.com/loft-sh/vcluster/pkg/platform/kube"
 	"github.com/loft-sh/vcluster/pkg/platform/sleepmode"
+	standaloneutil "github.com/loft-sh/vcluster/pkg/util/standalone"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/discovery"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -49,6 +53,7 @@ type VCluster struct {
 	Status                 Status
 	Context                string
 	Version                string
+	IsStandalone           bool
 }
 
 type Status string
@@ -128,6 +133,35 @@ func GetPlatformVCluster(ctx context.Context, platformClient platform.Client, na
 	}
 
 	return nil, fmt.Errorf("unexpected error searching for selected virtual cluster")
+}
+
+// GetVClusterStandaloneOrVCluster supports both regular host-cluster lookup and
+// local standalone lookup. The CLI selector "local-standalone" explicitly targets
+// the standalone service on this host and makes standalone probe errors fatal.
+// For real runtime names, standalone lookup is only an opportunistic shortcut:
+// if the local standalone runtime name matches the requested name, use it;
+// otherwise fall back to regular host-cluster discovery so in-cluster vClusters
+// with arbitrary names keep working.
+func GetVClusterStandaloneOrVCluster(ctx context.Context, context, name, namespace string, log log.Logger) (*VCluster, error) {
+	if name == constants.VClusterStandaloneCLISelector {
+		vClusterStandalone, err := getVClusterStandalone()
+		if err != nil {
+			return nil, err
+		}
+		if vClusterStandalone == nil {
+			return nil, fmt.Errorf("local standalone vCluster not found on this host")
+		}
+		return vClusterStandalone, nil
+	}
+
+	vClusterStandalone, err := getVClusterStandalone()
+	if err != nil {
+		log.Debugf("standalone runtime-name probe skipped: %v", err)
+	} else if vClusterStandalone != nil && name == vClusterStandalone.Name {
+		return vClusterStandalone, nil
+	}
+
+	return GetVCluster(ctx, context, name, namespace, log)
 }
 
 func GetVCluster(ctx context.Context, context, name, namespace string, log log.Logger) (*VCluster, error) {
@@ -705,6 +739,56 @@ func isScaledDown(object client.Object) bool {
 		return o.Spec.Replicas != nil && *o.Spec.Replicas == 0
 	}
 	return false
+}
+
+// getVClusterStandalone returns a vCluster for a standalone installation.
+// Detection relies on the systemd service file existing on the local filesystem,
+// so this only works when the CLI runs on the same host as the vCluster standalone.
+// Returns nil, nil when standalone is not detected.
+func getVClusterStandalone() (*VCluster, error) {
+	unitData, found, err := standaloneutil.DetectStandaloneHost()
+	if err != nil {
+		return nil, fmt.Errorf("detect standalone host: %w", err)
+	}
+	if !found {
+		//nolint:nilnil
+		return nil, nil
+	}
+
+	vConfig, err := vclusterconfig.LoadLocalStandaloneConfig("", nil)
+	if err != nil {
+		return nil, fmt.Errorf("load standalone config: %w", err)
+	}
+	kubeClientConfig, err := getStandaloneKubeClientConfig(vConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	version := standaloneutil.ParseEnvFromSystemdUnit(unitData, "VCLUSTER_VERSION")
+
+	// Use the systemd unit file's modification time as a proxy for cluster creation time.
+	fi, err := os.Stat(constants.VClusterStandaloneSystemdUnitFile)
+	if err != nil {
+		return nil, fmt.Errorf("stat standalone unit file: %w", err)
+	}
+	created := metav1.NewTime(fi.ModTime())
+
+	// Check if the systemd service is actually running.
+	status := StatusUnknown
+	out, err := exec.Command("systemctl", "is-active", constants.VClusterStandaloneSystemdServiceName).Output()
+	if err == nil && strings.TrimSpace(string(out)) == "active" {
+		status = StatusRunning
+	}
+
+	return &VCluster{
+		Name:          vConfig.Name,
+		Namespace:     constants.VClusterStandaloneSnapshotNamespace,
+		ClientFactory: kubeClientConfig,
+		Created:       created,
+		Version:       version,
+		Status:        status,
+		IsStandalone:  true,
+	}, nil
 }
 
 // isVirtualClusterInstanceResourceAvailable checks if VirtualClusterInstance resources from storage.loft.sh/v1 exist
--- a/pkg/cli/restore_helm.go
+++ b/pkg/cli/restore_helm.go
@@ -3,14 +3,21 @@ package cli
 import (
 	"context"
 	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/loft-sh/log"
 	"github.com/loft-sh/vcluster/pkg/cli/find"
 	"github.com/loft-sh/vcluster/pkg/cli/flags"
+	vclusterconfig "github.com/loft-sh/vcluster/pkg/config"
+	"github.com/loft-sh/vcluster/pkg/constants"
 	"github.com/loft-sh/vcluster/pkg/lifecycle"
+	"github.com/loft-sh/vcluster/pkg/pro"
 	"github.com/loft-sh/vcluster/pkg/snapshot"
 	"github.com/loft-sh/vcluster/pkg/snapshot/pod"
+	standaloneutil "github.com/loft-sh/vcluster/pkg/util/standalone"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -21,17 +28,29 @@ const (
 	RestoreResourceQuota = "vcluster-restore"
 )
 
-func Restore(ctx context.Context, args []string, globalFlags *flags.GlobalFlags, snapshotOpts *snapshot.Options, pod *pod.Options, newVCluster, restoreVolumes bool, log log.Logger) error {
+func Restore(ctx context.Context, args []string, globalFlags *flags.GlobalFlags, snapshotOpts *snapshot.Options, podOpts *pod.Options, newVCluster, restoreVolumes bool, log log.Logger) error {
 	// init kube client and vCluster
 	vCluster, kubeClient, restConfig, err := initSnapshotCommand(ctx, args, globalFlags, snapshotOpts, log, true)
 	if err != nil {
 		return err
 	}
 
-	return restoreVCluster(ctx, kubeClient, restConfig, vCluster, snapshotOpts, pod, newVCluster, restoreVolumes, log)
+	return restoreVCluster(ctx, kubeClient, restConfig, vCluster, snapshotOpts, podOpts, newVCluster, restoreVolumes, log)
 }
 
 func restoreVCluster(ctx context.Context, kubeClient *kubernetes.Clientset, restConfig *rest.Config, vCluster *find.VCluster, snapshotOpts *snapshot.Options, podOptions *pod.Options, newVCluster bool, restoreVolumes bool, log log.Logger) error {
+	cmdArgs := []string{"restore"}
+	if newVCluster {
+		cmdArgs = append(cmdArgs, "--new-vcluster")
+	}
+	if restoreVolumes {
+		cmdArgs = append(cmdArgs, "--restore-volumes")
+	}
+
+	if vCluster.IsStandalone {
+		return restoreStandaloneVCluster(ctx, vCluster, snapshotOpts, cmdArgs, log)
+	}
+
 	// pause vCluster
 	log.Infof("Pausing vCluster %s", vCluster.Name)
 	err := pauseVCluster(ctx, kubeClient, vCluster, log)
@@ -49,15 +68,71 @@ func restoreVCluster(ctx context.Context, kubeClient *kubernetes.Clientset, rest
 	}()
 
 	// set missing pod options and run snapshot restore pod
-	command := []string{"/vcluster", "restore"}
-	if newVCluster {
-		command = append(command, "--new-vcluster")
-	}
-	if restoreVolumes {
-		command = append(command, "--restore-volumes")
+	command := append([]string{"/vcluster"}, cmdArgs...)
+	return pod.RunSnapshotPod(ctx, restConfig, kubeClient, command, vCluster, podOptions, snapshotOpts, log)
+}
+
+// restoreStandaloneVCluster stops the standalone service, invokes the vcluster binary
+// directly to perform the restore, and restarts the service. The CLI must run on the
+// same host as the standalone installation because it needs filesystem access to the
+// binary and config.
+func restoreStandaloneVCluster(ctx context.Context, vCluster *find.VCluster, snapshotOpts *snapshot.Options, cmdArgs []string, log log.Logger) error {
+	vClusterConfig, err := vclusterconfig.LoadLocalStandaloneConfig("", nil)
+	if err != nil {
+		return fmt.Errorf("load standalone config: %w", err)
 	}
 
-	return pod.RunSnapshotPod(ctx, restConfig, kubeClient, command, vCluster, podOptions, snapshotOpts, log)
+	sm, err := standaloneutil.NewServiceManager()
+	if err != nil {
+		return err
+	}
+
+	if err := pro.CheckStandaloneHA(ctx, vClusterConfig); err != nil {
+		return err
+	}
+
+	log.Infof("Stopping vCluster service")
+	if err := sm.Stop(); err != nil {
+		return fmt.Errorf("stop vCluster service: %w", err)
+	}
+
+	restoreErr := runRestoreBinary(vClusterConfig, snapshotOpts, cmdArgs)
+
+	log.Infof("Starting vCluster service")
+	if startErr := sm.Start(); startErr != nil {
+		return fmt.Errorf("restore succeeded but failed to restart vCluster service (cluster is stopped): %w", startErr)
+	}
+	return restoreErr
+}
+
+func runRestoreBinary(vClusterConfig *vclusterconfig.VirtualClusterConfig, snapshotOpts *snapshot.Options, args []string) error {
+	binaryPath := filepath.Join(vClusterConfig.ControlPlane.Standalone.DataDir, "bin", "vcluster")
+	if _, err := os.Stat(binaryPath); os.IsNotExist(err) {
+		// Fall back to the currently executing binary (e.g. during development or
+		// non-standard installs where the binary is not in the data directory).
+		self, err := os.Executable()
+		if err != nil {
+			return fmt.Errorf("binary not found at %s and cannot determine current executable: %w", binaryPath, err)
+		}
+		binaryPath = self
+	}
+
+	optionsString, err := pod.ToOptionsString(snapshotOpts)
+	if err != nil {
+		return fmt.Errorf("serialise snapshot options: %w", err)
+	}
+
+	env := append(os.Environ(),
+		"VCLUSTER_NAME="+vClusterConfig.Name,
+		constants.VClusterStandaloneEnvVar+"=true",
+		constants.VClusterStorageOptionsEnv+"="+optionsString,
+	)
+
+	cmd := exec.Command(binaryPath, args...)
+	cmd.Env = env
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
 }
 
 func pauseVCluster(ctx context.Context, kubeClient *kubernetes.Clientset, vCluster *find.VCluster, log log.Logger) error {
--- a/pkg/cli/snapshot_helm.go
+++ b/pkg/cli/snapshot_helm.go
@@ -12,6 +12,7 @@ import (
 	"github.com/loft-sh/vcluster/pkg/cli/find"
 	"github.com/loft-sh/vcluster/pkg/cli/flags"
 	vclusterconfig "github.com/loft-sh/vcluster/pkg/config"
+	"github.com/loft-sh/vcluster/pkg/constants"
 	"github.com/loft-sh/vcluster/pkg/helm"
 	"github.com/loft-sh/vcluster/pkg/snapshot"
 	"github.com/loft-sh/vcluster/pkg/snapshot/pod"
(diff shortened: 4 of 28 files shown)


pkg/cli/restore_helm.go
```diff
@@ -49,15 +71,126 @@ func restoreVCluster(ctx context.Context, kubeClient *kubernetes.Clientset, rest
...
+	log.Infof("Stopping vCluster service")
+	if err := sm.Stop(); err != nil {
+		return fmt.Errorf("stop vCluster service: %w", err)
+	}
+
+	restoreErr := runRestoreBinary(vClusterConfig, snapshotOpts, cmdArgs)
+
+	log.Infof("Starting vCluster service")
+	if startErr := sm.Start(); startErr != nil {
+		return fmt.Errorf("restore succeeded but failed to restart vCluster service (cluster is stopped): %w", startErr)
```
jjaferson, line 103: We haven't checked whether the restore succeeded or not here, we should say that it failed to restart vCluster or check if the restore succeeded before returning this message here

pkg/cli/find/find.go
```diff
@@ -464,6 +465,15 @@ func findInContext(ctx context.Context, kubeClient kube.Interface, context, name
...
 	}
 
+	// Probe for a standalone vCluster. Errors are non-fatal: a normal cluster where
+	// the user lacks permission to GET default/kubernetes must not break vcluster list.
+	standaloneVCluster, err := getStandaloneVCluster(ctx, kubeClient, kubeClientConfig)
+	if err != nil {
+		log.GetInstance().Debugf("standalone probe skipped: %v", err)
+	} else if standaloneVCluster != nil && (name == "" || name == standaloneVCluster.Name) {
+		vclusters = append(vclusters, *standaloneVCluster)
+	}
```
FabianKramm: We want to get rid of these annotations and things inside the vCluster that tell too much about the configuration

pkg/cli/restore_helm.go
```diff
@@ -49,15 +71,126 @@ func restoreVCluster(ctx context.Context, kubeClient *kubernetes.Clientset, rest
...
+	name string
+}
+
+func (s *systemdServiceManager) Stop() error {
+	return exec.Command("systemctl", "stop", s.name).Run()
+}
+
+func (s *systemdServiceManager) Start() error {
+	return exec.Command("systemctl", "start", s.name).Run()
+}
```
FabianKramm: Why do we need a struct, an interface here for just stop / start, can we not just have this inline in the above function? Seems unnecessarily complex to me

pkg/constants/constants.go
```diff
@@ -46,6 +46,10 @@ var (
 	// DefaultVClusterConfigLocation is the default location of the vCluster config within the container
 	DefaultVClusterConfigLocation = "/var/lib/vcluster/config.yaml"
 
+	// StandaloneDefaultConfigPath is the config location for a standalone binary installation.
+	// Kept outside the data directory so it survives a data wipe or re-install.
+	StandaloneDefaultConfigPath = "/etc/vcluster/vcluster.yaml"
```
FabianKramm: We are then removing that constant from pro right? Right?

pkg/constants/standalone.go
```diff
@@ -3,4 +3,13 @@ package constants
 const (
 	VClusterStandaloneEndpointsAnnotation = "vcluster.loft.sh/standalone-endpoints"
 	VClusterStandaloneIPAddressEnvVar     = "VCLUSTER_STANDALONE_IP_ADDRESS"
+
+	// Set on default/kubernetes Service by the syncer so the CLI can detect
+	// standalone vClusters without any host-cluster context.
+	VClusterStandaloneNameAnnotation    = "vcluster.loft.sh/standalone-name"
+	VClusterStandaloneVersionAnnotation = "vcluster.loft.sh/standalone-version"
```
FabianKramm: Lets leave those two out

pkg/snapshot/client.go
```diff
@@ -29,7 +31,33 @@ func (c *Client) Run(ctx context.Context) error {
...
+			return fmt.Errorf("parse standalone vCluster config: %w", err)
+		}
+		// config.ParseConfig does not apply Helm value defaults, so Standalone.Enabled,
+		// PrivateNodes.Enabled, and DataDir may be unset even though we are clearly in
+		// standalone mode. Mirror what vcluster-pro/pkg/standalone/config does.
+		vConfig.ControlPlane.Standalone.Enabled = true
+		vConfig.PrivateNodes.Enabled = true
+		if vConfig.ControlPlane.Standalone.DataDir == "" {
+			vConfig.ControlPlane.Standalone.DataDir = "/var/lib/vcluster"
+		}
```
FabianKramm: This should reuse the proper config loading from standalone here so that we are not maintaining all of this in 2 places

pkg/snapshot/client.go
```diff
@@ -29,7 +31,33 @@ func (c *Client) Run(ctx context.Context) error {
...
+		// Embedded etcd requires VCLUSTER_STANDALONE_IP_ADDRESS to identify the local peer.
+		if vConfig.BackingStoreType() == vclusterconfig.StoreTypeEmbeddedEtcd {
+			if _, ok := os.LookupEnv(constants.VClusterStandaloneIPAddressEnvVar); !ok {
+				hostIP, err := apinet.ChooseHostInterface()
+				if err != nil {
+					return fmt.Errorf("determine host IP for embedded etcd peer (set %s to override): %w", constants.VClusterStandaloneIPAddressEnvVar, err)
+				}
+				os.Setenv(constants.VClusterStandaloneIPAddressEnvVar, hostIP.String())
+			}
+		}
```
FabianKramm, line 40: Same here, lets reuse it from pro instead of copying

pkg/snapshot/restoreclient.go
```diff
@@ -129,7 +135,41 @@ func (o *RestoreClient) Run(ctx context.Context) (retErr error) {
...
+			return fmt.Errorf("parse standalone vCluster config: %w", err)
+		}
+		// config.ParseConfig does not apply Helm value defaults, so Standalone.Enabled,
+		// PrivateNodes.Enabled, and DataDir may be unset even though we are clearly in
+		// standalone mode. Mirror what vcluster-pro/pkg/standalone/config does.
+		vConfig.ControlPlane.Standalone.Enabled = true
+		vConfig.PrivateNodes.Enabled = true
+		if vConfig.ControlPlane.Standalone.DataDir == "" {
+			vConfig.ControlPlane.Standalone.DataDir = "/var/lib/vcluster"
+		}
```
FabianKramm: Same as above, lets reuse, not copy

pkg/snapshot/restoreclient.go
```diff
@@ -129,7 +135,41 @@ func (o *RestoreClient) Run(ctx context.Context) (retErr error) {
...
+		// Embedded etcd requires VCLUSTER_STANDALONE_IP_ADDRESS to identify the local peer.
+		if vConfig.BackingStoreType() == vclusterconfig.StoreTypeEmbeddedEtcd {
+			if _, ok := os.LookupEnv(constants.VClusterStandaloneIPAddressEnvVar); !ok {
+				hostIP, err := apinet.ChooseHostInterface()
+				if err != nil {
+					return fmt.Errorf("determine host IP for embedded etcd peer (set %s to override): %w", constants.VClusterStandaloneIPAddressEnvVar, err)
+				}
+				os.Setenv(constants.VClusterStandaloneIPAddressEnvVar, hostIP.String())
+			}
+		}
```
FabianKramm: Same as above

FabianKramm (changes requested):

pkg/cli/snapshot_helm.go
```diff
@@ -175,11 +180,17 @@ func checkIfVClusterSupportsSnapshotRequests(vCluster *find.VCluster, log log.Lo
...
 	var vClusterConfig *vclusterconfig.VirtualClusterConfig
-	if snapshotOpts.Release.Values != nil {
+	if snapshotOpts.Release != nil && snapshotOpts.Release.Values != nil {
 		vClusterConfig, err = vclusterconfig.ParseConfigBytes(snapshotOpts.Release.Values, vCluster.Name, nil)
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse vcluster config: %w", err)
 		}
+	} else if vCluster.IsStandalone {
+		// Standalone config lives on disk, not in a Kubernetes Secret.
+		vClusterConfig, err = vclusterconfig.ParseConfig(constants.StandaloneDefaultConfigPath, vCluster.Name, nil)
```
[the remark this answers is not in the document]
jjaferson: This is expected

pkg/snapshot/restoreclient.go
```diff
@@ -295,10 +338,53 @@ func (o *RestoreClient) createRestoreRequest(ctx context.Context, vConfig *confi
...
+		secret.Name = fmt.Sprintf("%s-restore-request-%s", vConfig.Name, rand.String(5))
+		secret.TypeMeta = metav1.TypeMeta{APIVersion: "v1", Kind: "Secret"}
+		secret.UID = uuid.NewUUID()
+		secret.CreationTimestamp = metav1.Now()
+
+		secretBuf := &bytes.Buffer{}
+		if err = encoder.Encode(secret, secretBuf); err != nil {
+			return fmt.Errorf("encode restore secret: %w", err)
+		}
+		if _, err = o.etcdClient.Put(ctx, fmt.Sprintf("/registry/secrets/%s/%s", namespace, secret.Name), secretBuf.Bytes()); err != nil {
```
jjaferson, line 328: it's more of a question/curiosity: I am guessing vcluster service is stopped kube-apiserver will be down and etcd would be too?

jjaferson (commented):
Code looks good, I've just left a few suggestions

pkg/cli/restore_helm.go
jjaferson: IMO it makes the code more readable and portable in case we want to support other types that is not `systemctl` as well it's easier to mock for unit tests

FabianKramm: When we want to do that we can still do it, but adding all of this complexity without needing it is what we shouldn't do

pkg/cli/restore_helm.go
```diff
@@ -49,15 +71,126 @@ func restoreVCluster(ctx context.Context, kubeClient *kubernetes.Clientset, rest
...
+
+// newServiceManager returns a systemd-based service manager when on Linux with systemd
+// available. Returns an error on other platforms or when the service is not running.
+func newServiceManager() (serviceManager, error) {
+	if runtime.GOOS != "linux" {
+		return nil, fmt.Errorf("standalone restore is only supported on Linux (current OS: %s)", runtime.GOOS)
+	}
+
+	out, err := exec.Command("systemctl", "is-active", standaloneServiceName).Output()
```
mfranczy: I think you should execute `systemctl cat` ([or just check the file](https://github.com/loft-sh/vcluster/blob/main/pkg/cli/standalone/platform.go#L72)). Otherwise you require the standalone to be up and running for the snapshot restore. Unless there is a specific reason you need the standalone to be up and running?

To my understanding we should handle two use cases, restore when:
* vCluster service is down (e.g., manually shutdown by the user)
* vCluster service is up and we stop it for the restore procedure

pkg/cli/snapshot_helm.go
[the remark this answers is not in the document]
mfranczy: But users can set a different config path than `/etc/vcluster/vcluster.yaml`.

https://github.com/loft-sh/vcluster-pro/blob/main/pkg/standalone/config/config.go#L57-L60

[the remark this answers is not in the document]
jjaferson: The comment from codex is related to remote access for standalone snapshot which is not supported at the moment. 

Regarding to users being able to set a different config path where is this information stored? We'll need to know it when running the snapshot command via the CLI

pkg/cli/restore_helm.go
nprokopic: Basically what Jose wrote, plus to try to contain systemd references in a single place, so it doesn't escalate and becomes harder to refactor / add other ways to start/stop later.

I wanted to add unit tests here (that would actually make use of this), but didn't get to that since we're on the clock here.

I'll simplify this now and remove the interface, and get back to it in some follow-up PR when I work on tests more.

pkg/cli/restore_helm.go
nprokopic: Generally speaking, I think restore should be able to run even when the cluster is down (not sure if it really can do that today for non-standalone). Good catch, thanks!

For now I am removing the whole service manager struct, see comments [here](https://github.com/loft-sh/vcluster/pull/3658/changes#r2913785642), but I'll have this in mind for later when/if I add this back.

pkg/cli/snapshot_helm.go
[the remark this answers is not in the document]
nprokopic: > This is expected

Yeah, user must run commands from the same host where the standalone is running.

> But users can set a different config path than /etc/vcluster/vcluster.yaml.

True, will fix this.

> Regarding to users being able to set a different config path where is this information stored? We'll need to know it when running the snapshot command via the CLI

In general IIRC it's passed here https://github.com/loft-sh/vcluster/blob/e214d59551d7bcc68da3fcc4597497035d6563ea/cmd/vcluster/cmd/start.go#L60, but gotta check if I can get the correct value somehow from the snapshot command.

pkg/constants/constants.go
nprokopic: Yeah, good catch, I didn't remove it from the other side yet.

pkg/snapshot/client.go
nprokopic: Yeah 💯 I initially thought that existing `config.ParseConfig` would do everything, but then I found the additional standalone-specific config loading later. Will update.

pkg/snapshot/restoreclient.go
nprokopic: When restore runs, it stops vCluster (which stops api-server, etcd, etc), but then (depending on backing store type) it can also start backing store (kine/etcd) so it can write into it, see [here](https://github.com/loft-sh/vcluster/blob/e214d59551d7bcc68da3fcc4597497035d6563ea/pkg/snapshot/store.go#L23).

pkg/cli/restore_helm.go
mfranczy: I forgot that currently we restore snapshots by writing to db. So for now it must stay active.

pkg/cli/find/find.go
nprokopic: Updated standalone check so it doesn't use annotations.

pkg/constants/standalone.go
nprokopic: Removed.

pkg/snapshot/client.go
jjaferson: > This should reuse the proper config loading from standalone

Hey @FabianKramm when you say reuse the config loading from standalone, do you mean move things from vcluster pro to OSS and then reusing it in vcluster pro? I see ParseConfig in vcluster pro already imports vcluster OSS https://github.com/loft-sh/vcluster-pro/blob/main/pkg/standalone/config/config.go#L19

pkg/cli/snapshot_helm.go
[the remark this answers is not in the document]
jjaferson: That might be an issue with the entire snapshot package as the snapshot client reads the vcluster config from the default path [snapshot client](https://github.com/loft-sh/vcluster/blob/main/pkg/snapshot/client.go#L30) and [restore client](https://github.com/loft-sh/vcluster/blob/main/pkg/snapshot/restoreclient.go#L130)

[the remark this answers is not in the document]
nprokopic: So currently snapshot and restore effectively works only when the cluster beign snapshotted/restored is using the default config 🙈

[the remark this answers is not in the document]
jjaferson: Yes, question here is, do we want to change it?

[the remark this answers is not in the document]
nprokopic: Oh, wait, for a non-standalone cluster (e.g. shared nodes, private nodes), this path should be inside of a pod, so this is fine for non-standalone.

For figuring out config path for standalone, you could parse it from the systemd service file, similar to [this](https://github.com/loft-sh/vcluster/blob/3cb8f81fb3fa081ee0593775b24b53617c8abc15/pkg/cli/find/find.go#L770), just parse another line where `--config <some path>` is set.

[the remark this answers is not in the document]
jjaferson: >Oh, wait, for a non-standalone cluster (e.g. shared nodes, private nodes), this path should be inside of a pod, so this is fine for non-standalone.

Not really, if we use the new [create](https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/snapshot/create.go#L49) snapshot command it doesn't run from a pod afaik see [here](https://github.com/loft-sh/vcluster/blob/main/pkg/cli/snapshot_helm.go#L54)

>For figuring out config path for standalone, you could parse it from the systemd service file, similar to [this](https://github.com/loft-sh/vcluster/blob/3cb8f81fb3fa081ee0593775b24b53617c8abc15/pkg/cli/find/find.go#L770), just parse another line where --config <some path> is set.

~~We can just get the value from the `--config` flag from the `globalFlags *flags.GlobalFlags` and set it to the [vCluster](https://github.com/loft-sh/vcluster/blob/main/pkg/cli/find/find.go#L37) struct that is passed to [createSnapshotRequest](https://github.com/loft-sh/vcluster/blob/main/pkg/cli/snapshot_helm.go#L149C6-L149C27)~~

Disregard my comment better go with your approach 👍

pkg/snapshot/client.go
mfranczy: Yes, since vcluster commands now rely on standalone config detection, standalone config path resolution and standalone defaults should be implemented in OSS.

I see you have already pushed the changes, but they are incomplete or invalid. More details in another comment.

mfranczy (changes requested):
The issues I found:

1. `config.ParseConfig()` now auto-detects the standalone config path in OSS (`vcluster/pkg/config/parse.go:27`), but standalone normalization is still done later in command-specific fallback code (`vcluster/pkg/snapshot/restoreclient.go:145`, `vcluster/pkg/snapshot/client.go:31`). That means parsing can succeed without returning a fully normalized standalone config. Standalone path resolution and standalone defaults should be centralized in OSS parsing, and `vcluster-pro` should only add its own extra logic on top.

2. `VCLUSTER_STANDALONE_IP_ADDRESS` is a runtime detail of standalone embedded etcd, but OSS snapshot/restore code currently knows about it explicitly and fails when it is missing (`vcluster/pkg/cli/restore_helm.go:174`, `vcluster/pkg/snapshot/restoreclient.go:162`, `vcluster/pkg/snapshot/client.go:42`). These checks are only early guards and do not actually resolve or prepare anything. That leaks a pro-specific prerequisite into OSS orchestration. A cleaner design would be to remove these OSS guards and let `vcluster-pro` resolve standalone peer identity internally, while still honoring the env var as an override if it is set. This has to be fixed in pro.

pkg/cli/find/find.go
```diff
@@ -699,6 +714,82 @@ func GetPodStatus(pod *corev1.Pod) string {
...
+	name := rawConfig.CurrentContext
+	if idx := strings.Index(name, "@"); idx >= 0 {
+		name = name[idx+1:]
+	}
+
+	// 3. Version: parse VCLUSTER_VERSION from the systemd unit file.
+	version := parseEnvFromSystemdUnit(unitData, "VCLUSTER_VERSION")
+
+	// 4. Created: use the kubernetes service creation timestamp.
+	svc, err := kubeClient.CoreV1().Services("default").Get(ctx, "kubernetes", metav1.GetOptions{})
```
mfranczy: Remove this. Take mtime from the systemd unit file. I know that won't be a real creation time. In the future we can improve this and dump some statistics at the start if needed. For now mtime is good enough.

The function must be free of any calls to vCluster, if you face API / etcd issues, you won't be able to find standalone cluster.

jjaferson:
> 2. `VCLUSTER_STANDALONE_IP_ADDRESS` is a runtime detail of standalone embedded etcd, but OSS snapshot/restore code currently knows about it explicitly and fails when it is missing (`vcluster/pkg/cli/restore_helm.go:174`, `vcluster/pkg/snapshot/restoreclient.go:162`, `vcluster/pkg/snapshot/client.go:42`). These checks are only early guards and do not actually resolve or prepare anything. That leaks a pro-specific prerequisite into OSS orchestration. A cleaner design would be to remove these OSS guards and let `vcluster-pro` resolve standalone peer identity internally, while still honoring the env var as an override if it is set. This has to be fixed in pro.

I don't see any harm in vcluster CLI failing snapshot and restore commands if the `VCLUSTER_STANDALONE_IP_ADDRESS` env var is not set, that would prevent a request that will eventually fail.

mfranczy:
> I don't see any harm in vcluster CLI failing snapshot and restore commands if the VCLUSTER_STANDALONE_IP_ADDRESS env var is not set, that would prevent a request that will eventually fail.

The standard path will always fail because the env is set in the standalone start. Let's remove this check and handle this in pro.

pkg/cli/find/find.go
```diff
@@ -464,6 +470,15 @@ func findInContext(ctx context.Context, kubeClient kube.Interface, context, name
 		}
 	}
 
+	// Probe for a standalone vCluster. Errors are non-fatal: a normal cluster where
+	// the user lacks permission to GET default/kubernetes must not break vcluster list.
+	standaloneVCluster, err := getStandaloneVCluster(ctx, kubeClient, kubeClientConfig)
```
mfranczy: `getStandaloneVCluster()` is currently too deep in the generic discovery flow.
The path is `GetVCluster() -> ListVClusters() -> ListOSSVClusters() -> findInContext()`, and inside it is only probed after the normal StatefulSet/Deployment discovery paths have already run. That means a standalone lookup is not a local fast path today, even though standalone is a special local only mode for snapshots.

The practical problem is that before standalone is even considered, the code already performs many k8s client requests and goes through multiple failure paths that are unrelated to standalone detection at all. If one of those earlier discovery steps fails, we may never even reach the standalone probe.

I would advise to create a new `GetVClusterOrLocalStandalone` (can be a different name) that probes if there is a local standalone over getStandaloneVCluster(), if detection fails it should fallback to GetVCluster.

jjaferson: It makes sense, but we could reuse `GetVCluster()` by trying to get a standalone vcluster first, and if the detection fails we can fallback to get a normal vcluster `ListVClusters() -> ListOSSVClusters() -> findInContext()`?

mfranczy: I thought about this, but GetVCluster is used by many commands that don’t necessarily support standalone. It's better if a command explicitly asks for standalone if it supports it. I don't want to change the behavior for all commands that use GetVCluster.

mfranczy:
@codex review

pkg/cli/find/find.go
```diff
@@ -49,6 +53,8 @@ type VCluster struct {
 	Status                 Status
 	Context                string
 	Version                string
+	IsStandalone           bool
+	VClusterConfigPath     string
```
mfranczy: `VClusterConfigPath` should be removed. Code in find should only determine whether this is a standalone vCluster. Config path resolution is currently spread across multiple places, which mixes standalone detection with path guessing. The flow should be: detect standalone -> resolve possible standalone config paths (including user provided env and args) -> normalize standalone config.

mfranczy: As discussed in the call, I will take care of the config changes.

pkg/cli/find/find.go
jjaferson: I've pushed the changes

cmd/vcluster/cmd/snapshot/restore.go
```diff
@@ -18,12 +20,17 @@ func NewRestoreCommand() *cobra.Command {
...
+				return err
+			}
+
 			envOptions, err := snapshot.ParseOptionsFromEnv()
 			if err != nil {
 				return fmt.Errorf("failed to parse options from environment: %w", err)
 			}
 			restoreClient := snapshot.NewRestoreClient(*envOptions, restoreVolumes, newVCluster)
-			return restoreClient.Run(cmd.Context())
+			return restoreClient.Run(cmd.Context(), vConfig)
```
jjaferson, line 33: I'd pass the vCluster config as part of the NewRestoreClient as we do with the `snapshotClient` and the restoreClient already a [vCluster field](https://github.com/loft-sh/vcluster/blob/main/pkg/snapshot/restoreclient.go#L51)

pkg/config/load.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+// resolveStandaloneConfigFromDefaults resolves the config path from shared
+// standalone default locations only. This intentionally does not inspect systemd.
+// The order mirrors the previous pro standalone parser behavior:
+// standalone default first, then container default.
+func resolveStandaloneConfigFromDefaults() (string, error) {
+	return resolveStandaloneConfigFromCandidates([]string{
+		constants.StandaloneDefaultConfigPath,
+		constants.DefaultVClusterConfigLocation,
```
jjaferson, line 156: if this is standalone specific, why are we passing the vCluster in cluster path?

pkg/config/load.go
```diff
@@ -0,0 +1,203 @@
...
+// back to local standalone host detection via the systemd unit marker.
+func LoadRuntimeConfig(vClusterName string) (*VirtualClusterConfig, error) {
+	// Respect explicit runtime context from the caller before attempting any
+	// local host detection. This allows parent processes to force standalone or
+	// non-standalone behavior for child vcluster commands.
+	if standaloneEnv, ok := os.LookupEnv(constants.VClusterStandaloneEnvVar); ok {
+		if standaloneEnv == "true" {
+			vConfig, err := LoadStandaloneConfig(vClusterName, "", nil)
+			if err != nil {
+				return nil, fmt.Errorf("parse standalone vCluster config: %w", err)
```
jjaferson: nit: it should say error loading standalone vcluster config

pkg/config/load.go
```diff
@@ -0,0 +1,203 @@
...
+			}
+			return vConfig, nil
+		}
+
+		vConfig, err := LoadConfig(vClusterName, "", nil)
+		if err != nil {
+			return nil, fmt.Errorf("parse vCluster config: %w", err)
```
jjaferson: Nit: same here error loading vCluster config

pkg/config/load.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+		return vConfig, nil
+	}
+
+	_, found, err := standaloneutil.DetectStandaloneHost()
```
jjaferson: Do we need this check here, if it's already done by `resolveStandaloneConfigFromSystemd` [here](https://github.com/loft-sh/vcluster/pull/3658/changes/3d0afae126657619de997a9e8b747ea37f787cee#diff-c8b9958e5d26ca9d8e0b3c614965c7cb4a6995b7d49042d0f4cb333ebe86d474R133)

pkg/config/load.go
```diff
@@ -0,0 +1,203 @@
...
+		vConfig, err := LoadStandaloneConfigFromSystemd(vClusterName, nil)
+		if err != nil {
+			return nil, fmt.Errorf("parse standalone vCluster config from systemd: %w", err)
+		}
+		return vConfig, nil
+	}
+
+	vConfig, err := LoadConfig(vClusterName, "", nil)
+	if err != nil {
+		return nil, fmt.Errorf("parse vCluster config: %w", err)
```
jjaferson: nit: error message here should be error loading vcluster config path

jjaferson (commented):
Code looks good but I have a few comments there

cmd/vcluster/cmd/snapshot/restore.go
mfranczy: I don't think vConfig should be a part of restoreClient.

It is only needed for Run(). RestoreClient is also used for GetSnapshotRequest(), and that path does not need config at all. If we move it into the constructor, we either force callers to pass config where it is unused, or we allow nil there which makes the contract weaker. Keeping it on Run(ctx, vConfig) keeps that dependency explicit and only where it is actually needed.

But but... I have to completely remove it from the restoreClient so the comment is actually useful after all.

pkg/config/load.go
mfranczy: It's possible that standalone may have a config in this default. That's the same logic as it was in pro.

pkg/config/load.go
mfranczy: Fixed.

pkg/config/load.go
mfranczy: Fixed.

pkg/config/load.go
mfranczy: Good catch, we don't. Fixed.

pkg/config/load.go
mfranczy: Fixed.

cmd/vcluster/cmd/snapshot/restore.go
jjaferson: hmmm :thinking: shouldn't we remove the vcluster config from the snapshot client struct too?

mfranczy: We should. Fixed.

pkg/cli/find/find.go
```diff
@@ -130,6 +135,35 @@ func GetPlatformVCluster(ctx context.Context, platformClient platform.Client, na
...
+// GetVClusterStandaloneOrVCluster supports both regular host-cluster lookup and
+// local standalone lookup. The CLI selector "local-standalone" explicitly targets
+// the standalone service on this host and makes standalone probe errors fatal.
+// For real runtime names, standalone lookup is only an opportunistic shortcut:
+// if the local standalone runtime name matches the requested name, use it;
+// otherwise fall back to regular host-cluster discovery so in-cluster vClusters
+// with arbitrary names keep working.
+func GetVClusterStandaloneOrVCluster(ctx context.Context, context, name, namespace string, log log.Logger) (*VCluster, error) {
+	if name == constants.VClusterStandaloneCLISelector {
+		vClusterStandalone, err := getVClusterStandalone()
```
jjaferson: Do we need to call `getVClusterStandalone()` twice here? I think it could be simplified to have something like 


```
	vClusterStandalone, err := getVClusterStandalone()
	if err != nil && name == constants.VClusterStandaloneCLISelector {
		if vClusterStandalone == nil{
			return nil, fmt.Errorf("local standalone vCluster not found on this host")
		} else {
			return nil, err
		}
	} else if err != nil {
		log.Debugf("standalone runtime-name probe skipped: %v", err)
	} else if vClusterStandalone != nil && name == vClusterStandalone.Name {
		return vClusterStandalone, nil
	}
```

mfranczy: You're right, we don't need two calls. However, this should be simplified to:
```
vClusterStandalone, err := getVClusterStandalone()
if name == constants.VClusterStandaloneCLISelector {
      if err != nil {
              return nil, err
      }
      if vClusterStandalone == nil {
              return nil, fmt.Errorf("local standalone vCluster not found on this host")
      }
      return vClusterStandalone, nil
}

if err != nil {
      log.Debugf("standalone runtime-name probe skipped: %v", err)
} else if vClusterStandalone != nil && name == vClusterStandalone.Name {
      return vClusterStandalone, nil
}
```
Otherwise there is fallback to names search which should not happen for a selector.

pkg/config/load.go
```diff
@@ -117,7 +123,7 @@ func loadNormalizedStandaloneConfig(name, path string, setValues []string) (*Vir
 	cfg.ControlPlane.Standalone.Enabled = true
 	cfg.PrivateNodes.Enabled = true
 	if cfg.ControlPlane.Standalone.DataDir == "" {
-		cfg.ControlPlane.Standalone.DataDir = constants.StandaloneDefaultDataDir
+		cfg.ControlPlane.Standalone.DataDir = constants.VClusterStandaloneDefaultDataDir
 	}
 
 	err = ValidateConfigAndSetDefaults(cfg)
```
jjaferson: I know we are setting some values for standalone and private nodes but do we still need to call this function again? it's called in [Parse](https://github.com/loft-sh/vcluster/blob/main/pkg/config/parse.go#L62)

mfranczy: We don't, we can remove it after standalone normalization.

mfranczy (approved):

FabianKramm (approved):

PR #3698, opened 2026-03-16
chore(e2e-next): Migrate e2e_certs to e2e-next framework

pascalbreuninger:
## Summary
- Migrates `test/e2e_certs/` (certificate rotation tests) to the `e2e-next/test_certs/` framework
- Consolidates three old `Describe` blocks into a single `Ordered` `Describe` with four `Context` blocks
- Adds `CertsVCluster` cluster definition with dedicated values YAML
- Registers suite in `e2e_suite_test.go` with proper setup/cleanup wiring

## Test plan
- [ ] `go build ./e2e-next/...` compiles cleanly
- [ ] `go vet ./e2e-next/...` passes
- [ ] Run certs e2e suite against a kind cluster
- [ ] Verify old `test/e2e_certs/` files are fully removed

---

<details>
<summary>📋 E2E Migration Validator Notes</summary>

### 🔴 Must Fix

- **Background proxy container leak in `connectVCluster`** (lines 342-387):
  `cli.ConnectHelm` with `BackgroundProxy: true` starts a Docker container.
  The `cleanupFn` only calls `os.Remove(tmpPath)` on the kubeconfig file — it never stops
  the Docker container. This function is called 4 times, potentially leaking 4 containers
  per test run. Fix: add container cleanup to the cleanup function, or use `BackgroundProxy: false`.

### 🟡 Should Fix

- **Dropped "certs check" coverage**: The old test called `certs check <name>` to verify the check subcommand works both before and after rotation. Add at least one `It` that calls the check command and verifies it succeeds.

- **`labels.PR` on a slow, destructive test** (line 40): This suite performs multiple cert rotations with pod-readiness waits (2+ minutes). Consider removing `labels.PR` and running only with `labels.Certs` in a dedicated job.

### 🟢 Nice to Have

- **`os.Setenv` defer scope** (lines 161-166): The `defer` inside the `By()` closure fires at the end of that anonymous function, not at the end of the `It()`. A comment would prevent future confusion.
- `vclusterNS` and `vclusterName` could be `const` instead of `var`.
- The `connectVCluster` helper (45 lines) could be moved to a shared helper package if reused.

</details>


e2e-next/test_certs/test_rotate.go
```diff
@@ -0,0 +1,411 @@
...
+			labelSelector = "app=vcluster,release=" + clusters.CertsVClusterName
+
+			// Fingerprints tracked across contexts
+			apiserverFingerprintBefore string
+			apiserverCertBefore        *x509.Certificate
+			caFingerprintBefore        string
+			caCertBefore               *x509.Certificate
+		)
+
+		BeforeAll(func(ctx context.Context) {
```
adriankabala, line 58: BeforeAll does not return context.Context. Inside an Ordered Describe, BeforeAll should return the enriched context so that It blocks receive any context modifications. In this case the hostClient is a captured var so it works, but the signature is inconsistent with the framework convention.

e2e-next/test_certs/test_rotate.go
```diff
@@ -0,0 +1,411 @@
...
+		AfterAll(func(ctx context.Context) {
+			By("Reconnecting to vCluster to restore suite proxy for subsequent tests", func() {
+				_, cleanup := connectVCluster(ctx, vclusterName, vclusterNS)
+				DeferCleanup(cleanup)
+			})
+		})
+	},
+)
+
+// waitForVClusterReady polls until all vCluster pods are running with all containers ready.
```
adriankabala, line 316: Missing GinkgoHelper() - waitForVClusterReady is a shared assertion helper that should call GinkgoHelper() so failure stack traces point to the call site. 
Also it can be moved to some common global helpers so other tests can use it.

e2e-next/test_certs/test_rotate.go
```diff
@@ -0,0 +1,411 @@
...
+			}
+		}
+	}).WithPolling(constants.PollingInterval).
+		WithTimeout(constants.PollingTimeoutLong).
+		Should(Succeed())
+}
+
+// connectVCluster establishes a fresh connection to the vCluster using a background proxy.
+// Returns a rest.Config and a cleanup function that removes the temp kubeconfig.
+func connectVCluster(ctx context.Context, name, namespace string) (*rest.Config, func()) {
```
adriankabala, line 342: Missing GinkgoHelper(), also can be moved to some Helper file

e2e-next/test_certs/test_rotate.go
```diff
@@ -0,0 +1,411 @@
...
+	tmpFile, err := os.CreateTemp("", "vcluster-kubeconfig-*.yaml")
+	Expect(err).NotTo(HaveOccurred())
+	tmpPath := tmpFile.Name()
+	tmpFile.Close()
+
+	cleanupFn := func() {
+		os.Remove(tmpPath)
```
adriankabala, line 349: Background proxy container leak - cleanupFn only removes the temp kubeconfig file via os.Remove(tmpPath). cli.ConnectHelm with BackgroundProxy: true starts a Docker container. The cleanup
  never stops it. This function is called up to 4 times per run, potentially leaking 4 Docker containers.
  Fix (option A - stop the container):
```
  func connectVCluster(ctx context.Context, name, namespace string) (*rest.Config, func(context.Context)) {
      // ... setup ...
      err = cli.ConnectHelm(ctx, options, globalFlags, name, nil, loftlog.Discard)
      Expect(err).NotTo(HaveOccurred())


      cleanupFn := func(ctx context.Context) {
          // Stop the background proxy container
          disconnectOptions := &cli.DisconnectOptions{}
          _ = cli.DisconnectHelm(ctx, disconnectOptions, name)
          os.Remove(tmpPath)
      }
      // ...
      return cfg, cleanupFn
  }
```
 or avoid containers entirely:

 ```
 options := &cli.ConnectOptions{
      BackgroundProxy: false,
      KubeConfig:      tmpPath,
  }
```

File: e2e-next/test_certs/test_rotate.go [hunk 5]
```diff
@@ -0,0 +1,411 @@
+package test_certs
```
adriankabala, line 1: Dropped "certs check" subcommand coverage - old test called `certs check` twice

Explanation: 

The old test in test/e2e_certs/certs/rotate.go called vcluster certs check <name> twice via the CLI subcommand:

  1. Before rotation (line 703): certsCmd.SetArgs([]string{"check", f.VClusterName}) - inside the It("checking current validity date of CA cert of vCluster") block
  2. After rotation (line 797): certsCmd.SetArgs([]string{"check", f.VClusterName}) - inside the It("priniting new expiry date and time of vCluster CA cert") block (note the typo "priniting" in the original)

  Both of these instantiate certscmd.NewCertsCmd(...), set args to ["check", vclusterName], call Execute(), and assert no error. This exercises the vcluster certs check CLI subcommand - verifying it can inspect a vCluster's certificates and
  report them as valid.

  The new migrated test in e2e-next/test_certs/test_rotate.go dropped both of these calls entirely. It validates certificate fingerprints and expiry dates directly by reading Kubernetes secrets, but it never invokes the certs check CLI
  subcommand.

e2e-next/test_certs/test_rotate.go
```diff
@@ -0,0 +1,411 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// Ordered justification: cert rotation is destructive and cumulative —
+// each context modifies the vCluster's certificates, and subsequent
+// contexts depend on the state left by prior ones.
+var _ = Describe("Certificate Rotation",
+	Ordered,
+	labels.PR,
```
adriankabala, line 40: labels.PR on a slow/destructive suite (multiple rotations + pod restarts) - maybe this should be in nightly build only?

PR #3706, opened 2026-03-16
test(e2e-next): add vind test spec

roehrijn:
/kind enhancement

resolves #

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

### DIFF
--- a/e2e-next/test_vind/test_docker.go
+++ b/e2e-next/test_vind/test_docker.go
@@ -17,22 +17,13 @@ import (
 )
 
 var _ = Describe("Docker driver", labels.Vind, labels.PR, func() {
-	var (
-		clusterName string
-		ctx         context.Context
-	)
+	var clusterName string
 
 	BeforeEach(func() {
-		ctx = context.Background()
 		clusterName = "e2e-vind-" + random.String(6)
-
-		DeferCleanup(func() {
-			cleanupCtx := context.Background()
-			_, _ = runVCluster(cleanupCtx, "delete", clusterName, "--driver", "docker", "--ignore-not-found")
-		})
 	})
 
-	It("should create, connect, and delete a vcluster", func() {
+	It("should create, connect, and delete a vcluster", func(ctx context.Context) {
 		var (
 			cpContainer  string
 			network      string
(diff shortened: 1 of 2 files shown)


e2e-next/test_vind/test_docker.go
```diff
@@ -0,0 +1,116 @@
...
+	)
+
+	BeforeEach(func() {
+		ctx = context.Background()
+		clusterName = "e2e-vind-" + random.String(6)
+
+		DeferCleanup(func() {
+			cleanupCtx := context.Background()
+			_, _ = runVCluster(cleanupCtx, "delete", clusterName, "--driver", "docker", "--ignore-not-found")
+		})
```
adriankabala: `_, _ = runVCluster(...)` swallows the error silently. Even in cleanup, we should at least log it - the `--ignore-not-found` flag handles the expected case on the CLI side, so any error here is unexpected and worth surfacing.

e2e-next/test_vind/test_docker.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+var _ = Describe("Docker driver", labels.Vind, labels.PR, func() {
+	var (
+		clusterName string
+		ctx         context.Context
+	)
+
+	BeforeEach(func() {
+		ctx = context.Background()
```
adriankabala: `context.Background()` bypasses Ginkgo's deadline propagation - the test timeout won't be able to cancel long-running operations. Use the Ginkgo-provided `ctx` parameter instead: `BeforeEach(func(ctx context.Context) { ... })`

e2e-next/test_vind/test_docker.go
roehrijn: grr, this whole `DeferCleanup` makes no sense at all. I removed it already in the morning but then asked Claude to rebase and it was there again. :-(

e2e-next/test_vind/test_docker.go
roehrijn: same. Only needed for the `DeferCleanup`  - whole block doesn't make any sense

PR #3712, opened 2026-03-17
ci: add scheduled check for homebrew tap version drift

Piotr1215:
## Summary

- Adds a daily GitHub Action that detects homebrew tap version drift for vCluster
- Checks both `vcluster` and `vcluster-experimental` formulas against the latest stable release
- Sends Slack alert to `#eng-releases` when versions diverge

## Why

PR #3613 added retry + fail-hard logic to guard against tap downgrades during releases, but drift can also happen outside the release pipeline (manual tap edits, goreleaser bugs). This scheduled check catches those cases.

## Details

- **Schedule**: Daily at 08:00 UTC (+ manual `workflow_dispatch`)
- **Release version**: Uses `gh release list` to find the latest *stable* release (skips pre-releases, matching goreleaser's `skip_upload: auto`)
- **Tap fetch**: Curl with retry + exponential backoff (same pattern as release.yaml)
- **Slack**: Uses `slackapi/slack-github-action@v2.1.1` with `SLACK_WEBHOOK_URL_ENG_RELEASES` secret
- **Org guard**: Only runs when `repository_owner == 'loft-sh'`

## New secret required

`SLACK_WEBHOOK_URL_ENG_RELEASES` — incoming webhook URL for the `#eng-releases` Slack channel.

## Test plan

- [x] `.github/tests/test-brew-tap-drift.sh` — all 7 assertions pass (version comparison logic)
- [x] Verified formula version extraction against live URLs:
  - `curl` on `loft-sh/homebrew-tap/main/Formula/vcluster.rb` → `0.33.0`
  - `curl` on `loft-sh/homebrew-tap/main/Formula/vcluster-experimental.rb` → `0.33.0`
- [x] Verified release tag fetch: `gh release list --repo loft-sh/vcluster` → `v0.33.0` (latest stable)
- [x] Confirmed versions match (`v0.33.0` == `v0.33.0`) — no false drift on current state
- [ ] End-to-end: trigger via `workflow_dispatch` after `SLACK_WEBHOOK_URL_ENG_RELEASES` secret is added

Closes DEVOPS-654

### DIFF
--- a/.github/tests/test-brew-tap-drift.sh
+++ b/.github/tests/test-brew-tap-drift.sh
@@ -1,55 +1,92 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
-# Tests the version comparison logic used in check-brew-tap-drift.yaml.
-# Validates that drift is correctly detected between release tags and tap versions.
+# Tests for .github/scripts/brew-tap-drift.sh
+# Sources the library and exercises its functions directly.
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${SCRIPT_DIR}/../scripts/brew-tap-drift.sh"
 
 PASS=0
 FAIL=0
 
-check_drift() {
+assert_drift() {
   local release="$1" tap="$2" expected_drifted="$3" description="$4"
 
-  # Strip v prefix for consistent comparison (mirrors workflow logic)
-  release="${release#v}"
-  tap="${tap#v}"
+  compare_versions "$release" "$tap"
 
-  if [[ "$release" != "$tap" ]]; then
-    actual_drifted="true"
-  else
-    actual_drifted="false"
-  fi
-
-  if [[ "$actual_drifted" == "$expected_drifted" ]]; then
-    printf "  PASS: %s (release=%s tap=%s drifted=%s)\n" "$description" "$release" "$tap" "$actual_drifted"
+  if [[ "$DRIFTED" == "$expected_drifted" ]]; then
+    printf "  PASS: %s (release=%s tap=%s drifted=%s)\n" \
+      "$description" "$(normalize_version "$release")" "$(normalize_version "$tap")" "$DRIFTED"
     PASS=$((PASS + 1))
   else
-    printf "  FAIL: %s (release=%s tap=%s expected_drifted=%s got=%s)\n" "$description" "$release" "$tap" "$expected_drifted" "$actual_drifted"
+    printf "  FAIL: %s (release=%s tap=%s expected=%s got=%s)\n" \
+      "$description" "$(normalize_version "$release")" "$(normalize_version "$tap")" "$expected_drifted" "$DRIFTED"
     FAIL=$((FAIL + 1))
   fi
 }
 
-printf "Testing brew tap drift detection logic\n\n"
+assert_experimental_drift() {
+  local release="$1" tap="$2" experimental="$3" expected_drifted="$4" description="$5"
+
+  compare_versions "$release" "$tap" "$experimental"
+
+  if [[ "$DRIFTED" == "$expected_drifted" ]]; then
+    printf "  PASS: %s (drifted=%s)\n" "$description" "$DRIFTED"
+    PASS=$((PASS + 1))
+  else
+    printf "  FAIL: %s (expected=%s got=%s)\n" "$description" "$expected_drifted" "$DRIFTED"
+    FAIL=$((FAIL + 1))
+  fi
+}
+
+assert_normalize() {
+  local input="$1" expected="$2" description="$3"
+  local actual
+  actual=$(normalize_version "$input")
+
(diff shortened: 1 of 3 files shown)


.github/workflows/check-brew-tap-drift.yaml
```diff
@@ -0,0 +1,131 @@
...
+            {
+              echo "details<<EOF"
+              echo -e "$drift"
+              echo "EOF"
+            } >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Notify eng-releases
+        if: steps.compare.outputs.drifted == 'true'
+        uses: slackapi/slack-github-action@v2.1.1
```
sydorovdmytro:
``` 
INFO audit: zizmor: 🌈 completed .github/workflows/check-brew-tap-drift.yaml
error[unpinned-uses]: unpinned action reference
   --> .github/workflows/check-brew-tap-drift.yaml:108:15
    |
108 |         uses: slackapi/slack-github-action@v2.1.1
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
    |
    = note: audit confidence → High
```

.github/workflows/check-brew-tap-drift.yaml
```diff
@@ -0,0 +1,131 @@
...
+                {
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": "${{ steps.compare.outputs.details }}\n\nLatest stable release: `${{ steps.release.outputs.tag }}`\nTap formula: <https://github.com/loft-sh/homebrew-tap/blob/main/Formula/vcluster.rb|vcluster.rb>\nWorkflow run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+                  }
+                }
+              ]
+            }
+          webhook: ${{ secrets.SLACK_WEBHOOK_URL_ENG_RELEASES }}
```
sydorovdmytro, line 94:
```warning[secrets-outside-env]: secrets referenced without a dedicated environment
   --> .github/workflows/check-brew-tap-drift.yaml:130:24
    |
 12 |   check-drift:
    |   ----------- this job
...
130 |           webhook: ${{ secrets.SLACK_WEBHOOK_URL_ENG_RELEASES }}
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ secret is accessed outside of a dedicated environment
    |
    = note: audit confidence → High```

File: .github/tests/test-brew-tap-drift.sh [hunk 3]
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
sydorovdmytro, line 1: would be nice to run this script on CI when we change the pipeline

.github/workflows/check-brew-tap-drift.yaml
```diff
@@ -0,0 +1,131 @@
...
+      - name: Get latest stable release tag
+        id: release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # List recent releases and pick the latest stable (non-prerelease) one.
+          # gh release view returns the latest *any* release, which may be a
+          # pre-release. We need the latest stable release to compare against
+          # the tap, since goreleaser's skip_upload:auto skips pre-releases.
+          TAG=$(gh release list --repo loft-sh/vcluster --limit 20 --json tagName,isPrerelease \
```
sydorovdmytro: `gh release list` supports some native filtering:
```
FLAGS
      --exclude-drafts         Exclude draft releases
      --exclude-pre-releases   Exclude pre-releases
```
So it could be 
```
TAG=$(gh release list --repo loft-sh/vcluster --limit 1 --exclude-pre-releases --exclude-drafts --json tagName --jq '.[0].tagName')
```

.github/workflows/check-brew-tap-drift.yaml
```diff
@@ -0,0 +1,131 @@
...
+          fi
+          echo "Tap versions: vcluster=v${VCLUSTER_VERSION}, experimental=${EXPERIMENTAL_VERSION:-(fetch failed)}"
+
+      - name: Compare versions
+        id: compare
+        env:
+          RELEASE_TAG: ${{ steps.release.outputs.tag }}
+          TAP_VCLUSTER: ${{ steps.tap.outputs.vcluster }}
+          TAP_EXPERIMENTAL: ${{ steps.tap.outputs.vcluster_experimental }}
+        run: |
```
sydorovdmytro: If I understand it correctly, the logic here assumes that the tap version should always contain `v` as a prefix. Is it always true?
If the tag name is v0.33.0 but the tap version is 0.33.0, it will alert the drift.

.github/tests/test-brew-tap-drift.sh
Piotr1215: Good idea, done.

PR #3719, opened 2026-03-19
fix(linear-sync): skip duplicate stable release comments

Piotr1215:
## Summary

- Before this change, every stable release triggered a "Now available in stable release" comment on already-released Linear issues, flooding inboxes with duplicates
- Adds `ListIssueComments` to query existing comments via GraphQL, and skips posting when a stable release comment already exists
- Includes `hasStableReleaseComment` helper with unit tests

Closes DEVOPS-669

### DIFF
--- a/hack/linear-sync/linear.go
+++ b/hack/linear-sync/linear.go
@@ -341,29 +341,46 @@ func (l *LinearClient) updateIssueState(ctx context.Context, issueID, releasedSt
 const stableReleaseCommentPrefix = "Now available in stable release"
 
 // ListIssueComments returns the body text of all comments on the given issue.
+// It paginates through all comments to avoid missing any beyond the default page size.
 func (l *LinearClient) ListIssueComments(ctx context.Context, issueID string) ([]string, error) {
-	var query struct {
-		Issue struct {
-			Comments struct {
-				Nodes []struct {
-					Body string
-				}
-			}
-		} `graphql:"issue(id: $id)"`
+	var bodies []string
+	var cursor *string
+
+	for {
+		var query struct {
+			Issue struct {
+				Comments struct {
+					Nodes []struct {
+						Body string
+					}
+					PageInfo struct {
+						HasNextPage bool
+						EndCursor   string
+					}
+				} `graphql:"comments(first: 100, after: $cursor)"`
+			} `graphql:"issue(id: $id)"`
+		}
+
+		variables := map[string]any{
+			"id":     graphql.String(issueID),
+			"cursor": (*graphql.String)(cursor),
+		}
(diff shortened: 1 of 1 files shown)


hack/linear-sync/linear.go
```diff
@@ -326,6 +337,64 @@ func (l *LinearClient) updateIssueState(ctx context.Context, issueID, releasedSt
 	return nil
 }
 
+// stableReleaseCommentPrefix is the prefix used to identify stable release comments.
+const stableReleaseCommentPrefix = "Now available in stable release"
```
sydorovdmytro, line 341: I think we could make sure this constant and `fmt.Sprintf("Now available in stable release %v (released %v)", releaseTagName, releaseDate)` in Line 304 are consistent. Maybe some global constant or at least a code comment to note this dependency.

hack/linear-sync/linear.go
```diff
@@ -326,6 +337,64 @@ func (l *LinearClient) updateIssueState(ctx context.Context, issueID, releasedSt
...
+			Issue struct {
+				Comments struct {
+					Nodes []struct {
+						Body string
+					}
+					PageInfo struct {
+						HasNextPage bool
+						EndCursor   string
+					}
+				} `graphql:"comments(first: 100, after: $cursor)"`
```
sydorovdmytro, line 360: This query retrieves all comments in the issue. For longer discussions it can be multiple API calls to find a specific comment. We could try to improve it by getting latest comments first as the needed comment is more likely among latest ones:
``graphql:"comments(first: 100, after: $cursor, orderBy: createdAt, sortDirection:
  DESC)"``

and then exit faster from the pagination loop if the comment has been found.

hack/linear-sync/linear.go
```diff
@@ -326,6 +337,64 @@ func (l *LinearClient) updateIssueState(ctx context.Context, issueID, releasedSt
 	return nil
 }
 
+// stableReleaseCommentPrefix is the prefix used to identify stable release comments.
+const stableReleaseCommentPrefix = "Now available in stable release"
+
+// ListIssueComments returns the body text of all comments on the given issue.
+// It paginates through all comments to avoid missing any beyond the default page size.
+func (l *LinearClient) ListIssueComments(ctx context.Context, issueID string) ([]string, error) {
```
sydorovdmytro, line 345: it's exported but used internally only, so it can be changed to `listIssueComments`

sydorovdmytro (changes requested):

Piotr1215:
Superseded by #3758 which includes this change plus a tag-scoped improvement from code review.

PR #3743, opened 2026-03-28
Fix Azure Blob support for auto-snapshots

nprokopic:
> [!WARNING]
> Do NOT merge before releasing related platform changes and releasing github.com/loft-sh/api/v4 v4.8.2.

/kind bugfix

fixes ENGOX-322

Fix Azure Blob support for snapshots so it works for auto-snapshots in vCluster platform.

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,9 +29,9 @@ require (
 	github.com/invopop/jsonschema v0.12.0
 	github.com/kubernetes-csi/external-snapshotter/client/v8 v8.2.0
 	github.com/loft-sh/admin-apis v0.0.0-20260311181619-506013b79ba0
-	github.com/loft-sh/agentapi/v4 v4.8.2-next.0
+	github.com/loft-sh/agentapi/v4 v4.8.2-rc.1
 	github.com/loft-sh/analytics-client v0.0.0-20240219162240-2f4c64b2494e
-	github.com/loft-sh/api/v4 v4.8.2-next.0
+	github.com/loft-sh/api/v4 v4.8.2-rc.1
 	github.com/loft-sh/e2e-framework v0.0.0-20260226211029-dc642849b244
 	github.com/loft-sh/image v0.0.0-20250625154753-87447a6ad364
 	github.com/loft-sh/utils v0.0.29
--- a/go.sum
+++ b/go.sum
@@ -331,12 +331,12 @@ github.com/lithammer/dedent v1.1.0 h1:VNzHMVCBNG1j0fh3OrsFRkVUwStdDArbgBWoPAffkt
 github.com/lithammer/dedent v1.1.0/go.mod h1:jrXYCQtgg0nJiN+StA2KgR7w6CiQNv9Fd/Z9BP0jIOc=
 github.com/loft-sh/admin-apis v0.0.0-20260311181619-506013b79ba0 h1:12fy1LhhyuQ46ls272yWn6HpAMLNUClsBYqRvu+qTC0=
 github.com/loft-sh/admin-apis v0.0.0-20260311181619-506013b79ba0/go.mod h1:pp+2PgOsRCTb1DXnEev/HCc0gvj5t0nHuTIMU8C8b90=
-github.com/loft-sh/agentapi/v4 v4.8.2-next.0 h1:4nV2OydxQ5lrZHoGAeQ7wmr+JLDKUeZAkf7z6B1bXEc=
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -28,10 +28,10 @@ require (
 	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/invopop/jsonschema v0.12.0
 	github.com/kubernetes-csi/external-snapshotter/client/v8 v8.2.0
-	github.com/loft-sh/admin-apis v0.0.0-20260219192040-a66d50310311
-	github.com/loft-sh/agentapi/v4 v4.8.0-alpha.1
+	github.com/loft-sh/admin-apis v0.0.0-20260311181619-506013b79ba0
+	github.com/loft-sh/agentapi/v4 v4.8.1-next.0
```
nprokopic: Update after releasing v4.8.1 (or v4.8.2) github.com/loft-sh/agentapi/v4

go.mod
```diff
@@ -28,10 +28,10 @@ require (
...
 	github.com/loft-sh/analytics-client v0.0.0-20240219162240-2f4c64b2494e
-	github.com/loft-sh/api/v4 v4.8.0-alpha.1
+	github.com/loft-sh/api/v4 v4.8.1-next.0
```
nprokopic: Update after releasing v4.8.1 (or v4.8.2) github.com/loft-sh/api/v4

pkg/snapshot/azure/client.go
```diff
@@ -74,66 +73,105 @@ func getBlobInfo(blobURL string) (BlobInfo, error) {
...
 
-	// Build container URL with SAS token (reconstruct account URL from account name)
-	accountURL := fmt.Sprintf("%s://%s.blob.core.windows.net", parsedURL.Scheme, info.AccountName)
-	containerURL := fmt.Sprintf("%s/%s?%s", accountURL, info.ContainerName, parsedURL.RawQuery)
+// newContainerClient creates an Azure container client for listing blobs using the following priority:
+// 1. Storage key (from options or env var) → shared key credential
+// 2. Service principal (from options) → client secret credential
+// 3. Fallback → DefaultAzureCredential (CLI path)
+func newContainerClient(options *Options, accountName, containerName string) (*container.Client, error) {
+	containerURL := fmt.Sprintf("https://%s.blob.core.windows.net/%s", accountName, containerName)
```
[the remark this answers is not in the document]
nprokopic: We currently do not have case where SAS will be used for listing.

pkg/snapshot/azure/options.go
```diff
@@ -22,13 +22,41 @@ const (
...
+
+	// ClientSecretEnvVar is the name of the environment variable that contains the Azure client secret
+	ClientSecretEnvVar = "AZURE_CLIENT_SECRET"
 )
 
 type Options struct {
-	BlobURL        string `json:"blob-url,omitempty"`
-	SAS            string `json:"sas,omitempty"`
+	// BlobURL is the full Azure Blob Storage URL
+	BlobURL string `json:"blob-url,omitempty"`
```
johannesfrey, line 38: nit, non-blocking: any reason why all json tags here are kebab-case? IIRC that's uncommon in Go.

nprokopic: Basically the reason here is for the sake of consistency with other Options structs that are used for snapshot-related CLI flags, e.g. see S3 flags here [here](https://github.com/loft-sh/vcluster/blob/8249443a8e8bf0c87882eae9ce4b7f4f1ffe7eb4/pkg/snapshot/s3/store.go#L42-L67) 

Didn't question this TBH, I thought it was related to CLI flag parsing, but didn't really look into it. Now when you commented, I don't see any reason why we use kebab-case in all these structs.

We can look into changing this, just we should be careful not to break some behavior, since I think that the platform has some equivalent structs which it uses when it's serializing Options (when the platform is passing these Options to vcluster binary).

In any case, I would keep the style here, as inconsistency would we even more awkward 🙈 And we can consider changing this in all Options structs, but that would be out of scope of this PR.

nprokopic: Ah, just remembered, IIRC these options can be also passed as a part of snapshot URL, so maybe we use kebab-case because of that.

nprokopic: It could be that Azure Blob case is missing the option to pass Azure options via snapshot URL, but again, that's not a change for this PR (here it's just about making changes so Azure Blob works for auto-snapshots) :)

PR #3759, opened 2026-04-01
chore(e2e-next): Add custom linter to E2E-next tests

adriankabala:
and extend logging for setup func
ENGQA-521

/kind bugfix
/kind enhancement
/kind feature
/kind documentation
/kind test

resolves #
Resolves ENGQA-521

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,15 +4,15 @@
 
 /.github/workflows/ @loft-sh/devops-team
 /.github/workflows/backport.yaml @loft-sh/devops-team
-/.github/workflows/e2e*.yaml @loft-sh/eng-qa @loft-sh/cto-office
-/.github/workflows/release.yaml @loft-sh/devops-team @loft-sh/cto-office
-/.github/CODEOWNERS @loft-sh/cto-office
+/.github/workflows/e2e*.yaml @loft-sh/eng-qa @loft-sh/eng-tech-leads
+/.github/workflows/release.yaml @loft-sh/devops-team @loft-sh/eng-tech-leads
+/.github/CODEOWNERS @loft-sh/eng-tech-leads
 
-/chart/templates/ @loft-sh/cto-office
+/chart/templates/ @loft-sh/eng-tech-leads
 
-/cmd/vcluster/ @loft-sh/cto-office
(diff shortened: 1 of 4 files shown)


.github/workflows/lint.yaml
```diff
@@ -81,7 +83,15 @@ jobs:
         env:
           GOFLAGS: ""
 
+      - name: Install golangci-lint
+        run: go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.7.0
+        env:
+          GOFLAGS: ""
+
+      - name: Build custom golangci-lint with plugins
```
sowmyav27, line 91: **Blocking:** This step downloads a private module (`github.com/loft-sh/e2e-framework`) but has no git auth configured. Fork PRs will fail here since they don't have access to `GH_ACCESS_TOKEN`.

The loft-enterprise lint workflow ([golangci-lint.yaml](https://github.com/loft-sh/loft-enterprise/blob/main/.github/workflows/golangci-lint.yaml)) solves this with two things:

1. Skip the entire workflow for forks:
```yaml
changes:
  if: github.repository_owner == 'loft-sh'  # do not run on forks
```

2. Configure git auth before this step:
```yaml
- name: Configure git
  run: git config --global url.https://"$GH_ACCESS_TOKEN"@github.com/.insteadOf https://github.com/
  env:
    GH_ACCESS_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
```

Please apply the same pattern here.

adriankabala: Fied!

.github/workflows/lint.yaml
```diff
@@ -81,7 +83,21 @@ jobs:
         env:
           GOFLAGS: ""
 
-      - name: Run golangci-lint
-        uses: golangci/golangci-lint-action@v9
-        with:
-          version: v2.4
+      - name: Install golangci-lint
```
sydorovdmytro, line 86: There is a dedicated action for you https://github.com/golangci/golangci-lint-action

adriankabala: We use `go install` + `golangci-lint custom` instead of golangci/golangci-lint-action because the action does not support building a custom binary with plugins.

PR #3768, opened 2026-04-01
Add osimages and sshkeys to platform destroy resources

zerbitx:
/kind bugfix

resolves #ENGPLAT-401

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```


zerbitx:
> Is this something we could e2e test?

I believe so, [here](https://linear.app/loft/issue/ENGQA-565/add-e2e-tests-for-destroy)'s a ticket to investigate/implement that.

PR #3781, opened 2026-04-02
ENGPLAT-399 Add --secure flag for TLS verification

rlmcpherson:
Primary Change: 
Add a --secure flag and thread config.Platform.Insecure through all call sites so TLS verification is on if configured or set explicitly via the CLI. To preserve current behavior the default is to not verify to enable bootstrapping, when certificates are not available. 

This also fixes a linting issue that occurs in PRs from forks:
CI runs the "fork PRs, without custom linters" lint step. But .golangci.yml defines 4 custom plugin linters (defercleanupcluster, defercleanupctx, ginkgoreturnctx,
  describefunc) in linters.settings.custom. Stock golangci-lint can't load these — they require the custom binary built by golangci-lint custom. This was broken when #3759 added these plugins.
  

/kind bugfix

resolves ENGPLAT-399

Fixed an issue where vcluster defaults to insecure TLS connections in some cases.

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

### DIFF
--- a/cmd/vclusterctl/cmd/platform/start.go
+++ b/cmd/vclusterctl/cmd/platform/start.go
@@ -82,7 +82,7 @@ before running this command:
 	startCmd.Flags().StringVar(&cmd.ChartRepo, "chart-repo", "https://charts.loft.sh/", "The chart repo to deploy vCluster platform")
 	startCmd.Flags().StringVar(&cmd.ChartName, "chart-name", "vcluster-platform", "The chart name to deploy vCluster platform")
 	startCmd.Flags().BoolVar(&cmd.Docker, "docker", false, "If true, vCluster platform will be installed in Docker")
-	startCmd.Flags().BoolVar(&cmd.Insecure, "insecure", false, "If true, skip TLS certificate verification when connecting to the platform (use for self-signed certificates)")
+	startCmd.Flags().BoolVar(&cmd.Secure, "secure", false, "If true, verify TLS certificates when connecting to the platform (by default, TLS verification is skipped during bootstrap because the platform starts with a self-signed certificate)")
 
 	return startCmd
 }
@@ -90,8 +90,9 @@ before running this command:
 func (cmd *StartCmd) Run(ctx context.Context) error {
 	cfg := cmd.LoadedConfig(cmd.Log)
 
-	// Apply --insecure flag to config so all downstream TLS checks respect it.
-	if cmd.Insecure {
+	// Bootstrap defaults to insecure because the platform starts with a
+	// self-signed certificate. Pass --secure to enforce TLS verification.
+	if !cmd.Secure {
 		cfg.Platform.Insecure = true
 	}
 
--- a/cmd/vclusterctl/cmd/platform/start_test.go
+++ b/cmd/vclusterctl/cmd/platform/start_test.go
@@ -8,25 +8,25 @@ import (
 	"github.com/loft-sh/vcluster/pkg/cli/start"
 )
 
-func TestNewStartCmd_InsecureFlag(t *testing.T) {
+func TestNewStartCmd_SecureFlag(t *testing.T) {
 	globalFlags := &flags.GlobalFlags{}
 	cmd := NewStartCmd(globalFlags)
 
-	// Verify --insecure flag exists and defaults to false.
-	f := cmd.Flags().Lookup("insecure")
+	// Verify --secure flag exists and defaults to false (insecure by default).
+	f := cmd.Flags().Lookup("secure")
 	if f == nil {
-		t.Fatal("--insecure flag not registered on start command")
+		t.Fatal("--secure flag not registered on start command")
 	}
 	if f.DefValue != "false" {
-		t.Errorf("expected --insecure default to be 'false', got %q", f.DefValue)
+		t.Errorf("expected --secure default to be 'false', got %q", f.DefValue)
 	}
 
-	// Simulate passing --insecure on the command line.
-	if err := cmd.Flags().Set("insecure", "true"); err != nil {
-		t.Fatalf("failed to set --insecure flag: %v", err)
+	// Simulate passing --secure on the command line.
+	if err := cmd.Flags().Set("secure", "true"); err != nil {
+		t.Fatalf("failed to set --secure flag: %v", err)
 	}
 	if f.Value.String() != "true" {
-		t.Errorf("expected --insecure value to be 'true' after set, got %q", f.Value.String())
+		t.Errorf("expected --secure value to be 'true' after set, got %q", f.Value.String())
 	}
 }
 
--- a/pkg/cli/start/start.go
+++ b/pkg/cli/start/start.go
@@ -70,7 +70,7 @@ type StartOptions struct { //nolint:revive // linter suggests renaming to option
 	Upgrade          bool
 	ReuseValues      bool
 	Docker           bool
-	Insecure         bool
+	Secure           bool
 }
 
 func NewLoftStarter(options StartOptions) *LoftStarter {


pkg/cli/start/port_forwarding.go
```diff
@@ -21,7 +21,7 @@ func (l *LoftStarter) startPortForwarding(ctx context.Context, loftPod *corev1.P
 
 	// wait until loft is reachable at the given url
 	httpClient := &http.Client{
-		Transport: utilhttp.InsecureTransport(),
+		Transport: utilhttp.Transport(l.LoadedConfig(l.Log).Platform.Insecure),
```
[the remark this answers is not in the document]
rlmcpherson: Updated to use fallback mechanism in case of bootstrap flow, when the platform uses self-signed certs.

pkg/cli/start/login.go
```diff
@@ -66,51 +67,87 @@ func (l *LoftStarter) loginViaCLI(url string) error {
...
+	l.Log.WriteString(logrus.InfoLevel, "\n")
+	l.Log.Donef(product.Replace("Successfully logged in via CLI into Loft instance %s"), ansi.Color(url, "white+b"))
+
+	return nil
+}
+
+// passwordLogin posts the admin credentials to the platform's password login
+// endpoint and returns the access key. The insecure parameter controls whether
+// TLS certificate verification is skipped.
+func (l *LoftStarter) passwordLogin(url string, loginRequestBytes []byte, insecure bool) (string, bool, error) {
```
rmweir: Why are we returning insecure? It doesn't look like this function mutates it at all so we can set the variable outside of it and then pass it in to this function and `LoginWithAccessKey()`.

pkg/cli/start/login.go
```diff
@@ -66,51 +67,87 @@ func (l *LoftStarter) loginViaCLI(url string) error {
...
+	accessKey, insecure, err := l.passwordLogin(url, loginRequestBytes, false)
+	if err != nil {
+		if !isTLSError(err) {
+			return err
+		}
+		l.Log.Infof("TLS verification failed, retrying without verification (self-signed certificate expected during bootstrap)")
+		accessKey, insecure, err = l.passwordLogin(url, loginRequestBytes, true)
+		if err != nil {
+			return err
+		}
```
rmweir: I think falling back on insecure if TLS verification fails, defeats the purpose of TLS verification. If TLS verification is desired, as a secure default, then failing to verify the cert should return an error and cancel the process.

rlmcpherson: If we don't it will cause bootstrapping to fail. So the choice is to use the current behavior and always use it or break bootstrapping.

One approach is to have users explicitly disable it during bootstrapping although that does make setup more of a PITA process.

rlmcpherson: Reverted the fallback change. It now requires using `--inescure` on bootstrap.

pkg/cli/start/login.go
rlmcpherson: Good call. This is now removed.

.github/workflows/lint.yaml
```diff
@@ -98,6 +98,19 @@ jobs:
...
+        run: |
+          # Remove custom plugin definitions and their enable/exclusion entries
+          # so stock golangci-lint can run without the compiled plugin binary.
+          CUSTOM_LINTERS=$(yq '.linters.settings.custom | keys | .[]' .golangci.yml)
+          cp .golangci.yml .golangci-fork.yml
+          for linter in $CUSTOM_LINTERS; do
+            yq -i "del(.linters.settings.custom.\"$linter\")" .golangci-fork.yml
+            yq -i ".linters.enable -= [\"$linter\"]" .golangci-fork.yml
+            yq -i "del(.linters.exclusions.rules[] | select(.linters[] == \"$linter\"))" .golangci-fork.yml
+          done
```
FabianKramm, line 112: Why is this needed? Why do we need this part of the github action?

pkg/cli/start/login.go
```diff
@@ -66,8 +66,9 @@ func (l *LoftStarter) loginViaCLI(url string) error {
 		return err
 	}
 
+	config := l.LoadedConfig(l.Log)
 	httpClient := &http.Client{Transport: &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: config.Platform.Insecure},
```
FabianKramm, line 71: Does this actually take the current one or is this using the old config.Platform.Insecure from a previous logged in session?

pkg/cli/start/port_forwarding.go
FabianKramm, line 24: When starting a new loft isntance, how would I be able to even specify insecure?

pkg/cli/start/success.go
```diff
@@ -123,10 +124,11 @@ func (l *LoftStarter) pingLoftRouter(ctx context.Context, loftPod *corev1.Pod) (
 	loftRouterDomain := string(loftRouterSecret.Data["domain"])
 
 	// wait until loft is reachable at the given url
+	insecure := l.LoadedConfig(l.Log).Platform.Insecure
 	httpClient := &http.Client{
 		Transport: &http.Transport{
 			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: true,
+				InsecureSkipVerify: insecure,
```
FabianKramm, line 131: When starting a new loft isntance, how would I be able to even specify insecure?

pkg/cli/start/success.go
```diff
@@ -203,7 +206,7 @@ func (l *LoftStarter) successRemote(ctx context.Context, host string) error {
 
 	l.Log.Info("Waiting for you to configure DNS, so loft can be reached on https://" + host)
 	err = wait.PollUntilContextTimeout(ctx, 5*time.Second, clihelper.Timeout(), true, func(ctx context.Context) (done bool, err error) {
-		return clihelper.IsLoftReachable(ctx, host)
+		return clihelper.IsLoftReachable(ctx, host, insecure)
```
FabianKramm, line 209: When starting a new loft isntance, how would I be able to even specify insecure?

pkg/platform/clihelper/clihelper.go
```diff
@@ -344,10 +344,10 @@ func GetLoftDefaultPassword(ctx context.Context, kubeClient kubernetes.Interface
 	return string(loftNamespace.UID), nil
 }
 
-func IsLoftReachable(ctx context.Context, host string) (bool, error) {
+func IsLoftReachable(ctx context.Context, host string, insecure bool) (bool, error) {
 	// wait until loft is reachable at the given url
 	client := &http.Client{
-		Transport: utilhttp.InsecureTransport(),
+		Transport: utilhttp.Transport(insecure),
```
FabianKramm, line 350: When starting a new loft isntance, how would I be able to even specify insecure?

.github/workflows/lint.yaml
rlmcpherson: This was to fix an issue introduced in a a previous PR that broke linting in fork PRs. See this slack thread for more information. https://loft-labs-inc.slack.com/archives/C07G0T9HS4U/p1775139516620279

pkg/cli/start/login.go
rlmcpherson: It was reading the stale value from ~/.vcluster/config.json from a previous login session, not the "current" one. With the change to require an `--insecure` flag, this is no longer the case as it's set via the CLI.

pkg/cli/start/port_forwarding.go
rlmcpherson: Pushed a change to add a `--insecure` flag to the cli and ensure it is loaded into config at runtime.

pkg/cli/start/success.go
rlmcpherson: addressed above.

pkg/cli/start/success.go
rlmcpherson: addressed above.

pkg/platform/clihelper/clihelper.go
rlmcpherson: addressed above.

PR #3793, opened 2026-04-07
ENGCP-593: auto-rotate certificates at runtime before expiry

rlmcpherson:
vCluster leaf certificates (1-year lifetime) were only checked for renewal at pod startup. If a pod ran continuously for more than 275 days without restarting, certificates would expire because the 90-day renewal window was never
  re-evaluated at runtime.

  Additionally, the API server serving certificate syncer had a bug where expiry-triggered cert regeneration was silently discarded because the syncer compared SAN lists instead of cert bytes.

  ### Changes

  **Syncer SANs bug fix** (`pkg/server/cert/syncer.go`)
  - The syncer's `regen()` method always updates the cert/key fields and notifies listeners unconditionally. `GenAPIServerServingCerts` already returns unchanged bytes when no regeneration is needed, so no comparison is necessary.

  **Periodic cert watcher** (`pkg/certs/watcher.go`)
  - Background goroutine checks all kubeadm leaf certificates every 24 hours (configurable via `VCLUSTER_CERT_CHECK_INTERVAL` when `DEVELOPMENT=true`).
  - When any cert is within the 90-day renewal threshold, rotates all leaf certs preserving CA and SA keys.
  - **Clustered mode**: generates new certs in a temp directory, syncs to the cert secret, then triggers rolling restarts of the control-plane workload and deployed etcd StatefulSet (if enabled) by patching a
  `vcluster.loft.sh/cert-rotation-at` template annotation.
  - **Standalone mode**: rotates certs on disk, removes standalone etcd server/peer certs for dynamic regeneration, then exits after a random delay (5s–5m) to avoid synchronized restarts.
  - **HA coordination**: uses a Kubernetes Lease (`<name>-cert-rotation`) so only one replica performs the rotation. Single-replica deployments skip the lease (may not have `coordination.k8s.io/leases` RBAC).
  - **Retry logic**: if the rollout patch fails, retries with exponential backoff without regenerating certs again.
  - Backs up the PKI directory before rotation (consistent with the manual `vcluster certs rotate` command).
  - Deployed etcd rollout is guarded by `controlPlane.backingStore.etcd.deploy.enabled` config.

  **Serving cert duration override** (`pkg/util/certhelper/cert.go`, `pkg/server/cert/cert.go`)
  - `Config.Duration` field allows overriding the 365-day default cert validity.
  - `servingCertDuration()` respects `DEVELOPMENT=true` + `VCLUSTER_CERTS_VALIDITYPERIOD` for the syncer's serving cert, enabling e2e testing with short-lived certs.

  **Chart RBAC** (`chart/templates/role.yaml`)
  - Added `patch` and `update` verbs for `apps` group resources (`statefulsets`, `deployments`) to support the rollout annotation patching.

  ### Certs covered

  | Cert | Lifetime | How rotated |
  |------|----------|-------------|
  | API server serving cert (proxy) | 1 year | Syncer hot-reload every 2s (bug fix) |
  | apiserver.crt, apiserver-kubelet-client.crt | 1 year | Watcher + workload rollout |
  | front-proxy-client.crt | 1 year | Watcher + workload rollout |
  | etcd/server.crt, etcd/peer.crt, etcd/healthcheck-client.crt | 1 year | Watcher + workload rollout (+ etcd StatefulSet rollout if deployed) |
  | scheduler.conf, controller-manager.conf, admin.conf | 1 year | Watcher + workload rollout |
  | CA certs | 10 years | Manual only (`vcluster certs rotate-ca`) |

  ### E2e test coverage

  | Test | Cluster | What it verifies |
  |------|---------|-----------------|
  | `ServingCertRotationSpec` | 1 replica, embedded etcd, 3m certs | Serving cert serial changes via TLS handshake, no pod restart |
  | `SingleReplicaWatcherSpec` | 1 replica, embedded etcd, 15s interval | Watcher rotates without lease, workload rolled, CA preserved, all leaf certs renewed, no etcd StatefulSet rolled |
  | `HACertRotationSpec` | 2 replicas, deployed etcd, 15s interval | Lease created, both control-plane + etcd rolled, all pods replaced, CA preserved, all leaf certs renewed |
  | `CertAutoRotationSpec` | 1 replica, deployed etcd | Startup EnsureCerts detects expiring cert in secret, rotates all leaf certs, CA preserved |
  | `CertTestsSpec` | 1 replica, deployed etcd | Existing manual rotation tests (leaf + CA fingerprint verification) |

/kind bugfix
/kind enhancement

resolves ENGCP-593

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
/certs-vcluster/ && !non-default
```


pkg/server/cert/syncer.go
```diff
@@ -86,7 +86,12 @@ func (s *syncer) regen(ctx context.Context) (bool, error) {
 		return false, err
 	}
 
-	if reflect.DeepEqual(s.currentSANs, extraSANs) {
+	// Compare cert bytes rather than SANs. GenAPIServerServingCerts returns
+	// the same cert/key unchanged when no regeneration is needed, so a byte
+	// comparison correctly detects both SAN changes and expiry-triggered
+	// regeneration. The previous SAN-only comparison silently discarded
+	// certs that were regenerated due to expiry with unchanged SANs.
+	if bytes.Equal(s.currentCert, cert) {
```
FabianKramm: This makes not much sense, we should just replace it after a while because this is basically always replacing it

pkg/setup/initialize.go
```diff
@@ -61,6 +61,15 @@ func initialize(ctx context.Context, options *config.VirtualClusterConfig) error
 		if err != nil {
 			return err
 		}
+
+		// Start background watcher that periodically checks for expiring leaf
+		// certificates and rotates them. This covers long-running pods that
+		// might not restart before certs expire (1-year lifetime, 90-day window).
+		kubeadmConfig, err := certs.GenerateInitKubeadmConfig(serviceCIDR, certificatesDir, options)
```
FabianKramm: Can we move this into the certs.StartCertWatcher function?

FabianKramm (changes requested):

pkg/setup/initialize.go
rlmcpherson: Moved into function.

pkg/server/cert/syncer.go
rlmcpherson: Agreed. It always returns the cert now from this function without doing the byte comparison.

pkg/certs/watcher.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+
+			klog.Infof("Leaf certificates are expiring soon, rotating and restarting")
+			if err := rotateLeafCerts(ctx, currentNamespace, currentNamespaceClient, certificateDir, kubeadmConfig, options); err != nil {
+				klog.Errorf("Error rotating certificates: %v", err)
+				continue
+			}
+
+			klog.Infof("Leaf certificates rotated successfully, exiting for pod restart")
+			osutil.Exit(0)
```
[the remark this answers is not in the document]
rlmcpherson: Addressing this with jitter and a lease for HA deployments for coordinating rotation.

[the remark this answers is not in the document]
FabianKramm: Do we really have to restart as well?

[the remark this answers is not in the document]
rlmcpherson: Looking into whether that's possible.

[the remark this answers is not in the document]
rlmcpherson: The restart approach is at least consistent with the current rotation logic. Some of the certs are only read from disk once at pod start up too: 

### Certs that cannot hot-reload (require pod restart)

  | Cert | CLI Flag | Reason |
  |------|----------|--------|
  | `apiserver-kubelet-client.crt` | `--kubelet-client-certificate` | kube-apiserver loads once at startup |
  | `apiserver-etcd-client.crt` | `--etcd-certfile` | kube-apiserver loads once at startup |
  | `front-proxy-client.crt` | `--proxy-client-cert-file` | kube-apiserver loads once at startup |
  | `controller-manager.conf` | `--kubeconfig` | client-go loads embedded cert once |
  | `scheduler.conf` | `--kubeconfig` | client-go loads embedded cert once |
  | `admin.conf` | N/A | client-go loads embedded cert once |

  All leaf certs share the same 1-year validity (generated together by kubeadm), so they expire simultaneously. Hot-reloading only the serving/etcd certs doesn't help if the client certs above expire at the same time.

pkg/certs/watcher.go
```diff
@@ -0,0 +1,126 @@
...
+
+// checkCertsExpiring checks whether any leaf cert is within the renewal window.
+// Both standalone and platform-managed modes check certs on disk — in
+// platform-managed mode, certs were downloaded from the secret at startup.
+func checkCertsExpiring(certificateDir string) (bool, error) {
+	return diskCertsExpiringSoon(certificateDir), nil
+}
+
+// rotateLeafCerts rotates all leaf certificates, preserving CA and SA keys.
+func rotateLeafCerts(
```
mfranczy, line 407: As discussed, this should align with the [Rotate func](https://github.com/loft-sh/vcluster/blob/main/pkg/certs/rotate.go#L35) in terms of:
* backing up PKI before deletion
* handling standalone ETCD

If needed, we should extract some helpers from the Rotate func.

pkg/certs/watcher.go
```diff
@@ -0,0 +1,126 @@
...
+		return fmt.Errorf("remove expiring leaf certs: %w", err)
+	}
+
+	// Regenerate missing certs
+	if err := generateCertificates(certificateDir, kubeadmConfig); err != nil {
+		return fmt.Errorf("regenerate certs: %w", err)
+	}
+
+	// For non-standalone mode, sync renewed certs back to the host secret
+	if currentNamespaceClient != nil {
```
mfranczy: Let's be explicit about standalone mode. It's safer than just relying on the presence of the host client.
```
if options.ControlPlane.Standalone.Enabled {
        return nil
}
if currentNamespaceClient == nil {
        return errors.New("non-standalone cert rotation requires a host cluster client")
}
````
Similar to https://github.com/loft-sh/vcluster/blob/main/pkg/certs/ensure.go#L83-L86

**Review by mfranczy — CHANGES_REQUESTED** [review]

File: pkg/certs/watcher.go [hunk 6]
```diff
@@ -0,0 +1,126 @@
...
+			if err != nil {
+				klog.Errorf("Error checking certificate expiry: %v", err)
+				continue
+			}
+			if !expiring {
+				continue
+			}
+
+			klog.Infof("Leaf certificates are expiring soon, rotating and restarting")
+			if err := rotateLeafCerts(ctx, currentNamespace, currentNamespaceClient, certificateDir, kubeadmConfig, options); err != nil {
```
johannesfrey: As this is now happening during runtime, should the extra expiry checks [here](https://github.com/loft-sh/vcluster/blob/main/pkg/certs/ensure.go#L100-L112) and [here](https://github.com/loft-sh/vcluster/blob/main/pkg/certs/ensure.go#L129-L149) that happen during pod startup be removed?

pkg/certs/watcher.go
```diff
@@ -0,0 +1,126 @@
...
+// rotateLeafCerts rotates all leaf certificates, preserving CA and SA keys.
+func rotateLeafCerts(
+	ctx context.Context,
+	currentNamespace string,
+	currentNamespaceClient kubernetes.Interface,
+	certificateDir string,
+	kubeadmConfig *kubeadmapi.InitConfiguration,
+	options *config.VirtualClusterConfig,
+) error {
+	// Remove only leaf certs; preserve SA keys and CA keys
```
johannesfrey: If we decide to remove the extra expiry checks during pod start up (like mentioned in my [comment above](https://github.com/loft-sh/vcluster/pull/3793/changes#r3056696753)), we should download the certs from the secret first in case of non-standalone before we remove them, as it is currently done [here](https://github.com/loft-sh/vcluster/blob/main/pkg/certs/ensure.go#L129-L149).

pkg/certs/watcher.go
mfranczy: I would not remove it. The watcher is only a runtime safeguard, it does not check immediately (the first check is done after 24h by default), and even with an immediate first run it would still execute asynchronously. I think it's good to have the synchronous startup guard that makes sure PKI is valid before control plane components come up.

pkg/certs/watcher.go
mfranczy: Answered here: https://github.com/loft-sh/vcluster/pull/3793#discussion_r3057025319

johannesfrey: Thinking about it again, even when keeping the start up expiry check around, in case of non-standalone we should download the certs from the secret first (as it's done [here](https://github.com/loft-sh/vcluster/blob/main/pkg/certs/ensure.go#L131-L134)) before removing the files.

mfranczy: So what's the reason behind this? In the code you shared, you download the secrets after the expiry check, meaning you download them just to remove it. Is it to make sure that CA and SA is locally available?

johannesfrey: Yeah exactly. I know the existing func also unnecessarily downloads the leaf certs that are about to be deleted, but just wanted to reuse it initially.

mfranczy: I would say we don't necessarily need this in the watcher, since during startup we already ensure the certificates are available, even if they are not expired: https://github.com/loft-sh/vcluster/blob/main/pkg/certs/ensure.go#L125-L127.

pkg/certs/watcher.go
rlmcpherson: Per our discussion, extracted backupPKI() and removeStandaloneEtcdCerts() helpers from Rotate so both the manual rotation command and the runtime watcher share the same logic. rotateLeafCerts now backs up PKI before deletion and handles standalone etcd cert cleanup, matching Rotate's behavior.

pkg/certs/watcher.go
rlmcpherson: Done. Changed to check options.ControlPlane.Standalone.Enabled explicitly rather than inferring from a nil client. A nil client in non-standalone mode now returns errors.New("non-standalone cert rotation requires a host cluster client"), matching the pattern in ensure.go:83-86.

pkg/certs/watcher.go
rlmcpherson: Agreed with @mfranczy. The startup check in EnsureCerts is a synchronous guard that ensures PKI is valid before control plane components start. The watcher is an async safety net for long-running pods (first check is after 24h by default). They serve different purposes.

mfranczy:
@codex review

pkg/certs/watcher.go
rlmcpherson: `rotateLeafCerts` now downloads the secret to disk before removeFiles, ensuring CA and SA keys are present for generateCertificates. This matches the startup rotation path in ensure.go:131-134.

pkg/certs/watcher.go
```diff
@@ -0,0 +1,244 @@
...
+				klog.Errorf("Error checking certificate expiry: %v", err)
+				continue
+			}
+			if !expiring {
+				continue
+			}
+
+			// In HA deployments, acquire a lease so only one replica rotates
+			// and restarts at a time.
+			if currentNamespaceClient != nil && !tryAcquireRotationLease(ctx, currentNamespaceClient, currentNamespace, options.Name) {
```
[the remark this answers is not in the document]
rlmcpherson: Addressed by [acquiring lease only for multiple-replicas.](https://github.com/loft-sh/vcluster/pull/3793/changes/2691539ae8b9b8c78879f5030dfa5c9a13494789)

rlmcpherson:
@codex re-review, taking PR comments into account

rlmcpherson:
Nightly test run with passing e2e tests on this branch: https://github.com/loft-sh/vcluster/actions/runs/24200930480

pkg/server/cert/syncer.go
```diff
@@ -76,43 +75,33 @@ func (s *syncer) RunOnce(ctx context.Context) error {
...
 			return
 		}
 
-		if changed {
-			for _, l := range s.listeners {
-				l.Enqueue()
-			}
+		for _, l := range s.listeners {
+			l.Enqueue()
 		}
```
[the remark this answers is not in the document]
rlmcpherson: This was intentional, per @FabianKramm feedback.

pkg/certs/watcher.go
```diff
@@ -0,0 +1,257 @@
...
+	currentNamespace string,
+	currentNamespaceClient kubernetes.Interface,
+	certificateDir string,
+	options *config.VirtualClusterConfig,
+	kubeadmConfig *kubeadmapi.InitConfiguration,
+) {
+	// Add random jitter (up to 10% of interval) so HA replicas don't all
+	// check at exactly the same time.
+	jitter := time.Duration(rand.Int64N(int64(interval / 10)))
+	ticker := time.NewTicker(interval + jitter)
```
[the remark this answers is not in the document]
rlmcpherson: This only applies to development mode, so it is ok.

mfranczy (approved):
These changes are acceptable to me as an interim solution.

That said, we should add a follow up PR in this release cycle to generalize the rolling upgrade mechanism and move it from pro into oss, with an interface that supports:
* k8s based deployments, coordinated via leases
* standalone, coordinated directly via etcd (as we do today in pro).

With the current default interval, jitter can spread restarts across a window of up to 2h24m. This reduces the probability of simultaneous restarts, but it does not guarantee that one control plane instance is fully back before the next one is restarted.

@FabianKramm, I think it's better to get this as an interim step to avoid mixing it with the work of moving the rolling upgrade mechanism from pro.

e2e-next/test_security/certs/test_ha_cert_rotation.go
```diff
@@ -0,0 +1,194 @@
...
+				})
+			})
+
+			// Spec 5: cleanup the lease
+			It("should clean up the cert rotation lease", func(ctx context.Context) {
+				leaseName := translate.SafeConcatName("vcluster", vClusterName, "cert-rotation")
+				err := hostClient.CoordinationV1().Leases(vClusterNamespace).Delete(ctx,
+					leaseName, metav1.DeleteOptions{})
+				if err != nil {
+					klog.Infof("Lease cleanup: %v", err)
```
adriankabala:
```suggestion
err := hostClient.CoordinationV1().Leases(vClusterNamespace).Delete(ctx,
          leaseName, metav1.DeleteOptions{})
      Expect(clientpkg.IgnoreNotFound(err)).To(Succeed())
```

e2e-next/test_security/certs/test_serving_cert_rotation.go
```diff
@@ -0,0 +1,108 @@
...
+
+						for _, pod := range pods.Items {
+							for _, container := range pod.Status.ContainerStatuses {
+								g.Expect(container.State.Running).NotTo(BeNil(),
+									"container %s in pod %s should still be running", container.Name, pod.Name)
+								g.Expect(container.Ready).To(BeTrue(),
+									"container %s in pod %s should still be ready", container.Name, pod.Name)
+							}
+						}
+					}).WithPolling(constants.PollingInterval).WithTimeout(4 * time.Minute).Should(Succeed())
```
adriankabala: Add a constant like constants.PollingTimeoutCertLifetime or use constants.PollingTimeoutVeryLong

e2e-next/test_security/certs/test_serving_cert_rotation.go
```diff
@@ -0,0 +1,108 @@
...
+			// server would start rejecting TLS connections and the pod would become
+			// unready. We wait past the cert lifetime and verify the pod is still ready.
+			It("should remain healthy after serving cert would have expired", func(ctx context.Context) {
+				By("Verifying the pod stays ready past the cert lifetime", func() {
+					// With VCLUSTER_CERTS_VALIDITYPERIOD=3m, the serving cert expires
+					// in 3 minutes. If the syncer weren't continuously regenerating it,
+					// the API server would reject TLS connections and the pod would fail
+					// health checks. We verify the pod stays ready for 4 minutes —
+					// past the cert lifetime.
+					Consistently(func(g Gomega) {
```
adriankabala: The test waits 4 minutes polling pod readiness every 2 seconds. It does NOT verify that the cert was actually regenerated - only that the pod stays healthy. A pod can stay healthy for reasons unrelated to cert rotation (TLS session cache, graceful degradation). A better approach: record the cert fingerprint at the start, wait 30-60 seconds, then verify the fingerprint changed. This would save ~3 minutes per run while providing stronger proof that rotation actually happened.

e2e-next/test_security/certs/test_serving_cert_rotation.go
```diff
@@ -0,0 +1,108 @@
...
+				By("Verifying all container restart counts are zero", func() {
+					pods, err := hostClient.CoreV1().Pods(vClusterNamespace).List(ctx, metav1.ListOptions{
+						LabelSelector: "app=vcluster,release=" + vClusterName,
+					})
+					Expect(err).To(Succeed())
+					Expect(pods.Items).NotTo(BeEmpty())
+
+					for _, pod := range pods.Items {
+						for _, container := range pod.Status.ContainerStatuses {
+							Expect(container.RestartCount).To(BeNumerically("==", 0),
```
adriankabala: If the pod restarted for an unrelated reason before this test ran (OOM kill, liveness probe timeout during setup), this assertion fails as a false negative. Fix: record the restart count in BeforeAll or spec 1, then assert in spec 3 that it did not increase.

e2e-next/test_security/certs/test_single_replica_watcher.go
```diff
@@ -0,0 +1,151 @@
...
+					leaseName, metav1.GetOptions{})
+				Expect(kerrors.IsNotFound(err) || kerrors.IsForbidden(err)).To(BeTrue(),
+					"rotation lease should not exist for single-replica deployment, got: %v", err)
+			})
+		},
+	)
+}
+
+// execWriteFileSingleReplica writes data to a file inside a container using kubectl exec.
+func execWriteFileSingleReplica(ctx context.Context, restConfig *rest.Config, client kubernetes.Interface, namespace, podName, container, filePath string, data []byte) error {
```
adriankabala: `execWriteFile` and `execWriteFileSingleReplica` are byte-for-byte identical functions in two different files. Extract to a single shared helper in the certs package (e.g. `helpers.go`).

e2e-next/test_security/certs/test_auto_rotation.go
```diff
@@ -0,0 +1,201 @@
...
+
+					g.Expect(cert.NotAfter.After(time.Now().Add(90*24*time.Hour))).To(BeTrue(),
+						"apiserver cert should have been renewed, NotAfter=%s", cert.NotAfter.Format(time.RFC3339))
+				}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeoutVeryLong).Should(Succeed())
+			})
+
+			// Spec 5 depends on 3: verify CA was NOT rotated
+			It("should preserve the CA cert during auto-rotation", func(ctx context.Context) {
+				secret, err := hostClient.CoreV1().Secrets(vClusterNamespace).Get(ctx,
+					certs.CertSecretName(vClusterName), metav1.GetOptions{})
```
adriankabala: `CertAutoRotationSpec` (test_auto_rotation.go:127-141), `HACertRotationSpec` (test_ha_cert_rotation.go:127-145), and `SingleReplicaWatcherSpec` (test_single_replica_watcher.go:97-113) all do the same Eventually: decode apiserver.crt PEM → parse → assert NotAfter > now + 90 days. Extract to a shared helper like  `expectCertRenewed(ctx, hostClient, namespace, vClusterName)`.

mfranczy (changes requested):
After all, I must request changes.

I've tested the solution. Certificate rotation does not restart external etcd.
I also did some investigation and the simplest way to coordinate rollout in non-standalone setups is to patch the StatefulSet/DaemonSet spec.template.metadata.annotations.

This is a much better approach than using osutil.Exit, which we can address separately for standalone later.

e2e-next/test_security/certs/test_serving_cert_rotation.go
rlmcpherson: Added getServingCertSerial helper that port-forwards to the vcluster pod and inspects the TLS peer certificate. The test records the initial serial, waits for the syncer to regenerate (Every poll checks via a fresh TLS handshake), and asserts the serial changed. This provides direct proof of cert regeneration rather than inferring it from pod health.

e2e-next/test_security/certs/test_ha_cert_rotation.go
rlmcpherson: Done. Changed to kerrors.IsNotFound(err) guard with Expect(err).To(Succeed()) for non-NotFound errors.

e2e-next/test_security/certs/test_serving_cert_rotation.go
rlmcpherson: Replaced the entire 4-minute Consistently approach. The test now records the serving cert's serial number via TLS handshake (using port-forward), waits briefly, and verifies the serial changed.

e2e-next/test_security/certs/test_serving_cert_rotation.go
rlmcpherson: Done. Added containerRestartCounts helper that returns a map[string]int32 of container restart counts. Baseline is recorded in spec 1 (pods ready), and spec 3 asserts no container's count increased.

e2e-next/test_security/certs/test_single_replica_watcher.go
rlmcpherson: Thanks. Moved to shared helper.

e2e-next/test_security/certs/test_auto_rotation.go
rlmcpherson: Thanks. Extracted expectCertRenewed to helpers.go. It encapsulates the Eventually + PEM decode + NotAfter > 90 days assertion pattern. Replaced the inline versions in test_auto_rotation.go, test_ha_cert_rotation.go, and test_single_replica_watcher.go.

mfranczy:
@codex review

pkg/server/cert/syncer.go
```diff
@@ -76,43 +75,33 @@ func (s *syncer) RunOnce(ctx context.Context) error {
...
 			klog.Infof("Error regenerating certificate: %v", err)
 			return
 		}
 
-		if changed {
-			for _, l := range s.listeners {
-				l.Enqueue()
-			}
+		for _, l := range s.listeners {
+			l.Enqueue()
```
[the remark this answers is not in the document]
rlmcpherson: Already addressed above.

pkg/certs/watcher.go
```diff
@@ -0,0 +1,494 @@
...
+	interval time.Duration,
+	currentNamespace string,
+	currentNamespaceClient kubernetes.Interface,
+	certificateDir string,
+	options *config.VirtualClusterConfig,
+	kubeadmConfig *kubeadmapi.InitConfiguration,
+) {
+	// Add random jitter (up to 10% of interval) so HA replicas don't all
+	// check at exactly the same time.
+	jitter := time.Duration(rand.Int64N(int64(interval / 10)))
```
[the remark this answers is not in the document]
rlmcpherson: Already addressed above.

rlmcpherson:
Nightly test run against this branch: https://github.com/loft-sh/vcluster/actions/runs/24254936842

mfranczy (dismissed):

pkg/certs/watcher.go
```diff
@@ -0,0 +1,497 @@
...
+
+	klog.Infof("Leaf certificates rotated successfully, triggered workload rollout at %s", rolloutAt)
+	return nil
+}
+
+func rolloutControlPlaneWithRetry(ctx context.Context, client kubernetes.Interface, namespace, name, rolloutAt string) error {
+	return retryWithBackoff(ctx, "trigger control-plane rollout", func(ctx context.Context) error {
+		// Patching spec.template annotations triggers a rollout for Deployments
+		// and StatefulSets using RollingUpdate. StatefulSets configured with
+		// OnDelete will not restart pods automatically.
```
mfranczy, line 307: @rlmcpherson remeber to highlight this in our doc.

e2e-next/test_security/certs/test_single_replica_watcher.go
```diff
@@ -0,0 +1,144 @@
...
+			It("should not have created a rotation lease", func(ctx context.Context) {
+				leaseName := translate.SafeConcatName("vcluster", vClusterName, "cert-rotation")
+				_, err := hostClient.CoordinationV1().Leases(vClusterNamespace).Get(ctx,
+					leaseName, metav1.GetOptions{})
+				Expect(kerrors.IsNotFound(err) || kerrors.IsForbidden(err)).To(BeTrue(),
+					"rotation lease should not exist for single-replica deployment, got: %v", err)
+			})
+
+			// Spec 6 depends on 4: verify no deployed etcd rollout was attempted.
+			// This cluster uses embedded etcd (no deploy.enabled), so the config
```
sowmyav27, line 133: **Non-blocking (comment correctness):** This comment says "Spec 6 depends on 4" but this is actually the 8th `It` block in the container. The preceding spec 7's comment correctly says "Spec 7 depends on 4".

Suggested fix:
```suggestion
			// Spec 8 depends on 4: verify no deployed etcd rollout was attempted.
```

mfranczy (dismissed):

e2e-next/test_security/certs/test_single_replica_watcher.go
```diff
@@ -0,0 +1,144 @@
...
+				_, err := hostClient.CoordinationV1().Leases(vClusterNamespace).Get(ctx,
+					leaseName, metav1.GetOptions{})
+				Expect(kerrors.IsNotFound(err) || kerrors.IsForbidden(err)).To(BeTrue(),
+					"rotation lease should not exist for single-replica deployment, got: %v", err)
+			})
+
+			// Spec 6 depends on 4: verify no deployed etcd rollout was attempted.
+			// Spec 8 depends on 4: verify no deployed etcd rollout was attempted.
+			// guard should have skipped the etcd StatefulSet rollout entirely.
+			It("should not have attempted a deployed etcd rollout", func(ctx context.Context) {
```
sowmyav27: The fix for the spec number left both lines and dropped the middle context line. Currently:
```
// Spec 6 depends on 4: verify no deployed etcd rollout was attempted.
// Spec 8 depends on 4: verify no deployed etcd rollout was attempted.
// guard should have skipped the etcd StatefulSet rollout entirely.
```

Should be:
```suggestion
			// Spec 8 depends on 4: verify no deployed etcd rollout was attempted.
			// This cluster uses embedded etcd (no deploy.enabled), so the config
			// guard should have skipped the etcd StatefulSet rollout entirely.
```

Remove the old `Spec 6` line and restore the missing context line explaining *why* the etcd rollout shouldn't happen.

mfranczy (approved):

adriankabala (approved):

FabianKramm (approved):

PR #3823, opened 2026-04-10
feat: add GET /vcluster/features endpoint for vNode feature-flag checks

flomedja:
/kind feature

resolves #ENGNODE-316

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -253,6 +253,12 @@ func (s *Server) ServeOnListenerTLS(ctx *synccontext.ControllerContext) error {
 			)
 		}
 	}
+	redirectAuthNonResources = append(redirectAuthNonResources,
+		delegatingauthorizer.PathVerb{
+			Path: "/vcluster/features",
+			Verb: "*",
+		},
+	)
 	serverConfig.Authorization.Authorizer = union.New(
 		kubeletauthorizer.New(s.uncachedVirtualClient),
 		delegatingauthorizer.New(s.uncachedVirtualClient, redirectAuthResources, redirectAuthNonResources),


flomedja:
@janekbaraniewski

pkg/integrations/integrations.go
```diff
@@ -2,13 +2,15 @@ package integrations
...
 	"github.com/loft-sh/vcluster/pkg/integrations/metricsserver"
+	"github.com/loft-sh/vcluster/pkg/integrations/vclusterfeatures"
 	"github.com/loft-sh/vcluster/pkg/syncer/synccontext"
 )
 
 type Integration func(ctx *synccontext.ControllerContext) error
 
 var Integrations = []Integration{
 	metricsserver.Register,
+	vclusterfeatures.Register, // always register; returns {} without pro
```
janekbaraniewski: we should do it as PostServerHook in pro [pkg/setup/controller_context.go](https://github.com/loft-sh/vcluster-pro/blob/7682f4e6d23a9d6b3c3e1cae31d7b20125d5e51c/pkg/setup/controller_context.go#L32-L44)

pkg/pro/license.go
```diff
@@ -21,3 +21,8 @@ var LicenseStart = func(_ *synccontext.ControllerContext) error {
 var LicenseFeatures = func() map[string]bool {
 	return make(map[string]bool)
 }
+
+// LicenseFeaturesMap is used to retrieve the enabled feature map from the platform
+var LicenseFeaturesMap = func(_ context.Context) (map[string]bool, error) {
+	return make(map[string]bool), nil
```
janekbaraniewski: it's pretty much the same as `LicenseFeatures` above

pkg/integrations/vclusterfeatures/vclusterfeatures.go
```diff
@@ -0,0 +1,102 @@
...
+
+const (
+	// FeaturesPath is the non-resource URL served by this handler and registered
+	// in the delegatingauthorizer in pkg/server/server.go.
+	FeaturesPath = "/vcluster/features"
+	cacheTTL     = 5 * time.Minute
+)
+
+// Register hooks the /vcluster/features handler into the server via PostServerHooks.
+func Register(ctx *synccontext.ControllerContext) error {
```
janekbaraniewski: this whole thing should live in pro, there is no license in oss.

janekbaraniewski: but I also don't think we need it to be so complex. caching is already handled in [online license in pro](https://github.com/loft-sh/vcluster-pro/blob/7682f4e6d23a9d6b3c3e1cae31d7b20125d5e51c/pkg/license/online/license.go#L464-L473) - when getting features you are getting them from a local copy, not from platform. so this could be jkust a very lightweight http.HandlerFunc

janekbaraniewski (changes requested):

pkg/integrations/vclusterfeatures/vclusterfeatures.go
flomedja: Fine to remove the caching if overkill here.

flomedja: I thought about putting it in the `pro` but I was wondering if we would like to make the endpoint generic and implement the pro rewrite of the functions. But as you said this shouldn't never be use with the oss. If it's better I can move it in the pro. 
@FabianKramm what is your take on this ?

pkg/pro/license.go
flomedja: I agree. We don't really need the returned error there. If there is an issue and a empty list will be returned. Let just keep in mind this :

```
Stale data for online licenses: LicenseFeatures() reads from the in-memory cache that's updated by  
the license watcher (on secret changes, reconnects, etc.) — not on-demand. LicenseFeaturesMap makes a
live platform call, which is what makes the handler's own 5-minute TTL cache meaningful. Using         
LicenseFeatures() would make that cache redundant: vNodes would always get whatever happened to be in
memory, with no way to force a refresh within a cache window.
```

If we are fine to use the cache and not call the Platform, I agree with the change :).

pkg/integrations/integrations.go
flomedja: I was thinking about putting it [here](https://github.com/loft-sh/vcluster-pro/blob/75601e6a987ac14a9de0dc034c13e078c4a57e1d/pkg/pro/inject.go#L94)

pkg/integrations/vclusterfeatures/vclusterfeatures.go
FabianKramm: Yes I agree this should not be in OSS, lets do it like @janekbaraniewski says

pkg/integrations/integrations.go
flomedja: Done

pkg/pro/license.go
flomedja: Done

pkg/integrations/vclusterfeatures/vclusterfeatures.go
flomedja: Done

janekbaraniewski (approved):

pkg/server/server.go
```diff
@@ -253,6 +253,12 @@ func (s *Server) ServeOnListenerTLS(ctx *synccontext.ControllerContext) error {
 			)
 		}
 	}
+	redirectAuthNonResources = append(redirectAuthNonResources,
+		delegatingauthorizer.PathVerb{
+			Path: "/vcluster/features",
+			Verb: "get",
```
janekbaraniewski: this should be "*" - currently it only requires auth for GET requests leting all other requests through.

janekbaraniewski (changes requested):

pkg/server/server.go
flomedja: You're right. Fixed

janekbaraniewski (approved):

PR #3824, opened 2026-04-10
chore(e2e-next): Refactor snapshot/restore test

adriankabala:
/kind test

resolves #

  Changes:
  - "without volumes" restore: merged two separate It blocks into one that does all state mutations (add post-snapshot resources + delete pre-snapshot resources) before a single restore, then verifies everything in one pass. Reduces from 2 vCluster restores to 1 (~30s saved per run).
  - "with volumes": merged "Creates the snapshot" and "Verifies VolumeSnapshot cleanup" into a single It block since cleanup verification happens immediately after snapshot creation anyway.
  - canceling: increased time.Sleep between back-to-back snapshots from 1s to 2s for more reliable cancellation testing on slower CI runners.

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
snapshots
```


anvesh-loft (approved):
Looks good

PR #3825, opened 2026-04-10
fix: set UseLocationHost=true to fix kubectl logs/exec on Istio-fronted API servers

udayakp:
/kind bugfix

`kubectl logs`, `kubectl exec`, `kubectl port-forward`, and `kubectl attach` fail with:

```
Error from server (NotFound): the server could not find the requested resource
```

on any vcluster whose host cluster API server is fronted by an Istio proxy (e.g. Gardener shoots). The request returns HTTP 404 with an empty body — Istio rejects it before it reaches the API server.

**Root cause**

When `WithRedirect` (`pkg/server/filters/redirect.go`) proxies `pods/log|exec|portforward|attach` to the host cluster, it calls `handler.Handler()` which creates an `UpgradeAwareHandler`. The handler's `UseLocationHost` field defaults to `false`, so the outbound HTTP request carries the **original client's `Host` header** (the vcluster LoadBalancer hostname, e.g. `abc123.us-west-2.elb.amazonaws.com`) instead of the actual backend API server hostname.

On clusters with an Istio-fronted API server, Istio enforces virtual host routing and returns `404 Not Found` (Content-Length: 0) for any `Host` value it doesn't recognise. The fix is a single line.

`UseRequestLocation = true` (already set) only rewrites the URL path — it does **not** fix the `Host` header.

**The fix**

```go
// pkg/server/handler/handler.go
proxy.UseRequestLocation = true
proxy.UseLocationHost = true  // ← add this line
```

`UseLocationHost = true` tells the `UpgradeAwareHandler` to overwrite `req.Host` with `h.Location.Host` (the actual backend hostname) before forwarding. This is the correct behaviour for a transparent reverse proxy.

**Verification**

Confirmed by running `wget` from inside the syncer pod with both correct and wrong `Host` headers:

```bash
# Correct Host → 200 OK
wget --header="Host: api.vclster-hst...ondemand.com" "https://api.vclster-hst...ondemand.com/api/v1/.../log"

# Wrong Host (what vcluster currently sends) → 404, server: istio-envoy, Content-Length: 0
wget --header="Host: abc123.us-west-2.elb.amazonaws.com" "https://api.vclster-hst...ondemand.com/api/v1/.../log"
```

Patched image deployed and tested end-to-end on a Gardener shoot — `kubectl logs`, `kubectl exec`, and `kubectl port-forward` all work after the fix.

**Why this only surfaces on certain clusters**

| Environment | API server behind Istio? | Effect of wrong Host header |
|---|---|---|
| Gardener shoot | Yes (`server: istio-envoy` in every response) | `404 Not Found` — Istio rejects unknown virtual host |
| Kyma cluster | No (Istio only for workload traffic) | Ignored — plain kube-apiserver accepts any Host |
| Kind / vanilla k8s | No | Ignored — plain kube-apiserver accepts any Host |

The bug exists in all environments but is only observable where Istio sits in front of the API server.

Fixed `kubectl logs`, `kubectl exec`, `kubectl port-forward`, and `kubectl attach` returning `Error from server (NotFound)` on clusters with an Istio-fronted API server (e.g. Gardener shoots). The `UpgradeAwareHandler` was forwarding the client's original `Host` header to the backend instead of the backend's own hostname, causing Istio virtual host routing to reject the request.


udayakp:
Closing — branch was based on an old commit and includes unrelated changes. Reopening with a clean single-commit branch against current main.

PR #3832, opened 2026-04-10
chore(e2e-next): Fix lifecycle tests to do not use cluster.Use

adriankabala:
/kind test

resolves #

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

### DIFF
--- a/e2e-next/test_core/lifecycle/test_lifecycle.go
+++ b/e2e-next/test_core/lifecycle/test_lifecycle.go
@@ -16,7 +16,10 @@ import (
 	"github.com/loft-sh/vcluster/pkg/util/random"
 )
 
-var _ = Describe("Tenant cluster lifecycle - create and delete", labels.Core, labels.PR, func() {
+// TenantClusterLifecycleSpec registers tenant cluster create/list/delete tests.
+// These tests create their own vclusters via the CLI (not framework-provisioned).
+func TenantClusterLifecycleSpec() {
+	Describe("Tenant cluster lifecycle - create and delete", labels.Core, labels.PR, func() {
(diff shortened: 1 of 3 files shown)


e2e-next/test_core/lifecycle/test_lifecycle.go
```diff
@@ -16,133 +16,137 @@ import (
...
-			Expect(err).To(Succeed())
-		})
-
-		It("should create a tenant cluster", func(ctx context.Context) {
-			By("Creating a tenant cluster", func() {
-				_, err := runVClusterCmd(ctx, createArgs(clusterName, namespace)...)
-				Expect(err).To(Succeed())
+// TenantClusterLifecycleSpec registers tenant cluster create/list/delete tests.
+// These tests create their own vclusters via the CLI (not framework-provisioned).
+func TenantClusterLifecycleSpec() {
```
sowmyav27, line 21: Why is this function named TenantCluster? If its shared vCluster, should we make it more specific?

sowmyav27 (approved):

e2e-next/test_core/lifecycle/test_lifecycle.go
adriankabala: TenantCluster" here refers to the vcluster instance being tested, not the shared framework cluster. This test creates its own vcluster via CLI (vcluster create) and tests the full create/list/delete lifecycle. The name distinguishes it from ConnectSpec/PauseResumeSpec which operate on framework-provisioned clusters. That said, renaming to CLILifecycleSpec would also work if "TenantCluster" is confusing. Let me know please if I should change it

PR #3863, opened 2026-04-16
Explain TLS verification during platform start

cbron:


rmweir (approved):
I was thinking to enable it by default and print warning but this PR looks fine if this is what we want.

PR #3868, opened 2026-04-17
Refactor config load for standalone snapshots

mfranczy:
/kind enhancement

resolves #

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

### DIFF
--- a/pkg/cli/snapshot_helm.go
+++ b/pkg/cli/snapshot_helm.go
@@ -12,7 +12,6 @@ import (
 	"github.com/loft-sh/vcluster/pkg/cli/find"
 	"github.com/loft-sh/vcluster/pkg/cli/flags"
 	vclusterconfig "github.com/loft-sh/vcluster/pkg/config"
-	"github.com/loft-sh/vcluster/pkg/constants"
 	"github.com/loft-sh/vcluster/pkg/helm"
 	"github.com/loft-sh/vcluster/pkg/snapshot"
 	"github.com/loft-sh/vcluster/pkg/snapshot/pod"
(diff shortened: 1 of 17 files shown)


pkg/cli/snapshot_helm.go
```diff
@@ -124,9 +124,20 @@ func initSnapshotCommand(
...
+		vCluster, err = find.GetStandaloneVCluster()
+		if err != nil {
+			return nil, nil, nil, err
+		}
+		if vCluster == nil {
+			return nil, nil, nil, fmt.Errorf("local standalone vCluster not found on this host")
+		}
+	} else {
+		vCluster, err = find.GetVCluster(ctx, globalFlags.Context, vClusterName, globalFlags.Namespace, log)
+		if err != nil {
```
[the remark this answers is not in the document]
mfranczy: That's fine, we never released this.

FabianKramm (dismissed):

jjaferson (dismissed):

pkg/config/load.go
```diff
@@ -108,32 +83,79 @@ func LoadLocalStandaloneConfig(name string, setValues []string) (*VirtualCluster
...
+// applies set values, overlays the chart default config values, and normalizes
+// it as standalone. When the path is one of the default standalone config
+// locations and the file does not exist, it falls back to an in-memory default
+// config before merging chart defaults. Missing custom config paths still
+// return an error.
+func LoadStandaloneRuntimeConfig(name, path string, setValues []string) (*VirtualClusterConfig, error) {
+	rawConfig, err := os.ReadFile(path)
 	if err != nil {
-		return nil, err
+		if !os.IsNotExist(err) || !isDefaultStandaloneConfigPath(path) {
```
jjaferson, line 98: Shouldn't this be `&&` (and)? so we return an error if path is not default and it doesn't exist? 

```suggestion
		if !os.IsNotExist(err) && !isDefaultStandaloneConfigPath(path) {
```

pkg/config/load_test.go
```diff
@@ -32,6 +32,39 @@ func TestLoadNormalizedStandaloneConfig(t *testing.T) {
...
+func TestLoadStandaloneRuntimeConfig_MissingDefaultPath(t *testing.T) {
+	cfg, err := LoadStandaloneRuntimeConfig("test", "", nil)
+	if err != nil {
+		t.Fatalf("LoadStandaloneRuntimeConfig() error = %v", err)
+	}
+
+	if cfg.Name != "test" {
+		t.Fatalf("expected config name to be preserved, got %q", cfg.Name)
+	}
+	if cfg.Path != constants.VClusterStandaloneDefaultConfigPath {
```
jjaferson: here should be equals to default standalone path, right?

```suggestion
	if cfg.Path == constants.VClusterStandaloneDefaultConfigPath {
```

pkg/config/load.go
mfranczy: It shouldn't be `&&`. We want to fall back only in one narrow case: when the file is missing and the path is one of the default standalone paths.
 
 The current || enforces exactly that:
 - if the error is anything other than not found, fail immediately
 - if the error is not found, then only default standalone paths are allowed to continue
 - for a custom path, not found must still fail

pkg/config/load_test.go
mfranczy: No, the test was wrong, I forgot to update it.

FabianKramm (approved):

PR #3872, opened 2026-04-20
ci(e2e): migrate run-ginkgo-e2e to centralized composite action

sydorovdmytro:
/kind enhancement

Replace the local `run-ginkgo-e2e` action with `loft-sh/github-actions` `run-ginkgo/v1`. vcluster-specific setup (artifact downloads, syncer image load, vcluster CLI install, bridge netfilter modules, docker prune) moves into the workflow files. The `|| pr` label append is now caller-side.

The nightly workflow drops the manual ginkgo install and inline label-filter logic in favor of the shared action.

Matches the migration pattern already landed in loft-enterprise#6697.

Related to DEVOPS-769

**Release note**
```
NONE
```

### Additional test suites

```label-filter
none
```


.github/workflows/e2e-ginkgo-nightly.yaml
```diff
@@ -100,18 +95,12 @@ jobs:
...
-            --label-filter="${LABEL_FILTER}" \
-            ./e2e-next -- \
-            --vcluster-image="${{ env.REPOSITORY_NAME }}:${{ env.TAG_NAME }}" \
-            --teardown=false
+        uses: loft-sh/github-actions/.github/actions/run-ginkgo@run-ginkgo/v1
+        with:
+          ginkgo-label: ${{ inputs.ginkgo-label }}
+          timeout: "0"
+          additional-args: "--vcluster-image=${{ env.REPOSITORY_NAME }}:${{ env.TAG_NAME }} --teardown=false"
+          additional-ginkgo-flags: "-v"
```
adriankabala, line 98: can we add:
`--poll-progress-after=20s --poll-progress-interval=10s to additional-ginkgo-flags in both workflows so hang diagnosis remains intact.`

roehrijn: @adriankabala, if you want this, up to you. I removed that in platform a long time ago because it makes Githubs Workflow UI practically unusable due to the amount of logs being processed on client side in the browser. And IMHO it rather adds than reduces friction when diagnosing something, but this is my personal opinion. 
I would rather handle it like in platform where you have the option to give `additional-ginko-flags` when triggering the workflow manually. So you always have the option to add it in particular cases when its needed.

.github/workflows/e2e-ginkgo-nightly.yaml
```diff
@@ -100,18 +95,12 @@ jobs:
...
-            --poll-progress-after=20s \
-            --poll-progress-interval=10s \
-            --label-filter="${LABEL_FILTER}" \
-            ./e2e-next -- \
-            --vcluster-image="${{ env.REPOSITORY_NAME }}:${{ env.TAG_NAME }}" \
-            --teardown=false
+        uses: loft-sh/github-actions/.github/actions/run-ginkgo@run-ginkgo/v1
+        with:
+          ginkgo-label: ${{ inputs.ginkgo-label }}
+          timeout: "0"
```
roehrijn: we should add a reasonable timeout, which is slightly lower than the overall workflow timeout, so Ginkgo has a chance to teardown in case its timeout hits.

.github/workflows/e2e-ginkgo-nightly.yaml
adriankabala: Ok, lets give it a try then. I can always later change it

.github/workflows/e2e-ginkgo.yaml
```diff
@@ -208,18 +208,41 @@ jobs:
...
+        with:
+          name: vcluster
+
+      - name: Download syncer image
+        uses: actions/download-artifact@v6
+        with:
+          name: vcluster_syncer
+
+      - name: Clean up docker images
+        run: docker image prune --all --force --filter="label!=PRUNE=false"
```
roehrijn: not needed IMHO.

adriankabala (dismissed):

roehrijn (approved):

PR #3884, opened 2026-04-22
feat: add sanitise pod name for events

flomedja:
/kind bugfix
/kind enhancement
/kind feature
/kind documentation
/kind test

resolves #ENGNODE-324

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

### DIFF
--- a/e2e-next/test_core/sync/test_sync_error_events.go
+++ b/e2e-next/test_core/sync/test_sync_error_events.go
@@ -1,0 +1,202 @@
+package test_core
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/loft-sh/e2e-framework/pkg/setup/cluster"
+	"github.com/loft-sh/vcluster/e2e-next/constants"
+	"github.com/loft-sh/vcluster/e2e-next/labels"
+	"github.com/loft-sh/vcluster/pkg/util/random"
+	"github.com/loft-sh/vcluster/pkg/util/translate"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
+)
+
+// SyncErrorSanitisationSpec verifies that SyncError warning events recorded on
+// virtual pods never expose the host-side translated pod name.
+//
+// The host pod name (e.g. "my-pod-x-default-x-my-vc") must not appear in event
+// messages visible from inside the virtual cluster. The sanitising event recorder
+// (pkg/syncer/translator/sanitising_event_recorder.go) rewrites it back to the
+// virtual name before the event is persisted.
+func SyncErrorSanitisationSpec() {
+	Describe("SyncError event sanitisation",
+		labels.PR, labels.Core, labels.Events, labels.Pods, labels.Sync,
+		func() {
+			var (
+				hostClient     kubernetes.Interface
+				vClusterClient kubernetes.Interface
+				vClusterName   string
+			)
+
+			BeforeEach(func(ctx context.Context) {
+				hostClient = cluster.KubeClientFrom(ctx, constants.GetHostClusterName())
+				Expect(hostClient).NotTo(BeNil())
+				vClusterClient = cluster.CurrentKubeClientFrom(ctx)
+				Expect(vClusterClient).NotTo(BeNil())
+				vClusterName = cluster.CurrentClusterNameFrom(ctx)
+			})
+
+			It("should not expose the host pod name in SyncError events", labels.PR, func(ctx context.Context) {
+				suffix := random.String(6)
+				ns := "sync-err-test-" + suffix
+				podName := "test-pod-" + suffix
+
+				By("creating a test namespace in the virtual cluster", func() {
+					_, err := vClusterClient.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+						ObjectMeta: metav1.ObjectMeta{Name: ns},
+					}, metav1.CreateOptions{})
+					Expect(err).To(Succeed())
+					DeferCleanup(func(ctx context.Context) {
+						err := vClusterClient.CoreV1().Namespaces().Delete(ctx, ns, metav1.DeleteOptions{})
+						if !kerrors.IsNotFound(err) {
+							Expect(err).To(Succeed())
+						}
+					})
+				})
+
+				// hostPodName is the translated host-side name that MUST NOT appear
+				// in any virtual pod event (e.g. "test-pod-abc123-x-sync-err-test-abc123-x-my-vc").
+				hostPodName := translate.SingleNamespaceHostName(podName, ns, vClusterName)
+				hostNS := vClusterHostNamespace(vClusterName)
+
+				By("creating a pod in the virtual cluster", func() {
+					_, err := vClusterClient.CoreV1().Pods(ns).Create(ctx, &corev1.Pod{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      podName,
+							Namespace: ns,
+						},
+						Spec: corev1.PodSpec{
+							Containers: []corev1.Container{
+								{
+									Name:            testingContainerName,
+									Image:           testingContainerImage,
+									ImagePullPolicy: corev1.PullIfNotPresent,
+									SecurityContext: &corev1.SecurityContext{
+										Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
+										RunAsNonRoot:             boolPtr(true),
+										RunAsUser:                int64Ptr(12345),
+										AllowPrivilegeEscalation: boolPtr(false),
+										SeccompProfile:           &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault},
+									},
+								},
+							},
+						},
+					}, metav1.CreateOptions{})
+					Expect(err).To(Succeed())
+					DeferCleanup(func(ctx context.Context) {
+						err := vClusterClient.CoreV1().Pods(ns).Delete(ctx, podName, metav1.DeleteOptions{})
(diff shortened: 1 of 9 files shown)


pkg/syncer/translator/sanitising_event_recorder.go
```diff
@@ -0,0 +1,83 @@
...
+// host-translated object names, then forwards to the underlying recorder.
+func (s *sanitisingEventRecorder) Eventf(
+	regarding runtime.Object,
+	related runtime.Object,
+	eventtype, reason, action, note string,
+	args ...any,
+) {
+	message := fmt.Sprintf(note, args...)
+
+	acc, accErr := meta.Accessor(regarding)
```
zerbitx: Should we be surfacing non nil `accErr`s?  Either logging if we get one, or putting it in the event message maybe?

pkg/syncer/translator/generic_translator.go
zerbitx, line 10: You should be able to pass this mapper to `newSanitisingEventRecorder` here, and make one then, pass it in `from_config_translator.go`.  You'll need a *synccontext.SyncContext from the RegisterContext as well to use  `VirtualToHost` on the mapper, but that could help address the codex comment about using `strings.ReplaceAll` since you can use the mapper methods directly.

zerbitx: Something along the lines of 

```go
func newSanitisingEventRecorder(underlying events.EventRecorder, syncCtx *synccontext.SyncContext, mapper synccontext.Mapper) events.EventRecorder {
	return &sanitisingEventRecorder{underlying: underlying, syncCtx: syncCtx, mapper: mapper}
}

type sanitisingEventRecorder struct {
	underlying events.EventRecorder
	syncCtx    *synccontext.SyncContext
	mapper     synccontext.Mapper
}

```

flomedja: Thanks. Good idea. However not sure It would help for the codex comment. In the case of the codex comment we don't really know the gvk of the secondary resource. And the comment identity a edge case that I am not even sure is relevant in practice.

pkg/syncer/translator/sanitising_event_recorder.go
flomedja: I add a log :)

zerbitx (dismissed):

janekbaraniewski:
@codex please review

pkg/syncer/translator/from_config_translator.go
```diff
@@ -37,15 +37,21 @@ func NewFromHostTranslatorForGVK(ctx *synccontext.RegisterContext, gvk schema.Gr
...
 		namespace:      ctx.Config.HostNamespace,
 		translatorName: "from-host-" + strings.ToLower(gvk.Kind),
 		skipFuncs:      skipFuncs,
-	}, nil
+	}
+	syncCtx := ctx.ToSyncContext("from-host-" + strings.ToLower(gvk.Kind))
+	t.eventRecorder = newSanitisingEventRecorder(
+		syncCtx,
+		ctx.VirtualManager.GetEventRecorder("from-host-"+strings.ToLower(gvk.Kind)+"-syncer"),
+		t,
```
janekbaraniewski: this cyclic dependency seems wrong

pkg/syncer/translator/sanitising_event_recorder.go
```diff
@@ -0,0 +1,97 @@
...
+	// When syncCtx is nil, VirtualToHostFromStore and RecordMapping nil-check ctx
+	// and skip gracefully, falling back to formula-based translation.
+	var (
+		vName    string
+		hostName types.NamespacedName
+	)
+	if s.mapper != nil {
+		vName = acc.GetName()
+		vObj, _ := regarding.(client.Object)
+		hostName = s.mapper.VirtualToHost(s.syncCtx, types.NamespacedName{Name: vName, Namespace: acc.GetNamespace()}, vObj)
```
janekbaraniewski: can't we use `translate.Default.HostName(nil, baseConfigMap.Name, baseConfigMap.Namespace).Name` instead?

janekbaraniewski (changes requested):

pkg/syncer/translator/from_config_translator.go
flomedja: `t` is initialized before it is passed to the event recorder. Is is referencing itself as the mapper. 

The only real concern would be if someone later added logic to VirtualToHost that emitted an event, which would cause infinite recursion. Details from claude : 

```
The self-reference creates this call chain when an event is recorded:
                                                                                                                                                                                 
  Eventf(regarding, ...)                                                                                                                                                       
    └─ mapper.VirtualToHost(...)   ← calls back into t                                                                                                                           
         └─ matchesVirtualObject(...) ← map lookup, returns
                                                                                                                                                                                 
  It terminates because VirtualToHost is a pure map lookup — it does nothing else.                                                                                               
                                                                                                                                                                                 
  Now imagine someone later modifies VirtualToHost to log a warning when a name isn't found:                                                                                     
                                                                   
  func (c *fromHostTranslate) VirtualToHost(...) types.NamespacedName {                                                                                                          
      nn, ok := matchesVirtualObject(...)                                                                                                                                        
      if !ok {                                                                                                                                                                   
          c.eventRecorder.Eventf(...)  // ← added for observability                                                                                                              
      }                                                                                                                                                                          
      return nn                                                    
  }                                                                                                                                                                              
                                                                   
  Now the chain becomes:

  Eventf → VirtualToHost → Eventf → VirtualToHost → Eventf → ...
                                                                                                                                                                                 
  Stack overflow. The future developer has no obvious signal this is dangerous — fromHostTranslate looks like a normal struct with an event recorder field.                      
                                                                                                                                                                                 
  The risk is that the self-reference is invisible at the call site. When you write c.eventRecorder.Eventf(...) inside VirtualToHost, you don't see that this very struct is the 
  mapper the recorder will call back into.                         
                                                                                                                                                                                 
  This is exactly the kind of thing that passes code review and breaks at runtime under a specific error path (the !ok branch), making it hard to reproduce.  
 ```
 
 I can add clear comments to enlight that this should be avoided.

File: pkg/syncer/translator/sanitising_event_recorder.go (hunk 3 above)
Comment (flomedja) [reply pkg/syncer/translator/sanitising_event_recorder.go]: I used it first. Then @zerbitx pointed that I can use [mapper](https://github.com/loft-sh/vcluster/pull/3884#discussion_r3134180658). The advantage with the mapper is that it covers all use cases of a name translation no matter which Translation method was originally used.

File: pkg/syncer/translator/from_config_translator.go (hunk 2 above)
Comment (flomedja) [reply pkg/syncer/translator/from_config_translator.go]: @janekbaraniewski refactored

File: pkg/syncer/translator/sanitising_event_recorder.go (hunk 3 above)
Comment (flomedja) [reply pkg/syncer/translator/sanitising_event_recorder.go]: @janekbaraniewski refactored

**Review by janekbaraniewski — DISMISSED** [review]

File: e2e-next/test_core/sync/test_sync_error_events.go [hunk 4]
```diff
@@ -0,0 +1,202 @@
...
+// SyncErrorSanitisationSpec verifies that SyncError warning events recorded on
+// virtual pods never expose the host-side translated pod name.
+//
+// The host pod name (e.g. "my-pod-x-default-x-my-vc") must not appear in event
+// messages visible from inside the virtual cluster. The sanitising event recorder
+// (pkg/syncer/translator/sanitising_event_recorder.go) rewrites it back to the
+// virtual name before the event is persisted.
+func SyncErrorSanitisationSpec() {
+	Describe("SyncError event sanitisation",
+		labels.PR, labels.Core, labels.Events, labels.Pods, labels.Sync,
```
adriankabala: Remove labels.PR from the It - it's inherited from the Describe

e2e-next/test_core/sync/test_sync_error_events.go
```diff
@@ -0,0 +1,202 @@
...
+				suffix := random.String(6)
+				ns := "sync-err-test-" + suffix
+				podName := "test-pod-" + suffix
+
+				By("creating a test namespace in the virtual cluster", func() {
+					_, err := vClusterClient.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+						ObjectMeta: metav1.ObjectMeta{Name: ns},
+					}, metav1.CreateOptions{})
+					Expect(err).To(Succeed())
+					DeferCleanup(func(ctx context.Context) {
```
adriankabala: Hoist DeferCleanup out of By() closures into the It body:


      By("creating a test namespace in the virtual cluster", func() {
          _, err := vClusterClient.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
              ObjectMeta: metav1.ObjectMeta{Name: ns},
          }, metav1.CreateOptions{})
          Expect(err).To(Succeed())
      })
      DeferCleanup(func(ctx context.Context) {
          err := vClusterClient.CoreV1().Namespaces().Delete(ctx, ns, metav1.DeleteOptions{})
          if !kerrors.IsNotFound(err) {
              Expect(err).To(Succeed())
          }
      })

e2e-next/test_core/sync/test_sync_error_events.go
```diff
@@ -0,0 +1,202 @@
...
+						}
+					})
+				})
+
+				By("waiting for the syncer to create the host pod", func() {
+					Eventually(func(g Gomega) {
+						_, err := hostClient.CoreV1().Pods(hostNS).Get(ctx, hostPodName, metav1.GetOptions{})
+						g.Expect(err).NotTo(HaveOccurred(),
+							"host pod %s/%s should exist after sync", hostNS, hostPodName)
+					}).WithPolling(constants.PollingInterval).
```
adriankabala: Add `.WithContext(ctx)` and accept ctx

flomedja: Fixed

e2e-next/test_core/sync/test_sync_error_events.go
flomedja: Fixed

e2e-next/test_core/sync/test_sync_error_events.go
flomedja: Fixed

adriankabala (approved):

janekbaraniewski (approved):

PR #3889, opened 2026-04-24
fix: use PV/SC mapper for volumeName and storageClassName translation

### LINKED ISSUE #3810: PVC syncer unconditionally mangles volumeName and storageClassName, ignoring host-pv annotation
### What happened?

When sync.toHost.persistentVolumes.enabled: true, creating a PVC inside a vCluster with an explicit volumeName pointing to a pre-existing host PV results in the PVC syncer mangling the volumeName to vcluster-<pv-name>-x--<hash>. The storageClassName is also mangled similarly. No PV or SC with the mangled names exist on the host, so the PVC stays Pending permanently and the syncer enters a recreate loop. This happens even when the PV was synced from the host and has the vcluster.loft.sh/host-pv annotation, which the PV mapper (pkg/mappings/resources/persistentvolumes.go:18-29) is designed to check. The PVC syncer bypasses this mapper and calls translate.Default.HostNameCluster() directly.

When a PVC specifies a volumeName pointing to a PV that was synced from the host (identifiable by the vcluster.loft.sh/host-pv annotation), the PVC syncer should pass the volumeName through to the host unchanged, allowing the PVC to bind directly to the existing host PV. The storageClassName should also be handled through the SC mapper instead of being unconditionally translated. 

1. Deploy a vCluster with sync.toHost.persistentVolumes.enabled: true
2. Create a PVC inside the vCluster without volumeName (dynamic provisioning with a Retain SC) — PV is provisioned on host and synced into vCluster with vcluster.loft.sh/host-pv annotation
3. Delete the PVC — PV survives with Retain policy
4. Clear claimRef on the PV: kubectl patch pv <pv-name> --type merge -p '{"spec":{"claimRef":null}}'
5. Create a new PVC with explicit volumeName pointing to the existing PV: 
apiVersion: v1                                                                                                                                                                                                                                 
  kind: PersistentVolumeClaim                                                                                                                                                                                                                    
  metadata:                                                                                                                                                                                                                                      
    name: test-rebind-pvc                                                                                                                                                                                                                        
    namespace: default                                                                                                                                                                                                                           
  spec:                                                                                                                                                                                                                                          
    accessModes: [ReadWriteOnce]                                                                                                                                                                                                                 
    resources:                                                                                                                                                                                                                                   
      requests:                                                                                                                                                                                                                                  
        storage: 256Mi                                                                                                                                                                                                                        
    storageClassName: <sc-name>                                                                                                                                                                                                                  
    volumeName: <pv-name>   

6. Observe PVC stays Pending. The syncer logs:
recreate persistent volume claim because volumeName differs between physical and virtual pvc:                                                                                                                                                 pvc-<uid> != vcluster-pvc-<uid>-x--<hash>                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                 
 
  

_No response_

<details>

```console
$ kubectl version
# v1.35.2+k3s1 
```

</details>


### vcluster version

<details>

```console
$ vcluster --version
# v0.27.1 (also confirmed on v0.22.0)
```

</details>


### VCluster Config

<details>

```
sync:                                                                                                                                                                                                                                          
    toHost:                                                                                                                                                                                                                                      
      persistentVolumes:                                                                                                                                                                                                                         
        enabled: true                                                                                                                                                                                                                            
      persistentVolumeClaims:                                                                                                                                                                                                                    
        enabled: true
```

</details>

ChellaVigneshKP:
/kind bugfix

The PVC syncer's translateSelector function calls translate.Default.HostNameCluster() directly to translate volumeName and storageClassName, bypassing the PV and SC mappers. This unconditionally mangles these names even when the PV was synced from the host (has vcluster.loft.sh/host-pv annotation), making it impossible to bind a PVC to a pre-existing host PV by name.
Additionally, the PV syncer's translate function has the same pattern for storageClassName, causing SC name mismatches between PVs and PVCs on the host when toHost.storageClasses.enabled is false.
This fix replaces 5 instances of HostNameCluster() with mappings.VirtualToHostName() across 2 files, routing through the proper mappers which conditionally translate based on resource origin — host-synced resources retain their original names, vCluster-created resources still get translated normally.
resolves #3810

Fixed an issue where vCluster unconditionally mangled volumeName and storageClassName when syncing PVCs to the host, even for PVs synced from the host with the vcluster.loft.sh/host-pv annotation. PVCs with explicit volumeName pointing to a pre-existing host PV now bind correctly.

The PV mapper (pkg/mappings/resources/persistentvolumes.go:18-29) already had the correct annotation-aware logic — this fix simply ensures the PVC and PV syncers use it instead of bypassing it.
The fix also addresses the PV syncer's storageClassName translation (persistentvolumes/translate.go:20) to prevent SC name mismatches between PVs and PVCs on the host when toHost.storageClasses.enabled is false (mirror mapper scenario).

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

### DIFF
--- a/pkg/controllers/resources/persistentvolumeclaims/translate.go
+++ b/pkg/controllers/resources/persistentvolumeclaims/translate.go
@@ -55,7 +55,7 @@ func (s *persistentVolumeClaimSyncer) translateSelector(ctx *synccontext.SyncCon
 
 	// translate storage class if we manage those in vcluster
 	if s.storageClassesEnabled && storageClassName != "" {
-		translated := translate.Default.HostNameCluster(storageClassName)
+		translated := mappings.VirtualToHostName(ctx, storageClassName, "", mappings.StorageClasses())
 		delete(vPvc.Annotations, deprecatedStorageClassAnnotation)
 		vPvc.Spec.StorageClassName = &translated
 	}
@@ -67,7 +67,7 @@ func (s *persistentVolumeClaimSyncer) translateSelector(ctx *synccontext.SyncCon
 				vPvc.Spec.Selector = translate.HostLabelSelector(vPvc.Spec.Selector)
 			}
 			if vPvc.Spec.VolumeName != "" {
-				vPvc.Spec.VolumeName = translate.Default.HostNameCluster(vPvc.Spec.VolumeName)
+				vPvc.Spec.VolumeName = mappings.VirtualToHostName(ctx, vPvc.Spec.VolumeName, "", mappings.PersistentVolumes())
 			}
 			// check if the storage class exists in the physical cluster
 			if !s.storageClassesEnabled && storageClassName != "" {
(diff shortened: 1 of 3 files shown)


ChellaVigneshKP:
@codex review

ChellaVigneshKP:
@FabianKramm could you take a look when you have a moment? This resolves #3810 . Happy to address any feedback.

sadath-12 (approved):

PR #3890, opened 2026-04-24
chore(e2e-next): inline cluster lifecycle

adriankabala:
/kind test

resolves #
ENGQA-609

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```


sowmyav27:
**Question on label changes — intentional?**

Heads up: a few specs lost `labels.PR` in this refactor and don't pick it up from any ancestor container, so they'll drop out of the mandatory PR gate (default filter `pr || pr` in `.github/workflows/e2e-ginkgo.yaml:240`):

- `test_security/isolation/test_isolation.go` — `IsolationModeSpec` (parent `isolation-mode-vcluster` carries only `labels.Isolation`)
- `test_security/rootless/test_rootlessmode.go` — `RootlessModeSpec` (parent `rootless-vcluster` carries only `labels.Rootless`)
- `test_core/lifecycle/test_lifecycle.go` — `TenantClusterLifecycleSpec` (top-level `init()` registration, no ancestor)
- `test_core/lifecycle/test_pause_resume.go` — `PauseResumeScaledDownSpec` (same)

After this merges they'll only run when someone opts in via `core`, `security`, `rootless`, or `isolation`. Was that intentional, or should `labels.PR` go back on those four (or onto the outer suite Describes)?

sowmyav27:
**Nit — stale comment in `e2e-next/clusters/registry.go`**

The retained comment ("Tests in other repos may reference them") is now misleading — all the previously-exported `*VCluster` identifiers have been removed from this file.

Quick check on vcluster-pro: it imports `github.com/loft-sh/vcluster/e2e-next/clusters` (aliased `ossclusters`), but defines its own `SchedulerVCluster` etc. via a local `register(...)` pattern, so it doesn't appear to consume the deleted vars directly — looks safe. But worth confirming what vcluster-pro actually pulls from `ossclusters` now, and either updating or removing the comment to match reality.

adriankabala:
> **Question on label changes — intentional?**
> 
> Heads up: a few specs lost `labels.PR` in this refactor and don't pick it up from any ancestor container, so they'll drop out of the mandatory PR gate (default filter `pr || pr` in `.github/workflows/e2e-ginkgo.yaml:240`):
> 
> * `test_security/isolation/test_isolation.go` — `IsolationModeSpec` (parent `isolation-mode-vcluster` carries only `labels.Isolation`)
> * `test_security/rootless/test_rootlessmode.go` — `RootlessModeSpec` (parent `rootless-vcluster` carries only `labels.Rootless`)
> * `test_core/lifecycle/test_lifecycle.go` — `TenantClusterLifecycleSpec` (top-level `init()` registration, no ancestor)
> * `test_core/lifecycle/test_pause_resume.go` — `PauseResumeScaledDownSpec` (same)
> 
> After this merges they'll only run when someone opts in via `core`, `security`, `rootless`, or `isolation`. Was that intentional, or should `labels.PR` go back on those four (or onto the outer suite Describes)?

Good catch - regression, not intentional. Restored labels.PR directly on the four inner Describes.

sowmyav27 (approved):

mfranczy (approved):

PR #3900, opened 2026-04-28
fix: validate TLS certificate when downloading Kubernetes bundle

pascalbreuninger:
/kind bugfix

Adds the OSS-side wiring for an `--insecure` flag on the node upgrade flow:

- `vcluster node upgrade --insecure` (in-pod binary)
- `vclusterctl node upgrade <node> --insecure` (user-facing CLI; threaded through to the upgrade pod's command)
- `privateNodes.autoUpgrade.insecure` config field for the auto-upgrade controller

All paths default to **verifying TLS**. Operators using a non-publicly-trusted bundle repository (most commonly the in-cluster `/node/download` fallback that's served by a self-signed control plane certificate) must opt in explicitly.

The actual gating in the download path lives in vcluster-pro: https://github.com/loft-sh/vcluster-pro/pull/1730

Fixed an issue where the private-node upgrade flow skipped TLS verification when downloading the Kubernetes bundle from `github.com/loft-sh/kubernetes`. TLS verification is now enabled by default; pass `--insecure` (manual upgrade) or set `privateNodes.autoUpgrade.insecure: true` (auto-upgrade) when the bundle repository is served by a non-publicly-trusted CA.

- **Behavior change** for `privateNodes.autoUpgrade` against a self-signed control plane endpoint: auto-upgrades will fail until `insecure: true` is set. Worth a release-notes call-out.
- The implementation that gates the actual TLS verification lives in vcluster-pro; this PR is the OSS wiring (struct fields, flags, pod command threading). vcluster-pro PR: https://github.com/loft-sh/vcluster-pro/pull/1730

### Default Test Execution

### Additional test suites

```label-filter
none
```

### DIFF
--- a/chart/values.schema.json
+++ b/chart/values.schema.json
@@ -72,6 +72,10 @@
         "containerSecurityContext": {
           "type": "object",
           "description": "ContainerSecurityContext specifies security context options on the container level for the upgrade container."
+        },
+        "insecure": {
+          "type": "boolean",
+          "description": "Insecure skips TLS certificate verification when the upgrade pod downloads\nthe Kubernetes bundle. Set to true only when the bundle repository (typically\nthe in-cluster /node/download endpoint) is served by a non-publicly-trusted CA."
         }
       },
       "additionalProperties": false,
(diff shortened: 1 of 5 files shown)


roehrijn (approved):

FabianKramm (changes requested):
Lets do a different approach for this

PR #3907, opened 2026-04-29
feat: uptake etcd v3 snapshot facilities for embedded etcd

cbalan:
/kind feature

Part of  #ENGCP-655

Introduce native etcd v3 snapshot/restore for embedded etcd via etcdutl, while preserving the existing key-value tar-stream approach for other backing stores (SQLite, deployed/external etcd).

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
snapshots
```


cbalan:
@codex review

mfranczy (commented):
As agreed at the meeting, please review if we can safely modify the bbolt db offline, in the meantime I will take a closer look into the peers problem.

pkg/snapshot/client.go
```diff
@@ -110,7 +124,115 @@ func (c *Client) Delete(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to create gzip writer: %w", err)
+	}
+	defer gzipWriter.Close()
+
+	// stamp snapshot archive with type meta
+	gzipWriter.Comment = EtcdSnapshotTypeMetaStamp
+
+	tarWriter := tar.NewWriter(gzipWriter)
+	defer tarWriter.Close()
```
jjaferson, line 165: Why is the close being deferred and then closed later [here](https://github.com/loft-sh/vcluster/pull/3907/changes#diff-1c7eb707c246c3684b8e154bd0f05fd46c1f23760e0cae72fef282c56efedee1R209)

pkg/snapshot/restoreclient.go
```diff
@@ -165,24 +172,248 @@ func (o *RestoreClient) Run(ctx context.Context, vConfig *config.VirtualClusterC
...
+
+	log.Info("Restoring etcd snapshot", "dbPath", dbPath)
+	if err := snapshot.NewV3(zap.L().Named("restore-etcd")).Restore(snapshot.RestoreConfig{
+		SnapshotPath:        dbPath,
+		Name:                "default",
+		OutputDataDir:       constants.EmbeddedEtcdData,
+		OutputWALDir:        datadir.ToWALDir(constants.EmbeddedEtcdData),
+		PeerURLs:            []string{"http://127.0.0.1:2380"},
+		InitialCluster:      "default=http://127.0.0.1:2380",
+		InitialClusterToken: "etcd-cluster",
```
jjaferson: shouldn't these hardcoded values be configured or moved into constants?

pkg/snapshot/restoreclient.go
```diff
@@ -165,24 +172,248 @@ func (o *RestoreClient) Run(ctx context.Context, vConfig *config.VirtualClusterC
...
+			log.Info("Deleting skipped key", "key", key)
+			if _, err := etcdClient.Delete(ctx, key); ignoreKeyNotFound(err) != nil {
+				return fmt.Errorf("failed to delete key %s: %w", key, err)
+			}
+		}
+	}
+
+	if o.NewVCluster {
+		log.Info("Deleting old kube-root-ca.crt")
+		cmsCh, cmsErrCh := mirror.NewSyncer(etcdClient, "/registry/configmaps/", 0).SyncBase(ctx)
```
jjaferson: Could we replace `SyncBase` with `etcdClient.Get(...)`?

pkg/snapshot/restoreclient.go
```diff
@@ -165,24 +172,248 @@ func (o *RestoreClient) Run(ctx context.Context, vConfig *config.VirtualClusterC
...
+	defer os.Remove(snapshotPath)
+
+	typeMetaStamp, err := gzipCommentFromFile(snapshotPath)
+	if err != nil {
+		return fmt.Errorf("failed to extract type meta stamp from snapshot: %w", err)
+	}
+
+	if typeMetaStamp == EtcdSnapshotTypeMetaStamp {
+		if o.RestoreVolumes {
+			return fmt.Errorf("restore volumes is not supported for etcd snapshots")
```
jjaferson, line 205: just curious why is volumes not supported for etcd snapshots? If we are not supporting it we should add this to the docs

pkg/snapshot/restoreclient.go
```diff
@@ -165,24 +173,241 @@ func (o *RestoreClient) Run(ctx context.Context, vConfig *config.VirtualClusterC
...
+		Name:            etcdlocal.DefaultName,
+		OutputDataDir:   restoreDataDir,
+		OutputWALDir:    datadir.ToWALDir(restoreDataDir),
+		PeerURLs:        []string{etcdlocal.DefaultListenPeerURL},
+		InitialCluster:  etcdlocal.DefaultName + "=" + etcdlocal.DefaultListenPeerURL,
+		SkipHashCheck:   false,
+		InitialMmapSize: backend.InitialMmapSize,
+		RevisionBump:    uint64(BumpRevision),
+		MarkCompacted:   true,
+	}); err != nil {
```
mfranczy, line 333: Restore should be done with the real etcd member metadata. We should avoid restoring with placeholders and then fixing the member at runtime via `ensurePeerUrls()`.

For in-cluster embedded etcd, the peer URL is predictable and should use the headless DNS name, e.g.: `https://<pod-name>.<vcluster-name>-headless.<namespace>:2380`
So restore can set the correct `Name`, `PeerURLs`, and `InitialCluster` up front.

For standalone embedded etcd, we can resolve the real peer identity with [ResolveStandaloneIPAddress](https://github.com/loft-sh/vcluster/blob/main/pkg/util/standalone/metadata.go#L34)

That makes `ensurePeerUrls()` unnecessary for the normal restore path.

mfranczy (changes requested):

mfranczy:
@codex review

pkg/snapshot/restoreclient.go
cbalan: Fixed.

pkg/snapshot/restoreclient.go
cbalan: Yes. The cluster membership related values were moved to out to use actual member name.

pkg/snapshot/restoreclient.go
cbalan: Good point. Here we only need the keys. I'll do so now.

cbalan: Fixed. The config maps listing is using a regular get operation with keys only flag. 

The pods listing section is still using SyncBase as it provides pagination out of the box.

pkg/snapshot/client.go
cbalan: The deferred close is used here to prevent potential resources leak in case of a failure. 
On the success path, the second close should be a no-op.  

I believe the same approach is used on the key-value snapshot implementation - https://github.com/loft-sh/vcluster/blob/24d397fa5e20a786a2c5feb0e398ca264114cb04/pkg/snapshot/client.go#L202

pkg/snapshot/restoreclient.go
```diff
@@ -165,24 +173,250 @@ func (o *RestoreClient) Run(ctx context.Context, vConfig *config.VirtualClusterC
...
+	defer os.Remove(snapshotPath)
+
+	typeMetaStamp, err := gzipCommentFromFile(snapshotPath)
+	if err != nil {
+		return fmt.Errorf("failed to extract type meta stamp from snapshot: %w", err)
+	}
+
+	if typeMetaStamp == EtcdSnapshotTypeMetaStamp {
+		if o.RestoreVolumes {
+			return fmt.Errorf("restore volumes is not supported for etcd snapshots")
```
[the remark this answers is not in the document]
cbalan: This is intentional. The volume snapshotting feature, currently in Beta, is considered to be scheduled for deprecation.

pkg/snapshot/restoreclient.go
cbalan: Good point. The documentation will be updated to reflect this change. The volume snapshotting feature, currently in Beta, is considered to be scheduled for deprecation.

mfranczy:
@codex review

pkg/snapshot/restoreclient.go
```diff
@@ -165,24 +175,258 @@ func (o *RestoreClient) Run(ctx context.Context, vConfig *config.VirtualClusterC
...
+		SnapshotPath:        dbPath,
+		Name:                name,
+		OutputDataDir:       constants.EmbeddedEtcdData,
+		OutputWALDir:        datadir.ToWALDir(constants.EmbeddedEtcdData),
+		PeerURLs:            []string{peerURL},
+		InitialCluster:      name + "=" + peerURL,
+		InitialClusterToken: "vcluster",
+		SkipHashCheck:       false,
+		InitialMmapSize:     backend.InitialMmapSize,
+		RevisionBump:        uint64(BumpRevision),
```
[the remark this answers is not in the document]
cbalan: I do not believe this is true. The etcdutl snapshot restore is bumping the revision using latestRevision + BumpRevision - https://github.com/etcd-io/etcd/blob/f4f0a238762de22026f2e8ff0fd7c040d0667441/etcdutl/snapshot/v3_snapshot.go#L403

pkg/snapshot/restoreclient.go
```diff
@@ -165,24 +175,258 @@ func (o *RestoreClient) Run(ctx context.Context, vConfig *config.VirtualClusterC
...
+	defer os.Remove(snapshotPath)
+
+	typeMetaStamp, err := gzipCommentFromFile(snapshotPath)
+	if err != nil {
+		return fmt.Errorf("failed to extract type meta stamp from snapshot: %w", err)
+	}
+
+	if typeMetaStamp == EtcdSnapshotTypeMetaStamp {
+		if o.RestoreVolumes {
+			return fmt.Errorf("restore volumes is not supported for etcd snapshots")
```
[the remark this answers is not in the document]
cbalan: This is intentional. The volume snapshotting feature, currently in Beta, is considered to be scheduled for deprecation.

pkg/snapshot/restoreclient.go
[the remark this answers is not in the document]
cbalan: I was wrong. The revision bump is applied against the snapshot revision. Not the /data/etcd latest revision before the restore call. This comment is valid.

pkg/snapshot/client.go
```diff
@@ -110,7 +124,115 @@ func (c *Client) Delete(ctx context.Context) error {
...
+	if c.Request != nil {
+		log.Info("Adding snapshot request to snapshot archive")
+		requestBytes, err := json.Marshal(c.Request)
+		if err != nil {
+			return fmt.Errorf("failed to marshal snapshot request: %w", err)
+		}
+		key := fmt.Sprintf("%s/%s", RequestStoreKey, APIVersion)
+		err = writeKeyValue(tarWriter, []byte(key), requestBytes)
+		if err != nil {
+			return fmt.Errorf("failed to snapshot snapshot request: %w", err)
```
mfranczy: should be `failed to snapshot request`

pkg/snapshot/client.go
```diff
@@ -110,7 +124,115 @@ func (c *Client) Delete(ctx context.Context) error {
...
+
+	// don't compress with default level as this can get quite cpu heavy
+	gzipWriter, err := gzip.NewWriterLevel(snapshotFileWrite, 3)
+	if err != nil {
+		return fmt.Errorf("failed to create gzip writer: %w", err)
+	}
+	defer gzipWriter.Close()
+
+	// stamp snapshot archive with type meta
+	gzipWriter.Comment = EtcdSnapshotTypeMetaStamp
```
mfranczy: I don't think this is a reliable way to check if a snapshot is an etcd snapshot. Could you validate the snapshot over the [Status](https://pkg.go.dev/go.etcd.io/etcd/etcdutl/v3/snapshot#Status) func?

pkg/snapshot/client.go
```diff
@@ -110,7 +124,115 @@ func (c *Client) Delete(ctx context.Context) error {
...
+	defer tarWriter.Close()
+
+	if c.Options.Release != nil {
+		log.Info("Adding vCluster config to snapshot archive")
+		releaseBytes, err := json.Marshal(c.Options.Release)
+		if err != nil {
+			return fmt.Errorf("failed to marshal vCluster release: %w", err)
+		}
+
+		err = writeKeyValue(tarWriter, []byte(SnapshotReleaseKey), releaseBytes)
```
mfranczy: I know you didn't add this but could you please rename:

writeKeyValue       -> writeArchiveEntry
writeKeyValueFile -> writeArchiveFileEntry
readKeyValue        -> readArchiveEntry

this KeyValue is very confusing

pkg/snapshot/client.go
```diff
@@ -110,7 +124,115 @@ func (c *Client) Delete(ctx context.Context) error {
 	return nil
 }
 
+// writeSnapshot pulls a point-in-time snapshot from etcd, wraps it into a tar.gz archive, and stores it in the object store.
 func (c *Client) writeSnapshot(ctx context.Context, etcdClient etcd.Client, objectStore types.Storage) error {
```
mfranczy: Can we call it writeEtcdSnapshot?

mfranczy (changes requested):

pkg/snapshot/restoreclient.go
```diff
@@ -165,24 +175,273 @@ func (o *RestoreClient) Run(ctx context.Context, vConfig *config.VirtualClusterC
...
+	return nil
+}
+
+// restoreSnapshot restores the vCluster embedded etcd backing store as follows:
+//  1. read snapshot data from the snapshot archive
+//  2. restore the etcd snapshot via etcdutl snapshot package into etcd dataDir
+//  3. mutate etcd data via embedded-etcd instance:
+//     - remove skipped keys
+//     - reset pods nodeName and status
+func (o *RestoreClient) restoreSnapshot(ctx context.Context, vConfig *config.VirtualClusterConfig, snapshotPath string) (retErr error) {
```
mfranczy: Could you name it restoreEtcdSnapshot?

cbalan: Fixed.

pkg/snapshot/client.go
cbalan: Fixed.

pkg/snapshot/client.go
cbalan: Fixed.

pkg/snapshot/client.go
cbalan: Fixed.

pkg/snapshot/client.go
cbalan: The challenge here is the fact that the snapshot tar.gz file is not accompanied by any type metadata. The gzip comment header field is used to fill in this gap. It allows making the distinction between the tar.gz file that may contain key-value dump or an etcd binary snapshot, without unpacking the archive. 

To run the status func, we would need to unzip the archive, look for the snapshot key, write the file, and then run the status function. This logic is already executed in the restoreEtcdSnapshot archive.

cbalan: Refactored the snapshot archive type detection logic to open the snapshot archive and look for the DBKey entry.

pkg/snapshot/restoreclient.go
```diff
@@ -855,3 +1131,91 @@ func getTranslatedPVCName(pvcName string) string {
...
+	gzipReader, err := gzip.NewReader(f)
+	if err != nil {
+		return UnknownSnapshotKind, fmt.Errorf("create gzip reader: %w", err)
+	}
+	defer gzipReader.Close()
+
+	tarReader := tar.NewReader(gzipReader)
+
+	// look for the etcd snapshot key in the first 50 keys.
+	for range 50 {
```
mfranczy: Why not 51 keys? :P

cbalan: We need a limit to avoid scanning the whole snapshot archive. 50 provides enough room for potential snapshot archive structure changes. Should we look for other options for determining the snapshot archive type?

mfranczy (changes requested):
Following your idea with metadata. What if we define a small preamble for the tar file?
Right now we already rely on `/vcluster/snapshot/release` being the first entry if it exists. We could extend that and say that `/vcluster/snapshot/metadata` comes right after the optional release entry and contains the snapshot type.

Then we don't need the `for 50` check and we also don't need to unpack/read the archive twice. We can read the tar once, check `release? -> metadata?`, and if metadata is missing treat it as legacy key-value. Just need to make sure we don't lose the first non-metadata header that was already read.

mfranczy:
As agreed in the meeting, we are not going to introduce a preamble. We will check whether the DB file is in the second position.

pkg/snapshot/restoreclient.go
cbalan: Per our chat, this function was refactored to look for the exact position of the db key to determine the snapshot archive kind.

mfranczy (approved):

PR #3910, opened 2026-04-30
feat(metrics): expose kine metrics via /metrics/kine

pascalbreuninger:
## Summary

- Bind kine's metrics listener to `127.0.0.1:2381` instead of `0` (disabled), so kine's Prometheus metrics are reachable inside the pod.
- Proxy them through the control plane on `/metrics/kine`, gated to non-etcd backing stores (embedded sqlite, embedded database, external database).
- Add a ServiceMonitor endpoint scraping `/metrics/kine` for non-etcd deployments, with corresponding chart unit tests.
- Disable the listener in the snapshot restore path to avoid a port conflict with the running kine.

Closes ENGCP-400

## Test plan

- [x] go build ./... in vcluster
- [x] go test ./pkg/server/... ./pkg/k8s/... ./pkg/snapshot/...
- [x] helm unittest against chart/tests/service-monitor_test.yaml (default sqlite, virtual scheduler, embedded etcd, deployed etcd, external etcd, external database)
- [x] In-cluster verification: curl -k https://<vcluster-pod>:<port>/metrics/kine returns Prometheus metrics; ServiceMonitor scrape succeeds in Prometheus

### DIFF
--- a/chart/templates/service-monitor.yaml
+++ b/chart/templates/service-monitor.yaml
@@ -108,4 +108,25 @@ spec:
         name: vc-{{ .Release.Name }}
         key: client-key
   {{- end }}
+  {{- if not (or .Values.controlPlane.backingStore.etcd.embedded.enabled .Values.controlPlane.backingStore.etcd.deploy.enabled .Values.controlPlane.backingStore.etcd.external.enabled) }}
+  - interval: 30s
+    port: https
+    path: /metrics/kine
+    scheme: https
+    relabelings:
+      - targetLabel: endpoint
+        replacement: "kine"
+    tlsConfig:
+      ca:
+        secret:
+          name: vc-{{ .Release.Name }}
+          key: certificate-authority
+      cert:
+        secret:
+          name: vc-{{ .Release.Name }}
+          key: client-certificate
+      keySecret:
+        name: vc-{{ .Release.Name }}
+        key: client-key
+  {{- end }}
 {{- end }}
--- a/pkg/server/filters/k8s_metrics.go
+++ b/pkg/server/filters/k8s_metrics.go
@@ -3,6 +3,8 @@ package filters
 import (
 	"net/http"
 
+	"github.com/loft-sh/vcluster/config"
+	"github.com/loft-sh/vcluster/pkg/constants"
 	"github.com/loft-sh/vcluster/pkg/server/handler"
 	"github.com/loft-sh/vcluster/pkg/syncer/synccontext"
 	requestpkg "github.com/loft-sh/vcluster/pkg/util/request"
@@ -12,7 +14,7 @@ import (
 const (
 	controllerManagerMetricsHost = "https://127.0.0.1:10257"
 	schedulerMetricsHost         = "https://127.0.0.1:10259"
-	embeddedEtcdMetricsHost      = "http://127.0.0.1:2381"
+	localBackingStoreMetricsURL  = "http://" + constants.LocalBackingStoreMetricsHost
 )
 
 func WithK8sMetrics(h http.Handler, registerCtx *synccontext.RegisterContext) http.Handler {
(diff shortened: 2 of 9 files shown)


pascalbreuninger:
@codex review

chart/templates/service-monitor.yaml
```diff
@@ -108,4 +108,25 @@ spec:
         name: vc-{{ .Release.Name }}
         key: client-key
   {{- end }}
+  {{- if not (or .Values.controlPlane.backingStore.etcd.embedded.enabled .Values.controlPlane.backingStore.etcd.deploy.enabled .Values.controlPlane.backingStore.etcd.external.enabled) }}
```
cbalan, line 111: I believe this could be limited to database type backingStore. `Values.controlPlane.backingStore.database.embedded.enabled || Values.controlPlane.backingStore.database.exterla.enabled`.

pkg/server/filters/k8s_metrics.go
```diff
@@ -13,6 +14,7 @@ const (
 	controllerManagerMetricsHost = "https://127.0.0.1:10257"
 	schedulerMetricsHost         = "https://127.0.0.1:10259"
 	embeddedEtcdMetricsHost      = "http://127.0.0.1:2381"
+	kineMetricsHost              = "http://127.0.0.1:2381"
```
cbalan: Would it be possible to use a different port? Just to avoid possible port conflicts as mentioned below. 
Also, I believe we get a different set of metrics.

cbalan (changes requested):

pkg/server/filters/k8s_metrics.go
pascalbreuninger: It's mutually exclusive, Kine won't run when embedded etcd is enabled 🤔

chart/templates/service-monitor.yaml
pascalbreuninger: Sure, we can express it this way around, too

pkg/server/filters/k8s_metrics.go
cbalan: True, except in the case of migrations when we might need both to be up. 
If it is too awkward to change, we can start with this. Update it later if we do need it.

chart/templates/service-monitor.yaml
pascalbreuninger: Double checked and we can't really flip it because we don't set the `database` fields in the helm values - we just default to embedded SQlite if nothing is set: https://github.com/loft-sh/vcluster/blob/1836b50ed828670b09e25ec5485b7d0be04897aa/config/config.go#L918 
For the chart this means we wouldn't render the kine endpoint in the service monitor if we were to express it that way

cbalan (dismissed):

pkg/server/filters/k8s_metrics.go
mfranczy: We don't need both to be up during migration. We open sqlite db files and copy the data.

mfranczy: Let's have one constant `LocalBackingStoreMetricsHost` that is added to the [constants](https://github.com/loft-sh/vcluster/blob/main/pkg/constants/constants.go#L21) and used in [pkg/k8s/k8s.go](https://github.com/loft-sh/vcluster/pull/3910/changes#diff-27ce9a93c1aa125ec3a168f526c0f3c3da6b53a21e5360e1261f57a17840bc57R281) too.

mfranczy (changes requested):

pkg/server/filters/k8s_metrics.go
pascalbreuninger: Done

mfranczy (approved):

PR #3912, opened 2026-05-05
test: e2e test for large snapshot restore

jjaferson:
kind test

resolves ENGCP-503

Addes e2e test for large snapshot and restore

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
snapshot-large-restore
```


e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -800,6 +803,142 @@ func deletePVC(ctx context.Context, vClusterClient, _ kubernetes.Interface, _, _
 	}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
 }
 
+func describeSnapshotLargeObjectRestore(s *snapshotCtx) {
```
pascalbreuninger, line 815: How heavy is this test? Would it make sense to put it into an adjacent suite with a new label?

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -800,6 +803,142 @@ func deletePVC(ctx context.Context, vClusterClient, _ kubernetes.Interface, _, _
...
+			labelVal     string
+		)
+
+		BeforeAll(func(ctx context.Context) {
+			nsPrefix = "large-restore-" + random.String(6)
+			snapshotPath = "container:///snapshot-data/" + nsPrefix + ".tar.gz"
+			labelVal = nsPrefix
+			cleanupAllSnapshotArtifacts(ctx, s.hostClient, s.vClusterNS)
+
+			for i := range nsCount {
```
pascalbreuninger, line 839: Nit: Maybe a small "By" around this block?

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -800,6 +803,142 @@ func deletePVC(ctx context.Context, vClusterClient, _ kubernetes.Interface, _, _
...
+				}, metav1.CreateOptions{})
+				Expect(err).To(Or(Succeed(), Satisfy(kerrors.IsAlreadyExists)))
+			}
+		})
+
+		It("Writes 100,000 configmaps into the tenant cluster", func(ctx context.Context) {
+			By("Creating 1,000 configmaps in each of 100 namespaces concurrently", func() {
+				// The default client-go rate limiter (5 QPS, burst 10) would make 100k creates
+				// take ~20,000 seconds. Use a high-QPS client so the goroutines actually run in
+				// parallel and we hit the server's throughput ceiling instead.
```
pascalbreuninger, line 852: Do you roughly know how this performs on GH runners? I'm a little concerned about it taking forever there or just straight up failing

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -800,6 +803,142 @@ func deletePVC(ctx context.Context, vClusterClient, _ kubernetes.Interface, _, _
...
+						count, err := countConfigMapsByLabel(ctx, s.vClusterClient, ns, labelKey+"="+labelVal)
+						g.Expect(err).To(Succeed())
+						g.Expect(count).To(Equal(objectsPerNS),
+							"namespace %s should have %d configmaps after restore, got %d", ns, objectsPerNS, count)
+					}
+				}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeoutExtraLong).Should(Succeed())
+			})
+		})
+
+		AfterAll(func(ctx context.Context) {
```
pascalbreuninger, line 952: Could this go into a `DeferCleanup` block instead or does AfterAll also fire appropriately (I don't know)?

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -800,6 +803,142 @@ func deletePVC(ctx context.Context, vClusterClient, _ kubernetes.Interface, _, _
...
+
+		It("Restores snapshot and verifies the deleted namespaces are back", func(ctx context.Context) {
+			By("Restoring the tenant cluster from snapshot", func() {
+				restoreVCluster(ctx, s.hostClient, s.vClusterName, s.vClusterNS, snapshotPath, true, false)
+				s.refreshClient(ctx)
+			})
+
+			By("Verifying the deleted namespaces and their configmaps are restored", func() {
+				Eventually(func(g Gomega) {
+					for i := range deleteNSCount {
```
adriankabala, line 928: This loop only verifies the 4 deleted namespaces (`deleteNSCount`), leaving 96 of 100 namespaces (96,000 of 100,000 ConfigMaps) unverified after restore. The ticket says "checks if all 100,000 objects are restored" - a restore bug that drops objects past a certain threshold (pagination limit, memory pressure) would pass this test undetected.

**Suggested fix:** After this Eventually block, spot-check a few non-deleted namespaces to confirm the full dataset survived:

```go
By("Spot-checking non-deleted namespaces survived restore", func() {
    for _, i := range []int{nsCount / 4, nsCount / 2, nsCount - 1} {
        ns := fmt.Sprintf("%s-%d", nsPrefix, i)
        count, err := countConfigMapsByLabel(ctx, s.vClusterClient, ns, labelKey+"="+labelVal)
        Expect(err).To(Succeed())
        Expect(count).To(Equal(objectsPerNS),
            "non-deleted namespace %s should have %d configmaps after restore, got %d", ns, objectsPerNS, count)
    }
})
```

This adds lightweight verification (3 namespaces from different positions) without iterating all 100.

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -800,6 +803,142 @@ func deletePVC(ctx context.Context, vClusterClient, _ kubernetes.Interface, _, _
...
+						return nil
+					})
+				}
+				Expect(eg.Wait()).To(Succeed())
+			})
+		})
+
+		It("Takes a snapshot of the tenant cluster", func(ctx context.Context) {
+			createSnapshot(s.vClusterName, s.vClusterNS, true, snapshotPath, false)
+			waitForSnapshotToBeCreated(ctx, s.hostClient, s.vClusterNS)
```
adriankabala: `waitForSnapshotToBeCreated` hardcodes `PollingTimeoutVeryLong` (5 min), but this test uses `PollingTimeoutExtraLong` (15 min) for its other operations - acknowledging that 100k objects need more time. Snapshot creation (serializing 100k objects into a tarball) could exceed 5 minutes in CI, causing a misleading timeout failure.

**Suggested fix:** Call `waitForRequestToFinish` directly with the longer timeout:

```go
waitForRequestToFinish(ctx, s.hostClient, s.vClusterNS,
    pkgconstants.SnapshotRequestLabel, snapshot.UnmarshalSnapshotRequest,
    constants.PollingTimeoutExtraLong)
```

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -800,6 +803,142 @@ func deletePVC(ctx context.Context, vClusterClient, _ kubernetes.Interface, _, _
...
+
+			By("Waiting for the 4 namespaces to be fully terminated", func() {
+				Eventually(func(g Gomega) {
+					for i := range deleteNSCount {
+						ns := fmt.Sprintf("%s-%d", nsPrefix, i)
+						_, err := s.vClusterClient.CoreV1().Namespaces().Get(ctx, ns, metav1.GetOptions{})
+						g.Expect(kerrors.IsNotFound(err)).To(BeTrue(),
+							"namespace %s should be deleted but still exists", ns)
+					}
+				}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeoutExtraLong).Should(Succeed())
```
adriankabala: nit: This Eventually (and the one at line 927) is missing `.WithContext(ctx)`. Without it, Gomega won't respect Ginkgo spec timeout cancellation between poll iterations - the loop continues until its own 15-minute timeout expires even if the spec is already cancelled. The existing test at line 459 (`describeSnapshotRestore`) uses `.WithContext(ctx)` as a reference.

```go
Eventually(func(g Gomega, ctx context.Context) {
    // ...
}).WithContext(ctx).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeoutExtraLong).Should(Succeed())
```

adriankabala (commented):
**LGTM with concerns** — 0 blocking, 3 non-blocking.

Well-structured test that covers the core ENGCP-503 scenario (100k-object snapshot + restore). The concurrent creation with errgroup + high-QPS client is solid. See inline comments for:

1. Post-restore verification only checks 4 of 100 namespaces — spot-checking a few non-deleted ones would close the coverage gap against the ticket's "all 100k" requirement
2. `waitForSnapshotToBeCreated` uses a 5-min timeout while the rest of this test uses 15 min — could flake if snapshot serialization is slow at this scale
3. Two `Eventually` blocks missing `.WithContext(ctx)` for spec timeout propagation

e2e-next/test_storage/snapshot/test_snapshot.go
jjaferson: Good catch 👍 When I rewrote the test to only delete a few namespaces it didn't come to my mind that I should check random namespaces and objects are actually restored

e2e-next/test_storage/snapshot/test_snapshot.go
jjaferson: it's heavy and long, I'll move it to it's own suits and use a `snapshot-large-restore` label so it can run on it's own vcluster too

e2e-next/test_storage/snapshot/test_snapshot.go
jjaferson: Good question, I am using `AfterAll` as it's used in other parts of the e2e snapshot test suite, @adriankabala do you know which one should we use here?

AfterAll: splits creation and cleanup across two functions. 
DeferCleanup: runs inside BeforeAll

e2e-next/test_storage/snapshot/test_snapshot.go
jjaferson: Are we suppose to use `By` during setup phase? I've asked Claude and it says 

>By() is designed for It blocks — it creates timeline entries in the Ginkgo failure report that help distinguish which step within a spec failed and how long it took.
In a BeforeAll, Ginkgo already identifies failures by type ([BeforeAll]) and exact line number, which is sufficient. Adding By() in setup nodes doesn't improve the report meaningfully and creates visual noise that doesn't pay off.

e2e-next/test_storage/snapshot/test_snapshot.go
jjaferson: I've tested it locally and to run the requests in parallel without being blocked by the rate limiter in client-go, so using the default, the first 10 requests go through immediately then after that the next one goes every ~200 ms

e2e-next/test_storage/snapshot/test_snapshot.go
nprokopic: In this case `DeferCleanup` and `AfterAll` should work in the same way, both should get executed when the Ordered container completes (after all `It` specs).

> using AfterAll as it's used in other parts of the e2e snapshot test suite

➕ here, for the sake of consistency, it's probably better to use `AfterAll` in this PR.

The benefit of having `DeferCleanup` in `BeforeAll` is that it would place the cleanup code next to the setup code, which would be easier and more reliable to maintain in tests, e.g. when you setup something new in the test, the cleanup code is right next to the setup to remind you (or the agent) that your new resource should be also cleaned up.

All this being said, we could refactor cleanup code in tests and use `DeferCleanup` instead of `AfterAll` where it brings some value, but that's probably and idea for the @loft-sh/eng-qa team :)

mfranczy (changes requested):
This cannot be part of the standard e2e pipeline.

It extends the test runtime to over an hour without introducing any specific constraint or signal that would help us identify functionality regressions or performance degradation.

Please put this on hold and let's discuss it during the tech sync meeting.

.github/workflows/e2e-ginkgo.yaml
```diff
@@ -257,6 +257,7 @@ jobs:
           ginkgo-label: "${{ needs.parse_label-filter.outputs.label-filter }} || pr"
           additional-args: "--vcluster-image=${{ env.REPOSITORY_NAME }}:${{ env.TAG_NAME }} --teardown=false"
           additional-ginkgo-flags: "-v --show-node-events"
+          timeout: 90m
```
roehrijn, line 260: The matching `timeout: 90m` was not mirrored in `.github/workflows/e2e-ginkgo-nightly.yaml:117`, which still passes `timeout: "60m"` to `run-ginkgo`. If the `snapshot-large-restore` label is ever included in a nightly run (e.g. via `workflow_dispatch` or once added to a default label set), Ginkgo will abort after 60 minutes — before the three 90-minute `Eventually`/`waitForRequestToFinish` calls in `test_snapshot.go` (lines 894, 916, 938) can complete. Consider bumping the nightly workflow's timeout too, or documenting why it intentionally stays at 60m.

e2e-next/constants/timeouts.go
```diff
@@ -9,4 +9,5 @@ const (
 	PollingTimeout          = time.Second * 60
 	PollingTimeoutLong      = time.Second * 120
 	PollingTimeoutVeryLong  = time.Second * 300
+	PollingTimeoutExtraLong = time.Minute * 90
```
roehrijn, line 12: `PollingTimeoutExtraLong` (90m) is only used by the new large-restore spec, yet it lives in the shared `constants` package and jumps 18x from `PollingTimeoutVeryLong` (5m). Once a generic 90m option exists, future flaky tests are tempted to reach for it rather than diagnose. Consider either keeping it scoped (e.g. a package-local `const largeRestoreTimeout = 90 * time.Minute` inside `e2e-next/test_storage/snapshot/`), or renaming to convey scope so reuse outside this scenario is discouraged:

```suggestion
	SnapshotLargeRestoreTimeout = time.Minute * 90
```

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -800,6 +802,164 @@ func deletePVC(ctx context.Context, vClusterClient, _ kubernetes.Interface, _, _
...
+		s = *newSnapshotCtx(ctx)
+	})
+	describeSnapshotLargeObjectRestore(&s)
+}
+
+func describeSnapshotLargeObjectRestore(s *snapshotCtx) {
+	const (
+		objectCount   = 100_000
+		nsCount       = 100
+		objectsPerNS  = objectCount / nsCount // 1_000 per namespace
```
roehrijn, line 819: `objectsPerNS = objectCount / nsCount` carries an implicit invariant that `objectCount` is divisible by `nsCount`. If a future maintainer tweaks `objectCount` to e.g. 100_001, the channel-fed loop below still publishes `objectCount` indices and `i/objectsPerNS` will exceed `nsCount-1`, producing creates into a non-existent namespace. Either assert the invariant (`Expect(objectCount % nsCount).To(Equal(0))`) or iterate `for ns := range nsCount { for j := range objectsPerNS { ... } }` so the loop bound matches the namespace topology.

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -800,6 +802,164 @@ func deletePVC(ctx context.Context, vClusterClient, _ kubernetes.Interface, _, _
...
+						return nil
+					})
+				}
+				Expect(eg.Wait()).To(Succeed())
+			})
+		})
+
+		It("Takes a snapshot of the tenant cluster", func(ctx context.Context) {
+			createSnapshot(s.vClusterName, s.vClusterNS, true, snapshotPath, false)
+			waitForRequestToFinish(ctx, s.hostClient, s.vClusterNS,
```
roehrijn, line 894: This open-codes a `waitForRequestToFinish(...)` call with a custom timeout instead of using `waitForSnapshotToBeCreated`, leaking helper-internal knowledge (label key + unmarshal fn) into the spec. Consider extending the helper to accept a timeout — e.g. `waitForSnapshotToBeCreated(ctx, host, ns, timeout)` with existing callers passing `PollingTimeoutVeryLong` and this one passing the extra-long value — so all snapshot specs share the same abstraction.

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -800,6 +802,164 @@ func deletePVC(ctx context.Context, vClusterClient, _ kubernetes.Interface, _, _
...
+
+			By("Waiting for the 4 namespaces to be fully terminated", func() {
+				Eventually(func(g Gomega, ctx context.Context) {
+					for i := range deleteNSCount {
+						ns := fmt.Sprintf("%s-%d", nsPrefix, i)
+						_, err := s.vClusterClient.CoreV1().Namespaces().Get(ctx, ns, metav1.GetOptions{})
+						g.Expect(kerrors.IsNotFound(err)).To(BeTrue(),
+							"namespace %s should be deleted but still exists", ns)
+					}
+				}).WithContext(ctx).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeoutExtraLong).Should(Succeed())
```
roehrijn, line 916: Using a 90-minute polling timeout for namespace deletion conflates two concerns: the overall test budget and the expected operation latency. A real 90m wait on namespace termination usually masks a stuck finalizer rather than legitimately taking that long. Consider scoping each `Eventually` to its expected upper bound (a few minutes for ns deletion of 1000 cms) so failures fail fast with the right signal, and reserving the extra-long timeout only for the snapshot/restore RPC waits where 90m is plausible.

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -800,6 +802,164 @@ func deletePVC(ctx context.Context, vClusterClient, _ kubernetes.Interface, _, _
...
+			})
+		})
+
+		It("Restores snapshot and verifies the deleted namespaces are back", func(ctx context.Context) {
+			By("Restoring the tenant cluster from snapshot", func() {
+				restoreVCluster(ctx, s.hostClient, s.vClusterName, s.vClusterNS, snapshotPath, true, false)
```
roehrijn, line 922: The large-restore suite uses `PollingTimeoutExtraLong` (90m) elsewhere, but `restoreVCluster(...)` still goes through a helper that hardcodes `PollingTimeoutVeryLong` (5m) for the post-restore pod-ready wait (`helpers.go:74`). With a 100k-object restore the pods may take longer to become ready, and the suite will fail here long before the later 90m assertions run. Consider threading the timeout into `restoreVCluster`.

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -800,6 +802,164 @@ func deletePVC(ctx context.Context, vClusterClient, _ kubernetes.Interface, _, _
...
+				s.refreshClient(ctx)
```
roehrijn, line 923: `s.refreshClient(ctx)` hardcodes a `PollingTimeoutLong` (2m) wait for the restored API server to become usable (`test_snapshot.go:112`). After restoring 100k objects the pod may be Running while the tenant API is still replaying state, so this reconnect can time out even though the restore would eventually succeed. Worth letting the large-restore suite use a longer reconnect timeout.

e2e-next/vcluster-snapshot.yaml
```diff
@@ -1,4 +1,9 @@
 controlPlane:
+  distro:
+    k8s:
+      apiServer:
+        extraArgs:
+          - "--etcd-compaction-interval=1h"
```
roehrijn, line 6: Adding `--etcd-compaction-interval=1h` here changes the API server config for **all** snapshot suites that embed this YAML (both `suite_snapshot_test.go` and the new `suite_snapshot_large_restore_test.go`). The existing snapshot suite did not need this tuning. To avoid coupling unrelated suites to a perf knob added for one of them, consider either (a) moving this to a dedicated `vcluster-snapshot-large-restore.yaml` embedded only from the new suite, or (b) parameterizing via `lazyvcluster.WithExtraTemplateVars` so the existing suite keeps the default.

e2e-next/test_storage/snapshot/test_snapshot.go
roehrijn: I think the whole `AfterAll` is not needed at all. Why would we want to delete all these namespaces right before the tenant cluster is dropped anyway?

roehrijn (changes requested):
Only rather small explicit findings, but in general I fully agree with Marcin. We should discuss this as a team (likely also with QA team involved). This is nothing to just add to a regular e2e test suite. Maybe to the suite as such, but should not be executed on a daily base. For quite a while I have the idea of weekly executions in my mind. Maybe this is a viable option to label-filter it accordingly and run it once per week? Definitely we have to come to a common agreement what the goal is. This test is highly synthetic. 100k rather small config maps. But what about the object size limit in etcd? What if the 100k Configmaps are close to 4MB each? And is it enough if the test succeeds although it takes ages? We need to define a goal first and then adapt metric thresholds around that at least.

jjaferson:
I'll convert this to a draft PR as we might change it to the performance test framework

PR #3921, opened 2026-05-11
feat(standalone): add standalone-hosts annotation constant

pascalbreuninger:
## Summary

Adds the `VClusterStandaloneHostsAnnotation` constant (`vcluster.loft.sh/standalone-hosts`) used by vcluster-pro to track per-master `hostname -> last-known-IP` on the `default/kubernetes` service.

Paired with [vcluster-pro#1777](https://github.com/loft-sh/vcluster-pro/pull/1777), which consumes this constant to clean up stale IPs from the standalone-endpoints annotation when a master's IP changes across restarts — without depending on the on-disk runtime metadata file (which can be lost on data-dir wipe / VM replacement).

Linear: https://linear.app/loft/issue/ENGCP-733

### DIFF
--- a/pkg/constants/standalone.go
+++ b/pkg/constants/standalone.go
@@ -2,6 +2,7 @@ package constants
 
 const (
 	VClusterStandaloneEndpointsAnnotation = "vcluster.loft.sh/standalone-endpoints"
+	VClusterStandaloneHostsAnnotation     = "vcluster.loft.sh/standalone-hosts"
 	VClusterStandalonePortAnnotation      = "vcluster.loft.sh/standalone-port"
 	VClusterStandaloneEnvVar              = "VCLUSTER_STANDALONE"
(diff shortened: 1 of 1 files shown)


mfranczy (approved):

pkg/constants/standalone.go
```diff
@@ -2,6 +2,7 @@ package constants
 
 const (
 	VClusterStandaloneEndpointsAnnotation = "vcluster.loft.sh/standalone-endpoints"
+	VClusterStandaloneHostsAnnotation     = "vcluster.loft.sh/standalone-hosts"
```
mfranczy, line 5: Let's do it a bit differently, let's keep annotation per host. It will simplify the logic.

`VClusterStandaloneIdentityAnnotationPrefix = "vcluster.loft.sh/standalone-host"`, finally it should look like `vcluster.loft.sh/standalone-host.<host>=<ip>`

mfranczy (changes requested):

pascalbreuninger:
We'll need to convert this to a project, closing the PR(s)

PR #3928, opened 2026-05-11
adds: file protocol for snapshot and restore

jjaferson:
kind enhancement

resolves ENGCP-427

Adds a file protocol for snapshot and restore feature to enable users to use the filesystem as a backing storage

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
snapshots
```

### DIFF
--- a/pkg/cli/snapshot_local.go
+++ b/pkg/cli/snapshot_local.go
@@ -8,9 +8,7 @@ import (
 
 	"github.com/loft-sh/log"
 	vclusterconfig "github.com/loft-sh/vcluster/pkg/config"
-	"github.com/loft-sh/vcluster/pkg/lifecycle"
 	"github.com/loft-sh/vcluster/pkg/snapshot"
-	"github.com/loft-sh/vcluster/pkg/snapshot/container"
 	"github.com/loft-sh/vcluster/pkg/snapshot/pod"
 	"github.com/loft-sh/vcluster/pkg/util/podhelper"
 	corev1 "k8s.io/api/core/v1"
@@ -28,18 +26,16 @@ const dataMountPath = "/data"
 func snapshotToLocalFile(ctx context.Context, vCluster *find.VCluster,
 	kubeClient *kubernetes.Clientset, restConfig *rest.Config,
 	snapshotOpts *snapshot.Options, log log.Logger, vClusterConfig *vclusterconfig.VirtualClusterConfig) error {
-
-	localPath := snapshotOpts.LocalPath
 	tempPath := fmt.Sprintf("%s/vcluster-snapshot-%d.tar.gz", dataMountPath, time.Now().Unix())
-
-	// Translate to container:// for the in-cluster controller.
-	containerOpts := *snapshotOpts
-	containerOpts.Type = "container"
-	containerOpts.Container = container.Options{Path: tempPath}
+	localPath := snapshotOpts.File.Path
+	if !vCluster.IsStandalone {
+		// For non-standalone, we need to write the snapshot to the syncer PVC first, then download it via exec.
+		snapshotOpts.File.Path = tempPath
+	}
 
 	log.Infof("Creating snapshot request...")
 	snapshotRequest, err := snapshot.CreateSnapshotRequestResources(
-		ctx, vCluster.Namespace, vClusterConfig.Name, vClusterConfig, &containerOpts, kubeClient)
+		ctx, vCluster.Namespace, vClusterConfig.Name, vClusterConfig, snapshotOpts, kubeClient)
 	if err != nil {
 		return fmt.Errorf("create snapshot request: %w", err)
 	}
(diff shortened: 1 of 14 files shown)


jjaferson:
@codex review

jjaferson:
@codex review

jjaferson:
@codex review

pkg/cli/snapshot_local.go
seanschneeweiss, line 1: Is there any potential attack vector that you could think of for snapshot restores with vcluster CLI and path traversal?
Just thinking if we could improve file handling a bit with https://go.dev/blog/osroot - just in case it would be of interest in this code path.

pkg/cli/snapshot_local.go
```diff
@@ -0,0 +1,165 @@
...
+				}
+				return false, fmt.Errorf("snapshot %s: %s", req.Status.Phase, req.Status.Error.Message)
+			}
+			return false, nil
+		})
+}
+
+func findVClusterPod(vCluster *find.VCluster) (*corev1.Pod, error) {
+	for i := range vCluster.Pods {
+		p := &vCluster.Pods[i]
```
seanschneeweiss: For readability and index-less loops.
```suggestion
	for _, pod := range vCluster.Pods {
		p := &pod
```

pkg/cli/snapshot_local.go
```diff
@@ -0,0 +1,165 @@
...
+		return fmt.Errorf("open local snapshot %s: %w", localPath, err)
+	}
+	defer f.Close()
+
+	log.Infof("Uploading %s to pod %s at %s...", localPath, targetPod.Name, tempPath)
+	if err := podhelper.ExecStream(ctx, restConfig, &podhelper.ExecStreamOptions{
+		Pod:       targetPod.Name,
+		Namespace: vCluster.Namespace,
+		Container: "syncer",
+		Command:   []string{"/bin/sh", "-c", "cat > " + tempPath},
```
seanschneeweiss: potential shell injection, could do

```suggestion
		Command:   []string{"dd", fmt.Sprintf("of=%s", tempPath)},
```
or `[]string{"sh", "-c", "cat > \"$1\"", "--", tempPath}`

pkg/snapshot/options.go
```diff
@@ -151,6 +155,12 @@ func Parse(snapshotURL string, snapshotOptions *Options) error {
 		// Azure blob storage support
 		snapshotOptions.Type = "azure"
 		snapshotOptions.Azure.BlobURL = snapshotURL
+	case "file":
+		// file:///absolute/path — host is empty for absolute paths
+		if parsedURL.Path == "" {
+			return fmt.Errorf("path must be specified via file:///PATH")
+		}
+		snapshotOptions.File.Path = parsedURL.Path
```
seanschneeweiss, line 107: The path is user-supplied from the CLI and the user already controls the local machine. Nevertheless might be a good idea to clean/normalize the path? Or check with `filepath.Localize'.

```suggestion
		snapshotOptions.File.Path = filepath.Clean(parsedURL.Path)
```

pkg/snapshot/file/store.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (s *Store) GetObject(_ context.Context) (io.ReadCloser, error) {
+	return os.Open(s.path)
+}
+
+func (s *Store) PutObject(_ context.Context, body io.Reader) error {
+	if err := os.MkdirAll(filepath.Dir(s.path), 0755); err != nil {
+		return err
+	}
+	f, err := os.OpenFile(s.path, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0600)
```
seanschneeweiss, line 39: Use `0700` for the directory to match the restrictive intent of `0600` on the file.

pkg/cli/snapshot_local.go
jjaferson: I don't think there's a risk of shell injection here as the `tempPath` is generated in the server tho `"sh", "-c", "cat > \"$1\"", "--", tempPath` is more POSIX pattern 

`dd` is also cleaner but will require adding a dependency

pkg/cli/snapshot_local.go
jjaferson: Interesting topic but it doesn't have a potential risk AFAIK due to the way we read the snapshot tarball and the file path is controlled by the operator so I don't see any risk here

pkg/snapshot/options.go
jjaferson: Wound't `filepath.Lozalize` reject path that are not absolute? eg: `../mysnapshpt/`? same for `filepath.Clean`?

seanschneeweiss: Oh yes you are right for localize. Never mind, resolving.

pkg/cli/snapshot_local.go
```diff
@@ -0,0 +1,165 @@
...
+	if vCluster.IsStandalone {
+		// For standalone, we can read directly from the local file instead of going through the syncer PVC.
+		return restoreStandaloneVCluster(snapshotOpts, cmdArgs, log)
+	}
+
+	// For non-standalone, we need to upload the local file into the syncer container first, then point the restore command at it.
+	snapshotOpts.File.Path = tempPath
+
+	// Stream the local file into the syncer PVC via exec before pausing.
+	// The PVC (and the staged file) persist through scale-to-zero.
```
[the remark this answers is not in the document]
jjaferson: This would be an issue with any restore protocol as we restart the vcluster pod to run the restore.

mfranczy:
@codex review

pkg/cli/snapshot_local.go
```diff
@@ -0,0 +1,166 @@
...
+			}
+			return false, nil
+		})
+}
+
+func findVClusterPod(vCluster *find.VCluster) (*corev1.Pod, error) {
+	for _, pod := range vCluster.Pods {
+		p := &pod
+		if (vCluster.StatefulSet != nil || vCluster.Deployment != nil) && len(p.Name) > 0 {
+			return p, nil
```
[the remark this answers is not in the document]
jjaferson: Exposed the `findVClusterPod` to be reused by the RunSnapshotPod and local file snapshot and restore

jjaferson:
@codex review

PR #3942, opened 2026-05-18
test: e2e test for testing pagination in etcd snapshot and restore

jjaferson:
kind test

resolves ENGCP-828

Adds e2e test for pagination of object in etcd during snapshot and restore

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
snapshots
```

### DIFF
--- a/e2e-next/test_storage/snapshot/test_snapshot.go
+++ b/e2e-next/test_storage/snapshot/test_snapshot.go
@@ -483,6 +483,85 @@ func describeSnapshotRestore(s *snapshotCtx) {
 			deleteSnapshotRequestConfigMaps(ctx, s.hostClient, s.vClusterNS)
 		})
 	})
+
+	Describe("snapshot restore preserves objects across etcd page boundaries", Ordered, func() {
+		const (
+			labelKey    = "etcd-pagination"
+			objectCount = 1500 // > default etcd page limit of 1000 to ensure pagination occurs
+		)
+		var (
+			testNS       string
+			snapshotPath string
+			labelVal     string
+		)
+
+		BeforeAll(func(ctx context.Context) {
+			testNS = "etcd-pagination-" + random.String(6)
+			snapshotPath = "container:///snapshot-data/" + testNS + ".tar.gz"
+			labelVal = testNS
+			cleanupAllSnapshotArtifacts(ctx, s.hostClient, s.vClusterNS)
+		})
+
+		It("Creates a snapshot with enough objects to trigger etcd pagination and verifies they are restored", func(ctx context.Context) {
+			_, err := s.vClusterClient.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{Name: testNS},
+			}, metav1.CreateOptions{})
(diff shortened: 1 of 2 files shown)


jjaferson:
@codex review

cbalan (dismissed):

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -483,6 +483,85 @@ func describeSnapshotRestore(s *snapshotCtx) {
 			deleteSnapshotRequestConfigMaps(ctx, s.hostClient, s.vClusterNS)
 		})
 	})
+
+	Describe("snapshot restore preserves objects across etcd page boundaries", Ordered, func() {
```
roehrijn: [C12] `Ordered` container is missing a dependency comment explaining which spec depends on a prior spec's side effect.

**Fix:**
Add a comment immediately before the `Describe` that names the ordering dependency chain, for example:
```go
// Ordered: spec 1 creates namespace + configmaps -> spec 2 takes snapshot ->
// spec 3 deletes namespace -> spec 4 restores and verifies count.
Describe("snapshot restore preserves objects across etcd page boundaries", Ordered, func() {
```

Anyways, IMHO it would fit better to have a single `It()` block and no `Ordered` label.

e2e-next/test_storage/snapshot/test_snapshot.go
```diff
@@ -483,6 +483,85 @@ func describeSnapshotRestore(s *snapshotCtx) {
...
+		const (
+			labelKey    = "etcd-pagination"
+			objectCount = 1500 // > default etcd page limit of 1000 to ensure pagination occurs
```
roehrijn: The constant `1500` is structurally coupled to `clientv3.WithLimit(1000)` in `pkg/etcd/client.go` (etcd snapshot page size), but nothing links them. If that limit is ever tuned, this test silently stops exercising pagination while continuing to pass. Export the page size as a constant (e.g. `SnapshotPageLimit`) and use `objectCount = pkgetcd.SnapshotPageLimit + N`. 

Also, `SnapshotPageLimit + 1` would be sufficient to test. the other 499 additional ConfigMaps are adding no value to this particular test case. 

And FYI: It would be even better to make `clientv3.WithLimit(1000)` configurable through an env variable for example. With this, we could set it to 10 and test with 11 maps. But I see this out of scope of this PR. Just for next time.

e2e-next/test_storage/snapshot/test_snapshot.go
jjaferson: I'll move it to a single `It()` block, the only thing is that the single `It()` block will live inside the top level Describe("Snapshot and restore") block

e2e-next/test_storage/snapshot/test_snapshot.go
jjaferson: I can expose the etcd limit to a constant but the scope of this work is only to verify that etcd pagination functions correctly. I’m not planning to make any changes or improvements to the pagination implementation itself as part of this PR.

roehrijn (approved):

PR #3947, opened 2026-05-21
fix(snapshot): only send x-amz-tagging header when tagging is configured

### LINKED ISSUE #3882: Snapshot writer unconditionally sends x-amz-tagging header, breaks Cloudflare R2 (and other non-AWS backends)
### What happened?

`vcluster snapshot create` fails when writing to Cloudflare R2 with:

```
StatusCode: 501, api error NotImplemented: Header 'x-amz-tagging' with value '' not implemented
```

The S3 writer at [`pkg/snapshot/s3/store.go:249`](https://github.com/loft-sh/vcluster/blob/main/pkg/snapshot/s3/store.go#L249) unconditionally sets `Tagging: aws.String(o.tagging)` on every `PutObject` input. When `o.tagging` is empty, the AWS SDK v2 serializer still emits the `x-amz-tagging` header. R2 rejects the request because it does not implement object tagging at all — regardless of tag value.

```go
input := &s3.PutObjectInput{
    Bucket:  aws.String(o.bucket),
    Key:     aws.String(o.key),
    Body:    &wrapper{body},
    Tagging: aws.String(o.tagging), // always set, breaks R2
}
```

Snapshot writes successfully to R2 (or any S3-compatible backend that doesn't support object tagging), as it does against AWS S3. `Tagging` should only be sent when the user explicitly provides a value.

1. Configure an R2 bucket with S3-compat credentials.
2. Run:

   ```sh
   vcluster snapshot create -n my-vcluster my-vcluster \
     's3://my-r2-bucket/path/snap.tar.gz?access-key-id=<base64>&secret-access-key=<base64>&url=<base64 of https://<account>.r2.cloudflarestorage.com>&region=auto&force-path-style=true'
   ```

3. Inspect the snapshot-request ConfigMap:

   ```sh
   kubectl -n <host-ns> get cm -l vcluster.loft.sh/snapshot-request -o yaml
   ```

   Status reports `phase: Failed` with the 501 error above.

Setting a non-empty `tagging` URL query param doesn't help — R2 rejects the header regardless of value.

This file is forked from Velero's object store (the license header attributes `Copyright the Velero contributors`). Velero hit the same issue in [vmware-tanzu/velero#7940](https://github.com/vmware-tanzu/velero/issues/7940) and resolved it in v1.17; the vcluster fork never caught up.

Suggested one-line fix:

```go
input := &s3.PutObjectInput{
    Bucket: aws.String(o.bucket),
    Key:    aws.String(o.key),
    Body:   &wrapper{body},
}
if o.tagging != "" {
    input.Tagging = aws.String(o.tagging)
}
```

<details>

```console
$ kubectl version
Client Version: v1.35.4
Kustomize Version: v5.7.1
Server Version: v1.34.4+rke2r1
```

</details>

### vcluster version

<details>

```console
$ vcluster --version
0.33.1
```

</details>

### VCluster Config

<details>

Relevant excerpt (bug is config-independent; any vcluster snapshot write to an R2 `s3://` target fails identically):

```yaml
controlPlane:
  distro:
    k8s:
      enabled: true
  backingStore:
    etcd:
      deploy:
        enabled: true
        statefulSet:
          highAvailability:
            replicas: 3
  statefulSet:
    highAvailability:
      replicas: 3
  service:
    spec:
      type: LoadBalancer
```

</details>

vcluster-dv:
/kind bugfix

Guards the `Tagging` field on `s3.PutObjectInput` so it is only set when the user has actually configured a tagging value. Without this guard, the AWS SDK v2 serializer emits an `x-amz-tagging:` header with an empty value, which Cloudflare R2 and some other S3-compatible backends reject with `HTTP 501 NotImplemented` — regardless of value — because they do not implement object tagging at all.

resolves #3882 
resolves ENGCP-641

Fixed `vcluster snapshot create` failing against Cloudflare R2 and other S3-compatible backends with `StatusCode: 501, api error NotImplemented: Header 'x-amz-tagging' with value '' not implemented`. The empty `x-amz-tagging` header is no longer sent when no tagging value is configured.

- `pkg/snapshot/s3/store.go` is forked from Velero's `velero-plugin-for-aws` object store. The upstream still has the same unconditional `Tagging` assignment on `main`, `v1.13.x`, and `v1.14.x` ([object_store.go#L341](https://github.com/velero-io/velero-plugin-for-aws/blob/main/velero-plugin-for-aws/object_store.go#L341)), so this is an original fix, not an upstream sync.
- Reproduction: configure a vcluster snapshot to an R2 bucket via `s3://my-bucket/path/snap.tar.gz?force-path-style=true&region=auto&url=<base64 https://<acct>.r2.cloudflarestorage.com>&access-key-id=<base64>&secret-access-key=<base64>` and run `vcluster snapshot create`. The snapshot-request ConfigMap reports `phase: Failed` with the 501 error. Setting a non-empty `tagging` query param does not help — R2 rejects the header regardless of value.
- Follow-up PR (stacked on this one): tagging validation in `Init()` and unit tests covering this behavior.

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
snapshots
```

### DIFF
--- a/pkg/snapshot/s3/store.go
+++ b/pkg/snapshot/s3/store.go
@@ -216,10 +216,12 @@ func (o *ObjectStore) Target() string {
 
 func (o *ObjectStore) PutObject(ctx context.Context, body io.Reader) error {
 	input := &s3.PutObjectInput{
-		Bucket:  aws.String(o.bucket),
-		Key:     aws.String(o.key),
-		Body:    &wrapper{body},
-		Tagging: aws.String(o.tagging),
+		Bucket: aws.String(o.bucket),
+		Key:    aws.String(o.key),
+		Body:   &wrapper{body},
+	}
+	if o.tagging != "" {
+		input.Tagging = aws.String(o.tagging)
 	}
 
 	switch {


jjaferson (approved):

mfranczy:
@codex review

PR #3949, opened 2026-05-22
feat(pods): add sync.toHost.pods.dns.nameservers

roehrijn:
/kind feature

Closes ENGCP-861

Adds a new `sync.toHost.pods.dns.nameservers` list field. Each entry resolves a Service ClusterIP by namespace + label selector (scoped to either the host cluster or the tenant cluster). The resolved IPs are written into `dnsConfig.nameservers` on every synced pod and `dnsPolicy` is forced to `None`. Failing entries emit a Warning event per pod; an all-fail resolution blocks pod creation. Pods with user-supplied `dnsConfig.nameservers` or `hostNetwork: true` are left untouched.

This is the vCluster syncer portion of the Calico FQDN design: Calico's per-node FQDN cache only works when the DNS request originates from the same node as the workload, so synced pods need to be pointed at a node-local DNS endpoint via Service ClusterIPs configured by the operator.

Added `sync.toHost.pods.dns.nameservers`: a list of Service references (host or tenant scope) whose ClusterIPs are written into the `dnsConfig.nameservers` of every synced pod.

- New host-scoped resolver entries require the syncer ServiceAccount to have `get/list/watch` on Services in the target namespace; without it the controller blocks waiting for the informer cache to sync.
- Unit tests cover the validator, the translator, and the resolver fallback paths.
- E2E suite `pods-dns-nameservers-vcluster` covers happy path, hostNetwork opt-out, user-defined nameservers override, partial-failure with Warning events, and all-fail blocking pod creation.

### Default Test Execution

### Adding New Test Suites
- [x] **Add labels** to the test suite (`labels.PR`, `labels.Sync`, `labels.Core`, `labels.Pods`)
- [x] **Update label-filter section** below to execute the new test suite (carries `pr` so default mandatory suite picks it up)

### Additional test suites

```label-filter
none
```

### DIFF
--- a/chart/values.schema.json
+++ b/chart/values.schema.json
@@ -1121,7 +1121,7 @@
       "properties": {
         "scope": {
           "type": "string",
-          "description": "Scope chooses which cluster to look up the Service in.\n\n+kubebuilder:validation:Enum=host;tenant"
+          "description": "Scope chooses which cluster to look up the Service in.\n\n+kubebuilder:validation:Enum=control-plane-host;tenant-cluster"
         },
         "namespace": {
           "type": "string",
--- a/chart/values.schema.json
+++ b/chart/values.schema.json
@@ -1106,6 +1106,36 @@
       "additionalProperties": false,
       "type": "object"
     },
+    "DNSNameserverEntry": {
+      "properties": {
+        "service": {
+          "$ref": "#/$defs/DNSNameserverService",
+          "description": "Service selects the Service whose ClusterIP becomes one nameserver."
+        }
+      },
+      "additionalProperties": false,
+      "type": "object",
+      "description": "DNSNameserverEntry is one element of the nameservers list."
+    },
+    "DNSNameserverService": {
+      "properties": {
+        "scope": {
+          "type": "string",
+          "description": "Scope chooses which cluster to look up the Service in.\n\n+kubebuilder:validation:Enum=host;tenant"
+        },
+        "namespace": {
+          "type": "string",
+          "description": "Namespace containing the Service."
+        },
+        "labelSelector": {
+          "$ref": "#/$defs/StandardLabelSelector",
+          "description": "LabelSelector must match exactly one Service in Namespace. If zero or\nmore than one Service matches, a Warning event is emitted for this entry\nand the entry is skipped (other entries are still resolved)."
+        }
+      },
+      "additionalProperties": false,
+      "type": "object",
+      "description": "DNSNameserverService describes the Service lookup for a single nameserver."
+    },
     "Database": {
       "properties": {
         "embedded": {
@@ -5011,10 +5041,28 @@
         "hybridScheduling": {
           "$ref": "#/$defs/HybridScheduling",
           "description": "HybridScheduling is used to enable and configure hybrid scheduling for pods in the virtual cluster."
+        },
+        "dns": {
+          "$ref": "#/$defs/SyncPodsDNS",
+          "description": "DNS controls DNS-related behavior for synced pods."
         }
       },
       "additionalProperties": false,
       "type": "object"
+    },
+    "SyncPodsDNS": {
+      "properties": {
+        "nameservers": {
+          "items": {
+            "$ref": "#/$defs/DNSNameserverEntry"
+          },
+          "type": "array",
+          "description": "Nameservers is a list of Service-based nameserver sources. Each entry\nresolves independently to a Service ClusterIP which is written to\ndnsConfig.nameservers on every synced pod, in configuration order.\n\nEntries that fail to resolve emit a Warning event but do not block\nthe remaining entries. If ALL entries fail, pod sync returns an error\nand requeues -- the pod is not created until at least one Service is found.\n\nPods with spec.hostNetwork=true and pods that already pin\nspec.dnsConfig.nameservers are skipped entirely."
+        }
+      },
+      "additionalProperties": false,
+      "type": "object",
+      "description": "SyncPodsDNS controls DNS-related behavior for synced pods."
     },
     "SyncRewriteHosts": {
       "properties": {
--- a/config/config.go
+++ b/config/config.go
@@ -1547,7 +1547,58 @@ type SyncPods struct {
 
 	// HybridScheduling is used to enable and configure hybrid scheduling for pods in the virtual cluster.
 	HybridScheduling HybridScheduling `json:"hybridScheduling,omitempty"`
+
+	// DNS controls DNS-related behavior for synced pods.
+	DNS SyncPodsDNS `json:"dns,omitempty"`
 }
+
+// SyncPodsDNS controls DNS-related behavior for synced pods.
+type SyncPodsDNS struct {
+	// Nameservers is a list of Service-based nameserver sources. Each entry
+	// resolves independently to a Service ClusterIP which is written to
+	// dnsConfig.nameservers on every synced pod, in configuration order.
+	//
+	// Entries that fail to resolve emit a Warning event but do not block
+	// the remaining entries. If ALL entries fail, pod sync returns an error
+	// and requeues -- the pod is not created until at least one Service is found.
+	//
+	// Pods with spec.hostNetwork=true and pods that already pin
+	// spec.dnsConfig.nameservers are skipped entirely.
+	Nameservers []DNSNameserverEntry `json:"nameservers,omitempty"`
+}
+
+// DNSNameserverEntry is one element of the nameservers list.
+type DNSNameserverEntry struct {
+	// Service selects the Service whose ClusterIP becomes one nameserver.
+	Service DNSNameserverService `json:"service,omitempty"`
+}
+
+// DNSNameserverService describes the Service lookup for a single nameserver.
+type DNSNameserverService struct {
+	// Scope chooses which cluster to look up the Service in.
+	//
+	// +kubebuilder:validation:Enum=host;tenant
+	Scope DNSNameserverScope `json:"scope,omitempty"`
+
+	// Namespace containing the Service.
+	Namespace string `json:"namespace,omitempty"`
+
+	// LabelSelector must match exactly one Service in Namespace. If zero or
+	// more than one Service matches, a Warning event is emitted for this entry
+	// and the entry is skipped (other entries are still resolved).
+	LabelSelector *StandardLabelSelector `json:"labelSelector,omitempty"`
+}
+
+// DNSNameserverScope chooses where the nameserver Service is looked up.
+type DNSNameserverScope string
+
+const (
+	// DNSNameserverScopeHost looks the Service up in the Control Plane Cluster (host).
+	DNSNameserverScopeHost DNSNameserverScope = "host"
+
+	// DNSNameserverScopeTenant looks the Service up inside the tenant cluster.
+	DNSNameserverScopeTenant DNSNameserverScope = "tenant"
+)
 
 type SyncRewriteHosts struct {
 	// Enabled specifies if rewriting stateful set pods should be enabled.
--- a/config/config.go
+++ b/config/config.go
@@ -1577,7 +1577,7 @@ type DNSNameserverEntry struct {
 type DNSNameserverService struct {
 	// Scope chooses which cluster to look up the Service in.
 	//
-	// +kubebuilder:validation:Enum=host;tenant
+	// +kubebuilder:validation:Enum=control-plane-host;tenant-cluster
 	Scope DNSNameserverScope `json:"scope,omitempty"`
 
 	// Namespace containing the Service.
@@ -1594,10 +1594,10 @@ type DNSNameserverScope string
 
 const (
 	// DNSNameserverScopeHost looks the Service up in the Control Plane Cluster (host).
-	DNSNameserverScopeHost DNSNameserverScope = "host"
+	DNSNameserverScopeHost DNSNameserverScope = "control-plane-host"
 
 	// DNSNameserverScopeTenant looks the Service up inside the tenant cluster.
-	DNSNameserverScopeTenant DNSNameserverScope = "tenant"
+	DNSNameserverScopeTenant DNSNameserverScope = "tenant-cluster"
 )
 
 type SyncRewriteHosts struct {
--- a/e2e-next/suite_pods_dns_nameservers_test.go
+++ b/e2e-next/suite_pods_dns_nameservers_test.go
@@ -53,8 +53,9 @@ const podsDNSNameserversVClusterName = "pods-dns-nameservers-vcluster"
 func init() { suitePodsDNSNameserversVCluster() }
 
 func suitePodsDNSNameserversVCluster() {
-	// Ordered: BeforeAll must complete (vCluster + host namespace created)
-	// before any spec runs; specs share the lazily-created vCluster.
+	// Ordered: BeforeAll must complete before any spec runs so the vCluster
+	// and host namespace exist. Specs themselves are independent — each
+	// recreates its own pods via BeforeEach and shares no sequential state.
 	Describe("pods-dns-nameservers-vcluster", labels.PR, labels.Sync, Ordered,
 		cluster.Use(clusters.HostCluster),
 		func() {
--- a/e2e-next/suite_pods_dns_nameservers_test.go
+++ b/e2e-next/suite_pods_dns_nameservers_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/loft-sh/e2e-framework/pkg/setup/cluster"
 	"github.com/loft-sh/vcluster/e2e-next/clusters"
 	"github.com/loft-sh/vcluster/e2e-next/constants"
+	"github.com/loft-sh/vcluster/e2e-next/labels"
 	"github.com/loft-sh/vcluster/e2e-next/setup/lazyvcluster"
 	test_core "github.com/loft-sh/vcluster/e2e-next/test_core/sync"
 	. "github.com/onsi/ginkgo/v2"
@@ -52,13 +53,12 @@ const podsDNSNameserversVClusterName = "pods-dns-nameservers-vcluster"
 func init() { suitePodsDNSNameserversVCluster() }
 
 func suitePodsDNSNameserversVCluster() {
-	// Serial: every spec re-creates the same-named vCluster and host
-	// namespace in BeforeEach, so parallel processes would collide on the
-	// Helm release and the shared dns-test namespace.
-	Describe("pods-dns-nameservers-vcluster", Serial,
+	// Ordered: BeforeAll must complete (vCluster + host namespace created)
+	// before any spec runs; specs share the lazily-created vCluster.
+	Describe("pods-dns-nameservers-vcluster", labels.PR, labels.Sync, Ordered,
 		cluster.Use(clusters.HostCluster),
 		func() {
-			BeforeEach(func(ctx context.Context) context.Context {
+			BeforeAll(func(ctx context.Context) context.Context {
 				DeferCleanup(podsDNSNameserversCleanup)
 				return lazyvcluster.LazyVCluster(ctx,
 					podsDNSNameserversVClusterName,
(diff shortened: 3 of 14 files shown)


e2e-next/suite_pods_dns_nameservers_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+//go:embed vcluster-pods-dns-nameservers.yaml
+var podsDNSNameserversVClusterYAML string
+
+const podsDNSNameserversVClusterName = "pods-dns-nameservers-vcluster"
+
+func init() { suitePodsDNSNameserversVCluster() }
+
+func suitePodsDNSNameserversVCluster() {
+	Describe("pods-dns-nameservers-vcluster", labels.PR, labels.Sync, Ordered,
```
roehrijn, line 59: I don't think we need Ordered here.

pkg/controllers/resources/pods/dns_override.go
```diff
@@ -0,0 +1,130 @@
...
+	res := &dnsNameserversResolver{}
+	if len(cfgEntries) == 0 {
+		return res, nil
+	}
+	hostCaches := map[string]client.Reader{}
+
+	for i, ce := range cfgEntries {
+		svc := ce.Service
+		// selector validity is guaranteed by validateDNSNameservers at startup
+		selector, _ := svc.LabelSelector.ToSelector()
```
roehrijn: Silently discarding the selector error couples this code to `validateDNSNameservers` running first. If a future code path builds a resolver without prior validation, the nil selector will panic in `resolveEntryIP`. Consider returning the error explicitly rather than relying on a comment-documented invariant.

pkg/controllers/resources/pods/translate/translator.go
```diff
@@ -752,6 +772,50 @@ func (t *translator) translateDNSConfig(pPod *corev1.Pod, vPod *corev1.Pod, name
...
+	existing := pPod.Spec.DNSConfig
+	newCfg := &corev1.PodDNSConfig{
+		Nameservers: ips,
+	}
+	if existing != nil {
+		newCfg.Searches = existing.Searches
+		newCfg.Options = existing.Options
+	}
+	pPod.Spec.DNSPolicy = corev1.DNSNone
+	pPod.Spec.DNSConfig = newCfg
```
roehrijn, line 825: Pods with `spec.dnsPolicy=Default` are silently converted to `DNSNone` here, losing the node's `/etc/resolv.conf` defaults (search domains, options). The `Nameservers` doc comment only lists `hostNetwork` and pre-pinned `dnsConfig.nameservers` as skip cases. Consider documenting the `DNSPolicy=Default` behavior, skipping the override for `Default`, or at minimum emitting a Warning event so users can detect the conversion.

e2e-next/test_core/sync/test_pods_dns_nameservers.go
```diff
@@ -0,0 +1,293 @@
...
+
+			// createVPod creates a pod in the vCluster default namespace and
+			// registers cleanup. Returns the pod's name.
+			createVPod := func(ctx context.Context, name string, mutate func(*corev1.Pod)) {
+				GinkgoHelper()
+				pod := &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: "default"},
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{
+							{Name: "main", Image: "nginxinc/nginx-unprivileged:stable-alpine3.20-slim"},
```
roehrijn: use a busybox image like we do elswhere in the e2e tests.

e2e-next/test_core/sync/test_pods_dns_nameservers.go
```diff
@@ -0,0 +1,293 @@
...
+
+				secondarySvc, err := hostClient.CoreV1().Services(PodDNSNameserversNamespace).Get(ctx, PodDNSNameserversSecondaryService, metav1.GetOptions{})
+				Expect(err).To(Succeed())
+				secondaryIP = secondarySvc.Spec.ClusterIP
+				Expect(secondaryIP).NotTo(BeEmpty(), "secondary DNS Service has no ClusterIP")
+			})
+
+			// createVPod creates a pod in the vCluster default namespace and
+			// registers cleanup. Returns the pod's name.
+			createVPod := func(ctx context.Context, name string, mutate func(*corev1.Pod)) {
```
roehrijn: use GenerateName instead of the name parameter. The generated name is then in the returned Pod struct

e2e-next/test_core/sync/test_pods_dns_nameservers.go
```diff
@@ -0,0 +1,293 @@
...
+					} else {
+						svc.Labels[PodDNSNameserversLabelKey] = value
+					}
+					_, err = hostClient.CoreV1().Services(PodDNSNameserversNamespace).Update(ctx, svc, metav1.UpdateOptions{})
+					return err
+				})
+				Expect(err).To(Succeed())
+			}
+
+			It("writes all resolved ClusterIPs to synced pods", func(ctx context.Context) {
```
roehrijn, line 216: add the PR label here

e2e-next/test_core/sync/test_pods_dns_nameservers.go
```diff
@@ -0,0 +1,293 @@
...
+				podName := "dns-pod-hostnet-" + random.String(6)
+
+				By("Creating a hostNetwork pod in the vCluster", func() {
+					createVPod(ctx, podName, func(p *corev1.Pod) {
+						p.Spec.HostNetwork = true
+					})
+				})
+
+				By("Waiting for the host pod to appear without injected nameservers", func() {
+					Eventually(func(g Gomega) {
```
roehrijn, line 280: use also Consistently after the Eventuall to assert no nameserver change for PollingTimeout

e2e-next/test_core/sync/test_pods_dns_nameservers.go
```diff
@@ -0,0 +1,293 @@
...
+					createVPod(ctx, podName, func(p *corev1.Pod) {
+						p.Spec.DNSPolicy = corev1.DNSNone
+						p.Spec.DNSConfig = &corev1.PodDNSConfig{
+							Nameservers: []string{userIP},
+						}
+					})
+				})
+
+				By("Waiting for the host pod and verifying the user nameservers are preserved", func() {
+					Eventually(func(g Gomega) {
```
roehrijn, line 315: use also Consistently after the Eventuall to assert no nameserver change for PollingTimeout

e2e-next/test_core/sync/test_pods_dns_nameservers.go
```diff
@@ -0,0 +1,293 @@
...
+					}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeoutShort).Should(Succeed())
+				})
+			})
+
+			// The embedded CoreDNS variant exercises the same nameservers
+			// override but with controlPlane.coredns.embedded=true; it
+			// requires its own vcluster-pods-dns-nameservers-embedded.yaml
+			// and a dedicated suite_pods_dns_nameservers_embedded_test.go
+			// so the lifecycle stays parallelisable. Tracked separately.
+			PIt("works with embedded CoreDNS variant", Label("todo"), func(ctx context.Context) {})
```
roehrijn: remove this

e2e-next/test_core/sync/test_pods_dns_nameservers.go
roehrijn, line 1: we also need to test that the pod is rescheduled when the IP of a selected service changed and when the label selector selects a different service during runtime.

e2e-next/suite_pods_dns_nameservers_test.go
```diff
@@ -0,0 +1,183 @@
...
+			test_core.PodDNSNameserversSpec()
+		},
+	)
+}
+
+// podsDNSNameserversPreSetup provisions the host namespace and the two
+// label-selected ClusterIP Services that the vCluster's
+// sync.toHost.pods.dns.nameservers entries resolve to. Runs before the
+// vCluster starts so the syncer sees both Services on first reconcile.
+func podsDNSNameserversPreSetup(ctx context.Context) error {
```
roehrijn, line 81: the e2e tests need to not only add the services. We need to deploy a DNS server behind the services and make sure (by looking at the logs) that the dns request actually are answered by these servers. Anohter option is to add an arbitrary name to the DNS server's config which only this server is able to answer and in the pod we make sure, the name can be resolved.

e2e-next/vcluster-pods-dns-nameservers.yaml
```diff
@@ -0,0 +1,24 @@
...
+    pods:
+      enabled: true
+      dns:
+        nameservers:
+          - service:
+              scope: host
+              namespace: vcluster-dns-test
+              labelSelector:
+                matchLabels:
+                  vcluster.loft.sh/dns-ns: primary
```
roehrijn: use vcluster.com instead of vcluster.loft.sh

e2e-next/test_core/sync/test_pods_dns_nameservers.go
```diff
@@ -0,0 +1,579 @@
...
+						hostPod, err := getHostPod(ctx, pod.Name)
+						g.Expect(err).To(Succeed(), "host pod for %s vanished", pod.Name)
+						g.Expect(hostPod.Spec.DNSConfig).NotTo(BeNil(), "host pod dnsConfig is nil")
+						g.Expect(hostPod.Spec.DNSConfig.Nameservers).To(Equal([]string{userIP}),
+							"user-supplied nameservers were overwritten, got %v", hostPod.Spec.DNSConfig.Nameservers)
+					}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+				})
+			})
+
+			It("resolves partial list when one Service is missing", func(ctx context.Context) {
```
roehrijn: we need to implement a change here. All selectors in the nameservers array must resolve to 1 or more services and respectively 1 or more IP addresses. All the IP addresses must be added to the Pods nameservers list. In case a label selector is not able to resolve at least one service, the pod must not be scheduled.

mfranczy:
@codex review

e2e-next/vcluster-pods-dns-nameservers.yaml
```diff
@@ -0,0 +1,24 @@
...
+      dns:
+        nameservers:
+          - service:
+              scope: host
+              namespace: vcluster-dns-test
+              labelSelector:
+                matchLabels:
+                  vcluster.com/dns-ns: primary
+          - service:
+              scope: host
```
sowmyav27: **`scope: tenant` is supported but never exercised in e2e**

**What's the concern?**
- Both YAML entries here use `scope: host`. `buildDNSNameserversResolver` has a distinct `case vclusterconfig.DNSNameserverScopeTenant:` branch that wires `ctx.VirtualManager.GetClient()` as the reader.
- That branch ships uncaught by either unit or e2e tests.

**Why does it matter?**
- A regression in the tenant branch (wrong reader, wrong namespace lookup, nil deref) would not be detected.

**Proposed fix**
Add an `It` spec (or a parallel suite / YAML) that points at a tenant-side Service and asserts its ClusterIP is injected into `dnsConfig.nameservers`.

```yaml
nameservers:
  - service: {scope: host,   namespace: vcluster-dns-test, labelSelector: {matchLabels: {vcluster.com/dns-ns: primary}}}
  - service: {scope: tenant, namespace: kube-system,        labelSelector: {matchLabels: {k8s-app: kube-dns}}}
```

e2e-next/test_core/sync/test_pods_dns_nameservers.go
```diff
@@ -0,0 +1,543 @@
...
+					setServiceLabel(ctx, PodDNSNameserversSecondaryService, "")
+					DeferCleanup(func(ctx context.Context) {
+						setServiceLabel(ctx, PodDNSNameserversSecondaryService, "secondary")
+					})
+				})
+
+				pod := createVPod(ctx, nil)
+
+				By("Verifying the host pod is not created while one selector fails to resolve", func() {
+					Consistently(func(g Gomega) {
```
sowmyav27, line 346: **e2e "blocks pod creation" specs do not verify the Warning event**

**What's the concern?**
- Both `It("blocks pod creation when one/all Service(s) ...")` specs assert only that the host pod is absent (`kerrors.IsNotFound`).
- Neither lists events on the virtual pod to confirm a `Warning` event with reason `DNSNameservers` was emitted.
- The PR body explicitly lists *"partial-failure with Warning events"* as a covered scenario.

**Why does it matter?**
- If the event emission silently regressed (recorder dropped, reason renamed, type flipped to `Normal`), tests stay green and users lose all visibility into why their pod is stuck.

**Proposed fix**
After the `Consistently` block, list `v1.Events` for the virtual pod and assert at least one event has `Type=Warning` and `Reason=DNSNameservers`.

```go
By("Verifying a Warning event was emitted on the virtual pod", func() {
    Eventually(func(g Gomega) {
        events, err := vClusterClient.CoreV1().Events("default").List(ctx, metav1.ListOptions{
            FieldSelector: "involvedObject.name=" + pod.Name + ",involvedObject.kind=Pod",
        })
        g.Expect(err).To(Succeed())
        found := false
        for _, e := range events.Items {
            if e.Type == corev1.EventTypeWarning && e.Reason == "DNSNameservers" {
                found = true
                break
            }
        }
        g.Expect(found).To(BeTrue(), "no Warning DNSNameservers event on pod %s, events=%+v", pod.Name, events.Items)
    }).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeoutLong).Should(Succeed())
})
```

pkg/controllers/resources/pods/translate/translator_test.go
```diff
@@ -1048,3 +1049,177 @@ func TestTranslateEnforcedTolerationDedup(t *testing.T) {
...
+		select {
+		case e := <-ch:
+			out = append(out, e)
+		default:
+			return out
+		}
+	}
+}
+
+func TestApplyDNSNameservers(t *testing.T) {
```
sowmyav27, line 1107: **`DNSPolicy=Default` warning branch has no test**

**What's the concern?**
- The translator emits a `Warning` event when `vPod.Spec.DNSPolicy == corev1.DNSDefault` ("node resolv.conf search domains and options are not preserved").
- All 8 sub-tests in `TestApplyDNSNameservers` use the zero `DNSPolicy`, so this branch can be deleted without any unit test failing.

**Why does it matter?**
- This is the only signal users get that `Default` semantics are silently downgraded to `None`. Losing it means their search domains / options vanish without warning.

**Proposed fix**
Add a sub-test that sets `vPod.Spec.DNSPolicy = corev1.DNSDefault` and asserts a single Warning event whose message contains `resolv.conf`.

```go
t.Run("DNSPolicy=Default emits a Warning about lost resolv.conf options", func(t *testing.T) {
    recorder := events.NewFakeRecorder(10)
    tr, syncCtx := newDNSNameserversTranslator(t, recorder)
    vPod, pPod := basePodPair()
    vPod.Spec.DNSPolicy = corev1.DNSDefault
    resolver := &fakeNameserversResolver{enabled: true, ips: []string{"10.0.0.99"}}
    assert.NilError(t, tr.applyDNSNameservers(syncCtx, pPod, vPod, resolver))
    assert.Equal(t, pPod.Spec.DNSPolicy, corev1.DNSNone)
    evts := drainEvents(recorder.Events)
    assert.Equal(t, len(evts), 1)
    assert.Assert(t, strings.Contains(evts[0], "resolv.conf"), "got %q", evts[0])
})
```

chart/values.schema.json
```diff
@@ -1176,6 +1106,36 @@
...
+      },
+      "additionalProperties": false,
+      "type": "object",
+      "description": "DNSNameserverEntry is one element of the nameservers list."
+    },
+    "DNSNameserverService": {
+      "properties": {
+        "scope": {
+          "type": "string",
+          "description": "Scope chooses which cluster to look up the Service in.\n\n+kubebuilder:validation:Enum=host;tenant"
```
lizardruss: With the upcoming multi-tenant project, `tenant` alone is an overloaded term. So far it looks like "Platform Tenant" and "Tenant Cluster" will be the differentiating terms. Would you be opposed to making the enum values `control-plane-host` and `tenant-cluster`?

pkg/controllers/resources/pods/translate/translator.go
```diff
@@ -752,6 +772,60 @@ func (t *translator) translateDNSConfig(pPod *corev1.Pod, vPod *corev1.Pod, name
...
+	ips, errs := resolver.Resolve(ctx)
+	for _, err := range errs {
+		t.eventRecorder.Eventf(
+			vPod, nil, "Warning", "DNSNameservers", "PodDNSNameservers",
+			"could not resolve nameserver Service entry: %v", err,
+		)
+	}
+	if len(errs) > 0 {
+		return fmt.Errorf("one or more nameserver Service entries failed to resolve, requeuing: %v", errs)
+	}
```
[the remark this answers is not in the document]
roehrijn: This is intended behavior. See design doc.

pkg/controllers/resources/pods/dns_override.go
```diff
@@ -0,0 +1,145 @@
...
+	var svcList corev1.ServiceList
+	if err := e.reader.List(ctx, &svcList,
+		client.InNamespace(e.namespace),
+		client.MatchingLabelsSelector{Selector: e.selector},
+	); err != nil {
+		return nil, fmt.Errorf("list Services in %q: %w", e.namespace, err)
+	}
+	if len(svcList.Items) == 0 {
+		return nil, fmt.Errorf("no Service matches selector %q in namespace %q", e.selector, e.namespace)
+	}
```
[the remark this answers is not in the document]
roehrijn: also intended behavior

sowmyav27 (approved):

config/config.go
```diff
@@ -1553,8 +1547,59 @@ type SyncPods struct {
...
+	// +kubebuilder:validation:Enum=control-plane-host;tenant-cluster
+	Scope DNSNameserverScope `json:"scope,omitempty"`
+
+	// Namespace containing the Service.
+	Namespace string `json:"namespace,omitempty"`
+
+	// LabelSelector must match exactly one Service in Namespace. If zero or
+	// more than one Service matches, a Warning event is emitted for this entry
+	// and the entry is skipped (other entries are still resolved).
+	LabelSelector *StandardLabelSelector `json:"labelSelector,omitempty"`
```
FabianKramm, line 1589: Do we need to support this? Code would be a lot easier if we wouldn't and for the customer use case I think they can put a static name + namespace

config/config.go
```diff
@@ -1553,8 +1547,59 @@ type SyncPods struct {
...
+	// Service selects the Service whose ClusterIP becomes one nameserver.
+	Service DNSNameserverService `json:"service,omitempty"`
+}
+
+// DNSNameserverService describes the Service lookup for a single nameserver.
+type DNSNameserverService struct {
+	// Scope chooses which cluster to look up the Service in.
+	//
+	// +kubebuilder:validation:Enum=control-plane-host;tenant-cluster
+	Scope DNSNameserverScope `json:"scope,omitempty"`
```
FabianKramm, line 1581: Do we need this? Don't think thats required for now

config/config.go
```diff
@@ -1553,8 +1547,59 @@ type SyncPods struct {
...
+	// resolves independently to a Service ClusterIP which is written to
+	// dnsConfig.nameservers on every synced pod, in configuration order.
+	//
+	// Entries that fail to resolve emit a Warning event but do not block
+	// the remaining entries. If ALL entries fail, pod sync returns an error
+	// and requeues -- the pod is not created until at least one Service is found.
+	//
+	// Pods with spec.hostNetwork=true and pods that already pin
+	// spec.dnsConfig.nameservers are skipped entirely.
+	Nameservers []DNSNameserverEntry `json:"nameservers,omitempty"`
```
FabianKramm, line 1567: Do we really need multiple here? In my opinion this just adds complexity, lets just stay with a single one

FabianKramm (changes requested):

config/config.go
mfranczy: Let's keep multiple entries here since this translates directly to pod DNS config nameservers, which is already a list. Limiting it to one value would make the API less aligned with k8s and harder to extend if users need more than one nameserver later.

mfranczy:
Closing because we agreed to use a different path.

PR #3957, opened 2026-05-26
feat: add networking.advanced.overrideDNS config field

matskiv:
/kind feature

resolves #

feat: add networking.advanced.overrideDNS config field

New field allows users to override the dns server address set on synced pods.                                                                                                 Accepts an ip address or a hostname; hostnames are resolved once at startup and an error is returned if resolution fails.

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```


mfranczy:
Closing because we agreed to use a different path.

PR #3958, opened 2026-05-27
fix(snapshot): fail early when restore snapshot is inaccessible

vcluster-dv:
/kind bugfix

When `vcluster create --restore <url>` is used and the storage backend (e.g. S3) returns an error — expired credentials, missing object, permission denied — the CLI was only logging a warning and continuing. It would install the helm chart, create a resource quota, pause the vCluster, and launch the restore pod before the failure was surfaced.

The fix turns the warning in `buildExtraValues` into a hard error so the command stops before any cluster state is mutated.

resolves ENGCP-417

Fixed an issue where `vcluster create --restore` would proceed with creating and pausing the vCluster even when the snapshot was inaccessible (e.g. expired S3 credentials).

The change is a one-liner: `log.Warnf(...)` → `return nil, fmt.Errorf(...)` in `buildExtraValues`. The function `getVClusterConfigFromSnapshot` already returns `("", nil)` immediately when no `--restore` flag is set, so any error it returns implies the snapshot URL was provided and the backend rejected the access attempt.

The separate code path for restoring an already-existing vCluster (line ~177 in `create_helm.go`) does not call `buildExtraValues` and is a separate concern.

### Default Test Execution

### Adding New Test Suites

### Additional test suite(s) that will be executed before the mandatory PR suite:

```label-filter
snapshots
```

### DIFF
--- a/pkg/cli/create_helm.go
+++ b/pkg/cli/create_helm.go
@@ -403,7 +403,7 @@ func buildExtraValues(ctx context.Context, cmd *CreateOptions, log log.Logger) (
 	if len(cmd.Values) == 0 && len(cmd.SetValues) == 0 {
 		restoreValuesFile, err := getVClusterConfigFromSnapshot(ctx, cmd)
 		if err != nil {
-			log.Warnf("get vCluster config from snapshot: %w", err)
+			return nil, fmt.Errorf("get vCluster config from snapshot: %w", err)
 		} else if restoreValuesFile != "" {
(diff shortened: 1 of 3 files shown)


mfranczy:
@codex review

vcluster-dv:
@codex review

mfranczy:
@codex review

vcluster-dv:
@codex review

mfranczy (approved):

PR #3964, opened 2026-05-28
fix(snapshot): unblock PVC restore and warn on silent skip (#3951)

### LINKED ISSUE #3951: PVC restore from vcluster snapshot fails
### What happened?

We are usung vCluster OSS and want to be able to backup and restore vclusters properly. We are also using Rancher as k8s management tool.

Tests with velero in the pasts fails because of the rancher integration.

The snapshot runs fine and we get a file in our s3-Target. During the snapshot we see the volumesnapshots in the host cluster.

A snapshot restore without deleting deployments or pvcs inside the vcluster comes up but without restoring the old data, its using the existing pvc.

When we delete the deployment and pvcs in the vcluster and try to restore, the vcluster comes up but the pvcs stuck in "pending" state. 

Error in an impacted PVC describe:
`Warning  SyncError  7m51s  persistent-volume-claim-syncer  Error syncing to host cluster: update object status: persistentvolumeclaims "mysql-pv-claim-x-test-x-kw-test123" is forbidden: User "system:serviceaccount:kw-test123:vc-kw-test123" cannot update resource "persistentvolumeclaims/status" in API group "" in the namespace "kw-test123"`

PVCs will be restored sucessfully.

- create a vcluster
- create a stateful deployment (with ceph-rbd CSI driver StorageClass synced from Hostcluster)
- Run vcluster snapshot create <name> <s3-url> --include-volumes
- Verify the snapshot archive is written to S3
- Delete the workload and its PVCs
- Run vcluster snapshot restore <name> <s3-url>
- Observe that PVCs are created but remain Pending with no dataSource

- Host-Cluster OS is Talos OS
- Rancher Environment
- Ceph for CSI and S3-Endpoint

<details>

```console
$ kubectl version
Server Version: v1.33.4
```

</details>

### vcluster version

<details>

```console
$ vcluster --version
vcluster version 0.34.0
```

</details>

### VCluster Config

<details>

```
sync:
  toHost:
    ingresses:
      enabled: true
    secrets:
      enabled: true
      all: true


  fromHost:
    nodes:
      enabled: true
    storageClasses:
      enabled: true

controlPlane:
  backingStore:
    etcd:
      deploy:
        enabled: true
        statefulSet:
          highAvailability:
            replicas: 3
          resources:
            requests:
              cpu: 20m
              memory: 150Mi


  service:
    annotations:
      "loft.sh/uninstall-on-cluster-delete": "true"
  statefulSet:
    highAvailability:
      replicas: 3
    security:
      podSecurityContext:
        fsGroup: 65532
      containerSecurityContext:
        runAsUser: 65532
        runAsNonRoot: true

```

</details>

Roshr2211:
## Summary

Fixes #3951.

- **RBAC**: add `persistentvolumeclaims/status` to the syncer Role so the
  `Status().Update()` call in `patcher.ApplyObjectPatch` succeeds during
  PVC sync. Without this, restoring a PVC from a vcluster snapshot
  (after deleting the existing one) leaves it stuck `Pending` with
  `cannot update resource 'persistentvolumeclaims/status' in API group ''`.
- **Louder skip**: when a user runs restore without first deleting an
  existing PVC, the restorer silently marks it `Skipped` and emits a
  generic `Normal` event. Users assume the restore worked and lose the
  snapshot's data. Upgrade the event to `Warning`, expand the message to
  explain the cause and the required action, and add an `Infof` at the
  detection site so the skip appears in normal logs. Skip *semantics*
  are preserved as a safety default — replacing a bound PVC would
  destroy live data.

## Verification

- Live RBAC check via impersonation: \`PUT /api/v1/namespaces/X/persistentvolumeclaims/Y/status\`
  as the syncer ServiceAccount. Before: 403 Forbidden matching the exact error from #3951. After: 200 OK,
  \`status.phase=Bound\` set, \`subresource=status\` in managedFields.
- Real-running vcluster on rancher-desktop with the fixed image: PVC created in vcluster
  syncs Bound to host with no \`persistentvolumeclaims/status\` errors in syncer logs.
- \`helm unittest chart\`: 208/208 pass, including the new contains-assertion in
  \`chart/tests/role_test.yaml\`.
- \`go test ./pkg/snapshot/...\`: pass, including new unit tests for
  \`inProgressPVCReconcileFinished\` covering the Skipped / Completed / Failed branches.

## Test plan

- [x] Existing chart tests pass
- [x] New chart test pins the rule's resource list
- [x] New Go unit test asserts the Skipped event is Warning with actionable message;
      Completed/Failed branches unchanged
- [x] Live impersonated PUT to \`/status\` succeeds with the fixed Role
- [x] PVC sync works in a running vcluster pod built from this branch

### DIFF
--- a/chart/templates/role.yaml
+++ b/chart/templates/role.yaml
@@ -37,7 +37,7 @@ rules:
     resources: ["configmaps", "secrets", "services", "pods", "pods/attach", "pods/portforward", "pods/exec", "persistentvolumeclaims"]
     verbs: ["create", "delete", "patch", "update", "get", "list", "watch"]
   - apiGroups: [""]
-    resources: ["pods/status", "pods/ephemeralcontainers"]
+    resources: ["pods/status", "pods/ephemeralcontainers", "persistentvolumeclaims/status"]
     verbs: ["patch", "update"]
   {{- end }}
   {{- if ge (.Capabilities.KubeVersion.Minor|int) 35 }}
--- a/chart/tests/role_test.yaml
+++ b/chart/tests/role_test.yaml
@@ -111,6 +111,12 @@ tests:
       - equal:
           path: metadata.namespace
           value: my-namespace
+      - contains:
+          path: rules
+          content:
+            apiGroups: [""]
+            resources: ["pods/status", "pods/ephemeralcontainers", "persistentvolumeclaims/status"]
+            verbs: ["patch", "update"]
 
   - it: multi-namespace mode
     set:
--- a/pkg/snapshot/volumes/csi/restoreinprogress.go
+++ b/pkg/snapshot/volumes/csi/restoreinprogress.go
@@ -176,7 +176,12 @@ func (r *Restorer) reconcileInProgressPVC(ctx context.Context, requestObj runtim
 	originalPVC := &volumeRestoreRequest.PersistentVolumeClaim
 	_, err := r.kubeClient.CoreV1().PersistentVolumeClaims(originalPVC.Namespace).Get(ctx, originalPVC.Name, metav1.GetOptions{})
 	if err == nil {
-		// existing PVC found
+		// existing PVC found — skip restore to avoid data loss; user must delete the
+		// PVC and re-run the restore to recover data from the snapshot.
+		r.logger.Infof(
+			"PersistentVolumeClaim %s/%s already exists; skipping volume restore. "+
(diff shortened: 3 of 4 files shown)


mfranczy:
@codex review

mfranczy:
The volume snapshot feature has been removed.

PR #3965, opened 2026-05-28
fix(chart): grant lease verbs when embedded etcd is enabled

rlmcpherson:
/kind bugfix

The vCluster Pro embedded-etcd health checker (introduced in loft-sh/vcluster-pro#1815) uses a host-namespace `Lease` to leader-elect the polling replica, so health reporting survives an outage of any single pod including pod-0.

The current `Role` conditional in `chart/templates/role.yaml`:

```yaml
{{- if or .Values.privateNodes.enabled (gt (int .Values.controlPlane.statefulSet.highAvailability.replicas) 1) }}
- apiGroups: ["coordination.k8s.io"]
  resources: ["leases"]
{{- end }}
```

only grants `coordination.k8s.io/leases` verbs when `privateNodes.enabled` OR `replicas > 1`. A single-replica embedded-etcd deployment without `privateNodes` is left with no lease RBAC, and `leaderelection.RunOrDie` silently spins forever logging 403s.

This PR extends the conditional to also include `controlPlane.backingStore.etcd.embedded.enabled`, so any embedded-etcd deployment gets the lease verbs regardless of replica count.

Embedded etcd: ensure the host-namespace `Role` always grants `coordination.k8s.io/leases` verbs so the health checker can leader-elect on single-replica deployments without `privateNodes`.

- Companion change in vCluster Pro is loft-sh/vcluster-pro#1815, which uses the Lease for the new etcd health-checker leader election.
- Related Linear issue: ENGCP-882 (Standalone vCluster: add etcd health reporting).
- Chart test was extended with two cases: one verifying the rule is present when embedded etcd is enabled, and a counter-test verifying the rule is absent under default values. All 21 role tests pass locally via `helm unittest chart`.

### DIFF
--- a/chart/templates/role.yaml
+++ b/chart/templates/role.yaml
@@ -67,7 +67,7 @@ rules:
     resources: [ "endpointslices" ]
     verbs: [ "create", "list", "get", "delete", "patch", "update", "watch" ]
   {{- end }}
-  {{- if or .Values.privateNodes.enabled (gt (int .Values.controlPlane.statefulSet.highAvailability.replicas) 1) }}
+  {{- if or .Values.privateNodes.enabled (gt (int .Values.controlPlane.statefulSet.highAvailability.replicas) 1) .Values.controlPlane.backingStore.etcd.embedded.enabled }}
   - apiGroups: ["coordination.k8s.io"]
     resources: ["leases"]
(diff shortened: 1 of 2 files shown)


mfranczy:
@codex review

mfranczy (approved):

PR #3968, opened 2026-05-30
fix(storageclasses): log storage class name instead of struct pointer

chemicoholic21:
The host storage class syncer passed event.Virtual (a *storagev1.StorageClass) to a %s verb, dumping the whole struct instead of the name. Fixed to use event.Virtual.Name, matching the sibling csistoragecapacities syncer.


seanschneeweiss (approved):

seanschneeweiss:
@chemicoholic21 thanks for the contribution. Can you rebase the change to keep a linear history without merge commits?

PR #3970, opened 2026-06-01
fix(snapshot): suppress misleading etcd client warnings during restore

vcluster-dv:
/kind bugfix

Resolves ENGCP-428.

When restoring a snapshot, the key-value restore path probes and connects to a backing store that has not been started yet:
- `isEtcdReachable` runs a `MemberList` probe before kine is started (embedded/external database)
- `WaitForEtcd` polls while the store boots (deployed/external etcd)

Because vcluster installs a real global zap logger (`WithGlobalZap(true)` → `zap.ReplaceGlobals`) and every restore etcd client was created with `zap.L().Named("etcd-client")`, the etcd `clientv3` retry interceptor logged `retrying of unary invoker failed` warnings straight to the console. These retries are expected and ultimately succeed, but they look like the user did something wrong. The existing mitigation in `isEtcdReachable` only discarded grpclog output, not the clientv3 zap logger, so the warning still leaked.

This passes a no-op logger to the clientv3 client in both probe paths (still using the named logger under `-v=1` for debugging), so the expected retries no longer print false warnings. Real errors continue to propagate as returned errors. Embedded etcd restored with the new snapshot format was already unaffected, since it only talks to etcd after `WaitForEtcd` confirms readiness.

Fixed an issue where vcluster printed misleading etcd client warnings ("retrying of unary invoker failed") while restoring a snapshot.

Affects database, deployed etcd and external etcd backing stores. No functional change to restore behavior; logging only.

### Default Test Execution
The mandatory PR suite runs automatically.

```label-filter
snapshots
```

### DIFF
--- a/pkg/etcd/util.go
+++ b/pkg/etcd/util.go
@@ -25,19 +25,10 @@ type Certificates struct {
 }
 
 func WaitForEtcd(parentCtx context.Context, certificates *Certificates, endpoints ...string) error {
-	// This polls etcd before it is necessarily reachable, so transient dial
-	// failures are expected. Silence the etcd client logger (we log our own
-	// retry message below) unless verbose logging is enabled, so we don't print
-	// misleading "retrying of unary invoker failed" warnings.
-	log := zap.NewNop()
-	if klog.V(1).Enabled() {
-		log = zap.L().Named("wait-for-etcd")
-	}
-
 	var err error
 	waitErr := wait.PollUntilContextTimeout(parentCtx, time.Second, waitForClientTimeout, true, func(ctx context.Context) (bool, error) {
 		var etcdClient *clientv3.Client
-		etcdClient, err = GetEtcdClient(ctx, log, certificates, endpoints...)
+		etcdClient, err = GetEtcdClient(ctx, certificates, endpoints...)
 		if err == nil {
 			defer func() {
 				_ = etcdClient.Close()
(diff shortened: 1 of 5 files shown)


pkg/etcd/util.go
```diff
@@ -25,10 +25,19 @@ type Certificates struct {
...
+	log := zap.NewNop()
+	if klog.V(1).Enabled() {
+		log = zap.L().Named("wait-for-etcd")
+	}
+
 	var err error
 	waitErr := wait.PollUntilContextTimeout(parentCtx, time.Second, waitForClientTimeout, true, func(ctx context.Context) (bool, error) {
 		var etcdClient *clientv3.Client
-		etcdClient, err = GetEtcdClient(ctx, zap.L().Named("wait-for-etcd"), certificates, endpoints...)
+		etcdClient, err = GetEtcdClient(ctx, log, certificates, endpoints...)
```
mfranczy: I think it makes sense to move this logic into `GetEtcdClient`:
```
log := zap.NewNop()
	if klog.V(1).Enabled() {
		log = zap.L().Named("wait-for-etcd")
	}
```

That would allow us to remove the log argument.

vcluster-dv: Please re-check. Note, as GetEtcdClient is exported, we changed the contract due to this.

vcluster-dv:
@codex review

pkg/etcd/util.go
```diff
@@ -55,7 +55,16 @@ func WaitForEtcd(parentCtx context.Context, certificates *Certificates, endpoint
...
 // The returned client should be closed when no longer needed, in order to avoid leaking GRPC
 // client goroutines.
-func GetEtcdClient(ctx context.Context, log *zap.Logger, certificates *Certificates, endpoints ...string) (*clientv3.Client, error) {
+func GetEtcdClient(ctx context.Context, certificates *Certificates, endpoints ...string) (*clientv3.Client, error) {
+	// etcd clients frequently connect before etcd is reachable (reachability
+	// probes, restore, startup), so the clientv3 retry interceptor logs
+	// misleading "retrying of unary invoker failed" warnings. Silence the client
+	// logger unless verbose logging is enabled.
+	log := zap.NewNop()
+	if klog.V(1).Enabled() {
```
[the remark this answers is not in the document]
vcluster-dv: Already honored. `DEBUG=true` maps to `LOFT_LOG_LEVEL=debug`, `loftlogr` translates a debug level into `KUBERNETES_VERBOSITY_LEVEL=1`, which `SetGlobalKlog` (enabled via `WithGlobalKlog(true)`) applies with `klogFlagSet.Set("v", "1")`. So, under debug, `klog.V(1).Enabled()` is true and the `etcd-client` logger is re-enabled. `SetGlobalKlog` also calls `klog.InitFlags` in real code, not only tests.

mfranczy (approved):

PR #3983, opened 2026-06-05
Agent/cedar amber h7jq

chemicoholic21:
## Summary
The download-latest-cli job in e2e.yaml and e2e-ginkgo.yaml downloaded HTML error pages instead of the vCluster binary because curl didn't fail on HTTP errors.

## Changes
- Added `-f` flag to curl to fail on HTTP 4xx/5xx
- Added `--retry 3 --retry-delay 5` for transient failures
- Added ELF binary verification check


seanschneeweiss:
I cannot accept references to github users in the code, sorry.

PR #3985, opened 2026-06-06
Fix/test plan p0s

zerbitx:
/kind bugfix

Resolves ENGNODE-552
Resolves ENGNODE-553
Resolves ENGNODE-554
Resolves ENGNODE-556
Resolves ENGNODE-557
Resolves ENGNODE-558
Resolves ENGNODE-560

This covers several issues found in qa checks for PO cases and adds e2e tests for those P0 checks.

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

### DIFF
--- a/e2e-next/suite_gatewayapi_test.go
+++ b/e2e-next/suite_gatewayapi_test.go
@@ -25,6 +25,7 @@ const gatewayAPIVClusterName = "gatewayapi-vcluster"
 func init() { suiteGatewayAPIVCluster() }
 
 func suiteGatewayAPIVCluster() {
+	// Ordered so all specs share one lazyvcluster bring-up; specs are independent.
 	Describe("gatewayapi-vcluster", labels.PR, labels.GatewayAPI, labels.GatewayClasses, Ordered,
 		cluster.Use(clusters.HostCluster),
 		func() {
--- a/e2e-next/suite_gatewayapi_test.go
+++ b/e2e-next/suite_gatewayapi_test.go
@@ -1,0 +1,42 @@
+// Suite: gatewayapi-vcluster
+// vCluster: Gateway API sync coverage for GatewayClass visibility, Tenant
+// Gateway authorization, and Gateway/HTTPRoute sync.
+// Run:      just run-e2e 'pr && gatewayapi'
+package e2e_next
+
+import (
+	"context"
+	_ "embed"
+
+	"github.com/loft-sh/e2e-framework/pkg/setup/cluster"
+	"github.com/loft-sh/vcluster/e2e-next/clusters"
+	"github.com/loft-sh/vcluster/e2e-next/labels"
+	"github.com/loft-sh/vcluster/e2e-next/setup"
+	"github.com/loft-sh/vcluster/e2e-next/setup/lazyvcluster"
+	"github.com/loft-sh/vcluster/e2e-next/test_gatewayapi"
+	. "github.com/onsi/ginkgo/v2"
+)
+
+//go:embed vcluster-gatewayapi.yaml
+var gatewayAPIVClusterYAML string
+
+const gatewayAPIVClusterName = "gatewayapi-vcluster"
+
+func init() { suiteGatewayAPIVCluster() }
+
+func suiteGatewayAPIVCluster() {
+	Describe("gatewayapi-vcluster", labels.PR, labels.GatewayAPI, labels.GatewayClasses, Ordered,
+		cluster.Use(clusters.HostCluster),
+		func() {
+			BeforeAll(func(ctx context.Context) context.Context {
+				return lazyvcluster.LazyVCluster(ctx,
+					gatewayAPIVClusterName,
+					gatewayAPIVClusterYAML,
+					lazyvcluster.WithPreSetup(setup.GatewayAPIPreSetup()),
+				)
+			})
+
+			test_gatewayapi.GatewayAPISyncSpec()
+		},
+	)
+}
--- a/e2e-next/test_gatewayapi/test_gatewayapi_import.go
+++ b/e2e-next/test_gatewayapi/test_gatewayapi_import.go
@@ -1,0 +1,393 @@
+package test_gatewayapi
+
+import (
+	"context"
+
+	"github.com/loft-sh/e2e-framework/pkg/setup/cluster"
+	"github.com/loft-sh/vcluster/e2e-next/constants"
+	"github.com/loft-sh/vcluster/e2e-next/labels"
+	"github.com/loft-sh/vcluster/pkg/util/random"
+	"github.com/loft-sh/vcluster/pkg/util/translate"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+const importClassSelectorValue = "gatewayapi-import"
+
+// GatewayAPIImportSpec registers fromHost imported Gateway tests (P0 cases
+// 01-08, 16, control-plane behavior only).
+func GatewayAPIImportSpec() {
+	Describe("Gateway API import", labels.GatewayAPI, labels.GatewayClasses, func() {
+		var (
+			hostClient     ctrlclient.Client
+			vClusterClient ctrlclient.Client
+			vClusterName   string
+			vClusterHostNS string
+		)
+
+		BeforeEach(func(ctx context.Context) {
+			var err error
+			scheme := runtime.NewScheme()
+			Expect(corev1.AddToScheme(scheme)).To(Succeed())
+			Expect(gatewayv1.Install(scheme)).To(Succeed())
+
+			hostClient, err = ctrlclient.New(cluster.From(ctx, constants.GetHostClusterName()).KubernetesRestConfig(), ctrlclient.Options{Scheme: scheme})
+			Expect(err).To(Succeed())
+			vClusterClient, err = ctrlclient.New(cluster.CurrentClusterFrom(ctx).KubernetesRestConfig(), ctrlclient.Options{Scheme: scheme})
+			Expect(err).To(Succeed())
+			vClusterName = cluster.CurrentClusterNameFrom(ctx)
+			vClusterHostNS = "vcluster-" + vClusterName
+
+			ensureHostNamespace(ctx, hostClient, "gwapi-host")
+			Eventually(func(g Gomega) {
+				g.Expect(vClusterClient.List(ctx, &gatewayv1.GatewayClassList{})).To(Succeed())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+		})
+
+		It("imports a host Gateway into the tenant and syncs attached routes to the host Gateway", labels.PR, func(ctx context.Context) {
+			suffix := random.String(6)
+			class := createGatewayClass(ctx, hostClient, "gwc-import-"+suffix, importClassSelectorValue, "import class")
+
+			By("creating a host Gateway in the wildcard-mapped source namespace")
+			hostGW := hostGateway("gwapi-host", "edge-"+suffix, class.Name)
+			createHostGateway(ctx, hostClient, hostGW)
+
+			By("verifying the tenant mirror, namespace and GatewayClass exist")
+			Eventually(func(g Gomega) {
+				g.Expect(vClusterClient.Get(ctx, types.NamespacedName{Name: class.Name}, &gatewayv1.GatewayClass{})).To(Succeed())
+				mirror := &gatewayv1.Gateway{}
+				g.Expect(vClusterClient.Get(ctx, types.NamespacedName{Namespace: "gwapi-import", Name: hostGW.Name}, mirror)).To(Succeed())
+				g.Expect(mirror.Labels).To(HaveKeyWithValue("vcluster.loft.sh/imported-gateway", "true"))
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+
+			By("creating a tenant HTTPRoute attached to the imported Gateway")
+			routeNS := createTenantNamespace(ctx, vClusterClient, "gwapi-app-"+suffix)
+			svc := &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: "backend-" + suffix, Namespace: routeNS.Name}, Spec: corev1.ServiceSpec{Ports: []corev1.ServicePort{{Port: 80}}}}
+			Expect(vClusterClient.Create(ctx, svc)).To(Succeed())
+			route := importRoute(routeNS.Name, "route-"+suffix, "gwapi-import", hostGW.Name, svc.Name, "app.apps.example.com")
+			Expect(vClusterClient.Create(ctx, route)).To(Succeed())
+			DeferCleanup(func(ctx context.Context) {
+				Expect(ctrlclient.IgnoreNotFound(vClusterClient.Delete(ctx, route))).To(Succeed())
+			})
+
+			By("verifying the host route parentRef points at the host Gateway")
+			hostRouteName := translate.SafeConcatName(route.Name, "x", routeNS.Name, "x", vClusterName)
+			Eventually(func(g Gomega) {
+				got := &gatewayv1.HTTPRoute{}
+				g.Expect(hostClient.Get(ctx, types.NamespacedName{Namespace: vClusterHostNS, Name: hostRouteName}, got)).To(Succeed())
+				g.Expect(got.Spec.ParentRefs).To(HaveLen(1))
+				g.Expect(got.Spec.ParentRefs[0].Name).To(Equal(gatewayv1.ObjectName(hostGW.Name)))
+				g.Expect(got.Spec.ParentRefs[0].Namespace).NotTo(BeNil())
+				g.Expect(*got.Spec.ParentRefs[0].Namespace).To(Equal(gatewayv1.Namespace("gwapi-host")))
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+
+			By("confirming no tenant-created host Gateway object exists")
+			Consistently(func(g Gomega) {
+				err := hostClient.Get(ctx, types.NamespacedName{Namespace: vClusterHostNS, Name: translate.SafeConcatName(hostGW.Name, "x", "gwapi-import", "x", vClusterName)}, &gatewayv1.Gateway{})
+				g.Expect(kerrors.IsNotFound(err)).To(BeTrue())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeoutShort).Should(Succeed())
+		})
+
+		It("mirrors the GatewayClass with controllerName preserved and parametersRef stripped", labels.PR, func(ctx context.Context) {
+			suffix := random.String(6)
+			class := createGatewayClass(ctx, hostClient, "gwc-sanitize-"+suffix, importClassSelectorValue, "sanitize class")
+
+			Eventually(func(g Gomega) {
+				mirror := &gatewayv1.GatewayClass{}
+				g.Expect(vClusterClient.Get(ctx, types.NamespacedName{Name: class.Name}, mirror)).To(Succeed())
+				g.Expect(mirror.Spec.ControllerName).To(Equal(gatewayControllerName))
+				g.Expect(mirror.Spec.ParametersRef).To(BeNil())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+		})
+
+		It("does not sync routes that attach to a tenant-local non-imported Gateway", labels.PR, func(ctx context.Context) {
+			suffix := random.String(6)
+			class := createGatewayClass(ctx, hostClient, "gwc-local-"+suffix, importClassSelectorValue, "local class")
+			Eventually(func(g Gomega) {
+				g.Expect(vClusterClient.Get(ctx, types.NamespacedName{Name: class.Name}, &gatewayv1.GatewayClass{})).To(Succeed())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+
+			routeNS := createTenantNamespace(ctx, vClusterClient, "gwapi-local-"+suffix)
+			svc := &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: "backend-" + suffix, Namespace: routeNS.Name}, Spec: corev1.ServiceSpec{Ports: []corev1.ServicePort{{Port: 80}}}}
+			Expect(vClusterClient.Create(ctx, svc)).To(Succeed())
+
+			By("creating a tenant-local Gateway that is not imported from host")
+			localGW := tenantGateway(routeNS.Name, "local-gw-"+suffix, class.Name)
+			Expect(vClusterClient.Create(ctx, localGW)).To(Succeed())
+			DeferCleanup(func(ctx context.Context) {
+				Expect(ctrlclient.IgnoreNotFound(vClusterClient.Delete(ctx, localGW))).To(Succeed())
+			})
+			route := importRoute(routeNS.Name, "route-"+suffix, routeNS.Name, localGW.Name, svc.Name, "app.example.com")
+			Expect(vClusterClient.Create(ctx, route)).To(Succeed())
+			DeferCleanup(func(ctx context.Context) {
+				Expect(ctrlclient.IgnoreNotFound(vClusterClient.Delete(ctx, route))).To(Succeed())
+			})
+
+			hostRouteName := translate.SafeConcatName(route.Name, "x", routeNS.Name, "x", vClusterName)
+			Consistently(func(g Gomega) {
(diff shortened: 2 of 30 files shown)


zerbitx:
@codex review

e2e-next/test_gatewayapi/test_gatewayapi_sync.go
```diff
@@ -0,0 +1,265 @@
...
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+
+			Expect(vClusterClient.Get(ctx, ctrlclient.ObjectKeyFromObject(good), good)).To(Succeed())
+			good.Spec.GatewayClassName = gatewayv1.ObjectName(hidden.Name)
+			Expect(vClusterClient.Update(ctx, good)).To(Succeed())
+			Eventually(func(g Gomega) {
+				err := hostClient.Get(ctx, types.NamespacedName{Namespace: vClusterHostNS, Name: goodHostName}, &gatewayv1.Gateway{})
+				g.Expect(kerrors.IsNotFound(err)).To(BeTrue())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+		})
```
sowmyav27, line 167: Non-blocking refinement: when the tenant Gateway points to an unavailable GatewayClass, this spec only asserts host-side absence. Add tenant-side status assertions so the user-facing observability path is covered:

```go
g.Expect(meta.IsStatusConditionPresentAndEqual(got.Status.Conditions, "Programmed", metav1.ConditionUnknown)).To(BeTrue())
```

Also assert the `SyncWarning` event (`GatewayClass "<name>" is not visible in the virtual cluster`) is emitted on the tenant Gateway.

e2e-next/test_gatewayapi/test_gatewayapi_import.go
```diff
@@ -0,0 +1,393 @@
...
+			By("creating a host Gateway in the wildcard-mapped source namespace")
+			hostGW := hostGateway("gwapi-host", "edge-"+suffix, class.Name)
+			createHostGateway(ctx, hostClient, hostGW)
+
+			By("verifying the tenant mirror, namespace and GatewayClass exist")
+			Eventually(func(g Gomega) {
+				g.Expect(vClusterClient.Get(ctx, types.NamespacedName{Name: class.Name}, &gatewayv1.GatewayClass{})).To(Succeed())
+				mirror := &gatewayv1.Gateway{}
+				g.Expect(vClusterClient.Get(ctx, types.NamespacedName{Namespace: "gwapi-import", Name: hostGW.Name}, mirror)).To(Succeed())
+				g.Expect(mirror.Labels).To(HaveKeyWithValue("vcluster.loft.sh/imported-gateway", "true"))
```
sowmyav27: Nit: extract the imported-gateway marker label `"vcluster.loft.sh/imported-gateway"` into an exported constant rather than a string literal here. Unsure where it should live — likely alongside the syncer that writes the label so the test imports the same constant the producer emits.

e2e-next/test_gatewayapi/test_gatewayapi_tohost.go
```diff
@@ -0,0 +1,213 @@
...
+			backend := createTenantNamespace(ctx, vClusterClient, "backend-"+suffix)
+			gw := tenantGateway(frontend.Name, "gw-"+suffix, class.Name)
+			Expect(vClusterClient.Create(ctx, gw)).To(Succeed())
+			DeferCleanup(func(ctx context.Context) {
+				Expect(ctrlclient.IgnoreNotFound(vClusterClient.Delete(ctx, gw))).To(Succeed())
+			})
+			svc := &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: "backend-svc-" + suffix, Namespace: backend.Name}, Spec: corev1.ServiceSpec{Ports: []corev1.ServicePort{{Port: 80}}}}
+			Expect(vClusterClient.Create(ctx, svc)).To(Succeed())
+
+			By("creating a route whose backendRef crosses into another namespace")
```
sowmyav27: `By(...)` calls in this file are bare log annotations — missing the closure form. Without `func() { ... }` Ginkgo attributes failures to the enclosing It, not the specific step. Affects lines 71, 84, 129, 140, 180, 191 in this file. Wrap each block: `By("msg", func() { ... })`. sync.go uses the closure form correctly throughout.

e2e-next/test_gatewayapi/test_gatewayapi_import.go
```diff
@@ -0,0 +1,393 @@
...
+			Eventually(func(g Gomega) {
+				g.Expect(vClusterClient.List(ctx, &gatewayv1.GatewayClassList{})).To(Succeed())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+		})
+
+		It("imports a host Gateway into the tenant and syncs attached routes to the host Gateway", labels.PR, func(ctx context.Context) {
+			suffix := random.String(6)
+			class := createGatewayClass(ctx, hostClient, "gwc-import-"+suffix, importClassSelectorValue, "import class")
+
+			By("creating a host Gateway in the wildcard-mapped source namespace")
```
sowmyav27: `By(...)` calls in this file are bare log annotations — missing the closure form. Without `func() { ... }` Ginkgo attributes failures to the enclosing It, not the specific step. Affects lines 59, 63, 71, 81, 92, 122, 152, 159, 178, 189, 197, 214, 220, 239, 254, 282, 294, 313, 318. Wrap each block: `By("msg", func() { ... })`.

e2e-next/test_gatewayapi/test_gatewayapi_tohost.go
```diff
@@ -0,0 +1,213 @@
...
+
+			hostClient, err = ctrlclient.New(cluster.From(ctx, constants.GetHostClusterName()).KubernetesRestConfig(), ctrlclient.Options{Scheme: scheme})
+			Expect(err).To(Succeed())
+			vClusterClient, err = ctrlclient.New(cluster.CurrentClusterFrom(ctx).KubernetesRestConfig(), ctrlclient.Options{Scheme: scheme})
+			Expect(err).To(Succeed())
+			vClusterName = cluster.CurrentClusterNameFrom(ctx)
+			vClusterHostNS = "vcluster-" + vClusterName
+		})
+
+		It("does not sync cross-namespace backendRef routes until a ReferenceGrant permits it", labels.PR, func(ctx context.Context) {
```
sowmyav27: `labels.PR` is already on the suite-level Describe in `suite_gatewayapi_test.go:28` — Ginkgo inherits labels, so re-adding at every It is redundant. sync.go It blocks correctly omit it. Drop `labels.PR` from It blocks at lines 54, 101, 148 in this file (and 55, 99, 111, 141, 167, 203, 228, 267, 306 in import.go).

e2e-next/suite_gatewayapi_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+//go:embed vcluster-gatewayapi.yaml
+var gatewayAPIVClusterYAML string
+
+const gatewayAPIVClusterName = "gatewayapi-vcluster"
+
+func init() { suiteGatewayAPIVCluster() }
+
+func suiteGatewayAPIVCluster() {
+	Describe("gatewayapi-vcluster", labels.PR, labels.GatewayAPI, labels.GatewayClasses, Ordered,
```
sowmyav27, line 29: `Ordered` is used to share the lazyvcluster bring-up across specs (not because specs depend on each other). Add a one-line comment so the next reader does not have to guess whether removing `Ordered` is safe:
```go
// Ordered so all specs share one lazyvcluster bring-up; specs are independent.
```
Same applies to `suite_gatewayapi_import_test.go:28`.

e2e-next/test_gatewayapi/test_gatewayapi_tohost.go
```diff
@@ -0,0 +1,213 @@
...
+	"k8s.io/utils/ptr"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gatewayv1alpha3 "sigs.k8s.io/gateway-api/apis/v1alpha3"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+// GatewayAPIToHostSpec registers tenant-authored route/policy sync tests
+// (P0 cases 13, 18, 19).
```
sowmyav27: Doc-comment says `(P0 cases 13, 18, 19)` but TC-13 is sleep, TC-18 is private nodes, TC-19 is standalone — none of those are covered here. This file actually covers TC-07 (TLSRoute), TC-09 (BackendTLSPolicy), TC-11 (cross-ns refGrant). Suggest dropping the TC numbers and letting the spec descriptions speak — TC references in code rot quickly.

e2e-next/test_gatewayapi/test_gatewayapi_import.go
```diff
@@ -0,0 +1,393 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+const importClassSelectorValue = "gatewayapi-import"
+
+// GatewayAPIImportSpec registers fromHost imported Gateway tests (P0 cases
+// 01-08, 16, control-plane behavior only).
```
sowmyav27: Doc-comment says `(P0 cases 01-08, 16, control-plane behavior only)` but the file actually covers TC-30/32/42/43/45 (fromHost Gateway import). Suggest dropping the TC numbers — they rot quickly.

e2e-next/test_gatewayapi/test_gatewayapi_import.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+func ensureHostNamespace(ctx context.Context, c ctrlclient.Client, name string) {
+	GinkgoHelper()
+	ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: name}}
+	err := c.Create(ctx, ns)
+	if err != nil && !kerrors.IsAlreadyExists(err) {
+		Expect(err).To(Succeed())
+	}
+}
```
sowmyav27, line 369: This helper creates host-side namespaces (`gwapi-host`, `gwapi-host-policy`, `gwapi-host-rename`) and never registers `DeferCleanup`. Reruns on the same Control Plane Cluster (CI or local dev) inherit stale namespaces. Suggest: only Create when not pre-existing AND register DeferCleanup for the ones this suite created.
```go
err := c.Create(ctx, ns)
if kerrors.IsAlreadyExists(err) {
    return // pre-existing, not ours to delete
}
Expect(err).To(Succeed())
DeferCleanup(func(ctx context.Context) {
    Expect(ctrlclient.IgnoreNotFound(c.Delete(ctx, ns))).To(Succeed())
})
```

e2e-next/test_gatewayapi/test_gatewayapi_tohost.go
```diff
@@ -0,0 +1,213 @@
...
+
+			route := &gatewayv1alpha2.TLSRoute{
+				ObjectMeta: metav1.ObjectMeta{Name: "app-tls-" + suffix, Namespace: ns.Name},
+				Spec: gatewayv1alpha2.TLSRouteSpec{
+					Hostnames: []gatewayv1.Hostname{"app.apps.example.com"},
+					Rules: []gatewayv1alpha2.TLSRouteRule{{BackendRefs: []gatewayv1.BackendRef{{
+						BackendObjectReference: gatewayv1.BackendObjectReference{Name: gatewayv1.ObjectName(svc.Name), Port: ptr.To(gatewayv1.PortNumber(443))},
+					}}}},
+				},
+			}
```
sowmyav27, line 141: This TLSRoute has `Hostnames` and `Rules` but no `CommonRouteSpec.ParentRefs`. A parentless TLSRoute is non-functional in production. The test only proves spec-copy and hostname propagation — the syncers primary correctness path for routes (parentRef.namespace/name translation to the host-translated Gateway) is never exercised. Add a tenant Gateway, add ParentRefs to the route, then assert the host TLSRoutes parentRef is the translated host gateway name.

e2e-next/test_gatewayapi/test_gatewayapi_import.go
```diff
@@ -0,0 +1,393 @@
...
+			By("deleting the host Gateway")
+			Expect(hostClient.Delete(ctx, hostGW)).To(Succeed())
+			Eventually(func(g Gomega) {
+				err := vClusterClient.Get(ctx, mirrorKey, &gatewayv1.Gateway{})
+				g.Expect(kerrors.IsNotFound(err)).To(BeTrue())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+
+			By("recreating the host Gateway")
+			recreated := hostGateway("gwapi-host", hostGW.Name, class.Name)
+			Expect(hostClient.Create(ctx, recreated)).To(Succeed())
```
sowmyav27: Bypasses the `createHostGateway` helper, so no `DeferCleanup` is registered for the recreated gateway — it leaks if a later assertion in this or an appended spec fails. Use the helper: `createHostGateway(ctx, hostClient, recreated)`.

e2e-next/test_gatewayapi/test_gatewayapi_import.go
```diff
@@ -0,0 +1,393 @@
...
+			Expect(hostClient.Get(ctx, types.NamespacedName{Namespace: "gwapi-host", Name: hostGW.Name}, current)).To(Succeed())
+			current.Status.Addresses = []gatewayv1.GatewayStatusAddress{{Type: ptr.To(gatewayv1.IPAddressType), Value: "203.0.113.10"}}
+			Expect(hostClient.Status().Update(ctx, current)).To(Succeed())
+
+			By("verifying the tenant mirror hides addresses")
+			Consistently(func(g Gomega) {
+				got := &gatewayv1.Gateway{}
+				g.Expect(vClusterClient.Get(ctx, mirrorKey, got)).To(Succeed())
+				g.Expect(got.Status.Addresses).To(BeEmpty())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeoutShort).Should(Succeed())
```
sowmyav27: `Consistently` runs immediately after the host `Status().Update` with no `Eventually` between them. If the syncer has not yet observed the host status change, Consistently trivially passes — a regression that *would* leak addresses on observation still passes. Gate this with an `Eventually` first (e.g. assert another mutable status field DOES propagate, then assert addresses stay empty).

e2e-next/test_gatewayapi/test_gatewayapi_tohost.go
```diff
@@ -0,0 +1,213 @@
...
+					To:   []gatewayv1beta1.ReferenceGrantTo{{Group: gatewayv1.Group(""), Kind: gatewayv1.Kind("Service")}},
+				},
+			}
+			Expect(vClusterClient.Create(ctx, grant)).To(Succeed())
+			DeferCleanup(func(ctx context.Context) {
+				Expect(ctrlclient.IgnoreNotFound(vClusterClient.Delete(ctx, grant))).To(Succeed())
+			})
+			Eventually(func(g Gomega) {
+				g.Expect(hostClient.Get(ctx, types.NamespacedName{Namespace: vClusterHostNS, Name: hostRouteName}, &gatewayv1.HTTPRoute{})).To(Succeed())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
```
sowmyav27: Test proves `no RG → no sync → create RG → sync appears`. Does not prove `delete RG → host route removed`. RG is the security gate — the revoke path is the regression-prone half. Add a delete-RG step and Eventually-assert the host route disappears.

e2e-next/test_gatewayapi/test_gatewayapi_import.go
```diff
@@ -0,0 +1,393 @@
...
+			Expect(vClusterClient.Get(ctx, mirrorKey, toDelete)).To(Succeed())
+			Expect(vClusterClient.Delete(ctx, toDelete)).To(Succeed())
+			Eventually(func(g Gomega) {
+				g.Expect(vClusterClient.Get(ctx, mirrorKey, &gatewayv1.Gateway{})).To(Succeed())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+
+			By("confirming the host Gateway is never mutated")
+			got := &gatewayv1.Gateway{}
+			Expect(hostClient.Get(ctx, types.NamespacedName{Namespace: "gwapi-host", Name: hostGW.Name}, got)).To(Succeed())
+			Expect(got.Spec.Listeners[0].Port).To(Equal(gatewayv1.PortNumber(80)))
```
sowmyav27: Single point-in-time `Get`. A misbehaving syncer that propagates the tenant mutation *later* (race) would not be caught. Wrap in `Consistently` over a short window so any late mutation fails the spec.

e2e-next/test_gatewayapi/test_gatewayapi_tohost.go
```diff
@@ -0,0 +1,213 @@
...
+			Expect(gatewayv1alpha3.Install(scheme)).To(Succeed())
+			Expect(gatewayv1beta1.Install(scheme)).To(Succeed())
+
+			hostClient, err = ctrlclient.New(cluster.From(ctx, constants.GetHostClusterName()).KubernetesRestConfig(), ctrlclient.Options{Scheme: scheme})
+			Expect(err).To(Succeed())
+			vClusterClient, err = ctrlclient.New(cluster.CurrentClusterFrom(ctx).KubernetesRestConfig(), ctrlclient.Options{Scheme: scheme})
+			Expect(err).To(Succeed())
+			vClusterName = cluster.CurrentClusterNameFrom(ctx)
+			vClusterHostNS = "vcluster-" + vClusterName
+		})
```
sowmyav27, line 46: sync.go BeforeEach (lines 52-56) waits for the tenant API to respond to GatewayClass/Gateway/HTTPRoute List calls. tohost.go has no such wait — the first spec hits cold tenant API and is flake-prone. Add an equivalent `Eventually` listing HTTPRouteList, TLSRouteList, BackendTLSPolicyList.

e2e-next/test_gatewayapi/test_gatewayapi_sync.go
```diff
@@ -0,0 +1,265 @@
...
+				}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+			})
+
+			By("deleting the Host GatewayClass and waiting for the Tenant mirror to disappear", func() {
+				Expect(ctrlclient.IgnoreNotFound(hostClient.Delete(ctx, allowed))).To(Succeed())
+				Eventually(func(g Gomega) {
+					err := vClusterClient.Get(ctx, types.NamespacedName{Name: allowed.Name}, &gatewayv1.GatewayClass{})
+					g.Expect(kerrors.IsNotFound(err)).To(BeTrue())
+				}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+			})
```
sowmyav27, line 115: This spec proves selector decision at creation time (allowed appears, hidden absent) but does not test runtime selector transitions: (1) relabel a hidden class to matching → mirror should appear; (2) delabel a matching class → mirror should disappear. These are common label-driven syncer regression vectors — a syncer may evaluate the selector once at creation but miss re-evaluating on label updates (cache, watch-predicate). Worth adding two `By(...)` steps.

e2e-next/test_gatewayapi/test_gatewayapi_import.go
```diff
@@ -0,0 +1,393 @@
...
+			Expect(vClusterClient.Create(ctx, route)).To(Succeed())
+			DeferCleanup(func(ctx context.Context) {
+				Expect(ctrlclient.IgnoreNotFound(vClusterClient.Delete(ctx, route))).To(Succeed())
+			})
+
+			hostRouteName := translate.SafeConcatName(route.Name, "x", routeNS.Name, "x", vClusterName)
+			Consistently(func(g Gomega) {
+				err := hostClient.Get(ctx, types.NamespacedName{Namespace: vClusterHostNS, Name: hostRouteName}, &gatewayv1.HTTPRoute{})
+				g.Expect(kerrors.IsNotFound(err)).To(BeTrue())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeoutShort).Should(Succeed())
```
sowmyav27: `vcluster-gatewayapi-import.yaml` only enables `toHost.gatewayApi.httpRoutes`, not `gateways`. So absence of the route on host could be from (a) the syncer correctly gating routes on imported parents, OR (b) the tenant Gateway never reaches the host because `gateways` toHost is off. A regression in (a) would not be caught by this assertion. Either add a companion spec in the non-import suite (where `gateways` toHost is enabled), or extend this assertion to also verify the tenant Gateway has no host counterpart — that rules out (a) being satisfied vacuously.

e2e-next/test_gatewayapi/test_gatewayapi_tohost.go
```diff
@@ -0,0 +1,213 @@
...
+				g.Expect(got.Spec.Validation.Hostname).To(Equal(gatewayv1.PreciseHostname("backend-updated.apps.example.com")))
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+
+			By("deleting the policy and expecting host deletion")
+			Expect(ctrlclient.IgnoreNotFound(vClusterClient.Delete(ctx, current))).To(Succeed())
+			Eventually(func(g Gomega) {
+				err := hostClient.Get(ctx, hostKey, &gatewayv1alpha3.BackendTLSPolicy{})
+				g.Expect(kerrors.IsNotFound(err)).To(BeTrue())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+		})
```
sowmyav27, line 258: Only positive sync + translated targetRef is covered. Add at least one negative companion: BackendTLSPolicy with cross-namespace targetRef and no ReferenceGrant — should not sync to host. Symmetric with the HTTPRoute cross-ns refGrant test above.

sowmyav27:
The PR body says this resolves ENGNODE-556 (TC-33: enabling both `fromHost.gateways: true` AND `toHost.gatewayApi.gateways: true` together caused crash-loop). I cannot find any test in the PR that exercises this combination:

- `vcluster-gatewayapi.yaml` has `toHost.gatewayApi.gateways: true` but no `fromHost.gateways`.
- `vcluster-gatewayapi-import.yaml` has `fromHost.gateways: true` but no `toHost.gatewayApi.gateways`.
- The unit tests (`TestTenantGatewaysRequireExplicitToggle`, `TestGatewayClassesEnabledWhenGatewaysAreImported`, etc.) do not set both flags together either.

Could a test case be added that creates a tenant vcluster with both toggles enabled and asserts the cluster comes up successfully (per TC-33)? Without it there is no regression coverage for ENGNODE-556 — and it would also let the `import.go:111-139` "tenant-local non-imported Gateway" spec be made unambiguous (today its assertion passes whether or not the imported-parent gating rule is actually working, because `toHost.gatewayApi.gateways` is off in that suite).

e2e-next/test_gatewayapi/test_gatewayapi_sync.go
zerbitx: Without a real gateway controller running, I won't be able to assert on the condition, but I added the SyncWarning event.

zerbitx:
@sowmyav27 I believe all your comments have been resolved.

e2e-next/test_gatewayapi/test_gatewayapi_tohost.go
sowmyav27: nit: `LocalPolicyTargetReference` has no namespace field, so this is actually a same-namespace policy with a dangling Name (not a cross-ns ref). Suggest renaming `badPolicy` / `btls-cross-ns-` and the `By(...)` to something like "policy with an unresolvable local targetRef does not sync" so it does not read as a cross-ns + RG test.

sowmyav27 (dismissed):

e2e-next/test_gatewayapi/test_gatewayapi_tohost.go
```diff
@@ -0,0 +1,309 @@
...
+				Eventually(func(g Gomega) {
+					g.Expect(hostClient.Get(ctx, types.NamespacedName{Namespace: vClusterHostNS, Name: hostRouteName}, &gatewayv1.HTTPRoute{})).To(Succeed())
+				}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+			})
+			By("deleting the ReferenceGrant and expecting the host route to be removed", func() {
+				Expect(ctrlclient.IgnoreNotFound(vClusterClient.Delete(ctx, grant))).To(Succeed())
+				Eventually(func(g Gomega) {
+					current := &gatewayv1.HTTPRoute{}
+					g.Expect(vClusterClient.Get(ctx, ctrlclient.ObjectKeyFromObject(route), current)).To(Succeed())
+					if current.Annotations == nil {
```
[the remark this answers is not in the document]
zerbitx: Yeah, this was a good one.

pkg/config/gateway_enabled.go
```diff
@@ -0,0 +1,33 @@
+package config
```
janekbaraniewski: can we move this to `pkg/util/gatewayapi/...`?

pkg/mappings/resources/register.go
```diff
@@ -83,28 +84,21 @@ func isEnabled[T any](enabled bool, fn T) T {
 }
 
 func gatewayGatewaysEnabled(ctx *synccontext.RegisterContext) bool {
-	return ctx.Config.Sync.ToHost.GatewayAPI.Gateways.Enabled
```
janekbaraniewski, line 86: we don't need those helpers now right? can just call our new util functions directly

janekbaraniewski (changes requested):
LGTM, 2 small cleanup comments

sowmyav27 (approved):

janekbaraniewski (approved):

PR #3988, opened 2026-06-06
fix(snapshot): refuse snapshot creation when vCluster is not in Running state

vcluster-dv:
/kind bugfix

Resolves [ENGCP-847](https://linear.app/loft/issue/ENGCP-847/log-errors-and-do-not-attempt-to-take-snapshots-if-vcluster-not-in).

When the virtual cluster is in any state other than `Running` (paused, sleeping, scaled down, crashing, unknown), `vcluster snapshot create` and the legacy `vcluster snapshot` now fail fast with an error naming the actual state instead of attempting the snapshot:

```
cannot create snapshot because virtual cluster "my-vcluster" is in state "Paused", snapshots can only be taken when the virtual cluster is in state "Running"
```

Previously the async path silently created a snapshot request that was never reconciled (the controller lives inside the stopped virtual cluster), leaving the user with no feedback.

Implementation: a `validateVClusterIsRunning` guard in `CreateSnapshot` (`pkg/cli/snapshot_helm.go`), following the existing precedent in `pkg/cli/certs/rotate_helm.go`. Intentionally untouched: `snapshot get`/list (must keep working for non-running clusters), `restore` (pauses the virtual cluster by design), the Docker driver (manages container state itself), and the in-pod snapshot CLI (only runs when the virtual cluster is up).

Fixed an issue where `vcluster snapshot create` on a paused or otherwise non-running virtual cluster would silently create a snapshot request that never completes. The CLI now fails fast with a clear error naming the current state.

- Unit test: `TestValidateVClusterIsRunning` covers all status values plus raw pod statuses.
- E2E: new spec `should refuse to create a snapshot while the tenant cluster is paused` added to the pause/resume lifecycle chain in `e2e-next/test_core/lifecycle/test_pause_resume.go` (labeled `pr`, runs in the mandatory PR suite). Verified RED (before the guard, snapshot create exited 0 and left a hanging request) and GREEN locally; full `core` label group and `golangci-lint` are clean.
- The Platform auto-snapshots half of ENGCP-847 needs an equivalent guard in the platform repo and will be a separate PR.

### Additional test suites

```label-filter
snapshots
```


vcluster-dv:
@codex review

pkg/cli/snapshot_helm.go
```diff
@@ -34,6 +34,12 @@ func CreateSnapshot(ctx context.Context, args []string, globalFlags *flags.Globa
 		return err
 	}
 
+	// abort in case the virtual cluster has a non-running status
+	err = validateVClusterIsRunning(vCluster)
```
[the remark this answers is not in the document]
vcluster-dv: The state check ran after initSnapshotCommand, which fills storage credentials (azure.FillCredentials) before the cluster is even found, so a paused cluster with Azure storage could hit a credential error instead of the clear state error.

Fixed by moving the lookup and the running-state check to the front of initSnapshotCommand, before SetURLAndFillCredentials. It's gated by a requireRunning flag: snapshot create passes true (fails fast with the state error), while snapshot get and restore pass false so they keep working on a non-running cluster.

pkg/cli/snapshot_helm.go
```diff
@@ -113,18 +115,16 @@ func initSnapshotCommand(
 	log log.Logger,
 	credentialsRequiredInCluster bool,
 	standalone bool,
+	requireRunning bool,
```
jjaferson, line 118: Do you really need this flag? From what I see, we only want to take snapshot of a running cluster so we can add this validation only for snapshot, restore should remain the same.

We can add this validation at the `CreateSnapshot` func or even at the cli layer [here](https://github.com/loft-sh/vcluster/blob/main/cmd/vclusterctl/cmd/snapshot/create.go#L67)

pkg/cli/snapshot_helm.go
```diff
@@ -209,6 +221,23 @@ func resolveSnapshotArgs(args []string, standalone bool) (string, string, error)
 	return args[0], args[1], nil
 }
 
+// validateVClusterIsRunning aborts snapshot creation if the virtual cluster is in
+// any non-running state (paused, scaled down, sleeping, unhealthy, etc.): the async
+// snapshot request would never be reconciled, and snapshotting a stopped control
+// plane is not supported.
+func validateVClusterIsRunning(vCluster *find.VCluster) error {
+	if vCluster.Status != find.StatusRunning {
```
jjaferson, line 229: it would be good to validate if vCluster status is correctly assigned in standalone and non standalone clusters

pkg/cli/snapshot_helm.go
vcluster-dv: restore/get are unchanged (both pass false), only create passes true. It lives in initSnapshotCommand rather than CreateSnapshot because of the Codex comment above: in CreateSnapshot the check would run after SetURLAndFillCredentials, so a paused cluster with Azure storage gets a credential error instead of the clear "state Paused" one. The cli layer has the same ordering issue and would need its own find.GetVCluster. Happy to drop the flag if you'd rather pull the lookup out of initSnapshotCommand instead, just more churn.

pkg/cli/snapshot_helm.go
vcluster-dv: Makes sense, done.

pkg/cli/snapshot_helm.go
```diff
@@ -209,6 +221,23 @@ func resolveSnapshotArgs(args []string, standalone bool) (string, string, error)
 	return args[0], args[1], nil
 }
 
+// validateVClusterIsRunning aborts snapshot creation if the virtual cluster is in
+// any non-running state (paused, scaled down, sleeping, unhealthy, etc.): the async
+// snapshot request would never be reconciled, and snapshotting a stopped control
+// plane is not supported.
+func validateVClusterIsRunning(vCluster *find.VCluster) error {
```
roehrijn, line 228: The `vCluster.Status != find.StatusRunning` guard here duplicates the same inline check in `pkg/cli/certs/check_helm.go:39` and `pkg/cli/certs/rotate_helm.go:48`. Consider promoting this to a shared helper on `find.VCluster` (e.g. `IsRunning()` / `EnsureRunning()`) in `pkg/cli/find/find.go` so all three call sites share one definition and consistent error wording. The new empty-status -> Unknown normalization would then benefit the certs paths too.

roehrijn (commented):
- The description's "What else do we need to know?" section lists only the unit test `TestValidateVClusterIsRunning` and the new e2e spec. It omits a material body of work that makes up roughly half the diff: the refactor in
  `pkg/cli/find/find.go` extracting a new `standaloneStatus` helper, and an entirely new test file `pkg/cli/find/status_test.go` (`TestGetVClusterStatus` and `TestStandaloneStatus`). Three of the six commits are dedicated to this find-package
  work, yet neither title nor body mentions vCluster status detection/coverage.
- The Implementation paragraph describes the guard as a `validateVClusterIsRunning` guard in `CreateSnapshot` (`pkg/cli/snapshot_helm.go`). In the final diff the guard was relocated into `initSnapshotCommand` and gated by a new `requireRunning` parameter, so the check runs enfore storage credentials are filled. The body wasn't updated to reflect that the check now lives in the shared init path rather than in `CreateSnapshot`, and doesn't mention the `requireRunning` flag that `restore`/`get` pass as `false`.

r0bnet:
Superseded by #4080

PR #3989, opened 2026-06-07
chore: removes volume snapshot and restore code

jjaferson:
kind enhancement

resolves ENGCP-782

Removes volume snapshot and restore code

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
snapshots
```

---

> [!NOTE]
> **High Risk**
> This is a breaking removal of storage snapshot sync and PVC volume restore paths; clusters or workflows that still set the old flags or rely on CSI snapshot RBAC/sync will silently lose behavior while configs may still parse.
> 
> **Overview**
> **Removes volume snapshot and restore** as a product feature in 0.36.0: CSI `VolumeSnapshot` / `VolumeSnapshotContent` / `VolumeSnapshotClass` sync controllers, mappings, Helm RBAC, snapshot-controller deploy wiring, and the `external-snapshotter` client dependency are deleted. vCluster **etcd/API snapshots** remain, but **`--include-volumes` / `--restore-volumes`** and automatic snapshot **volume** options are dropped from CLI and restore flows.
> 
> **Config and chart compatibility:** related `values.yaml` / schema fields (`sync.toHost.volumeSnapshots`, `enableVolumeSnapshotRules`, `deploy.volumeSnapshotController`, etc.) stay in place as **deprecated no-ops** so existing configs still validate. Helm cluster/role rules no longer grant `snapshot.storage.k8s.io` permissions unless something else enables a cluster role; Gateway API sync flags were added to the RBAC “needs cluster role” conditions in their place.
> 
> **Tests and tooling:** e2e CSI hostpath/snapshot-controller setup (`Justfile`, `e2e/setup/csi.go`) and the “snapshot with volumes” suite are removed; remaining snapshot e2e exercises controller snapshots without volume CSI. Loft `api`/`agentapi` bump to `v4.11.0-alpha.2`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 36d637feec0810ebea9135ba08be06938dad44d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/snapshot/controller.go
```diff
@@ -199,29 +177,9 @@ func (c *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
-	case snapshotapi.RequestPhaseCreatingVolumeSnapshots:
-		requeueAfter, err := c.reconcileVolumeSnapshots(ctx, &configMap, snapshotRequest)
-		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to reconcile volume snapshots for snapshot request %s/%s: %w", configMap.Namespace, configMap.Name, err)
-		}
-		if requeueAfter > 0 {
-			return ctrl.Result{
-				RequeueAfter: requeueAfter,
-			}, nil
-		}
```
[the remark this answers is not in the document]
jjaferson: This will be fixed when platform PR with the snapshot types gets merged https://github.com/loft-sh/loft-enterprise/pull/7068/changes#diff-47356a5ec3edef6df2b7be814a8ac7ebdd5bce97dc5238ba7eb1b00765c1d33b

pkg/controllers/resources/persistentvolumeclaims/translate.go
```diff
@@ -34,8 +32,6 @@ func (s *persistentVolumeClaimSyncer) translate(ctx *synccontext.SyncContext, vP
 			}
 
 			switch pPVC.Spec.DataSourceRef.Kind {
-			case "VolumeSnapshot":
```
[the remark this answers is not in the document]
jjaferson: PVCs backed by VolumeSnapshot should not be translated as volume snapshot are removed from the code

cmd/vclusterctl/cmd/snapshot/create.go
```diff
@@ -74,9 +74,6 @@ vcluster snapshot create my-vcluster --driver docker
 				return fmt.Errorf("--standalone cannot be used with --driver docker")
 			}
 
-			if cmd.Snapshot.IncludeVolumes {
```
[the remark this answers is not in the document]
jjaferson: This will be fixed when platform PR with the snapshot types gets merged https://github.com/loft-sh/loft-enterprise/pull/7068/changes#diff-47356a5ec3edef6df2b7be814a8ac7ebdd5bce97dc5238ba7eb1b00765c1d33b

chart/templates/clusterrole.yaml
```diff
@@ -113,26 +113,6 @@ rules:
     resources: ["deviceclasses"]
     verbs: ["get", "watch", "list"]
   {{- end }}
-  {{- if or .Values.sync.fromHost.volumeSnapshotClasses.enabled (include "vcluster.enableVolumeSnapshotRules" .) }}
```
mfranczy, line 116: Do we need to remove this? I think we can allow sync of volume snapshot classes since it's a k8s object.

config/config.go
```diff
@@ -339,6 +339,7 @@ type Deploy struct {
 	MetricsServer DeployMetricsServer `json:"metricsServer,omitempty"`
 
 	// VolumeSnapshotController holds dedicated CSI snapshot-controller configuration.
+	// Deprecated: Removed in 0.36.0. Retained as a no-op so existing configs continue to parse.
 	VolumeSnapshotController VolumeSnapshotController `json:"volumeSnapshotController,omitempty"`
```
mfranczy, line 343: Should not we just remove it and handle this in the parser?

e2e/setup/gatewayapi.go
```diff
@@ -35,3 +37,18 @@ func GatewayAPIPreSetup() func(ctx context.Context) error {
 		return kubectlApplyWithOptions(ctx, kubeContext, []string{"--server-side", "--force-conflicts"}, crds...)
 	}
 }
+
+func kubectlApplyWithOptions(ctx context.Context, kubeContext string, options []string, files ...string) error {
```
mfranczy, line 41: Did you rebase correctly? Why some gateway  api changes are included in this PR?

chart/templates/clusterrole.yaml
jjaferson: This was added as part of supporting volume snapshot [changes](https://github.com/jjaferson/vcluster/commit/5d8e06d8bf6069dac0a0eed169c2a6427f6263b5#diff-2aec030062f9a219c6d8341048900ca70650348c9975e7377f1248639e8a6fc9R66) that is why I removed it, I am wondering what would be the gain of granting permission to sync volume snapshot class from the host to the tenant cluster?

config/config.go
jjaferson: Config is parsed with `yaml.UnmarshalStrict` [here](https://github.com/loft-sh/vcluster/blob/main/pkg/config/parse.go#L49) and every removed key will fail and crash  so I just followed the same patterns other fields were removed and deprecated by annotating it https://github.com/loft-sh/vcluster/blob/main/config/config.go#L335

e2e/setup/gatewayapi.go
jjaferson: This `kubectlApplyWithOptions` func was part of the [csi.go](https://github.com/loft-sh/vcluster/blob/main/e2e/setup/csi.go#L139) setup that was removed and as it's only used in the gatewayapi.go I've moved it here

config/config.go
jjaferson: I spoke with @sowmyav27 yesterday and we will be validating these fields at the helm chart update but to avoid any crashloopbackoff I would keep these values as deprecated for while

chart/templates/_rbac.tpl
```diff
@@ -48,26 +46,19 @@
     .Values.sync.fromHost.secrets.enabled
     .Values.integrations.istio.enabled
     .Values.sync.toHost.namespaces.enabled
-    (include "vcluster.enableVolumeSnapshotRules" .)
+    .Values.sync.fromHost.gatewayClasses.enabled
```
jjaferson, line 49: These fields were added because `.Values.rbac.enableVolumeSnapshotRules.enabled` [default value was set to auto](https://github.com/loft-sh/vcluster/pull/3989/changes#diff-a9d07052701ab3b718c82ac0fe74b965a8260b06bb5906362081718b0ee593b4L1131) which would set the `vcluster.enableVolumeSnapshotRules` to true, now that `vcluster.enableVolumeSnapshotRules` is removed we need to check whether one of these values are enabled

mfranczy (approved):

adriankabala (approved):

PR #3991, opened 2026-06-08
e2e: Gateway API P0/P1 coverage on top of #3985

sowmyav27:
/kind test

Adds e2e coverage on top of #3985 for Gateway API test plan cases that can be automated without a real Gateway controller:

- TC-02a — umbrella vs broken-out `toHost.gatewayApi` sync forms
- TC-04d — `referenceGrants.enabled: false` keeps tenant RGs tenant-local
- TC-12 — HTTPRoute with non-existent `parentRefs.name` does not sync; patches recover
- TC-34a/b — fromHost + toHost gateway name reservation (ENGNODE-556)
- TC-38/39/40/41 — `fromHost.gateways` config validation (ENGNODE-554/555)

Only TC-02a (umbrella + selective) is PR-gated. The rest run via the `gatewayapi` label. The bug fixes these specs depend on (ENGNODE-554/555/556) have landed, so all gateway API specs currently pass.

**TODO before merging**
- [x] Rebase onto `main` once #3985 lands (this PR currently includes #3985's commits)

## E2E Tests

```label-filter
gatewayapi
```


e2e-next/suite_gatewayapi_fromhostboth_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+//go:embed vcluster-gatewayapi-fromhost-both.yaml
+var gatewayAPIFromHostBothVClusterYAML string
+
+const gatewayAPIFromHostBothVClusterName = "gatewayapi-fromhostboth-vcluster"
+
+func init() { suiteGatewayAPIFromHostBothVCluster() }
+
+func suiteGatewayAPIFromHostBothVCluster() {
+	Describe("gatewayapi-fromhostboth-vcluster", labels.GatewayAPI, labels.GatewayClasses, Ordered,
```
zerbitx, line 24: Should this and the others have the PR label as well?

e2e-next/test_gatewayapi/test_gatewayapi_fromhostboth.go
```diff
@@ -0,0 +1,83 @@
...
+			vClusterHostNS string
+		)
+
+		BeforeEach(func(ctx context.Context) {
+			var err error
+			scheme := runtime.NewScheme()
+			Expect(corev1.AddToScheme(scheme)).To(Succeed())
+			Expect(gatewayv1.Install(scheme)).To(Succeed())
+
+			hostClient, err = ctrlclient.New(cluster.From(ctx, constants.GetHostClusterName()).KubernetesRestConfig(), ctrlclient.Options{Scheme: scheme})
```
zerbitx: You can use [newGatewayApiClient](https://github.com/loft-sh/vcluster/blob/main/e2e-next/test_gatewayapi/clients.go#L26) here

e2e-next/suite_gatewayapi_fromhostboth_test.go
sowmyav27: Intentional — these suites cover unmerged fixes (ENGNODE-556/554/555) and would fail on every PR, so they run via the `gatewayapi` label only. (rgdisabled stays off PR too: `selective` already PR-gates the same referenceGrants-disabled assertion.)

zerbitx (dismissed):
lgtm

e2e-next/test_gatewayapi/test_gatewayapi_invalidcfg.go
```diff
@@ -0,0 +1,137 @@
...
+	Expect(hostKubeconfig).NotTo(BeEmpty())
+
+	tmpDir, err := os.MkdirTemp("", "vcluster-invalidcfg-*")
+	Expect(err).To(Succeed())
+	DeferCleanup(func() { _ = os.RemoveAll(tmpDir) })
+
+	yamlPath := filepath.Join(tmpDir, "values.yaml")
+	Expect(os.WriteFile(yamlPath, []byte(yaml), 0o600)).To(Succeed())
+
+	vclusterBin := filepath.Join(os.Getenv("GOBIN"), "vcluster")
```
anvesh-loft, line 92: **Non Blocking :**

`vclusterBin := filepath.Join(os.Getenv("GOBIN"), "vcluster")` with no guard. When `GOBIN` is unset, `filepath.Join("", "vcluster") = "vcluster" `(PATH lookup). If the binary isn't found, `exec.Command` errors with no relevant output, causing `ContainSubstring("sync.fromHost.gateways.mappings.byName")` to  fail with a confusing `actual output: "" `which looks like a validation regression, not a missing binary.
  
  **Fix:**
```
  gobin := os.Getenv("GOBIN")
  Expect(gobin).NotTo(BeEmpty(), "GOBIN must be set to locate the vcluster binary")
  vclusterBin := filepath.Join(gobin, "vcluster")
```
  
  **Becomes blocking** if: CI runs in an environment where GOBIN is unset (fresh container), hiding real TC-38–41 regressions.

e2e-next/test_gatewayapi/test_gatewayapi_invalidcfg.go
```diff
@@ -0,0 +1,137 @@
...
+	})
+
+	createCmd := exec.CommandContext(ctx, vclusterBin,
+		"create", name,
+		"--namespace", namespace,
+		"--context", hostContext,
+		"--driver", "helm",
+		"--connect=false",
+		"--upgrade",
+		"--local-chart-dir", "../chart",
```
anvesh-loft, line 113: **Non-Blocking:** `--local-chart-dir", "../chart" `is resolved against the test process CWD. `tenant_crds.go` in the same PR correctly anchors to the repo root via runtime.Caller(0). The `../chart pat`h only works when the test binary CWD is e2e-next/ — an implicit CI convention with no in-code anchor.

  **Fix: Mirror the tenant_crds.go pattern:**
```
  _, file, _, ok := runtime.Caller(0)
  Expect(ok).To(BeTrue())
  chartDir := filepath.Clean(filepath.Join(filepath.Dir(file), "..", "..", "..", "chart"))
  // pass: "--local-chart-dir", chartDir
```

e2e-next/test_gatewayapi/test_gatewayapi_invalidcfg.go
```diff
@@ -0,0 +1,137 @@
...
+var invalidCfgTC41OverrideOutsideMappingsYAML string
+
+func GatewayAPIInvalidConfigSpec() {
+	Describe("Gateway API invalid fromHost.gateways config", labels.GatewayAPI, labels.GatewayClasses, func() {
+		It("rejects empty fromHost.gateways.mappings.byName at deploy time (TC-38)", func(ctx context.Context) {
+			expectVClusterStartupError(ctx, "tc38", invalidCfgTC38EmptyMappingsYAML, "sync.fromHost.gateways.mappings.byName")
+		})
+
+		It("rejects invalid mapping keys at deploy time (TC-39)", func(ctx context.Context) {
+			for _, variant := range []struct {
```
anvesh-loft: **Non-Blocking :** **TC-39 (3 variants)** and **TC-40 (2 variants)** loop over `[]struct{...}` in a single It block. A Expect failure inside any By(variant.slug, ...) closure panics and unwinds the entire It — the remaining variants in the loop are silently skipped and never reported.

  **Fix:** Use DescribeTable/Entry so each variant gets its own independently-reported spec node. Apply to both TC-39 and TC-40.

 **Becomes blocking** if: A validation bug causes one variant to fail while other variants were also broken, the regression lands silently.

anvesh-loft (commented):
Hi @sowmyav27 , looks good to me !! But added a few **non-blocking** comments to review

e2e-next/test_gatewayapi/test_gatewayapi_invalidcfg.go
sowmyav27: Thanks! Going to leave this as-is for consistency: the unguarded `filepath.Join(os.Getenv("GOBIN"), "vcluster")` is the established pattern in the same test tree — `test_core/lifecycle/helpers.go:22`, `clusters/registry.go:36`, `test_migration/k3s_to_k8s.go:167`. `registry.go` uses it to provision every e2e-next cluster, so an unset `GOBIN` fails the whole suite at setup, well before these specs run — it cant silently mask a TC-38–41 regression. Adding an `Expect` here would diverge from those call sites.

e2e-next/test_gatewayapi/test_gatewayapi_invalidcfg.go
sowmyav27: Good eye, but Im keeping `"../chart"` to match the existing convention for this flag: `test_core/lifecycle/helpers.go:26` (`localChartDir() => filepath.Join("..", "chart")`) is CWD-relative the same way. The `runtime.Caller` anchor in `tenant_crds.go` is for locating repo-root CRD files, a different helper. The `../chart`-from-`e2e-next/` assumption is already baked into the lifecycle tests, so this stays consistent rather than introducing a second approach.

e2e-next/test_gatewayapi/test_gatewayapi_invalidcfg.go
sowmyav27: Agreed — fixed in 952432dcc. Rather than `DescribeTable` (not used anywhere in e2e-next), I split TC-39 (3) and TC-40 (2) into separate `It` blocks, matching the one-`It`-per-case house style. Each variant is now its own spec node, so a failure no longer aborts the `It` and skips the siblings.

e2e-next/test_gatewayapi/test_gatewayapi_invalidcfg.go
anvesh-loft: Okay make sense

anvesh-loft (approved):
Looks good !

PR #3994, opened 2026-06-10
fix: install tenant referencegrant crd whenever route sync is enabled

zerbitx:
- route controllers watch virtual ReferenceGrants for cross-namespace authorization regardless of sync.toHost.gatewayApi.referenceGrants.enabled; with the flag "false" the CRD was never installed and the watch failed forever (no matches for kind "ReferenceGrant"), silently blocking all HTTPRoute/TLSRoute sync
  - extract EnsureReferenceGrantCRD and call it from the HTTPRoute and TLSRoute mappers, independent of grant sync
  - keep "false" semantics: no host discovery check, no mapper, no syncer, grants never sync to the host; virtual grants stay authoritative for cross-namespace refs in single-namespace mode
  - add gatewayapi-grants-disabled e2e suite: same-namespace route syncs, cross-namespace backendRef denied until a virtual ReferenceGrant permits it, grant itself never syncs to the host
  - add unit test asserting route mappers ensure the grant CRD when grant sync is disabled

/kind bugfix

resolves #ENGNODE-566

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

### DIFF
--- a/e2e-next/test_gatewayapi/test_gatewayapi_grants_disabled.go
+++ b/e2e-next/test_gatewayapi/test_gatewayapi_grants_disabled.go
@@ -1,0 +1,154 @@
+package test_gatewayapi
+
+import (
+	"context"
+
+	"github.com/loft-sh/vcluster/e2e-next/constants"
+	"github.com/loft-sh/vcluster/e2e-next/labels"
+	"github.com/loft-sh/vcluster/pkg/util/random"
+	"github.com/loft-sh/vcluster/pkg/util/translate"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+const grantsDisabledGatewayClassSelectorValue = "gatewayapi-grants-disabled-vcluster"
+
+// GatewayAPIGrantsDisabledSpec registers route sync tests for a vCluster with
+// sync.toHost.gatewayApi.referenceGrants.enabled set to "false" while route
+// sync stays on. Disabling grant sync must not break the route controllers and
+// must keep virtual ReferenceGrants authoritative for cross-namespace refs.
+func GatewayAPIGrantsDisabledSpec() {
+	Describe("Gateway API toHost with ReferenceGrant sync disabled", labels.GatewayAPI, func() {
+		var (
+			hostClient     ctrlclient.Client
+			vClusterClient ctrlclient.Client
+			vClusterName   string
+			vClusterHostNS string
+		)
+
+		BeforeEach(func(ctx context.Context) {
+			clients := newGatewayAPIClients(ctx, true)
+			hostClient = clients.HostClient
+			vClusterClient = clients.VClusterClient
+			vClusterName = clients.VClusterName
+			vClusterHostNS = clients.VClusterHostNS
+
+			Eventually(func(g Gomega) {
+				g.Expect(vClusterClient.List(ctx, &gatewayv1.HTTPRouteList{}, ctrlclient.InNamespace("default"))).To(Succeed())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+		})
+
+		It("syncs HTTPRoutes to the host while ReferenceGrant sync is disabled", func(ctx context.Context) {
+			suffix := random.String(6)
+			class := createGatewayClass(ctx, hostClient, "gc-rgd-"+suffix, grantsDisabledGatewayClassSelectorValue, "grants disabled class")
+			Eventually(func(g Gomega) {
+				g.Expect(vClusterClient.Get(ctx, types.NamespacedName{Name: class.Name}, &gatewayv1.GatewayClass{})).To(Succeed())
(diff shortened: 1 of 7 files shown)


zerbitx:
@codex review

e2e-next/test_gatewayapi/test_gatewayapi_grants_disabled.go
```diff
@@ -0,0 +1,154 @@
...
+		BeforeEach(func(ctx context.Context) {
+			clients := newGatewayAPIClients(ctx, true)
+			hostClient = clients.HostClient
+			vClusterClient = clients.VClusterClient
+			vClusterName = clients.VClusterName
+			vClusterHostNS = clients.VClusterHostNS
+
+			Eventually(func(g Gomega) {
+				g.Expect(vClusterClient.List(ctx, &gatewayv1.HTTPRouteList{}, ctrlclient.InNamespace("default"))).To(Succeed())
+			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
```
adriankabala, line 45: nit: this readiness gate only Lists HTTPRoutes, whose CRD exists even under the bug, so it does not prove the actually-affected subsystem (the tenant ReferenceGrant watch) is healthy. The suite proves a route eventually syncs, but nothing pins that the fix installed the tenant ReferenceGrant CRD. Consider gating on the ReferenceGrant CRD directly here so every spec pins the fix:

```go
// The fix installs the tenant ReferenceGrant CRD even with grant sync off;
// gate on it directly so the suite pins the actually-affected subsystem.
Eventually(func(g Gomega) {
    g.Expect(vClusterClient.List(ctx, &gatewayv1beta1.ReferenceGrantList{}, ctrlclient.InNamespace("default"))).To(Succeed())
}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
```

The cross-namespace spec already does this `List(ReferenceGrantList)` inline; lifting it to BeforeEach covers the whole suite. Non-blocking.

pkg/mappings/resources/register_gateway_test.go
```diff
@@ -45,6 +49,45 @@ func TestReferenceGrantMapperChecksHostCRDWithoutNamespaceSync(t *testing.T) {
...
+	ctx.Config.Sync.ToHost.GatewayAPI.TLSRoutes.Enabled = true
+
+	if _, err := CreateHTTPRouteMapper(ctx); err != nil {
+		t.Fatalf("create HTTPRoute mapper: %v", err)
+	}
+	if _, err := CreateTLSRouteMapper(ctx); err != nil {
+		t.Fatalf("create TLSRoute mapper: %v", err)
+	}
+
+	if !ensured[mappings.ReferenceGrants()] {
```
adriankabala, line 86: nit: this asserts only the final state of a single `ensured` map shared across both `CreateHTTPRouteMapper` and `CreateTLSRouteMapper`, so dropping the `EnsureReferenceGrantCRD` call from `tlsroutes.go` alone would still pass (the HTTPRoute mapper already set the boolean). Since TLSRoute has no e2e coverage, this is the only TLSRoute guard. Reset the map between the two calls and assert each independently:

```go
if _, err := CreateHTTPRouteMapper(ctx); err != nil {
    t.Fatalf("create HTTPRoute mapper: %v", err)
}
if !ensured[mappings.ReferenceGrants()] {
    t.Fatalf("HTTPRoute mapper must ensure the tenant ReferenceGrant CRD with grant sync disabled")
}

ensured = map[schema.GroupVersionKind]bool{}
if _, err := CreateTLSRouteMapper(ctx); err != nil {
    t.Fatalf("create TLSRoute mapper: %v", err)
}
if !ensured[mappings.ReferenceGrants()] {
    t.Fatalf("TLSRoute mapper must ensure the tenant ReferenceGrant CRD with grant sync disabled")
}
```

Non-blocking.

pkg/mappings/resources/httproutes.go
```diff
@@ -25,5 +25,10 @@ func CreateHTTPRouteMapper(ctx *synccontext.RegisterContext) (synccontext.Mapper
 		return nil, err
 	}
 
+	err = EnsureReferenceGrantCRD(ctx)
```
adriankabala, line 28: nit: the key improvement this fix introduces is converting silent retry-forever into a loud registration failure (`return nil, err` here and in `tlsroutes.go`). That error path is untested - the unit test stubs `util.EnsureCRD` to always succeed. If a future change swallowed the error, the original silent-degradation bug returns unnoticed. Consider a small subtest pinning the contract:

```go
func TestRouteMappersPropagateReferenceGrantCRDError(t *testing.T) {
    wantErr := errors.New("ensure crd failed")
    restore := util.EnsureCRD
    util.EnsureCRD = func(context.Context, *rest.Config, []byte, schema.GroupVersionKind) error { return wantErr }
    restoreKind := util.KindExists
    util.KindExists = func(*rest.Config, schema.GroupVersionKind) (bool, error) { return true, nil }
    t.Cleanup(func() { util.EnsureCRD = restore; util.KindExists = restoreKind })
    // ... same grant-sync-disabled RegisterContext fixture as the sibling test
    if _, err := CreateHTTPRouteMapper(ctx); !errors.Is(err, wantErr) {
        t.Fatalf("HTTPRoute mapper must propagate the CRD-ensure error, got: %v", err)
    }
    if _, err := CreateTLSRouteMapper(ctx); !errors.Is(err, wantErr) {
        t.Fatalf("TLSRoute mapper must propagate the CRD-ensure error, got: %v", err)
    }
}
```

Non-blocking.

adriankabala (approved):
LGTM. Lint clean (e2e-next + pkg/mappings/resources), and the P0 regression - HTTPRoute sync with `referenceGrants.enabled: false` - is exercised directly against the failing config. The fix is correct and minimal.

Left 3 non-blocking `nit:` comments (test-robustness only): pin the regression mechanism via a ReferenceGrant CRD readiness gate, isolate the per-mapper unit assertion so TLSRoute is actually guarded, and cover the loud-fail-at-startup error path. None block merge.

janekbaraniewski:
@codex review please

zerbitx:
Combining with #3996

PR #3999, opened 2026-06-11
fix: Keep the GatewayCRD valid during translation when sanitization is on

zerbitx:
/kind bugfix

resolves #ENGNODE-571

### Problem: 
With sanitize.certificateRefs: true (the default), the import syncer stripped only certificateRefs from the mirror's listeners and left the rest of theTLS block intact. For an HTTPS listener with tls.mode: Terminate (also the CRD default when mode is omitted), this violates the Gateway API CEL rule "certificateRefs or options must be specified when mode is Terminate". The tenant API server rejected the mirror on every reconcile, so the imported Gateway never appeared in the tenant cluster and any HTTPRoute parented to it could never attach. Switching the mode is not an option either, since a second CEL rule requires HTTPS listeners to keep mode Terminate.

### Fix: 

When sanitizing strips the refs from a listener whose effective mode is Terminate and which has no tls.options, the syncer now adds a marker option so the
  mirror stays CRD-valid:

```yaml
  tls:
    mode: Terminate
    options:
      vcluster.loft.sh/certificate-refs-sanitized: "true"
```

  This keeps the https listener in the mirror (so tenant routes can still reference it by sectionName) and self-documents why the certs are absent. Host-provided tls.options already satisfy the rule and are preserved unchanged; Passthrough listeners need no marker.

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

### DIFF
--- a/e2e-next/test_gatewayapi/test_gatewayapi_import.go
+++ b/e2e-next/test_gatewayapi/test_gatewayapi_import.go
@@ -98,6 +98,34 @@ func GatewayAPIImportSpec() {
 			}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
 		})
 
+		It("imports a TLS-terminating host Gateway with certificateRefs sanitized and the mirror kept valid", func(ctx context.Context) {
+			suffix := random.String(6)
+			class := createGatewayClass(ctx, hostClient, "gwc-tls-"+suffix, importClassSelectorValue, "tls class")
+			hostGW := hostTLSGateway("gwapi-host", "tls-edge-"+suffix, class.Name, "edge-cert-"+suffix)
+
+			By("creating a TLS-terminating host Gateway in the wildcard-mapped source namespace", func() {
+				createHostGateway(ctx, hostClient, hostGW)
+			})
+			By("verifying the tenant mirror keeps the https listener with sanitized certificateRefs", func() {
+				Eventually(func(g Gomega) {
+					mirror := &gatewayv1.Gateway{}
+					g.Expect(vClusterClient.Get(ctx, types.NamespacedName{Namespace: "gwapi-import", Name: hostGW.Name}, mirror)).To(Succeed())
+					g.Expect(mirror.Spec.Listeners).To(HaveLen(1))
+					listener := mirror.Spec.Listeners[0]
+					g.Expect(listener.Protocol).To(Equal(gatewayv1.HTTPSProtocolType))
+					g.Expect(listener.TLS).NotTo(BeNil(), "sanitizing certificateRefs must not drop the TLS config")
+					g.Expect(listener.TLS.CertificateRefs).To(BeEmpty(), "host certificateRefs must not leak into the tenant mirror")
+					g.Expect(listener.TLS.Options).To(HaveKeyWithValue(
+						gatewayv1.AnnotationKey(gateways.SanitizedCertificateRefsTLSOption),
+						gatewayv1.AnnotationValue("true"),
+					), "sanitized Terminate listener needs the marker option to stay CRD-valid")
+					if listener.TLS.Mode != nil {
+						g.Expect(*listener.TLS.Mode).To(Equal(gatewayv1.TLSModeTerminate))
+					}
+				}).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeout).Should(Succeed())
+			})
+		})
+
 		It("does not sync routes that attach to a tenant-local non-imported Gateway", func(ctx context.Context) {
 			suffix := random.String(6)
 			class := createGatewayClass(ctx, hostClient, "gwc-local-"+suffix, importClassSelectorValue, "local class")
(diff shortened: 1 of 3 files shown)


janekbaraniewski:
@codex review please

zerbitx:
@codex review

janekbaraniewski (dismissed):

e2e-next/test_gatewayapi/test_gatewayapi_import.go
```diff
@@ -98,6 +98,34 @@ func GatewayAPIImportSpec() {
...
+					g.Expect(mirror.Spec.Listeners).To(HaveLen(1))
+					listener := mirror.Spec.Listeners[0]
+					g.Expect(listener.Protocol).To(Equal(gatewayv1.HTTPSProtocolType))
+					g.Expect(listener.TLS).NotTo(BeNil(), "sanitizing certificateRefs must not drop the TLS config")
+					g.Expect(listener.TLS.CertificateRefs).To(BeEmpty(), "host certificateRefs must not leak into the tenant mirror")
+					g.Expect(listener.TLS.Options).To(HaveKeyWithValue(
+						gatewayv1.AnnotationKey(gateways.SanitizedCertificateRefsTLSOption),
+						gatewayv1.AnnotationValue("true"),
+					), "sanitized Terminate listener needs the marker option to stay CRD-valid")
+					if listener.TLS.Mode != nil {
```
sowmyav27: The Mode assertion no-ops when `Mode` is nil, and `hostTLSGateway` never sets Mode (it's omitted), so this relies on API-server defaulting and may never actually assert the Terminate path the bug is about. Suggest setting `Mode: ptr.To(gatewayv1.TLSModeTerminate)` in the fixture and asserting it unconditionally.

e2e-next/test_gatewayapi/test_gatewayapi_import.go
```diff
@@ -98,6 +98,34 @@ func GatewayAPIImportSpec() {
...
+			By("verifying the tenant mirror keeps the https listener with sanitized certificateRefs", func() {
+				Eventually(func(g Gomega) {
+					mirror := &gatewayv1.Gateway{}
+					g.Expect(vClusterClient.Get(ctx, types.NamespacedName{Namespace: "gwapi-import", Name: hostGW.Name}, mirror)).To(Succeed())
+					g.Expect(mirror.Spec.Listeners).To(HaveLen(1))
+					listener := mirror.Spec.Listeners[0]
+					g.Expect(listener.Protocol).To(Equal(gatewayv1.HTTPSProtocolType))
+					g.Expect(listener.TLS).NotTo(BeNil(), "sanitizing certificateRefs must not drop the TLS config")
+					g.Expect(listener.TLS.CertificateRefs).To(BeEmpty(), "host certificateRefs must not leak into the tenant mirror")
+					g.Expect(listener.TLS.Options).To(HaveKeyWithValue(
```
sowmyav27, line 118: Optional coverage: the test confirms the mirror exists with the marker but not that the tenant API server actually accepts it. Consider asserting an `Accepted=True` status condition to bind directly to the reported symptom.

pkg/controllers/resources/gateways/syncer_test.go
```diff
@@ -55,6 +58,86 @@ func TestGatewaySpecToVirtualSanitizesTLSInfrastructureAndAppliesVirtualAllowedR
...
+
+			spec := gatewaySpecToVirtual(ctx, host)
+			got := spec.Listeners[0].TLS
+			if got == nil {
+				t.Fatalf("expected TLS config to be preserved")
+			}
+			if len(got.CertificateRefs) != tt.wantRefs {
+				t.Fatalf("expected %d certificateRefs, got %#v", tt.wantRefs, got.CertificateRefs)
+			}
+			if len(got.Options) != len(tt.wantOptions) {
```
sowmyav27, line 142: The table never asserts `TLS.Mode` is preserved (incl. the passthrough case). A regression that mutates Mode while adding the marker would still pass. Consider a `wantMode` column asserted per case.

pkg/controllers/resources/gateways/syncer_test.go
```diff
@@ -55,6 +58,86 @@ func TestGatewaySpecToVirtualSanitizesTLSInfrastructureAndAppliesVirtualAllowedR
...
+			wantRefs:    1,
+			wantOptions: nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			vcConfig := &pkgconfig.VirtualClusterConfig{}
+			vcConfig.Sync.FromHost.Gateways.Sanitize.CertificateRefs = tt.sanitize
+			ctx := &synccontext.SyncContext{Context: context.Background(), Config: vcConfig}
+			host := &gatewayv1.Gateway{Spec: gatewayv1.GatewaySpec{Listeners: []gatewayv1.Listener{{
```
sowmyav27, line 120: Optional: the fix loops `for i := range spec.Listeners`, but every fixture (here and in the e2e test) has exactly one listener, so the loop is only exercised at index 0. A multi-listener case (e.g. Terminate + passthrough + HTTP) would guard against the marker landing on the wrong listener. Minor hardening, not blocking.

janekbaraniewski (approved):

sowmyav27 (approved):

PR #4019, opened 2026-06-16
feat(standalone): add controlPlane.standalone.advertiseAddress field

rlmcpherson:
/kind feature

Adds a `controlPlane.standalone.advertiseAddress` config field that pins the IP the standalone control plane advertises to the `kubernetes` and `konnectivity` Services.

Today this is only controllable via the `VCLUSTER_STANDALONE_IP_ADDRESS` env var (set indirectly by `install-standalone.sh --node-ip`). On multi-homed hosts (cloud VMs with both a public and a private NIC), `ChooseHostInterface()` selects the default-route IP — typically the public address, which is not reachable from the pod network — so the `kubernetes`/`konnectivity` EndpointSlices end up with unreachable IPs. This exposes a discoverable, first-class config field so operators can pin the in-cluster-reachable address without systemd drop-ins.

The vcluster-pro standalone code consumes it (prefer the config field, then the env var, then default-route detection) in a follow-up PR.

Added `controlPlane.standalone.advertiseAddress` to pin the IP the standalone control plane advertises to the kubernetes/konnectivity Services (useful on multi-homed hosts).

Refs ENGCP-733

### DIFF
--- a/chart/values.schema.json
+++ b/chart/values.schema.json
@@ -4857,6 +4857,10 @@
           "type": "string",
           "description": "DataDir defines the data directory for the standalone mode."
         },
+        "advertiseAddress": {
+          "type": "string",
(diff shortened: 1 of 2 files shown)


lizardruss (approved):
lgtm

PR #4033, opened 2026-06-24
fix(e2e): deflake sync error event sanitisation test

adriankabala:
/kind test

resolves #

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
events
```

---

> [!NOTE]
> **Low Risk**
> Test-only changes in e2e; no production sync or event-sanitisation logic is modified.
> 
> **Overview**
> The **SyncError event sanitisation** e2e no longer relies on a fixed burst of 20 host/virtual pod updates with `RetryOnConflict`. It now runs two background mutators in a tight loop until verification finishes, so contention with the pod syncer stays open while `Eventually` waits for a `SyncError` event.
> 
> A shared `stop` channel and `runMutator` helper coordinate shutdown; **`DeferCleanup`** closes `stop` and **`wg.Wait()`** before pod/namespace teardown so goroutines do not race cleanup. Raw updates replace conflict retries because failed updates from syncer contention are the desired signal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dec1afcc9224387748b79233978abc01946437ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/e2e/test_core/sync/test_sync_error_events.go
+++ b/e2e/test_core/sync/test_sync_error_events.go
@@ -116,14 +116,12 @@ func SyncErrorSanitisationSpec() {
 				// event (pkg/controllers/resources/pods/syncer.go) before the conflict is
 				// silently requeued by the outer controller.
 				//
-				// Both goroutines keep mutating for the entire verification window below,
-				// rather than running a fixed number of iterations up front. A single short
-				// burst can finish before the syncer ever loses the race, after which the
-				// syncer settles and no further conflicts occur — that is what made this test
-				// flaky. Keeping the race window open until a SyncError event is observed
-				// makes the conflict reliable. Errors are intentionally ignored: an update
-				// that fails because the syncer just changed the pod is exactly the
-				// contention we want.
+				// Both goroutines mutate continuously until the verification step closes
+				// `stop` via DeferCleanup, keeping the race window open until a SyncError
+				// event is observed. A bounded burst can complete before the syncer ever
+				// loses the race, leaving no contention to verify. Errors are intentionally
+				// ignored: an update that fails because the syncer just changed the pod is
+				// exactly the contention we want.
 				stop := make(chan struct{})
 				var wg sync.WaitGroup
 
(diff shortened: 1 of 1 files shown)


e2e/test_core/sync/test_sync_error_events.go
```diff
@@ -109,59 +108,72 @@ func SyncErrorSanitisationSpec() {
...
-					var wg sync.WaitGroup
-					wg.Add(2)
+				// Both goroutines keep mutating for the entire verification window below,
+				// rather than running a fixed number of iterations up front. A single short
+				// burst can finish before the syncer ever loses the race, after which the
+				// syncer settles and no further conflicts occur — that is what made this test
+				// flaky. Keeping the race window open until a SyncError event is observed
+				// makes the conflict reliable. Errors are intentionally ignored: an update
+				// that fails because the syncer just changed the pod is exactly the
+				// contention we want.
```
sowmyav27: **Comment block narrates change history**

**What's the concern?**
- "rather than running a fixed number of iterations up front" and "that is what made this test flaky" both reference the old fixed-burst implementation that no longer exists in the file.
- A reader landing on this code only sees continuous mutation; the comparison to "the previous approach" has nothing to anchor against. Change context belongs in the commit message / PR description, not the source file.

**Why does it matter?**
- Comments that reference removed code rot — the next reader will spend cycles trying to find the "fixed number of iterations" being contrasted against.
- The preceding paragraph (lines 111–117) already explains the actual race mechanic at the API-server layer, which is the load-bearing rationale; the historical narration is additive noise.

**Proposed fix — pick one**

**Option A — rewrite to describe only the current design:**

```go
// Both goroutines mutate continuously until the verification step closes
// `stop` via DeferCleanup, keeping the race window open until a SyncError
// event is observed. A bounded burst can complete before the syncer ever
// loses the race, leaving no contention to verify. Errors are intentionally
// ignored: an update that fails because the syncer just changed the pod is
// exactly the contention we want.
```

**Option B — remove the entire block (lines 119–126).**

The preceding paragraph (lines 111–117) already explains the API-server race mechanic, and the `By("continuously racing host and virtual pod updates …")` label below tells the reader what the goroutines do. The error-swallowing is self-evident from `_ = bump(i)`. Inline rationale isn't strictly needed here — the *why* lives in the PR description.

adriankabala: Good call, went with Option A.

sowmyav27 (approved):

Issue #48, opened 2021-06-15
feat: context flag on vcluster


matdehaast:
It would be nice to have the ability to specify the kube-context to use when calling vcluster. Currently the default one is used so you have to ensure you are setup correctly before calling vcluster.

Example
```
vcluster --context my-cluster connect test --namespace test
```

would use the my-cluster context from `kubectl config get-contexts`

FabianKramm:
@matdehaast thanks for creating this issue! Thats a good idea and we will add that to the `vcluster create/connect/delete` commands.

FabianKramm:
@matdehaast this is working now with v0.3.0-beta.3

PR #4093, opened 2026-07-17
feat(e2e): consume shared specpriority scheduling fixture

roehrijn:
/kind test
/kind enhancement

Makes vcluster consume the shared `specpriority` scheduling fixture from
`e2e-framework` (`github.com/loft-sh/e2e-framework/pkg/e2e/specpriority`).

- `e2e-next/init/init.go` registers `specpriority.Transformer` as an additional
  tree-construction node-args transformer, alongside the existing
  `suite.NodeTransformer` and `e2e.ContextualNodeTransformer`.
- Adds an inline **Download spec duration stats** step to both ginkgo workflows
  (`.github/workflows/e2e-ginkgo.yaml` PR suite and
  `.github/workflows/e2e-ginkgo-nightly.yaml`). The step downloads the recorded
  spec-duration stats from `gs://${{ vars.E2E_INSIGHTS_BUCKET }}/stats/specs/v1/latest/<slug>.json.gz`,
  gunzips it, and writes `E2E_SPEC_STATS_FILE` and `E2E_SPEC_SPREAD_SEED` into
  `$GITHUB_ENV` so the ginkgo test process (in the same job) picks them up at
  runtime. It is `continue-on-error: true`, so it is a safe no-op when stats are
  unavailable.

> **Intermediate state:** `e2e-framework` is pinned to the framework PR branch
> `janroehrich/spec-priority-e2e-framework` (framework PR loft-sh/e2e-framework#48).
> Once #48 merges to `e2e-framework` `main`, a follow-up commit will move the pin
> from the branch pseudo-version back to `main`.

N/A (internal e2e tooling only).

Part of a coordinated multi-repo migration onto the shared specpriority fixture.

### Default Test Execution
The mandatory PR suite runs automatically.

```label-filter
none
```

### DIFF
--- a/.github/workflows/e2e-ginkgo-nightly.yaml
+++ b/.github/workflows/e2e-ginkgo-nightly.yaml
@@ -106,6 +106,24 @@ jobs:
         if: steps.gcp-auth.outcome == 'success'
         uses: google-github-actions/setup-gcloud@v2
 
+      - name: Download spec duration stats
+        continue-on-error: true
+        env:
+          E2E_INSIGHTS_BUCKET: ${{ vars.E2E_INSIGHTS_BUCKET }}
+        run: |
+          slug="${GITHUB_REPOSITORY//\//-}"
+          src="gs://${E2E_INSIGHTS_BUCKET}/stats/specs/v1/latest/${slug}.json.gz"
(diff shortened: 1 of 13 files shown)


roehrijn:
Closing: loft-sh/vcluster is being consolidated today; specpriority is delivered via e2e-framework and the vcluster-pro/loft-enterprise consumers instead.

PR #4097, opened 2026-07-20
fix(snapshot): delete existing data before restore for external database

r0bnet:
/kind bugfix

resolves #ENGCP-557

Clears the existing data before restoring a snapshot into an **external database** backing store.

Before a restore, vCluster clears the target store so that resources created *after* the snapshot was taken do not survive the restore. This worked for every backing store except external database, which was skipped entirely, so snapshot keys were replayed on top of stale data and post-snapshot resources survived the restore.

An external database is reached through a local kine process that speaks the etcd v3 protocol, so the same `etcdClient.DeletePrefix(ctx, "/")` already used for deployed/external etcd works against it. No file deletion or revision bump is needed, kine's delete operations advance the revision monotonically just like the deployed/external etcd branch.

- Add a `rangeDeleteBeforeRestore` helper (deployed etcd, external etcd, and now external database) and use it at the pre-restore delete site in `newRestoreEtcdClient`.
- Add `TestRangeDeleteBeforeRestore` covering all five store types.

Scope: only the deletion aspect. Backup & rollback for external database is intentionally out of scope (deployed/external etcd also lack it today) and tracked as follow-up.

Reported via Pylon #8006.

Closes ENGCP-557

Fixed an issue where restoring a snapshot into a vCluster backed by an external database did not delete the existing data first, leaving stale resources behind after the restore.

External database is a Pro-gated feature, so this path only executes in a Pro build against a real database. OSS behavior is unchanged.

### Default Test Execution

### Additional test suites

```label-filter
none
```

---

> [!NOTE]
> **Medium Risk**
> Pre-restore wipes all keys in external DB/etcd-backed stores; incorrect classification could delete the wrong store or skip cleanup, but scope is limited to snapshot restore and matches existing etcd behavior.
> 
> **Overview**
> Fixes snapshot restore for **external database** backing stores by running the same pre-restore **`DeletePrefix("/")`** path already used for deployed and external etcd. External DB is reached via kine’s etcd API, so stale keys are removed before snapshot keys are replayed instead of being layered on top.
> 
> The restore logic is refactored behind **`shouldDeleteBeforeRestore`**, which returns true for deployed etcd, external etcd, and external database (false for embedded etcd/SQLite, which still clear local files). Comments and log text are generalized accordingly. **`TestShouldDeleteBeforeRestore`** covers all five store types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2191490c25ccbf19022638e0c1d99c7c2f6de461. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/snapshot/restoreclient.go
+++ b/pkg/snapshot/restoreclient.go
@@ -546,7 +546,8 @@ func newRestoreEtcdClient(ctx context.Context, vConfig *config.VirtualClusterCon
 	// * embedded etcd: delete the files locally and make sure revision is not decreasing, this is important as otherwise watches will not work correctly
 	// * deploy etcd: range delete request
 	// * embedded database: delete the files locally and make sure revision is not decreasing, this is important as otherwise watches will not work correctly
-	// * external database: we can't so we skip and then check later if there are any already
+	// * external etcd: range delete request
+	// * external database: range delete request (reached through kine, which handles delete-by-prefix)
 	if vConfig.BackingStoreType() == vclusterconfig.StoreTypeEmbeddedDatabase {
 		// get latest revision from database
 		latestRevision, err := getLatestRevisionSQLite(ctx, constants.K8sSqliteDatabase)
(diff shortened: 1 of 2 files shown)


pkg/snapshot/restoreclient.go
```diff
@@ -633,6 +634,22 @@ func newRestoreEtcdClient(ctx context.Context, vConfig *config.VirtualClusterCon
 	return etcdClient, revertBackup, nil
 }
 
+// rangeDeleteBeforeRestore reports whether the backing store's existing data
+// must be cleared with an etcd range delete before restore (as opposed to
+// deleting local files). True for stores vCluster reaches over the etcd
+// protocol but does not own on disk: deployed etcd, external etcd, and
+// external database (via kine).
+func rangeDeleteBeforeRestore(storeType vclusterconfig.StoreType) bool {
```
cbalan: nit: naming wise, what do you think about using `shouldDeleteBeforeRestore` or similar.

r0bnet: Ack.

cbalan (dismissed):

mfranczy (approved):

mfranczy:
> Scope: only the deletion aspect. Backup & rollback for external database is intentionally out of scope (deployed/external etcd also lack it today) and tracked as follow-up

Could you update our doc? Please add a warning there that a failed restore may permanently damaged the database as there is no rollback.

PR #4101, opened 2026-07-20
fix(ci): route platform version bumps to coupled release branch

Piotr1215:
/kind bugfix
/kind enhancement

Closes DEVOPS-1108

Reworks `.github/workflows/update-platform-minimum-version.yaml`. Today every platform release, including old ones, opens its `MinimumVersionTag` bump PR against `main`. Each platform minor is coupled 1:1 with a vCluster release branch (platform v4.10.x lives on `v0.35`, v4.11.x on `v0.36`), so these PRs land on the wrong branch, pile up unmerged, and cannot target the branch that carries the platform version.

### What changed

1. Routes each bump to its coupled release branch. The branch is derived from the tag by matching each `v0.*` branch's committed `MinimumVersionTag` major.minor, so there is no hardcoded mapping table and it self-corrects as branches move. `main` is excluded, so older platform lines stop targeting `main`. Zero or multiple matches fail the run loudly instead of mis-routing.
2. Adds an explicit `branch` target (workflow_dispatch input or `client_payload.branch`) that skips derivation and pins the base directly. This covers cutting a platform and vCluster release together, where the branch does not yet carry the tag.
3. Closes the now-stale earlier bump PRs on the same branch when a newer platform tag supersedes them. The incoming tag's own PR and any newer PR are left untouched.

### How

Decision logic moves out of inline YAML into a sourced, unit-tested script (`.github/scripts/platform-version-bump.sh`) with a bash test (`.github/tests/test-platform-version-bump.sh`) wired into a `pull_request` job, matching the existing `brew-tap-drift` convention. The workflow splits into a `decide` job (reads on `main`, picks target + skip + stale-PR list) and an `apply` job (checks out the target branch, edits `version.go`, opens the PR with `base` set to the coupled branch, closes stale PRs).

This also fixes a latent bug in the old inline compare: `"${TAG//-/\~}"` embedded a literal backslash, so the `sort -V` pre-release normalization never applied and rc tags compared incorrectly. `version_gt` uses `tr` instead, and the unit tests lock in pre-release ordering.

The orphaned `test-version-comparison.sh` (never wired into CI) is removed; its coverage is superseded by the new suite.

### Note on existing stale PRs

The 5 bump PRs currently open against `main` (v4.11.0, three v4.11.0 rc tags, v4.10.5) were created under the old main-only behavior. Going forward nothing re-targets `main`, so the new stale-close logic will not touch them. They are a one-time cleanup, handled separately from this change.

N/A: internal CI workflow change, not user-facing.

The `test` job runs the new unit suite on this PR (22 cases: `platform_minor`, `version_gt`, branch derivation, stale-PR selection). Verified locally against the live repo: v4.10.x derives to `v0.35`, v4.11.0 skips as equal on `v0.36`, and an explicit `main` target flags only the strictly-older PR as stale while keeping newer ones.

### Additional test suites

```label-filter
none
```

---

> [!NOTE]
> **Medium Risk**
> Changes release automation that opens/closes PRs and writes `MinimumVersionTag` on release branches; mis-routing would affect dependency pins, though ambiguous/unknown mappings fail loudly and logic is unit-tested.
> 
> **Overview**
> **Platform `MinimumVersionTag` bumps no longer always open PRs against `main`.** The workflow derives the target `v0.*` release branch by matching each branch’s committed `MinimumVersionTag` major.minor to the incoming platform tag, with optional `branch` pinning via `workflow_dispatch` or `repository_dispatch` payload.
> 
> Decision logic moves into **`platform-version-bump.sh`**, with a **`test` PR job** running **`test-platform-version-bump.sh`** (replacing the unwired **`test-version-comparison.sh`**). The workflow splits into **`decide`** (target, skip if not newer, stale PR list) and **`apply`** (checkout target branch, bump `pkg/platform/version.go`, open PR with branch-scoped head name, close superseded bump PRs). **`version_gt`** fixes pre-release comparison (uses `tr` instead of broken bash substitution in the old inline compare).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit acfcf95c306e4da3e0e3dbf0a106395d41ea552b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Piotr1215:
This should be now in vclsuter-pro

PR #4037, opened 2026-06-26
fix: keep pods healthy when tenant and host run different K8s versions

flomedja:
**What issue type does this pull request address?** (keep at least one, remove the others) 
/kind bugfix

**What does this pull request do? Which issues does it resolve?** (use `resolves #<issue_number>` if possible) 
resolves #

Pods could get stuck at `Ready=False`, so their Deployments and StatefulSetsvnever became ready — when a tenant cluster ran an older Kubernetes version than the host. This showed up with a 1.32/1.33 tenant on a 1.34/1.35 host (e.g. EKS), and the syncer logged `field is immutable` or `object has been modified` errors.

Fixes #3578
Resolves ENGNODE-105

### Root causes

**1. QOS class (the "field is immutable" error)** : The pod syncer copied the tenant's QOS class onto the host object before
patching. A later status update then tried to write that QOS class to the host, which Kubernetes 1.32+ rejects because `status.qosClass` is immutable. We now leave the host QOS class untouched and keep the tenant's own value on the tenant
pod, so no immutable-field patch is ever produced.

**2. observedGeneration (the stuck `Ready=False`)**
A host kubelet on K8s 1.34+ sets `observedGeneration` on pod status and conditions. A tenant apiserver on K8s < 1.34 drops that field on write (`PodObservedGenerationTracking` is alpha/off in 1.33 and absent before). That made the object cache (what we sent) and the informer (what got stored) disagree on every reconcile. The syncer read this as a "conditions changed" event and kept overwriting the real `Ready=True` condition with stale data. We now ignore `observedGeneration` in that comparison when the tenant cluster strips it.

**Please provide a short message that should be published in the vcluster release notes**

Fixed an issue where vCluster pods stayed Ready=False (blocking their Deployments and StatefulSets) when the tenant cluster ran an older Kubernetes version than the control plane cluster.

**What else do we need to know?** 

### Default Test Execution
The mandatory PR suite runs automatically. Only specify additional test suites below if needed.

### Adding New Test Suites
When adding a new ginkgo test suite:
- [ ] **Add labels** to the test suite
- [ ] **Update label-filter section** below to execute the new test suite
- [ ] **Verify test suite runs** in CI/CD pipeline

### Additional test suites

Additional test suite(s) that will be executed before the mandatory PR suite:

```label-filter
none
```

---

> [!NOTE]
> **Medium Risk**
> Changes core pod status reconciliation and startup host-version discovery; behavior is version-gated but affects all shared-mode pod sync paths.
> 
[...]

PR #4122, opened 2026-07-29
fix(ci): stop no-op edited runs masking the real e2e result

Piotr1215:
/kind bugfix

Closes DEVOPS-1203

A PR-description edit fires a second run of the e2e workflow on the same head SHA. That run correctly skips the suite (`skip-edited=true`, DEVOPS-1057), but it published its check under the same name as the run that actually executed the tests. GitHub surfaces only the newest run per check name, so the skip overwrote the real pass/fail and the PR showed "skipping" for a suite that had already finished, sometimes red.

This names the skipped job differently, so both checks stay visible and the real result wins. The DEVOPS-1057 skip behaviour is untouched.

**Evidence**

Measured across recent PRs, counting cases where the newest run of the workflow skipped the test job while an earlier run on the same SHA executed it:

| Repo / workflow | Masked | Sampled |
| --- | --- | --- |
| vcluster `e2e-ginkgo.yaml` | 24 | 29 |
| vcluster-pro `e2e-oss-ginkgo.yaml` | 4 | 6 |
| vcluster-pro `e2e.yaml` | 7 | 18 |
| loft-enterprise `e2e-ginkgo.yaml` | 11 | 24 |

Three PRs merged with a failing suite displayed as skipping:

| PR | Real run | Visible run |
| --- | --- | --- |
| #4037 | [30466414251](https://github.com/loft-sh/vcluster/actions/runs/30466414251) failed 15:48:55 | 30466439411 skipped, merged 16:02:06 |
| #4121 | [30468864161](https://github.com/loft-sh/vcluster/actions/runs/30468864161) failed | 30468895732 skipped, merged |
| #4114 | [30351654374](https://github.com/loft-sh/vcluster/actions/runs/30351654374) failed | 30351690516 skipped, merged |

vcluster is the worst affected because Cursor Bugbot edits the description on nearly every PR.

**Why this shape and not the other**

The alternative is dropping `edited` from the `pull_request` types entirely and moving label-filter retargeting to an explicit trigger. That removes the duplicate run rather than renaming it, and remains available as a follow-up. I did not take it here because it changes how people request extra suites, which is a UX decision for the team rather than a bug fix. This change does not preclude it.

**Verification, and its limit**

`actionlint` passes and the YAML parses. Job-level `name:` accepts the `needs` context (confirmed against actionlint's context table, with a negative control showing it rejects `steps`). No repo gates merges on these checks: vcluster and vcluster-pro have no required status checks configured, and loft-enterprise requires only `Build Loft`, `lint`, `unit-tests` and `check`, so no branch protection breaks.

This PR cannot demonstrate the fix on itself. `e2e-ginkgo.yaml`'s `detect_changes` path list watches `.github/workflows/e2e.yaml` but not `e2e-ginkgo.yaml`, so a workflow-only change reports `has_changed=false` and the suite is skipped for an unrelated reason. The behaviour will be visible on the next code PR. That the e2e workflow does not watch itself is a separate issue and I have left it alone.

CI only, no release note needed.

Companion PRs, same change:

- loft-sh/vcluster-pro (`e2e.yaml`, `e2e-oss-ginkgo.yaml`)
- loft-sh/loft-enterprise (`e2e-ginkgo.yaml`)

Two things found while investigating and deliberately not fixed here: vcluster-pro `e2e-oss-ginkgo.yaml` inlines its own copy of the skip-edited shell instead of calling `parse-label-filter@parse-label-filter/v1`; and loft-enterprise has no `edited`/`code` concurrency split at all, where 7 of 24 sampled PRs ran the workflow repeatedly, one 36 times, without the test job ever executing.

### Additional test suites

```label-filter
none
```

---

> [!NOTE]
> **Low Risk**
> Workflow display-name only; no change to when tests run or application code.
> 
> **Overview**
> **Fixes misleading PR checks when a PR description edit retriggers e2e without running tests** (DEVOPS-1203).
> 
> The `ginkgo-e2e-tests` job now uses a **dynamic `name:`**: when `parse_label-filter` sets `skip-edited=true` on a no-op `edited` event, the check is published as **Execute test suites (skipped, PR description edit)** instead of sharing **Execute test suites** with the run that actually executed Ginkgo. That keeps both checks visible so the newest run no longer replaces a real pass/fail with “skipped.”
> 
> Skip logic from DEVOPS-1057 is unchanged; only the displayed check name differs on skipped edited runs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84d4a1107d3882a198f719d3c2de1b1c183b66cb. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/e2e-ginkgo.yaml
+++ b/.github/workflows/e2e-ginkgo.yaml
@@ -167,7 +167,11 @@ jobs:
           retention-days: 7
 
   ginkgo-e2e-tests:
-    name: Execute test suites
+    # A no-op `edited` run must not publish a check under the same name as the
+    # run that actually executed the suite. GitHub surfaces only the newest run
+    # per check name, so the skip silently overwrote the real pass/fail and the
+    # PR showed "skipping" for a suite that had already gone red. DEVOPS-1203.
+    name: "${{ needs.parse_label-filter.outputs.skip-edited == 'true' && 'Execute test suites (skipped, PR description edit)' || 'Execute test suites' }}"
     needs:
       - build
       - parse_label-filter


Piotr1215:
Closing: wrong repo. \`.github/workflows/e2e-ginkgo.yaml\` is owned by the monorepo staging tree at \`staging/github.com/loft-sh/vcluster/\` in loft-sh/vcluster-pro and replayed here by \`sync-to-oss\`, whose \`exclude-paths\` list covers seven producer workflows and not this one. Editing the mirror directly would have been absorbed back through \`sync-from-oss\` instead of landing in the monorepo.

The change now ships in loft-sh/vcluster-pro#2114, which carries \`Closes DEVOPS-1203\` and covers both the Pro workflows and the OSS staging copy.

Issue #3578, opened 2026-02-11
Sync issues with K8s hosts 1.34 or 1.35


chirauki:
Hello everyone!

I'm raising this issue as it may be related to https://github.com/loft-sh/vcluster/issues/2472.

I've found that on EKS host clusters running versions 1.34 or 1.35, vCluster fails to sync the state for some objects, rendering the guest clusters almost unusable. This happened to me using `k3s` as a guest with a version different than the host cluster. Ie. 1.34 `k3s` guests work fine in 1.34 hosts, and 1.35 `k3s` guests work fine in 1.35 hosts. Running a guest version different from the host one leads to this issue.

Within the guest cluster, you'd see situations like the one below:

```
❯ k get po
k get NAME                                        READY   STATUS    RESTARTS   AGE
gfn-p-1129rc1-1-33-t2-cp1-54d5d7cff-gzf85   1/1     Running   0          127m
❯ k get deployments.apps
NAME                        READY   UP-TO-DATE   AVAILABLE   AGE
gfn-p-1129rc1-1-33-t2-cp1   0/1     1            0           127m
```

Where the pod for a deployment is ready, but the deployment never shows ready.

Inspecting the logs, I can see lines like:

```
2026-02-11 12:47:25     ERROR   controller/controller.go:474    Reconciler error        {"component": "vcluster", "controller": "pod", "namespace": "cert-manager", "name": "p-1129rc1-1-33-t2-cp1-cert-manager-7cf4ddcbbf-k5m2n", "reconcileI
D": "fcdb4b71-240f-4ea5-ab40-1ca04de63e78", "error": "sync: patch host object: update object status: Pod \"p-1129rc1-1-33-t2-cp1-cert-manager-7cf4ddcbbf-k5m2n--b21ec34921\" is invalid: status.qosClass: Invalid value: \"BestEffort\": field
 is immutable"}
```

Restarting the API server pod sometimes fixes some of these issues, but they eventually come back after creating new deployments or updating existing ones.

This issue presented using vCluster 0.28.0, but is still happening with 0.31.0.

I'm not clear at this point if this is an issue with vCluster or with my config... I'd appreciate any pointers to be able to further pinpoint the cause of this.

vissersg:
I believe I am encountering the same problem with:

EKS Host: v1.34
virtual k8s : v1.32 / v1.33 
vCluster: v0.31.0

[...]

PR #4123, opened 2026-07-29
test(pods): add tests for running an older virtual cluster on a newer host

flomedja:
**What issue type does this pull request address?** (keep at least one, remove the others) <br>/kind bugfix<br>/kind enhancement<br>/kind feature<br>/kind documentation<br>/kind test

**What does this pull request do? Which issues does it resolve?** (use `resolves #<issue_number>` if possible) <br>resolves ENGNODE-696

**Please provide a short message that should be published in the vcluster release notes**<br>Fixed an issue where vcluster ...

### Default Test Execution

### Adding New Test Suites

### Additional test suites

```label-filter
none
```

---

> \[!NOTE\]<br>**Low Risk**<br>Test-only changes that exercise existing pod sync behavior; no production code paths modified in this PR.
>
> **Overview**<br>Adds **test coverage** for running an older virtual cluster on a newer host (issue loft-sh/vcluster#3578), without changing production sync logic in this diff.
>
> A new **e2e** case creates a pod with `spec.subdomain` so the host pod gets an extra init container and **QOS diverges** (virtual BestEffort vs host Burstable). It asserts the pod stays **Ready**, each side keeps its own `status.qosClass`, and no **immutable field** events appear from bad QOS writes.
>
> **Unit tests** extend pod translate `Diff` for **cache/informer divergence** on `PodReady` `observedGeneration` (stripped on virtual < 1.34 vs preserved on 1.34+). Another test locks `resizeHostPodContainerResourcesInPlace` so in-place resize is skipped when the host version is unknown or below 1.35, even when a resize patch would otherwise be built.
>
> Reviewed by [Cursor Bugbot](<https://cursor.com/bugbot>) for commit de0d8e898734cf0d91061bdd19718e660a528008. Bugbot is set up for automated code reviews on this repo. Configure [here](<https://www.cursor.com/dashboard/bugbot>).

---

> [!NOTE]
> **Low Risk**
> Only e2e and unit tests are added or reshaped; no production code paths change in this diff.
> 
> **Overview**
> Adds **test-only** coverage for older virtual clusters on newer hosts ([#3578](https://github.com/loft-sh/vcluster/issues/3578)); production sync logic is unchanged in this diff.
> 
> The **e2e** QOS regression case now creates a pod with `spec.subdomain` so virtual **BestEffort** and host **Burstable** QOS can diverge. It asserts each side keeps its own `status.qosClass`, **Ready** stays stable, and no **SyncError** events mention `qosClass`. The separate e2e case that only checked Ready flapping was removed from this file (version-skew coverage lives elsewhere).
> 
> **Unit tests** add `TestDiffConditionsCacheInformerDivergence` for virtual **cache vs informer** drift on `PodReady` `observedGeneration` (stripped below 1.34 vs preserved on 1.34+). They also lock **`resizeHostPodContainerResourcesInPlace`** to no-op when the host version is unknown or &lt; 1.35, **`TestSyncTerminatingHostPodStatus`** for phase propagation without overwriting virtual QOS or redundant writes, and a tiny rename in an existing diff table case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 15092be90471a12edab8984adafa4068846675e2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


e2e/test_core/sync/test_pods_versionskew.go
```diff
@@ -195,6 +195,84 @@ func PodVersionSkewSpec() {
...
+				suffix := random.String(6)
+				ns := "skew-qos-test-" + suffix
+				createTestNamespace(ctx, ns)
+
+				podName := "skew-qos-pod-" + suffix
+				By("Creating a pod without resource requests and with a subdomain", func() {
+					_, err := vClusterClient.CoreV1().Pods(ns).Create(ctx, &corev1.Pod{
+						ObjectMeta: metav1.ObjectMeta{Name: podName},
+						Spec: corev1.PodSpec{
+							Subdomain:  "skew-qos",
```
sowmyav27: **Necessity.** This `Subdomain` divergence trick is exactly what `e2e/test_core/sync/test_pods.go:899` (`should preserve virtual pod QOS class`) is missing, and that spec is the one the review comment was anchored on. Right now both run on every PR and only this one proves anything.

Can we apply the same fix there rather than leaving a no-op QOS spec beside a real one?

e2e/test_core/sync/test_pods_versionskew.go
```diff
@@ -195,6 +195,84 @@ func PodVersionSkewSpec() {
 					}).WithContext(ctx).WithPolling(constants.PollingInterval).WithTimeout(constants.PollingTimeoutShort).Should(Succeed())
 				})
 			})
+
+			It("should keep the pod Ready when the host and virtual QOS classes diverge", func(ctx context.Context) {
```
sowmyav27: **Placement.** The QOS invariant is version-independent: `diff.go:317` (`desired.QOSClass = virtualStatus.QOSClass`) has no version gate, and the rewrite-hosts injection at `translator.go:358` doesn't either. So this spec gains nothing from the pinned v1.33 tenant, and loses coverage the day that pin moves past 1.34.

Suggest moving it into `PodSyncSpec` (the default suite), which also fixes `test_pods.go:899` in the same edit.

e2e/test_core/sync/test_pods_versionskew.go
```diff
@@ -195,6 +195,84 @@ func PodVersionSkewSpec() {
...
+							"pod Ready condition flapped away from True: %v", vPod.Status.Conditions)
+
+						pPod, err := hostClient.CoreV1().Pods(hostNS).Get(ctx, pPodName, metav1.GetOptions{})
+						g.Expect(err).To(Succeed(), "failed to get host pod %s/%s", hostNS, pPodName)
+						g.Expect(pPod.Status.QOSClass).To(Equal(corev1.PodQOSBurstable),
+							"host pod QOS class was overwritten to %s", pPod.Status.QOSClass)
+
+						// This is the check that catches the bug: the apiserver rejects a
+						// bad QOS write, so the values above pass either way and only
+						// this error event reveals it.
```
sowmyav27: **Comment accuracy.** The pre-fix code already preserved QOS at every write site: `diff.go:35-39`, `syncer.go:318-320`, and `event.Host.Status.QOSClass = event.Virtual.Status.QOSClass` at `syncer.go:381`, which ran before `NewSyncerPatcher` snapshots the baseline so no `qosClass` ever reached a patch. There was no QOS bug for this to catch.

What it does do is guard against the preservation being removed, which is worth having. Could we say that instead?

```suggestion
						// Guards against removing the QOS preservation in DesiredVirtualStatus:
						// the apiserver rejects the write, so the stored values above stay
						// correct either way and only the sync error reveals it.
```

e2e/test_core/sync/test_pods_versionskew.go
```diff
@@ -195,6 +195,84 @@ func PodVersionSkewSpec() {
...
+						events, err := vClusterClient.CoreV1().Events(ns).List(ctx, metav1.ListOptions{
+							FieldSelector: "involvedObject.name=" + podName,
+						})
+						g.Expect(err).To(Succeed(), "failed to list events for pod %s/%s", ns, podName)
+						for _, event := range events.Items {
+							g.Expect(event.Message).NotTo(ContainSubstring("field is immutable"),
+								"the syncer tried to overwrite the immutable QOS class: %s", event.Message)
+						}
```
sowmyav27: **Assertion strength.** This keys on the apiserver's wording. `k8s.io/kubernetes` core pod validation isn't vendored here, so nothing in the repo pins `field is immutable` for `status.qosClass`. If upstream rewords it, the check goes quietly green instead of failing.

Keying on `Reason` is stable and still surfaces the message:

```suggestion
						for _, event := range events.Items {
							g.Expect(event.Reason).NotTo(Equal("SyncError"),
								"the syncer failed to patch the pod: %s", event.Message)
						}
```

sowmyav27:
**Ticket scope.** This covers Item 1 of ENGNODE-696 (the QOS spec). Items 2 and 3 are still open:

- **Item 2** — no unit coverage for the terminating-host-pod status path at `pkg/controllers/resources/pods/syncer.go:305`. `TestDesiredVirtualStatus` (`diff_test.go:453`) tests the helper, not the caller. Nothing in the repo calls the pods `Sync()` terminating branch; `TestSync` only calls `SyncToHost`.
- **Item 3** — the duplicate spec at `e2e/test_core/sync/test_pods.go:936` (`should keep pod Ready condition stable after reaching Ready=True`) is still there. Its own comment says the bug can't reproduce on a same-version cluster, and it runs in the PR gate.

Could we move Items 2 and 3 into a new ticket so they don't get lost when this one closes?

flomedja:
> Ticket scope. This covers Item 1 of ENGNODE-696 (the QOS spec). Items 2 and 3 are still open:
> 
> Item 2 — no unit coverage for the terminating-host-pod status path at pkg/controllers/resources/pods/syncer.go:305. TestDesiredVirtualStatus (diff_test.go:453) tests the helper, not the caller. Nothing in the repo calls the pods Sync() terminating branch; TestSync only calls SyncToHost.
> Item 3 — the duplicate spec at e2e/test_core/sync/test_pods.go:936 (should keep pod Ready condition stable after reaching Ready=True) is still there. Its own comment says the bug can't reproduce on a same-version cluster, and it runs in the PR gate.
> Could we move Items 2 and 3 into a new ticket so they don't get lost when this one closes?

I missed this. It is fixed.

e2e/test_core/sync/test_pods_versionskew.go
flomedja: Done

e2e/test_core/sync/test_pods_versionskew.go
flomedja: Done

e2e/test_core/sync/test_pods_versionskew.go
flomedja: Done

e2e/test_core/sync/test_pods_versionskew.go
flomedja: Done, with one adjustment: the QOS divergence doesn't depend on version skew (the field is immutable on any K8s 1.32+), so I moved the divergence spec into `test_pods.go` as the new body of `should preserve virtual pod QOS class`  instead of keeping copies in both suites. The version-skew suite now only holds the observedGeneration specs, which actually need virtual < 1.34 <= host.

e2e/test_core/sync/test_pods_versionskew.go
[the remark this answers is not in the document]
flomedja: Done

sowmyav27 (approved):

PR #4144, opened 2026-08-14
fix(docker): support running the docker driver on podman

### LINKED ISSUE #5: Podman: Priviledged port mapping not allowed despite Rootful
Hi!

I saw the vind announcement on Reddit and thought it sounded like a great project.

I use Podman 5.7.1 with a Rootful machine instance on a Mac M1.
When creating a first cluster with
`vcluster create my-first-cluster`

I get the following output:
```
17:39:18 info Ensuring environment for vCluster my-first-cluster...
17:39:18 info Docker is using the non-containerd image store, please use containerd image store to use the docker daemon registry proxy. For more information, see https://docs.docker.com/engine/storage/containerd/
17:39:21 warn Load balancer type services are not supported inside the vCluster because privileged port mapping is not allowed. If you are using Docker Desktop, please enable it in the Docker Desktop settings
```

But Podman Machine is rootful:
```
podman machine inspect --format "{{.Rootful}}"
true
```

Also I have other containers running using priviledged ports.
Do you have any suggestions what the issue might be?

Thanks!

saiyam1814:
## What does this PR do?

Makes `vcluster create --driver docker` (vind) work on Podman without manual workarounds. With a rootful podman machine running, `vcluster create my-cluster --driver docker` now works out of the box, no `DOCKER_HOST` export, no socket shims, no manual modprobe.

Four changes, each addressing a failure users currently work around by hand:

1. **Podman fallback for the container daemon** (`pkg/cli/docker_runtime.go`)
   Every docker driver entry point now verifies the docker CLI can reach a daemon. If not, and Podman is installed, the CLI resolves Podman's Docker-compatible API socket (`podman machine inspect` on macOS/Windows, `podman info` on Linux) and sets `DOCKER_HOST` for the process. An explicitly set `DOCKER_HOST` is never overridden; unreachable endpoints now produce a clear error instead of a confusing downstream failure.

2. **Socket detection fixed for Podman VMs** (`getDockerSocketPath`)
   The load balancer setup detected the docker socket by scanning `netstat -xlp` for `docker.sock`. In a podman machine, `/var/run/docker.sock` is a symlink to `/run/podman/podman.sock`, which never appears in netstat, so the whole create failed with `no docker socket path found`. Detection now checks well-known socket paths first (a symlink to a live socket passes `[ -S ]`) with netstat as the fallback, and a detection failure disables the load balancer with a warning instead of failing the create.

3. **Kernel modules loaded inside the runtime VM** (`ensureKernelModules`)
   Node join and Flannel need `overlay`, `bridge` and `br_netfilter`. Docker Desktop preloads them; podman machine (Fedora CoreOS) does not, and the existing modprobe only ran on Linux hosts. On non-Linux hosts the CLI now loads missing modules and sets the bridge netfilter sysctls inside the VM, using the same privileged `nsenter` container pattern the socket detection already uses. This also hardens Colima, Lima and Rancher Desktop setups. Failures stay warnings, since the modules can be built into the kernel.

4. **install-standalone.sh fetched on the host** (`getInstallStandaloneScript`)
   The control plane bootstrap used to `curl` the script from github.com inside the container, which corporate proxies commonly block. The CLI now downloads the version-matched script on the host (where proxy and CA configuration exist), caches it under `~/.vcluster/docker/install-standalone/<version>/`, and streams it into the container over stdin. Air-gapped environments can point `VCLUSTER_INSTALL_STANDALONE_SCRIPT` at a local copy.

Also updates the privileged port warning to explain the Podman limitation (fixes loft-sh/vind#5): Podman cannot publish privileged ports bound to a specific host IP through gvproxy, see containers/podman#28009. Load balancer support on Podman is blocked on that upstream issue and degrades gracefully.

## Why?

Users have been building elaborate manual workarounds to run vind on Podman (socat systemd units to fake a netstat-visible docker.sock, manual modprobe over `podman machine ssh`, pre-downloading and `podman cp`-ing the install script after an expected failure). Related reports: loft-sh/vind#5, and the kernel module gap likely explains "Flannel stuck / pods pending" style reports on non Docker Desktop runtimes (loft-sh/vind#16, loft-sh/vind#19).

## Testing

All flows tested live on macOS (arm64):

**Podman 6.1.0, rootful applehv machine, freshly restarted (no kernel modules loaded), Docker Desktop stopped, no `DOCKER_HOST` set:**
- `vcluster create podtest --driver docker --chart-version 0.36.0 -f worker.yaml`: CLI logged `Docker daemon not found, using Podman's Docker-compatible API at ...`, loaded bridge/br_netfilter inside the VM, streamed the install script over stdin, control plane and worker both reached Ready, Flannel and CoreDNS Running, pod scheduling and service DNS verified.
- `vcluster connect / list / delete` all work through the fallback.
- Script cache written and reused on subsequent creates.

**Docker Desktop (regression):**
- Same create succeeds unchanged; no fallback triggered.
- New socket detection command returns the identical path as the old netstat scan inside the Docker Desktop VM.

**Error paths:**
- `DOCKER_HOST` pointing at a dead socket: clear error, no silent podman override.
- Unit tests for the install script override and cache (`create_docker_test.go`).

**Not covered:** load balancer end-to-end on Podman (blocked upstream, degrades gracefully with an explanatory warning), Windows podman machine (the fallback returns an explicit "set DOCKER_HOST manually" error there since podman on Windows uses a named pipe), sudo load balancer path on macOS (socket detection verified manually inside both VMs instead).

## Hardening after adversarial review

A second commit addresses findings from an internal review of the first commit:

- Socket detection scans netstat listeners **first** (exact pre-existing behavior for docker hosts, so a stale socket file can never shadow a live listener) and only falls back to the well-known paths when netstat yields nothing, which is the podman machine case.
- The podman fallback notice logs at warn level (stderr), so `vcluster list --driver docker --output json` output stays clean, and the fallback is explicitly unsupported on Windows with an actionable error instead of a misleading one.
- The install script download has a 1 minute timeout, validates the payload starts with a shebang (an intercepting proxy returning an HTML error page is rejected instead of executed), and the cache is written atomically (temp file + rename) and self-heals if a corrupted entry is found.
- The VM kernel module helper only sets the bridge netfilter sysctls when it actually had to load a module, so docker daemons that already have everything in place (including remote `DOCKER_HOST` machines) are never modified.
- `EnsureDockerDaemon` is exported and wired into `vcluster node load-image`, which also shells out to docker.

Known trade-offs called out for reviewers: if the docker daemon is briefly down while podman runs, the CLI falls back for that invocation (visible via the warn log); a socket detection failure disables the load balancer in the persisted config the same way the existing privileged-port and sudo checks do (re-running `create --upgrade` re-evaluates it); other podman-compat gaps in `vcluster platform start --docker` and the kind/minikube helpers are pre-existing and out of scope here.

---

> [!NOTE]
> **Medium Risk**
> Changes container daemon selection, VM kernel networking, and control-plane bootstrap paths; failures degrade with warnings in several places but incorrect socket or module handling could still break cluster creation on non-Docker Desktop runtimes.
> 
> **Overview**
> Makes **`vcluster create --driver docker`** work on Podman without manual `DOCKER_HOST` or socket shims by centralizing runtime setup in **`EnsureDockerDaemon`**, which pings Docker and falls back to Podman’s Docker-compatible socket (respecting an explicit **`DOCKER_HOST`**).
> 
> **Runtime and networking:** **`ensureKernelModules`** loads `overlay`, `bridge`, and `br_netfilter` on Linux hosts or inside the container VM (Podman machine, Colima, etc.). **`getDockerSocketPath`** prefers live `netstat` listeners, then well-known paths (including Podman’s socket); missing socket disables load balancer with a warning instead of failing create. Privileged-port messaging mentions Podman’s upstream limitation.
> 
> **Bootstrap:** **`install-standalone.sh`** is fetched on the host (cache, **`VCLUSTER_INSTALL_STANDALONE_SCRIPT`**, shebang validation, atomic cache writes) and piped into the control-plane container instead of curling from inside the VM.
> 
> **Coverage:** Daemon check is wired into connect, list, delete, describe, pause/resume, snapshot/restore, and **`vcluster node load-image`**. Tests cover install-script override, cache, and validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 912689742f1c09db98f4bc43757910c6751356a2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/vclusterctl/cmd/node/load-docker-image.go
+++ b/cmd/vclusterctl/cmd/node/load-docker-image.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 
 	"github.com/loft-sh/log"
+	"github.com/loft-sh/vcluster/pkg/cli"
 	"github.com/loft-sh/vcluster/pkg/cli/flags"
 	"github.com/loft-sh/vcluster/pkg/snapshot/pod"
 	"github.com/spf13/cobra"
@@ -67,6 +68,11 @@ func (o *LoadImageOptions) Run(ctx context.Context, nodeName string) error {
 
 	// save image to archive
 	if o.Image != "" {
+		// make sure a docker daemon is reachable, falling back to podman if available
+		if err := cli.EnsureDockerDaemon(ctx, o.Log); err != nil {
+			return err
+		}
+
 		o.Log.Infof("Saving image %s to archive...", o.Image)
 		if err := runCommand("docker", "save", "-o", "image.tar.gz", o.Image); err != nil {
 			return fmt.Errorf("failed to save image: %w", err)
--- a/pkg/cli/connect_docker.go
+++ b/pkg/cli/connect_docker.go
@@ -52,6 +52,12 @@ type connectDocker struct {
 }
 
 func ConnectDocker(ctx context.Context, options *ConnectOptions, globalFlags *flags.GlobalFlags, vClusterName string, command []string, log log.Logger) error {
+	// make sure a docker daemon is reachable, falling back to podman if available
+	err := ensureDockerDaemon(ctx, log)
+	if err != nil {
+		return err
+	}
+
 	cmd := &connectDocker{
 		GlobalFlags:    globalFlags,
 		ConnectOptions: options,
--- a/pkg/cli/connect_docker.go
+++ b/pkg/cli/connect_docker.go
@@ -53,7 +53,7 @@ type connectDocker struct {
 
 func ConnectDocker(ctx context.Context, options *ConnectOptions, globalFlags *flags.GlobalFlags, vClusterName string, command []string, log log.Logger) error {
 	// make sure a docker daemon is reachable, falling back to podman if available
-	err := ensureDockerDaemon(ctx, log)
+	err := EnsureDockerDaemon(ctx, log)
 	if err != nil {
 		return err
 	}
--- a/pkg/cli/create_docker.go
+++ b/pkg/cli/create_docker.go
@@ -76,7 +76,7 @@ func CreateDocker(ctx context.Context, options *CreateOptions, globalFlags *flag
 	}
 
 	// make sure a docker daemon is reachable, falling back to podman if available
-	err := ensureDockerDaemon(ctx, log)
+	err := EnsureDockerDaemon(ctx, log)
 	if err != nil {
 		return err
 	}
@@ -477,10 +477,13 @@ func getInstallStandaloneScript(ctx context.Context, vClusterVersion string, glo
 		return script, nil
 	}
 
-	// check the cache first
+	// check the cache first, discarding a corrupted entry so it gets re-downloaded
 	cachePath := filepath.Join(filepath.Dir(globalFlags.Config), "docker", "install-standalone", "v"+vClusterVersion, "install-standalone.sh")
 	if script, err := os.ReadFile(cachePath); err == nil {
-		return script, nil
+		if err := validateInstallStandaloneScript(script); err == nil {
+			return script, nil
+		}
+		_ = os.Remove(cachePath)
 	}
 
 	scriptURL := fmt.Sprintf("https://github.com/loft-sh/vcluster/releases/download/v%s/install-standalone.sh", vClusterVersion)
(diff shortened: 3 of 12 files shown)


saiyam1814:
@codex review

